Issue #6, opened 2019-11-26
Screen viewing does not restart after screen sharing is toggled off then on.


dylonChime:
**Important:** Please attach INFO-level logs with timestamps to the report from *before* the session started up until the point at which the unexpected behavior occured. For multi-participant calls, attach a log file for each participant if possible.
[screenshare-toggeled-bug.log](https://github.com/aws/amazon-chime-sdk-js/files/3893704/screenshare-toggeled-bug.log)

**Meeting and attendee IDs:**
List the meeting and attendee IDs that the attached logs apply to.
MeetingId: ab05ded2-b910-4a8d-8404-2fac42330a99
AttendeeId: e19373bd-8d97-46c4-9204-40a46ae55bfe
**Timeline**
Outline the main actions the user performed and when:

* 2019-11-26T19:28:23.437Z DefaultScreenShareFacadeStop
* 2019-11-26T19:28:23.541Z received StreamEnd
* 2019-11-26T19:28:23.543Z DefaultScreenViewingDeltaRenderer: Closing

**Describe the bug**
Describe the bug and the timestamp at which it happened relative to the timestamp in the log file.

As seen above, DefaultScreenViewingDeltaRenderer is closed when screenshare stops. This closes screen viewing.

**Screenshots**
If applicable, add screenshots to help explain the problem.

**Platform**
 - OS: Mojave 10.14.6
 - Browser: Chrome
 - Browser Version: 78.0.3904.108
 - Amazon Chime SDK GitHub Commit: fbabb4601bd025171d43a090c28608dcab2235fd

PR #7, opened 2019-11-26
Fix bug that caused screenview to stay off when screenshare was toggled

dylonChime:
*Issue #:* 
#6
*Description of changes*
Added hide and reveal methods to deltaRenderer. This allows screen view to stay open but hidden when screen share closes.

### DIFF
--- a/src/screenviewing/deltarenderer/DefaultScreenViewingDeltaRenderer.ts
+++ b/src/screenviewing/deltarenderer/DefaultScreenViewingDeltaRenderer.ts
@@ -188,6 +188,18 @@ export default class DefaultScreenViewingDeltaRenderer implements ScreenViewingD
     }
   }
 
+  hideViewport(): void {
+    if (this.content) {
+      this.content.style.display = 'none';
+    }
+  }
+
+  revealViewport(): void {
+    if (this.content) {
+      this.content.style.display = 'block';
+    }
+  }
+
   changePresentationPolicy(policy: PresentationPolicy): void {
     this.policy = policy;
     this.updatePresentation();
--- a/test/screenviewing/deltarenderer/DefaultScreenViewingDeltaRenderer.test.ts
+++ b/test/screenviewing/deltarenderer/DefaultScreenViewingDeltaRenderer.test.ts
@@ -434,4 +434,47 @@ describe('DefaultScreenViewingDeltaRenderer', () => {
       deltaRenderer.zoomReset();
     });
   });
+
+  describe('hideViewport and revealViewport', () => {
+    it('hides and reveals the viewport', () => {
+      let style = {
+        display: 'none',
+      };
+
+      let canvas = {
+        style: style,
+      };
+
+      const document: SubstituteOf<Document> = Substitute.for();
+      document.createElement(Arg.any()).returns(canvas);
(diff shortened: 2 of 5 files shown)


src/screenviewing/deltarenderer/DefaultScreenViewingDeltaRenderer.ts
```diff
@@ -188,6 +188,18 @@ export default class DefaultScreenViewingDeltaRenderer implements ScreenViewingD
     }
   }
 
+  hideViewport(): void {
+    if (this.viewport) {
+      this.viewport.style.display = 'none';
```
aburnson: I don't think we should modify the CSS styling of the viewport since this is an element that is passed in by the SDK consumer. In particular, they may be controlling the `display` attribute themselves and so this could overwrite the attribute. Is there some other way to accomplish the objective here without using CSS styling?

dylonChime: As discussed offline, I modified the CSS for the canvas element rather than the exposed viewport.

src/screenviewing/deltarenderer/DefaultScreenViewingDeltaRenderer.ts
```diff
@@ -188,6 +188,18 @@ export default class DefaultScreenViewingDeltaRenderer implements ScreenViewingD
     }
   }
 
+  hideViewport(): void {
+    if (this.viewport) {
```
aburnson: Should this be `if (this.content) {`?

src/screenviewing/deltarenderer/DefaultScreenViewingDeltaRenderer.ts
```diff
@@ -188,6 +188,18 @@ export default class DefaultScreenViewingDeltaRenderer implements ScreenViewingD
...
+      this.content.style.display = 'none';
+    }
+  }
+
+  revealViewport(): void {
+    if (this.viewport) {
```
aburnson: Should this be `if (this.content) {`?

src/screenviewing/deltarenderer/DefaultScreenViewingDeltaRenderer.ts
dylonChime: Yes. Updated.

src/screenviewing/deltarenderer/DefaultScreenViewingDeltaRenderer.ts
dylonChime: Yes. Updated.

test/screenviewing/deltarenderer/DefaultScreenViewingDeltaRenderer.test.ts
```diff
@@ -434,4 +434,21 @@ describe('DefaultScreenViewingDeltaRenderer', () => {
       deltaRenderer.zoomReset();
     });
   });
+
+  describe('hideViewport and revealViewport', () => {
+    it('hides and reveals the viewport', () => {
```
aburnson: This test does not verify the change made by `hideViewport` or `revealViewport`. It looks like other tests in this file check properties of the canvas. Could we do the same here?

PR #12, opened 2019-12-03
Call track stops before removing tracks from media stream

XHatan:
*Issue #:* 

*Description of changes*
Call track stop before removing tracks from media stream.

### DIFF
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -114,6 +114,8 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
       this.trace('startVideoPreviewForVideoInput', element.id);
       return;
     }
+
+    // TODO: implement MediaDestroyer to provide single release MediaStream function
     this.releaseMediaStream(element.srcObject as MediaStream);
     DefaultVideoTile.disconnectVideoStreamFromVideoElement(element);
     navigator.mediaDevices
@@ -126,6 +128,7 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
           `Unable to reacquire video stream for preview to element ${element.id}: ${error}`
         );
       });
+
     this.trace('startVideoPreviewForVideoInput', element.id);
   }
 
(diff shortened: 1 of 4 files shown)


simmkyu:
typo in the commit message: disconnet => disconnect

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -114,18 +114,34 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+    if (this.activeDevices['video'].stream && this.activeDevices['video'].stream.active) {
+      // bind a cloned MediaStream of chosen activeDevices stream
+      // API users are responsible to managing video preview MediaStream. And ensure stopVideoPreviewForVideoInput is called when end the preview
+      DefaultVideoTile.connectVideoStreamToVideoElement(
+        this.activeDevices['video'].stream.clone(),
+        element,
+        true
+      );
+    } else {
+      this.logger.warn('startVideoPreviewForVideoInput before any videoInputDevice is chosen');
```
aburnson: I don't understand how this edge case is different than the one at the top of the function:

```
if (!this.activeDevices['video']) {
  this.logger.warn('cannot bind video preview since video input device has not been chosen');
  ...
```

How would we end up in this scenario given the above guard?

XHatan: This is more of a demo issue, when we choose synthesized device, we have `this.activeDevices['video'].stream` corresponds to a canvas captured stream. Before this PR, `getUserMedia` is always called here and return a actual video stream instead of canvas-captured stream.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -483,7 +499,12 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     if (restartVideo) {
       this.logger.info('restarting local video to switch to new device');
       this.boundAudioVideoController.restartLocalVideo(() => {
-        this.releaseMediaStream(oldStream);
+        // tracks of oldStream should be stopped when video tile is disconnected from MediaStream
+        // otherwise, camera is still being accessed and we need to stop it here.
+        if (oldStream && oldStream.active) {
+          this.logger.error('previous media stream is not stopped during restart video');
```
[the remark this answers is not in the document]
XHatan: I am being really cautious, `restartLocalVideo()` stops `oldStream`, and this callback of `releaseStream` should not be needed as mentioned in the comment.
When it happens, it indicates an implementation error, it's possible from us or it could be from a wrong API calling sequence. 
I am okay with just logging a warning.

PR #23, opened 2019-12-06
Start marking internal classes and interfaces, first pass

asnyatkov:


aburnson:
The scope of this PR is too narrow. We should find a generic, programmatic way to determine which classes are public or internal based on their relationship to the MeetingSession and then add a linter to enforce that those classes are marked internal. Closing for now until we have a better solve for this.

PR #28, opened 2019-12-12
Implement keyframe request handling

agowatch:
- Add handler for keyframe requests that recycles the MediaRecorder
  instance that wraps the MediaStream.  The first frame the new instance
  emits will always be a keyframe

*Issue #:*

*Description of changes*
Adds a keyframe request handler to enable the SDK to respond to keyframe requests


aws-geoff:
Does this need to have an impact on release version at all?

PR #29, opened 2019-12-12
Fix issue where Safari disconnects when backgrounded

justindarc:
This PR blocks Safari from measuring connection health when backgrounded which prevents it from disconnecting.

To put Safari in the background, it is not enough to simply focus another window to the front. You either need to minimize Safari to the dock or switch to another desktop on macOS.


XHatan:
Thanks for the investigation. I think #37 should resolve this. Close this for now.

PR #32, opened 2019-12-13
Move screen view data connection open and close

aws-geoff:
*Issue #:* 

*Description of changes*
This change ensures we only open the screen viewing data connection when the user requests to view the shared screen and closes the connection when the user requests to stop viewing the shared screen.

### DIFF
--- a/src/screenshareviewfacade/ScreenShareViewFacade.ts
+++ b/src/screenshareviewfacade/ScreenShareViewFacade.ts
@@ -18,12 +18,12 @@ export default interface ScreenShareViewFacade {
    * Starts viewing the screen share within an HTML element. Note that an
    * HTMLCanvas will be placed inside of this element.
    */
-  start(element: HTMLDivElement): void;
+  start(element: HTMLDivElement): Promise<void>;
 
   /**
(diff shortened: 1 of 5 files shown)


agowatch:
I've tested this successfully against the prod bithub.  the screen viewer starts receiving data in 2-3 seconds.

src/screenshareviewfacade/ScreenShareViewFacade.ts
```diff
@@ -18,12 +18,12 @@ export default interface ScreenShareViewFacade {
    * Starts viewing the screen share within an HTML element. Note that an
    * HTMLCanvas will be placed inside of this element.
    */
-  start(element: HTMLDivElement): void;
+  start(element: HTMLDivElement): Promise<void>;
```
asnyatkov: I think this is a breaking behavior change and would require at least minor SDK version bump.

aburnson: Versions are bumped when we call `script/publish`. You do not need to adjust the version as part of the PR. The next time we `npm publish`, we will run `script/publish` and can bump the minor version.

PR #37, opened 2019-12-15
Add alternative for audio device switching and add flag to disable the use of WebAudio

XHatan:
*Issue #:* 

*Description of changes*

We use `WebAudio` in `DefaultDeviceController` for audio stream/track management. It poses some problem: for example, in Safari,  audio will stop when apps running in background. 
This PR provides an alternative implementation where we use audio stream directly. 
When switching device, `replaceTrack` is used in RTCRtpSender to avoid rejoining meeting since triggering a re-subscription violates certain server assumptions. 

### DIFF
--- a/src/meetingsession/MeetingSessionConfiguration.ts
+++ b/src/meetingsession/MeetingSessionConfiguration.ts
@@ -54,6 +54,11 @@ export default class MeetingSessionConfiguration {
   connectionHealthPolicyConfiguration: ConnectionHealthPolicyConfiguration = new ConnectionHealthPolicyConfiguration();
 
   /**
+   * Feature flag to enable WebAudio processing
+   */
+  enableWebAudio: boolean = false;
+
+  /**
    * Constructs a MeetingSessionConfiguration optionally with a chime:CreateMeeting and
    * chime:CreateAttendee response. You can pass in either a JSON object containing the
    * responses, or a JSON object containing the information in the Meeting and Attendee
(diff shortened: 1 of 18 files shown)


aburnson:
Typo in commit message: "Add flag to disbale webaudio"

src/meetingsession/MeetingSessionConfiguration.ts
```diff
@@ -53,6 +53,11 @@ export default class MeetingSessionConfiguration {
    */
   connectionHealthPolicyConfiguration: ConnectionHealthPolicyConfiguration = new ConnectionHealthPolicyConfiguration();
 
+  /**
+   * Feature flag to enable WebAudio processing
+   */
+  enableWebAudio: boolean = true;
```
aburnson: I think we should set this to `false` by default so that it's opt-in rather than opt-out.

aburnson:
I haven't seen the Travis build fail with that error before. Any idea what's going with that?

XHatan:
> 
> 
> I haven't seen the Travis build fail with that error before. Any idea what's going with that?

 I noticed when I built it locally, after all tests passed showing `1112 tests passing (31s)`, it took really long time (minutes) to show `coverage summary`. I think it's due to the tests in 'DefaultAudioVideoController' that I added.

aburnson:
Typo in commit message "disbale" instead of "disable".

PR #42, opened 2019-12-18
Add source node to audio graph for silent devices

aburnson:
*Description of changes*

Add source node to audio graph for silent devices. Without a source node, the WebAudio MediaStreamDestination will not actually contain any audio. The server will not show the attendee as present unless audio data is received. To fix, it is enough to have a gain node attached to the MediaStreamDestinationNode with no input of its own and a gain of 0.

### DIFF
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -249,14 +249,15 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
   static synthesizeAudioDevice(toneHz: number): Device {
     const audioContext = DefaultDeviceController.createAudioContext();
     const outputNode = audioContext.createMediaStreamDestination();
+    const gainNode = audioContext.createGain();
+    gainNode.gain.value = 0.0;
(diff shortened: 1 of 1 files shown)


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -249,14 +249,15 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
   static synthesizeAudioDevice(toneHz: number): Device {
     const audioContext = DefaultDeviceController.createAudioContext();
     const outputNode = audioContext.createMediaStreamDestination();
+    const gainNode = audioContext.createGain();
+    gainNode.gain.value = 0.0;
+    gainNode.connect(outputNode);
```
XHatan: Should this `gainNode.connect(outputNode)` be executed after the following `if` or it does not matter ?

aburnson: It doesn't matter. It just needs to get hooked up to the outputNode. If using a tone generator then that will then get hitched to the gainNode in the if-block.

PR #49, opened 2019-12-20
Remove noSignalStatusReceived from ReconnectionHealthPolicy. Remove SignalStrengthBarsConnectionHealthPolicy.

anuranduttaroy:
…

*Issue #:* 

*Description of changes*
* Removed `noSignalStatusReceived` from ReconnectionHealthPolicy since it was not being used anywhere and clean up everything that went into generating that status.
* Removed SignalStrengthBarsConnectionHealthPolicy since it was not used anywhere.
* Removed related UnitTests


anuranduttaroy:
> 👍 There is some more clean-up we can do as a follow-up to remove the data sources altogether. In particular, take a look at this file and see what fields/methods can be removed:
> 
> https://github.com/aws/amazon-chime-sdk-js/blob/9e8b5e0adc561cf374cf10692a572583a904818a/src/connectionhealthpolicy/ConnectionHealthData.ts#L62

Will update the issue with the comment and will fix it in another commit.

PR #50, opened 2019-12-21
Fix chooseVideoInputDevice with null

XHatan:
*Issue #:* 

*Description of changes*

`chooseVideoInputDevice(null)` indicates no video device and it should just return.


XHatan:
> Nice! Prior to this PR, is SDK always stuck waiting for `navigator.mediaDevices.getUserMedia()` to resolve when a builder passes `null`?
> 
> https://github.com/aws/amazon-chime-sdk-js/blob/2ae6d8f35a5c991036a813103d077434a6ef624f/src/devicecontroller/DefaultDeviceController.ts#L495

This PR doesn't fix that. Prior to this PR, `getUserMedia(null)` rejects promise (instantly), `chooseInputDevice(null)` for video indicates no video device so the same effect (no video) is achieved. This PR only handles the promise rejection in a better way.

PR #61, opened 2020-01-07
Fix #60 - Set device to null when active device unplugged

### LINKED ISSUE #60: Unplugging selected audio device causes reconnect loop
**Important:** Please attach INFO-level logs with timestamps to the report from *before* the session started up until the point at which the unexpected behavior occurred. For multi-participant calls, attach a log file for each participant if possible.

Logs: [reconnect-loop.txt](https://github.com/aws/amazon-chime-sdk-js/files/4027428/reconnect-loop.txt)

**Describe the bug**
Unplugging explicitly selected audio device causes reconnect loop due to stats showing no receive packets for the audio track.

**Screenshots**
If applicable, add screenshots to help explain the problem.

Screen Shot 2020-01-06 at 12 43 12 PM

**Platform**
 - OS: macOS
 - Browser: Chrome
 - Browser Version: Chrome Version 79.0.3945.88 (Official Build) (64-bit)
 - Amazon Chime SDK GitHub Commit: https://github.com/aws/amazon-chime-sdk-js/commit/64e4da7e5150a9675a07144dc7d07d2d3b86d108

aburnson:
*Issue #:* #60 

*Description of changes*

Set device to null when active device unplugged. It is the consuming application's responsibility to monitor device changes and react accordingly with UI to prompt the user about selecting a new device. However, this fix prevents the app from getting into a bad state in the interim. It is possible the user may never select a new device, so the meeting should continue to function with other active inputs and outputs.

Tested by unplugging an active USB mic to see that the warning is generated and the meeting continues to function. When plugging back in the device, it is then possible to reselect it.

### DIFF
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -497,6 +497,14 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
         newDevice.constraints = proposedConstraints;
       } else {
         newDevice = await this.getDeviceFromBrowser(proposedConstraints);
+        newDevice.stream.getTracks()[0].addEventListener('ended', () => {
+          if (this.activeDevices[kind].stream === newDevice.stream) {
+            this.logger.warn(
+              `${kind} input device which was active is no longer available, resetting to null device`
+            );
+            this.chooseInputDevice(kind, null, false);
+          }
+        });
       }
     } catch (error) {
       this.logger.error(


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -497,6 +497,14 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
         newDevice.constraints = proposedConstraints;
       } else {
         newDevice = await this.getDeviceFromBrowser(proposedConstraints);
+        newDevice.stream.getTracks()[0].addEventListener('ended', () => {
```
XHatan: Does device change event fire in this case ?

aburnson: The device change event fires after the `ended` track listener fires. The consuming application can then optionally use that event (via the DeviceController) to prompt the user to select a new device.

PR #75, opened 2020-01-14
Adding docs folder

ltrung:
*Description of changes*
Check in docs folder output from typedoc


package.json
```diff
@@ -13,8 +13,7 @@
     "check": "node script/check-code-style.js",
     "prepublishOnly": "script/publish",
     "lint": "eslint . --ext .ts,.tsx,.js --fix",
-    "build": "npm install && node script/barrelize.js && tsc && npm run copy",
-    "doc": "npm run build && node node_modules/typedoc/bin/typedoc --options typedoc.js",
+    "build": "npm install && node script/barrelize.js && tsc && npm run copy && node node_modules/typedoc/bin/typedoc --options typedoc.js",
```
aburnson: Adding this to every build is going to slow things down during development. I would suggest creating `test:travis` which would do `test:retry` followed by `doc` and then use `test:travis` in [.travis.yml](https://github.com/aws/amazon-chime-sdk-js/blob/master/.travis.yml). If we are going to add it to the travis build, then we need to also confirm (via script) that regenerating docs produced no git diff. This will help catch cases where a PR changes documentation but does not include the generated documentation changes as well as provide extra assurance that checked-in documentation mirrors the inline code documentation.

`doc` also needs to make sure that any files that no longer exist after regenerating get removed. This could be done by wrapping the doc gen like this: `git rm -rf docs && ...typedoc... && git add docs`.

README.md
```diff
@@ -96,10 +96,10 @@ after running `npm run test`.
 To generate JavaScript API reference documentation run:
 
 ```
-npm run doc
```
aburnson: See note below. I think we may want to keep `npm run doc`.

ltrung: Add back npm run doc

package.json
ltrung: Add predoc and postdoc to remove the previous docs folder and  update the doc link to point to master

README.md
ltrung: Add back npm run doc

package.json
```diff
@@ -14,12 +14,15 @@
     "prepublishOnly": "script/publish",
     "lint": "eslint . --ext .ts,.tsx,.js --fix",
     "build": "npm install && node script/barrelize.js && tsc && npm run copy",
+    "predoc": "rm -rf docs",
```
aburnson: Are `predoc` or `postdoc` used anywhere?

script/check-uncommitted-changes
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env ruby
+Dir.chdir(File.expand_path(File.join(File.dirname(__FILE__), '..')))
+uncommited_files = `git status -s`.strip
```
aburnson: Typo: `s/uncommited_files/uncommitted_files/`

script/update-typedoc-link.js
```diff
@@ -0,0 +1,33 @@
...
+  return results;
+};
+
+walk('docs')
+  .filter(fn => fn.endsWith('.html'))
+  .forEach(file => {
+    fs.readFile(file, 'utf8', (err, data) => {
+      if (err) return console.error(err);
+
+      let result = data.replace(/blob\/[0-9a-f]{5,40}\/src/g, 'blob/master/src');
```
aburnson: Let's remove `src` from the regex if possible so that this rewrite will apply to any URL if in the future there are URLs to files outside of `src`.

script/check-uncommitted-changes
ltrung: Fixed

script/update-typedoc-link.js
ltrung: Fixed.

package.json
ltrung: predoc and postdoc are called before and after npm run doc. It is just npm convention.

PR #81, opened 2020-01-17
Updating audio and video test to create the meeting upfront to fix ro…

katayal:
Audio and video canaries were intermittently failing roster check. The behavior right now is that the canaries fail during the first try and pass during retry. Here is the reason:
- The canaries start 2 separate session on saucelabs and try to join a meeting with same title. Both the sessions are concurrently creating a meeting with the same title. Since there are 2 meetings being created with the same title , the participants don't show up on the first try. The retry works because by the second retry the database has information of a meeting [the last one that was created] mapped to the title.

This change will fix the issue as it will create the meeting before the test starts and will make sure there is only one meeting with the same title.

### DIFF
--- a/demos/browser/server.js
+++ b/demos/browser/server.js
@@ -83,6 +83,32 @@ const server = require(protocol).createServer(options, async (request, response)
       response.write(JSON.stringify(attendeeInfo), 'utf8');
       response.end();
       log(JSON.stringify(attendeeInfo, null, 2));
+    } else if (request.method === 'POST' && request.url.startsWith('/meeting?')) {
+      const query = url.parse(request.url, true).query;
+      const title = query.title;
+      if (!meetingCache[title]) {
+        meetingCache[title] = await chime
+          .createMeeting({
+            ClientRequestToken: uuid(),
+            // NotificationsConfiguration: {
+            //   SqsQueueArn: 'Paste your arn here',
+            //   SnsTopicArn: 'Paste your arn here'
+            // }
+          })
+          .promise();
+        attendeeCache[title] = {};
+      }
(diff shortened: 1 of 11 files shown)


demos/serverless/src/handlers.js
```diff
@@ -87,6 +87,45 @@ function getNotificaionsConfig() {
...
+    return;
+  }
+  const title = event.queryStringParameters.title;
+  const region = event.queryStringParameters.region || 'us-east-1';
+  let meetingInfo = await getMeeting(title);
+  if (!meetingInfo) {
+    const request = {
+      ClientRequestToken: uuid(),
+      MediaRegion: region,
+      NotificationsConfiguration: getNotificaionsConfig(),
```
aburnson: Let's fix the typo in getNotificationsConfig()

demos/browser/server.js
```diff
@@ -83,6 +83,32 @@ const server = require(protocol).createServer(options, async (request, response)
       response.write(JSON.stringify(attendeeInfo), 'utf8');
       response.end();
       log(JSON.stringify(attendeeInfo, null, 2));
+    } else if (request.method === 'POST' && request.url.startsWith('/createMeeting?')) {
```
aburnson: We haven't strictly been following REST semantics in the endpoints defined in this server, but `POST /createMeeting` could just be `POST /meeting`.

katayal: Will update it

demos/serverless/src/handlers.js
katayal: Oh yes, will fix it

PR #84, opened 2020-01-17
Update copyright year lint

aburnson:
*Description of changes*

Update copyright year lint so that it automatically determines the range of copyright years on a per file basis by looking at the git history. Also updates copyrights for files that were changed so far in 2020.


script/check-code-style.js
```diff
@@ -179,32 +209,32 @@ components().forEach(component => {
...
 const spdx = '// SPDX-License-Identifier: Apache-2.0';
 
+let allYears = [];
+
 allFiles().forEach(file => {
   if (file.endsWith('.d.ts') || file.includes('.DS_Store')) {
     return;
   }
+  let yearsForFileInGitHistory = [];
+  const stdout = execute(`git log --pretty=format:'%ad' --date=short '${file}'`);
```
simmkyu: Using git will fail if a builder downloads the repo from GitHub.

aburnson: This script is now only run for `build:release`:
https://github.com/aws/amazon-chime-sdk-js/blob/master/package.json#L18

The intention of `build:release` is that you are preparing a commit to be checked in, in which case there is an assumption that you have a Git repo.

If you are only running demos, you will only need `npm run build` which does not call this script.

PR #89, opened 2020-01-21
Update build and publish script

ltrung:
*Description of changes*
- Rename postbuild script and move it to travis 
- Update the publish script to use build:release since npm run test no longer runs build.


package.json
```diff
@@ -16,7 +16,6 @@
     "build": "npm install && node script/barrelize.js && tsc && npm run copy",
     "prebuild:release": "script/prebuild",
     "build:release": "npm run build && npm run lint && npm run check && npm run doc && npm run test:retry",
-    "postbuild:release": "script/postbuild",
```
aburnson: The postbuild script is useful to run locally as it tells you whether your build is likely to pass CI. What is the motivation behind removing it?

PR #102, opened 2020-01-27
Writing the latest canary completing time to a file

katayal:
The canary is getting stuck after completing all the tests. With this change the canary will emit a timestamp which would indicate the last time it ran and if it greater than some time, the canary can restart.


aburnson:
Rebased and updated changelog

PR #114, opened 2020-01-28
Enabling canary for safari and fixing audio checks

katayal:
1. Enabling integration tests for safari 12.
2. Fixing the audio check.


aburnson:
Rebased

PR #121, opened 2020-02-07
Show SDK version in Demo meeting app

ltrung:
*Description of changes*
Show amazon-chime-sdk-js version in the demo meeting app

### DIFF
--- a/demos/browser/app/meeting/meeting.ts
+++ b/demos/browser/app/meeting/meeting.ts
@@ -25,6 +25,7 @@ import {
   ScreenMessageDetail,
   ScreenShareFacadeObserver,
   TimeoutScheduler,
+  Versioning,
   VideoTileState,
 } from '../../../../src/index';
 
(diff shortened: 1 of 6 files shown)


demos/browser/app/meeting/meeting.ts
```diff
@@ -25,8 +25,9 @@ import {
   ScreenMessageDetail,
   ScreenShareFacadeObserver,
   TimeoutScheduler,
+  Versioning,
   VideoTileState,
-} from '../../../../src/index';
+} from 'amazon-chime-sdk-js';
```
aburnson: Why is this changing? I think the original idea was that for local development this should point to the index file directly in the source folder. However, the demo package.json does use a `file:` reference, so maybe that is equivalent?

ltrung: Nope it should not change. I tested with the npm version and forgot to revert it.

PR #123, opened 2020-02-10
Add mechanism to capture browser logs on CloudWatch

RidipDe:
Add mechanism to capture browser logs on CloudWatch

*Issue #:* 

*Description of changes*
Added cloudwatch implementation to capture the logs generated in the browser. Every 5 seconds batch size of 50 logs are written to the CloudWatch
LogGroup Name: BrowserLogs [has to be created manually from the console]
LogStream is meeting ID


demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  constructor(name: string, level = LogLevel.WARN, meetingID: string, attendeeID: string) {
+    this.name = name;
+    this.level = level;
+    this.meetingID = meetingID;
+    this.attendeeID = attendeeID;
+    this.logSeqNo = 0;
+  }
+
+  batchSizes = 50;
```
katayal: move these two fields to the top

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,135 @@
...
+  lock = false;
+  async publishToCloudWatch(base_url: string) {
+    var _this = this;
+
+    setInterval(async function () {
+
+      if (_this.lock == true)
+        return
+      _this.lock = true;
+      //if (_this.logCapture.length > 0 )
```
katayal: remove this commented line

demos/browser/app/meeting/CloudWatchLogger.ts
katayal: maybe make the batchSize a const

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,135 @@
...
+          method: 'POST',
+          body: JSON.stringify({
+            logs: batch
+          })
+        }
+      );
+
+      const json = await response.json();
+      if (json.error) {
+        throw new Error(`Server error: ${json.error}`);
```
katayal: Wouldn't the app crash if there is an error. Is that what you want? If not shouldn't we wait for next invocation.

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    private log(type: LogLevel, msg: string): void {
+        if (type < this.level) {
+            return;
+        }
+        const date = new Date();
+        const timestamp = date.toISOString();
+        const logMessage = `${timestamp} [${LogLevel[type]}] ${this.name} - ${msg}`;
+        //const logMessageCW = `${timestamp} MeetingId=${this.meetingID} AttendeeId=${this.attendeeID} [${LogLevel[type]}] ${this.name} - ${msg}`;
```
katayal: remove all commented code

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+  var sequenceToken = "";
+  const logGroupName = "BrowserLogs";
+  var response = {
+    "statusCode": 200,
+    "headers": {},
+    "body": '',
+    "isBase64Encoded": false
+  };
+  try {
+    if (typeof JSON.parse(event.body) === 'undefined' || typeof JSON.parse(event.body).logs === 'undefined') {
```
katayal: Maybe parse the body once before the if statement

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,134 @@
...
+                break;
+            case LogLevel.INFO:
+                console.info(logMessage);
+                break;
+        }
+        var logJSON = {
+           "logSeqNo": this.logSeqNo,
+           "logMessage" : msg,
+           "timestamp" : date.getTime(),
+           "meetingID" : this.meetingID,
```
katayal: I think you can send meetingID, attendeeID and appName in the POST request as they will be the same for every log message.

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+            response.body = "Empty";
+      return response;
+    } else {
```
katayal: nit: I don't think you need an else here. You are already returning from the if statement.

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+
+        var msg = `${logSeqNo} AttendeeID: ${attendeeID} [${logLevelType}] ${logMessage}`;
+
+        var par = {
+          "logGroupName": logGroupName,
+          "logStreamNamePrefix": meetingID
+        }
+        if (sequenceToken != "") {
+          par = {
+            "logGroupName": logGroupName,
```
katayal: ```par["nextToken"] = sequenceToken``` should be sufficient here

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+            "logStreamNamePrefix": meetingID,
+            "nextToken": sequenceToken
+          }
+        }
+
+        var logStreams = await cloudwatchlogs.describeLogStreams(par).promise();
```
katayal: What if this request fails

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+
+        //if meetingID does not exist as logStream
+        if (logStreams.logStreams.length == 0){
+          var params = {
+            "logGroupName": logGroupName,
+            "logStreamName": meetingID
+          };
+          var res = await cloudwatchlogs.createLogStream(params).promise();
```
katayal: what would happen if this request fails

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+        }
+        else if (logStreams.logStreams.length > 0){
+            sequenceToken = logStreams.logStreams[0].uploadSequenceToken;
```
katayal: If there are multiple log streams with the your meetingId as prefix, shouldn't you find the log stream whose ```name === meetingID``` and use the uploadSequenceToken from that stream.

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+
+        var par = {
+          "logGroupName": logGroupName,
+          "logStreamNamePrefix": meetingID
+        }
+        if (sequenceToken != "") {
+          par = {
+            "logGroupName": logGroupName,
+            "logStreamNamePrefix": meetingID,
+            "nextToken": sequenceToken
```
katayal: The sequenceToken you have defined in this handler is for `putLogEvents` api call. You cannot use it for `describeLogStreams` api.

katayal: The `describeLogStreams` api will return its own sequence token which you should use incase there are multiple logstreams with same prefix

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+          }
+        }
+
+        var logStreams = await cloudwatchlogs.describeLogStreams(par).promise();
+
+        //if meetingID does not exist as logStream
+        if (logStreams.logStreams.length == 0){
```
katayal: [**Unlikely scenario but should be handled**]: What would happen if there are log streams with meetinID as prefix but there is not logstream with name as meetingID (not just prefix). You should check for that as well and create a logstream.

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+          };
+          var res = await cloudwatchlogs.createLogStream(params).promise();
+        }
+        else if (logStreams.logStreams.length > 0){
+            sequenceToken = logStreams.logStreams[0].uploadSequenceToken;
+        }
+
+        //if meetingID exists as logStream
+        var params = {
+          "logEvents": [{
```
katayal: Why are you not sending all the logs in single request. Check the documentation for the putLogEvents api, it says: `Uploads a batch of log events to the specified log stream.` So you can send all the logs. Also, this loop should check the sequenceToken and should not loop over log messages.

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+            "message": msg,
+            "timestamp": timestamp
+          }, ],
+          "logGroupName": logGroupName,
+          "logStreamName": meetingID
+        };
+
+        if (sequenceToken != "") {
```
katayal: `params["sequenceToken"]= sequenceToken` should be sufficient here

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+            "logEvents": [{
+              "message": msg,
+              "timestamp": timestamp
+            }, ],
+            "logGroupName": logGroupName,
+            "logStreamName": meetingID,
+            "sequenceToken": sequenceToken
+          };
+        }
+        var res = await cloudwatchlogs.putLogEvents(params).promise();
```
katayal: what if the request fails here

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+        sequenceToken = res.sequenceToken;
+      }
+    }
+  } catch (ex) {
+    response.statusCode = -1;
```
katayal: maybe log the exception so that you can check it in the lambda cloudwatch logs

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+    response.body = JSON.stringify("\nFail " + ex);
+  }
+  response.body = JSON.stringify(logs_arr);
```
katayal: Why are you overriding the response body here? Why are you sending back the logs messages from the request ?

demos/serverless/template.yaml
```diff
@@ -44,6 +44,19 @@ Resources:
...
+      PolicyName: CloudWatchAccess
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Action:
+              - 'logs:*'
+            Resource: '*'
+      Roles:
+        - Ref: ChimeSdkBrowserLogsLambdaRole
```
katayal: Where is this role defined ?

demos/serverless/template.yaml
```diff
@@ -164,7 +177,7 @@ Resources:
             Path: /attendee
             Method: GET
   ChimeSQSQueueLambda:
-    Type: AWS::Serverless::Function
+    Type: 'AWS::Serverless::Function'
```
katayal: Why did you put this in quotes

demos/serverless/template.yaml
```diff
@@ -175,7 +188,7 @@ Resources:
             Queue: !GetAtt MeetingNotificationsQueue.Arn
             BatchSize: 10
   ChimeEventBridgeLambda:
-    Type: AWS::Serverless::Function
+    Type: 'AWS::Serverless::Function'
```
katayal: same

demos/serverless/template.yaml
```diff
@@ -189,8 +202,19 @@ Resources:
...
+      CodeUri: src/
+      Events:
+        Api1:
+          Type: Api
+          Properties:
+            Path: /logs
+            Method: POST
   ChimeNotificationsQueuePolicy:
-    Type: AWS::SQS::QueuePolicy
+    Type: 'AWS::SQS::QueuePolicy'
```
katayal: same

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "amazon-chime-sdk-js",
-  "version": "1.1.2",
+  "version": "1.1.3",
```
katayal: are you suppose to update this version ?

src/versioning/Versioning.ts
```diff
@@ -11,7 +11,7 @@ export default class Versioning {
    * Return string representation of SDK version
    */
   static get sdkVersion(): string {
-    return '1.1.2';
+    return '1.1.3';
```
katayal: same as above?

demos/browser/app/meeting/CloudWatchLogger.ts
RidipDe: I have un-commented this line now. It will not send requests to the cloudwatch when the logCapture[] is empty

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,105 @@ exports.end = async(event, context, callback) => {
...
+            "sequenceToken": sequenceToken
+          };
+        }
+        var res = await cloudwatchlogs.putLogEvents(params).promise();
+        sequenceToken = res.sequenceToken;
+      }
+    }
+  } catch (ex) {
+    response.statusCode = -1;
+    response.body = JSON.stringify("\nFail " + ex);
```
katayal: Don't need to stringify a string

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,135 @@
...
+           "attendeeID" : this.attendeeID,
+           "logLevelType" : LogLevel[type],
+           "appName" : this.name
+        };
+        this.logCapture.push(JSON.stringify(logJSON));
```
katayal: I don't think you need to stringify here. The stringify what your are sending the POST request will convert the entire json to string.

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,131 @@
...
+      const response = await fetch(
+        `${base_url}logs`, {
+          method: 'POST',
+          body: bodyString
+        }
+      );
+      if (response.status == 200){
+        // delete elements upto current_size(logCapture) from the array logCapture
+        _this.logCapture = _this.logCapture.slice(batch.length);
+        _this.lock = false;
```
katayal: I think you need to release the lock even when the response status is not 200. 
You should also handle non 200 response codes.

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,58 @@ exports.end = async(event, context, callback) => {
...
+    putLogEventsInput["logEvents"] = logEvents;
+    var res = await cloudWatchClient.putLogEvents(putLogEventsInput).promise();
+    if (res.nextSequenceToken){
+      response.body = "Log Write Positive";
+      return response;
+    }
+  } catch (ex) {
+    response.statusCode = 500;
+    response.body = ex;
+  }
```
katayal: don't you have to return a response here ?

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,58 @@ exports.end = async(event, context, callback) => {
...
+      logEvents.push({
+        "message": message,
+        "timestamp": log.timestamp
+      });
+    }
+    putLogEventsInput["logEvents"] = logEvents;
+    var res = await cloudWatchClient.putLogEvents(putLogEventsInput).promise();
+    if (res.nextSequenceToken){
+      response.body = "Log Write Positive";
+      return response;
```
katayal: maybe move this return to the end of the function

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,58 @@ exports.end = async(event, context, callback) => {
...
+      var log = JSON.parse(body.logs[i]);
+      var message = `${log.logSeqNo} AttendeeID: ${body.attendeeID} [${log.logLevelType}] ${log.logMessage} ${log.timestamp}`;
+      logEvents.push({
+        "message": message,
+        "timestamp": log.timestamp
+      });
+    }
+    putLogEventsInput["logEvents"] = logEvents;
+    var res = await cloudWatchClient.putLogEvents(putLogEventsInput).promise();
+    if (res.nextSequenceToken){
```
katayal: Not sure why you need this if statement

CHANGELOG.md
```diff
@@ -18,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fix publish script to use npm version
 - Add stage to saucelabs session name for integration tests
+- Optimized call requests to the cloudwatch
```
katayal: Isn't this a new thing you are adding to the demo app. So you are not fixing anything here right.

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,131 @@
...
+
+import Logger from '../../../../src/logger/Logger';
+import LogLevel from '../../../../src/logger/LogLevel';
+
+/**
+ * CloudWatchLogger writes logs with console
+ *
+ * ```typescript
+ *   // working with the CloudWatchLogger
+ *   const logger = new CloudWatchLogger('demo'); //default level is LogLevel.WARN
```
katayal: please fix this comment.

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,131 @@
...
+      );
+      if (response.status == 200){
+        // delete elements upto current_size(logCapture) from the array logCapture
+        _this.logCapture = _this.logCapture.slice(batch.length);
+      }
+      _this.lock = false;
+    }, 5000);
+  }
+
+    info(msg: string): void {
```
katayal: I wonder if you can just extend the ConsoleLogger and inherit most of the functions and override the once you want

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,86 @@
...
+import LogLevel from '../../../../src/logger/LogLevel';
+import {ConsoleLogger} from '../../../../src/index';
+
+export default class CloudWatchLogger extends ConsoleLogger {
+  name: string;
+  level: LogLevel;
+  logCapture: string[] = [];
+  meetingID: string;
+  attendeeID: string;
+  logSeqNo: any;
```
simmkyu: Please do not use abbreviations. e.g., => ```logSequenceNumber```.
(Sorry, I remember I've seen this convention somewhere, but I can't find that doc.)

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  async publishToCloudWatch(base_url: string) {
+    var _this = this;
+
+    setInterval(async function () {
+
+      if (_this.lock == true || _this.logCapture.length == 0 )
+        return
+      _this.lock = true;
+      var batch = _this.logCapture.slice(0, _this.batchSizes);
```
simmkyu: Use ```const``` since you do not assign a new value to ```batch```.

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,86 @@
...
+  name: string;
+  level: LogLevel;
+  logCapture: string[] = [];
+  meetingID: string;
+  attendeeID: string;
+  logSeqNo: any;
+  lock = false;
+  batchSizes = 50;
+
+  constructor(name: string, level = LogLevel.WARN, meetingID: string, attendeeID: string) {
```
simmkyu: - ```meetingID``` -> ```meetingId```
- ```attendeeID``` -> ```attendeeId```

Please review ```demos/browser/app/meeting/meeting.ts``` and Chime SDK files as references.

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,86 @@
...
+    super(name, level);
+    this.name = name;
+    this.level = level;
+    this.meetingID = meetingID;
+    this.attendeeID = attendeeID;
+    this.logSeqNo = 0;
+  }
+
+  async publishToCloudWatch(base_url: string) {
+    var _this = this;
```
simmkyu: I understand that we have used this pattern in the demo app, but a better pattern is to use an arrow function.

1. Remove this line.
2. Use an arrow function and replace ```_this``` with ```this```.
  ```
  setInterval(async () => {
     // In an arrow function, the ```this``` context is the same as the outer scope.
     if (this.lock == true || this.logCapture.length == 0 )
     ...
  }, 5000);
  ```

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export default class CloudWatchLogger extends ConsoleLogger {
+  name: string;
+  level: LogLevel;
+  logCapture: string[] = [];
+  meetingID: string;
+  attendeeID: string;
+  logSeqNo: any;
+  lock = false;
+  batchSizes = 50;
```
simmkyu: Please make some properties ```private```. Also, you can make this static constant like 
https://github.com/aws/amazon-chime-sdk-js/blob/a0cfc9611c848b69b2f1839dd91f867dc4d32934/demos/browser/app/meeting/meeting.ts#L33

```
private static BATCH_SIZE = 16;
```

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,86 @@
...
+  constructor(name: string, level = LogLevel.WARN, meetingID: string, attendeeID: string) {
+    super(name, level);
+    this.name = name;
+    this.level = level;
+    this.meetingID = meetingID;
+    this.attendeeID = attendeeID;
+    this.logSeqNo = 0;
+  }
+
+  async publishToCloudWatch(base_url: string) {
```
simmkyu: ```camelCase``` - Probably, ESLinter will fix this too.

simmkyu:
I found the number of syntax issues in committed file. e.g., a mix of single and double quotes, no blank line at the end, no consistent lines, and so on.

The root ```package.json``` has a ```lint``` script that reports syntax issues and automatically formats your files.
https://github.com/aws/amazon-chime-sdk-js/blob/a0cfc9611c848b69b2f1839dd91f867dc4d32934/package.json#L15

Please run this ESLint script on your committed files.

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,83 @@
...
+    super(name, level);
+    this.name = name;
+    this.level = level;
+    this.meetingId = meetingId;
+    this.attendeeId = attendeeId;
+    this.logSequenceNumber = 0;
+  }
+
+  async publishToCloudWatch(base_url: string) {
+    setInterval(async() => {
```
aburnson: Use scheduler/IntervalScheduler from the SDK

demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,83 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import LogLevel from '../../../../src/logger/LogLevel';
+import {ConsoleLogger} from '../../../../src/index';
+
+export default class CloudWatchLogger extends ConsoleLogger {
```
aburnson: This implementation should go in the SDK (logger/CloudWatchLogger) so it can be reused

PR #124, opened 2020-02-11
Makes DefaultBrowserBehavior implement interface and fix transceiverController async function signature

XHatan:
*Issue #:* 

*Description of changes*

1. Makes `DefaultBrowserBehavior` implements interface.
2. Several methods on RTCRtpTransceiver is async: make corresponding methods on TransceiverController async. Fix unit tests and address a few race conditions.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "amazon-chime-sdk-js",
-  "version": "1.1.2",
+  "version": "1.1.3",
```
simmkyu: @aburnson @vidya-mohan
I had the same question in another PR. Are we patching per PR? (Sorry, if I'm missing something.)

XHatan: Good call out. Need some clarifications.

aburnson: @simmkyu @XHatan 
Yes, we are bumping patch on a per-commit basis. This guarantees that each commit to master receives a unique version number. We adopted this approach in order to track version adoption down to the commit level and to make it easier to determine which version canaries are running (the version number is visible in the lower right corner of the meeting demo now).

PR #126, opened 2020-02-14
Add mechanism to capture browser logs on cloudwatch

RidipDe:
*Issue #:* 

*Description of changes*
Added cloudwatch implementation to capture the logs generated in the browser. Every 5 seconds batch size of 50 logs are written to the CloudWatch
LogGroup Name: BrowserLogs [has to be created manually from the console]
LogStream is meeting ID

### DIFF
--- a/demos/browser/app/meeting/meeting.ts
+++ b/demos/browser/app/meeting/meeting.ts
@@ -22,6 +22,7 @@ import {
   MeetingSessionStatus,
   MeetingSessionStatusCode,
   MeetingSessionVideoAvailability,
+  POSTRequestLogger,
   ScreenMessageDetail,
   ScreenShareFacadeObserver,
   TimeoutScheduler,
@@ -128,7 +129,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver 
 
   // feature flags
   enableWebAudio = false;
-
+  logger: any = null;
   constructor() {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     (global as any).app = this;
@@ -442,22 +443,26 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver 
       }
       new AsyncScheduler().start(async () => {
         (buttonMeetingEnd as HTMLButtonElement).disabled = true;
+        (buttonMeetingLeave as HTMLButtonElement).disabled = true;
         await this.endMeeting();
         this.leave();
-        (buttonMeetingEnd as HTMLButtonElement).disabled = false;
         // @ts-ignore
         window.location = window.location.pathname;
+        (buttonMeetingEnd as HTMLButtonElement).disabled = false;
+        (buttonMeetingLeave as HTMLButtonElement).disabled = false;
       });
     });
 
     const buttonMeetingLeave = document.getElementById('button-meeting-leave');
     buttonMeetingLeave.addEventListener('click', _e => {
       new AsyncScheduler().start(async () => {
+        (buttonMeetingEnd as HTMLButtonElement).disabled = true;
         (buttonMeetingLeave as HTMLButtonElement).disabled = true;
         this.leave();
-        (buttonMeetingLeave as HTMLButtonElement).disabled = false;
         // @ts-ignore
         window.location = window.location.pathname;
+        (buttonMeetingEnd as HTMLButtonElement).disabled = false;
+        (buttonMeetingLeave as HTMLButtonElement).disabled = false;
       });
     });
   }
@@ -548,11 +553,16 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver 
     }
   }
 
+
   async initializeMeetingSession(configuration: MeetingSessionConfiguration): Promise<void> {
-    const logger = new ConsoleLogger('SDK', LogLevel.DEBUG);
-    const deviceController = new DefaultDeviceController(logger);
+    if (!(DemoMeetingApp.BASE_URL.includes('localhost') || DemoMeetingApp.BASE_URL.includes("127.0.0.1"))) {
+      this.logger = new POSTRequestLogger('SDK', configuration, 75, 1750, `${DemoMeetingApp.BASE_URL}logs`, LogLevel.INFO);
+    } else {
+      this.logger = new ConsoleLogger('SDK', LogLevel.DEBUG);
+    }
+    const deviceController = new DefaultDeviceController(this.logger);
     configuration.enableWebAudio = this.enableWebAudio;
-    this.meetingSession = new DefaultMeetingSession(configuration, logger, deviceController);
+    this.meetingSession = new DefaultMeetingSession(configuration, this.logger, deviceController);
     this.audioVideo = this.meetingSession.audioVideo;
 
     this.audioVideo.addDeviceChangeObserver(this);
--- a/demos/serverless/src/handlers.js
+++ b/demos/serverless/src/handlers.js
@@ -210,8 +210,20 @@ exports.end = async(event, context, callback) => {
   callback(null, response);
 };
 
+async function ensureLogStream(cloudWatchClient, logStreamName) {
+  var describeLogStreamsParams = {
+    "logGroupName": logGroupName,
+    "logStreamNamePrefix": logStreamName
+  };
+  var response = await cloudWatchClient.describeLogStreams(describeLogStreamsParams).promise();
+  var foundStream = response.logStreams.find(s => s.logStreamName === logStreamName);
+  if (foundStream) {
+    return foundStream.uploadSequenceToken;
+  }
+  return null;
+}
+
 exports.logs = async (event, context) => {
-  var sequenceToken = "";
   var response = {
     "statusCode": 200,
     "headers": {},
@@ -222,27 +234,20 @@ exports.logs = async (event, context) => {
     var body = JSON.parse(event.body);
     if (!body.logs || !body.meetingId || !body.attendeeId || !body.appName) {
       response.body = "Empty Parameters Received";
+      response.statusCode = 400;
       return response;
     }
     const logStreamName = "ChimeSDKMeeting_" + body.meetingId.toString();
     var cloudWatchClient = new AWS.CloudWatchLogs({
       apiVersion: '2014-03-28'
     });
+    var uploadSequence = await ensureLogStream(cloudWatchClient, logStreamName);
     var putLogEventsInput = {
       "logGroupName": logGroupName,
       "logStreamName": logStreamName
     };
-    var describeLogStreamsParams = {
-      "logGroupName": logGroupName,
-      "logStreamNamePrefix": logStreamName
-    };
-    var describeLogStreamsResponse = await cloudWatchClient.describeLogStreams(describeLogStreamsParams).promise();
-    var logStreamsArr = describeLogStreamsResponse.logStreams;
-    var currentLogStream = logStreamsArr.find(logStreamJSON => logStreamJSON.logStreamName === logStreamName);
-    if (!currentLogStream) {
-      var res = await cloudWatchClient.createLogStream(putLogEventsInput).promise();
-    } else {
-      putLogEventsInput["sequenceToken"] = currentLogStream.uploadSequenceToken;
+    if (uploadSequence) {
+      putLogEventsInput["sequenceToken"] = uploadSequence;
     }
     var logEvents = [];
     for (let i = 0; i < body.logs.length; i++) {
--- a/demos/serverless/src/handlers.js
+++ b/demos/serverless/src/handlers.js
@@ -10,6 +10,7 @@ const meetingsTableName = process.env.MEETINGS_TABLE_NAME;
 const attendeesTableName = process.env.ATTENDEES_TABLE_NAME;
 const sqsQueueArn = process.env.SQS_QUEUE_ARN;
 const provideQueueArn = process.env.USE_EVENT_BRIDGE === 'false';
+const logGroupName = process.env.BROWSER_LOG_GROUP_NAME;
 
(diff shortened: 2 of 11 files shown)


demos/browser/app/meeting/CloudWatchLogger.ts
```diff
@@ -0,0 +1,105 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { Logger, LogLevel } from '../../../../src/index';
+
+export default class CloudWatchLogger implements Logger {
```
simmkyu: Sorry, I did not catch this in your previous PR, so I just force-pushed a fix on top of your commit. Please pull the latest change from the remote ```cwl```.

I don't see a good reason to change the implementation detail ```ConsoleLogger.log()``` to ```protected``` and extended ```ConsoleLogger``` because we did not use the ConsoleLogger's feature at all but duplicate writing console logging code.

In this PR, I updated two:
- Implement ```Logger``` instead of extending ```ConsoleLogger```.
- Revert ```ConsoleLogger``` back. Now, this PR does not contain  ```ConsoleLogger.ts```, ```consolelogger.html```, and ```search.js```.

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,55 @@ exports.end = async(event, context, callback) => {
   }).promise();
   callback(null, response);
 };
+
+exports.logs = async (event, context) => {
```
simmkyu: Don't need to update in this PR. We need to revisit ```handlers.js``` and fix syntax issues. 
- A mix of ```"``` and ```'```.
- Remove unused variables e.g., ```res```
- ```const```/```let``` over ```var```

demos/serverless/template.yaml
```diff
@@ -164,7 +177,7 @@ Resources:
             Path: /attendee
             Method: GET
   ChimeSQSQueueLambda:
-    Type: AWS::Serverless::Function
+    Type: 'AWS::Serverless::Function'
```
asnyatkov: Here and everywhere: remove single quotes.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "amazon-chime-sdk-js",
-  "version": "1.1.3",
+  "version": "1.1.4",
```
asnyatkov: This seems wrong. Why are we bumping Sdk version?

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,55 @@ exports.end = async(event, context, callback) => {
...
+  var sequenceToken = "";
+  const logGroupName = "BrowserLogs";
```
asnyatkov: If we want to make cloud watch logging the part of official demo, we need to move log group creation to the CF template, and pass its name as an argument to the lambda

demos/browser/app/meeting/meeting.ts
```diff
@@ -549,7 +550,8 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
   }
 
   async initializeMeetingSession(configuration: MeetingSessionConfiguration): Promise<void> {
-    const logger = new ConsoleLogger('SDK', LogLevel.DEBUG);
+    const logger = new CloudWatchLogger('SDK', LogLevel.INFO, configuration.meetingId, configuration.credentials.attendeeId);
```
asnyatkov: Logging to CW has to be optional and configurable.

asnyatkov: (And probably off by default)

demos/serverless/src/handlers.js
```diff
@@ -208,3 +208,55 @@ exports.end = async(event, context, callback) => {
...
+    if (!body.logs || !body.meetingId || !body.attendeeId || !body.appName) {
+      response.body = "Empty Parameters Received";
+      return response;
+    }
+    var cloudWatchClient = new AWS.CloudWatchLogs({
+      apiVersion: '2014-03-28'
+    });
+    var putLogEventsInput = {
+      "logGroupName": logGroupName,
+      "logStreamName": body.meetingId.toString()
```
asnyatkov: This should be 'ChimeSDKMeeting_' + meetingId

demos/serverless/template.yaml
```diff
@@ -44,6 +44,19 @@ Resources:
...
+  CloudWatchAccessPolicy:
+    Type: AWS::IAM::Policy
+    Properties:
+      PolicyName: CloudWatchAccess
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Action:
+              - 'logs:*'
```
asnyatkov: Follow 'minimal required privilege' principal: you only need couple of operations here.

demos/browser/app/meeting/meeting.ts
```diff
@@ -9,7 +9,7 @@ import {
   AudioVideoFacade,
   AudioVideoObserver,
   ClientMetricReport,
-  ConsoleLogger,
```
aburnson: Need to log ConsoleLogger by default and use POSTRequestLogger if a POST URL is supplied. Think about the case of someone checking out the repo for the first time and running the the meeting demo locally.

demos/serverless/template.yaml
```diff
@@ -19,6 +19,7 @@ Globals:
         MEETINGS_TABLE_NAME: !Ref Meetings
         ATTENDEES_TABLE_NAME: !Ref Attendees
         SQS_QUEUE_ARN: !GetAtt MeetingNotificationsQueue.Arn
+        BROWSER_LOG_GROUP_NAME: BrowserLogs
```
asnyatkov: You need to create CloudWatch log group as part of the stack

`
ChimeSDKBrowserLogs:
  Type: AWS::Logs::LogGroup
  Properties: 
    LogGroupName: BrowserLogs
`
and then refer it like this:

`
BROWSER_LOG_GROUP_NAME: !Ref ChimeSDKBrowserLogs
`

demos/serverless/template.yaml
```diff
@@ -44,8 +45,24 @@ Resources:
...
+  CloudWatchAccessPolicy:
+    Type: AWS::IAM::Policy
+    Properties:
+      PolicyName: CloudWatchAccess
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Action:
+              - 'logs:CreateLogGroup'
```
asnyatkov: Remove, the lambda won't create any log groups.

demos/serverless/src/handlers.js
```diff
@@ -208,3 +209,55 @@ exports.end = async(event, context, callback) => {
...
+    });
+    var putLogEventsInput = {
+      "logGroupName": logGroupName,
+      "logStreamName": logStreamName
+    };
+    var describeLogStreamsParams = {
+      "logGroupName": logGroupName,
+      "logStreamNamePrefix": logStreamName
+    };
+    var describeLogStreamsResponse = await cloudWatchClient.describeLogStreams(describeLogStreamsParams).promise();
```
asnyatkov: There's too much business logic in this function. Extract a helper - `ensureCloudWatchLogStream` which creates the stream if it doesn't exists.

src/meetingsession/MeetingSessionConfiguration.ts
```diff
@@ -26,6 +26,10 @@ export default class MeetingSessionConfiguration {
    * The URLs the session uses to reach the meeting service.
    */
   urls: MeetingSessionURLs | null = null;
+  /**
+   * Maximum amount of time in milliseconds to allow for connecting.
+   */
+  sendCloudWatchLogsIntervalMs: number = 10000;
```
asnyatkov: This should not be part of our public API.

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,153 @@
...
+  ) {
+    this.intervalScheduler = new IntervalScheduler(this.intervalMs);
+    var _this = this;
+    window.addEventListener('unload', function(event) {
+        _this.intervalScheduler.stop();
+        if(!navigator.sendBeacon){
+            return
+        }
+        // Send the beacon
+        while (_this.logCapture.length > 0) {
```
aburnson: I think Beacon is only intended to be called once. Why are we splitting up the logs with multiple requests? If the limit is known to be 64KB, then can we add as many logs to the body as will fit in 64KB and make one request?

RidipDe: i thought if we could break the remaining logs array into smaller chunks, we could avoid the 64kb limitation. So we will send multiple beacon calls with smaller log size.

aburnson: I think it would be better to append log lines keeping track of overall size and stop when some limit is reached. Otherwise this could potentially cause tens or hundreds of network requests which could end up being throttled by the server (which will defeat the purpose since they cannot be retried and will drop on the floor).

src/logger/ConsoleLogger.ts
```diff
@@ -59,7 +59,7 @@ export default class ConsoleLogger implements Logger {
     return this.level;
   }
 
-  private log(type: LogLevel, msg: string): void {
+  protected log(type: LogLevel, msg: string): void {
```
simmkyu: Keep this ```private```.

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,143 @@
...
+  constructor(
+    private name: string,
+    private meetingId: string,
+    private attendeeId: string,
+    private intervalMs: number,
+    private baseURL: string,
+    private level = LogLevel.WARN,
+  ) {
+    this.intervalScheduler = new IntervalScheduler(this.intervalMs);
+    var _this = this;
```
simmkyu: We prefer to use an arrow function ``() => {}``` over this approach. (See [arrow function doc.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions))
```
window.addEventListener('unload', () => {
  // "this" is the same as "this" outside an arrow function
});
```

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,143 @@
...
+    window.addEventListener('unload', function(event) {
+      _this.intervalScheduler.stop();
+      if(!navigator.sendBeacon){
+        return
+      }
+      let body = JSON.stringify({
```
simmkyu: Please go over all ```let``` in this file and replace them with ```const``` if you do not assign a new value to it.

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,143 @@
...
+      JSON.stringify({
+        logSequenceNumber: this.logSequenceNumber,
+        logMessage: msg,
+        timestamp: date.getTime(),
+        logLevelType: LogLevel[type],
+      })
+    );
+    this.logSequenceNumber += 1;
+  }
+}
```
simmkyu: Add a blank line at the end. (Need to enable the rule in ESLinter.)

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,143 @@
...
+      let body = JSON.stringify({
+        meetingId: _this.meetingId,
+        attendeeId: _this.attendeeId,
+        appName: _this.name,
+        logs: _this.logCapture
+      });
+      navigator.sendBeacon(_this.baseURL, body);
+    });
+  }
+  debug(debugFunction: () => string): void {
```
simmkyu: Add a blank line between methods.

simmkyu: Add a blank line between methods.

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,143 @@
...
+    });
+    try {
+      const response = await fetch(this.baseURL, {
+        method: 'POST',
+        body,
+      });
+      if (response.status === 200) {
+        this.logCapture = this.logCapture.slice(batch.length);
+      }
+    } catch (error) {
```
simmkyu: - Commented in the previous revision. ```error``` will be an Error object (JS object). Concatenating an object to a string will result in the following output:
```[POSTRequestLogger] [object Object] ```

- Try this:
```
this.warn(error && error.message ? error.message : `...custom message here...`);
```

File: package.json (hunk 4 above)
Comment (aburnson) [reply package.json]: @asnyatkov the SDK bumps patch version with every commit now as part of `npm run build:release`.

File: src/logger/POSTRequestLogger.ts [hunk 21]
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  getLogLevel(): LogLevel {
+    return this.level;
+  }
+
+  async publishLogsCaller(flushAll: boolean): Promise<void> {
+    if (flushAll === true && this.logCapture.length > 0) {
+      this.intervalScheduler.stop();
+      if (this.logCapture.length <= POSTRequestLogger.BATCH_SIZES){
```
simmkyu: Do not need this ```if``` statement because ```batchSize``` (line 71) will be 1 so it calls ```publishLogs``` depending on the capture length.

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,143 @@
...
+  setLogLevel(level: LogLevel): void {
+    this.level = level;
+  }
+
+  getLogLevel(): LogLevel {
+    return this.level;
+  }
+
+  async publishLogsCaller(flushAll: boolean): Promise<void> {
+    if (flushAll === true && this.logCapture.length > 0) {
```
simmkyu: This ```if``` block for flushAll is no more required since the beacon API sends all logs?

aburnson:
Closing this PR for now since it depends on #137

PR #132, opened 2020-02-20
Add ContentShareController APIs

ltrung:
*Description of changes*
Add ContentShareController APIs


src/contentsharecontroller/ContentShareMediaStreamBroker.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  async acquireScreenCaptureDisplayInputStream(sourceId?: string): Promise<MediaStream> {
+    return this.acquireDisplayInputStream(this.screenCaptureDisplayMediaConstraints(sourceId));
+  }
+
+  private screenCaptureDisplayMediaConstraints(sourceId?: string): MediaStreamConstraints {
+    return {
+      audio: false,
+      video: {
+        ...(!sourceId && {
```
simmkyu: optional: how about using the ternary operator, since there will be two shapes from one ```sourceId``` condition?

```
...(sourceId ? {
  mandatory: {
    chromeMediaSource: 'desktop',
    chromeMediaSourceId: sourceId,
    maxFrameRate: 3,
  }
} : {
  frameRate: {
    max: 3,
  }
}),
```

src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -0,0 +1,59 @@
...
+import MeetingSessionCredentials from '../meetingsession/MeetingSessionCredentials';
+import ContentShareConstants from './ContentShareConstants';
+import ContentShareController from './ContentShareController';
+import ContentShareMediaStreamBroker from './ContentShareMediaStreamBroker';
+
+export default class DefaultContentShareController implements ContentShareController {
+  static createContentShareMeetingSessionConfigure(
+    configuration: MeetingSessionConfiguration
+  ): MeetingSessionConfiguration {
+    let contentShareConfiguration = new MeetingSessionConfiguration();
```
simmkyu: nit: use ```const```

src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+    this.mediaStreamBroker.mediaStream = stream;
+    this.audioVideo.start();
+    if (this.mediaStreamBroker.mediaStream.getVideoTracks().length > 0) {
+      this.audioVideo.videoTileController.startLocalVideoTile();
+    }
+  }
+
+  async startContentShareFromScreenCapture(sourceId?: string): Promise<void> {
+    let mediaStream = await this.mediaStreamBroker.acquireScreenCaptureDisplayInputStream(sourceId);
```
simmkyu: Use ```const```.

src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.startContentShare(mediaStream);
```
simmkyu: Use ```await``` if you want to finish ```startContentShareFromScreenCapture()``` after executing ```startContentShare()```.

src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.mediaStreamBroker.toggleMediaStream(false);
+  }
+
+  unpauseContentShare(): void {
+    this.mediaStreamBroker.toggleMediaStream(true);
+  }
+
+  stopContentShare(): void {
+    this.audioVideo.stop();
+    this.mediaStreamBroker.cleanup();
```
simmkyu: ```this.audioVideo.stop()``` is a non-async call, but it performs the async "actionDisconnect" operation. Could you confirm that calling ```cleanup``` while running "actionDisconnect" has no conflict?

src/contentsharecontroller/ContentShareMediaStreamBroker.ts
```diff
@@ -0,0 +1,99 @@
...
+import DefaultDeviceController from '../devicecontroller/DefaultDeviceController';
+import Logger from '../logger/Logger';
+import MediaStreamBroker from '../mediastreambroker/MediaStreamBroker';
+
+export default class ContentShareMediaStreamBroker implements MediaStreamBroker {
+  private _mediaStream: MediaStream;
+
+  constructor(private logger: Logger) {}
+
+  get mediaStream(): MediaStream {
```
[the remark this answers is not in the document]
simmkyu: Ignore this comment. I noticed that we've been using a getter in the audio video controller.

src/contentsharecontroller/DefaultContentShareController.ts
aburnson: I think this is fine. There is a question as to who should be responsible for stopping the media tracks, the API consumer or the API implementation. I think what we may want to do later is provide an optional boolean flag that allows for running cleanup or not, but we can leave this as is for now since it's reasonable to clean up a MediaStream that is no longer being shared.

src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  constructor(
+    private mediaStreamBroker: ContentShareMediaStreamBroker,
+    private audioVideo: AudioVideoController
+  ) {}
+
+  async startContentShare(stream: MediaStream): Promise<void> {
+    if (!stream) {
+      return;
+    }
```
aburnson: This shouldn't be necessary. We should assume that a stream is passed in. I don't think there's a use case for passing in a null stream?

src/contentsharecontroller/ContentShareMediaStreamBroker.ts
aburnson: This code was copied from here: https://github.com/aws/amazon-chime-sdk-js/blob/904bb29e4da9ab7ef7ea4a64298f3df25c649d76/src/screensharingsession/ScreenSharingSessionContainer.ts#L56

Let's leave it as is since we know it works and refactor later.

src/contentsharecontroller/ContentShareController.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export default interface ContentShareController {
+  /**
+   * Start content sharing
```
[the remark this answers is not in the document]
aburnson: We should just remove`@param sourceId` from below unless you are going to explain the parameter. TypeDoc doesn't need it in order to generate the documentation.

src/audiovideofacade/DefaultAudioVideoFacade.ts
```diff
@@ -356,9 +358,38 @@ export default class DefaultAudioVideoFacade implements AudioVideoFacade {
...
+    this.contentShareController.stopContentShare();
+    this.trace('stopContentShare');
+  }
+
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private trace(name: string, input?: any, output?: any): void {
-    let s = `API/DefaultAudioVideoFacade/${name}`;
+    const meetingId = this.audioVideoController.configuration.meetingId;
+    const attendeeId = this.audioVideoController.configuration.credentials.attendeeId;
+    let s = `API/DefaultAudioVideoFacade/${meetingId}/${attendeeId}/${name}`;
```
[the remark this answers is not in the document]
aburnson: The `API/` logs are not logged by the task name wrapper though, so we need to add the IDs here too. The goal is to make sure that all the main lifecycle logs have the meeting and attendee IDs.

src/contentsharecontroller/DefaultContentShareController.ts
ltrung: Without null check we can run into null exception here. I think I ran into this in a unit test for another component. We also do null check in other API in content share.
```
if (this.mediaStreamBroker.mediaStream.getVideoTracks().length > 0) {
 this.audioVideo.videoTileController.startLocalVideoTile();
}
```

test/contentsharecontroller/ContentShareMediaStreamBroker.test.ts
```diff
@@ -0,0 +1,198 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import * as chai from 'chai';
+import * as sinon from 'sinon';
+import { SinonStub } from 'sinon';
```
simmkyu: We do not recommend using the Sinon stub. Since you used it only once, please replace it with a test class extending DefaultDeviceController.

https://github.com/aws/amazon-chime-sdk-js/blob/f8c7aa9bc1c6d42d510f72ca4f74bfb0f7aacccd/script/check-code-style.js#L113

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -328,6 +307,10 @@ export default class DefaultAudioVideoController implements AudioVideoController
     this._reconnectController.reset();
   }
 
+  private wrapTaskName(taskName: string): string {
```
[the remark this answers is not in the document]
ltrung: Fixed

src/contentsharecontroller/ContentShareController.ts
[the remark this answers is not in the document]
ltrung: Removed

src/contentsharecontroller/DefaultContentShareController.ts
ltrung: Fixed

src/contentsharecontroller/DefaultContentShareController.ts
ltrung: Fixed

PR #137, opened 2020-02-25
MeetingSessionPOSTLogger source and tests

RidipDe:
*Issue #:* 

*Description of changes*
MeetingSessionPOSTLogger source and tests


src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,127 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { Logger, LogLevel } from '../index';
```
aburnson: These should be imported from `../logger/Logger` and `../logger/LogLevel`.

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,127 @@
...
+  constructor(
+    private name: string,
+    private meetingId: string,
+    private attendeeId: string,
+    private intervalMs: number,
+    private baseURL: string,
+    private level = LogLevel.WARN,
+  ) {
+    this.intervalScheduler = new IntervalScheduler(this.intervalMs);
+    this.publishLogs(POSTRequestLogger.BATCH_SIZES);
```
aburnson: Suggest renaming to `startLogPublishScheduler`. `publishLogs` sounds like a one-time event.

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,127 @@
...
+    window.addEventListener('unload', () => {
+      this.intervalScheduler.stop();
+      if(!navigator.sendBeacon){
```
aburnson: This check should be unnecessary. All browsers that the SDK supports will have sendBeacon.

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,127 @@
...
+        return
+      }
+      const body = JSON.stringify({
```
aburnson: Move this to a function to be more DRY.

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (type < this.level) {
+      return;
+    }
+    const date = new Date();
+    const timestamp = date.toISOString();
+    const logMessage = `${this.logSequenceNumber} ${timestamp} [${LogLevel[type]}] ${this.name} - ${msg}`;
+
+    switch (type) {
+      case LogLevel.ERROR:
+        console.trace(logMessage);
```
aburnson: POSTRequestLogger should not log out to the console. "ConsoleLogger logs to console" "POSTRequestLogger logs via POST request".

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,127 @@
...
+        break;
+      case LogLevel.INFO:
+        console.info(logMessage);
+        break;
+    }
+    this.logCapture.push(
+      JSON.stringify({
+        logSequenceNumber: this.logSequenceNumber,
+        logMessage: msg,
+        timestamp: date.getTime(),
```
aburnson: Suggest renaming to `timestampMs` to make it more clear that this refers to Unix epoch millis.

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,127 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { Logger, LogLevel } from '../index';
+import IntervalScheduler from '../scheduler/IntervalScheduler';
+
+export default class POSTRequestLogger implements Logger {
+  private logCapture: string[] = [];
+  private logSequenceNumber: number = 0;
+  private lock = false;
+  private static BATCH_SIZES = 75;
```
aburnson: If intervalMs is configurable, the batch size should also be configurable (add as a param to constructor).

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,127 @@
...
+  private intervalScheduler: IntervalScheduler;
+
+  constructor(
+    private name: string,
+    private meetingId: string,
+    private attendeeId: string,
```
aburnson: The logger should be independent of the meeting session. Meeting and attendee IDs are scoped to meeting sessions. It is the responsibility of the thing doing the logging to ensure that the meeting ID and attendee ID are part of the log line. Suggest removing these.

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,127 @@
...
+    private intervalMs: number,
+    private baseURL: string,
+    private level = LogLevel.WARN,
```
aburnson: Suggest using INFO level as the default. WARN level will not be very helpful.

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  publishLogs(batchSize: number): void {
+    this.intervalScheduler.start(async () => {
+      if (this.lock === true || this.logCapture.length === 0) {
+        return;
+      }
+      this.lock = true;
+      const batch = this.logCapture.slice(0, batchSize);
+      const body = JSON.stringify({
```
aburnson: As noted above, move this to function.

src/logger/POSTRequestLogger.ts
```diff
@@ -0,0 +1,127 @@
...
+      try {
+        const response = await fetch(this.baseURL, {
+          method: 'POST',
+          body,
+        });
+        if (response.status === 200) {
+          this.logCapture = this.logCapture.slice(batch.length);
+        }
+      } catch (error) {
+        this.warn('[POSTRequestLogger] ' + error.message);
```
aburnson: Suggest adding an explicit `console.warn`. If the endpoint behind POSTRequestLogger is completely down, it will be difficult to detect since the warn log about that issue will never make it. (See note below about removing console logging for regular messages).

src/logger/POSTRequestLogger.ts
RidipDe: Should we pass  MeetingSessionConfiguration object as a parameter and then read the meetingId and the AttendeeId ?

aburnson: The logger should be independent of the meeting session. It should not care about meeting sessions or their properties.

src/logger/POSTRequestLogger.ts
RidipDe: ConsoleLogger.ts has also used "timestamp" as the variable

src/logger/MeetingSessionPOSTLogger.ts
```diff
@@ -0,0 +1,124 @@
...
+  private logSequenceNumber: number = 0;
+  private lock = false;
+  private intervalScheduler: IntervalScheduler;
+
+  constructor(
+    private name: string,
+    private configuration: MeetingSessionConfiguration,
+    private batchSize: number,
+    private intervalMs: number,
+    private baseURL: string,
```
aburnson: This should just be `url` as it is the actual URL that is being used in the fetch/sendBeacon requests.

src/logger/MeetingSessionPOSTLogger.ts
```diff
@@ -0,0 +1,124 @@
...
+        }
+      } catch (error) {
+        console.warn('[POSTRequestLogger] ' + error.message);
+      } finally {
+        this.lock = false;
+      }
+    });
+  }
+
+  private jsonStringify(batch: string[]) : string {
```
aburnson: Suggest renaming to makeRequestBody

src/logger/MeetingSessionPOSTLogger.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (type < this.level) {
+      return;
+    }
+    const date = new Date();
+    const timestamp = date.toISOString();
+    const logMessage = `${this.logSequenceNumber} ${timestamp} [${LogLevel[type]}] ${this.name} - ${msg}`;
+
+    switch (type) {
+      case LogLevel.ERROR:
+        console.trace(logMessage);
```
aburnson: As mentioned before, the MeetingSessionPOSTLogger should not console log.

src/logger/MeetingSessionPOSTLogger.ts
```diff
@@ -0,0 +1,124 @@
...
+import MeetingSessionConfiguration from '../meetingsession/MeetingSessionConfiguration';
+
+export default class MeetingSessionPOSTLogger implements Logger {
+  private logCapture: string[] = [];
+  private logSequenceNumber: number = 0;
+  private lock = false;
+  private intervalScheduler: IntervalScheduler;
+
+  constructor(
+    private name: string,
```
aburnson: Do we need `name`? Suggest inserting the meeting id/attendee id into the log message. That will be more useful than a `name`.

src/logger/MeetingSessionPOSTLogger.ts
```diff
@@ -0,0 +1,124 @@
...
+    });
+  }
+
+  private log(type: LogLevel, msg: string): void {
+    if (type < this.level) {
+      return;
+    }
+    const date = new Date();
+    const timestamp = date.toISOString();
+    const logMessage = `${this.logSequenceNumber} ${timestamp} [${LogLevel[type]}] ${this.name} - ${msg}`;
```
aburnson: As mentioned above, since we have access to the meeting session configuration, we might as well inject it into the log line here so that we can guarantee that all meeting session logs will have meeting id/attendee id.

src/logger/MeetingSessionPOSTLogger.ts
```diff
@@ -0,0 +1,124 @@
...
+        console.info(logMessage);
+        break;
+    }
+    this.logCapture.push(
+      JSON.stringify({
+        logSequenceNumber: this.logSequenceNumber,
+        logMessage: msg,
+        timestamp: date.getTime(),
+        logLevelType: LogLevel[type],
+      })
```
aburnson: This is going to get double stringified by makeRequestBody.

For field names I would suggest:

```
sequenceNumber
message
timestampMs
logLevel
```

RidipDe: the array is being stringified, the makeRequestBody() stringifies the JSON

src/logger/MeetingSessionPOSTLogger.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export default class MeetingSessionPOSTLogger implements Logger {
+  private logCapture: string[] = [];
+  private logSequenceNumber: number = 0;
+  private lock = false;
+  private intervalScheduler: IntervalScheduler;
+
+  constructor(
+    private name: string,
+    private configuration: MeetingSessionConfiguration,
```
asnyatkov: This is what I call "Inverse dependency": logger is inherently a low level component, and should not know anything about meetings or meeting session configuration. You should either pass required context fields separately (meeting id, attendee id), or use untyped options dictionary.

src/logger/MeetingSessionPOSTLogger.ts
asnyatkov: Another note: I'm confused by the purpose of logSequenceNumber. I believe that timestamp is sufficient.

src/logger/MeetingSessionPOSTLogger.ts
```diff
@@ -0,0 +1,124 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import Logger from '../logger/Logger';
+import LogLevel from '../logger/LogLevel';
+import IntervalScheduler from '../scheduler/IntervalScheduler';
+import MeetingSessionConfiguration from '../meetingsession/MeetingSessionConfiguration';
+
+export default class MeetingSessionPOSTLogger implements Logger {
```
asnyatkov: The name should be just POSTLogger. Following the comment I left below, logger should not be tied (or know) to a meeting session.

src/logger/MeetingSessionPOSTLogger.ts
aburnson: We should not double stringify, this will make it harder to parse by the endpoint.

src/logger/LogBody.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export default interface LogBody {
```
asnyatkov: This class should be a simple type instead of an interface. You just need to store four primitive data fields. There's no need for inheritance-based implementation.

src/logger/MeetingLog.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export default class MeetingLog {
+  sequenceNumber: number;
+  message: string;
+  timestampMs: number;
+  logLevel: string;
```
simmkyu: Use the `LogLevel` enum instead of `string`.

src/logger/MeetingLog.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  constructor(sequenceNumber: number, message: string, timestampMs: number, logLevel: string) {
```
simmkyu: - If you look at other "model" classes in SDK, many of them do not have a constructor unless you need a particular logic. e.g., [AudioVideoControllerState](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/audiovideocontroller/AudioVideoControllerState.ts) and [ConnectionHealthPolicyConfiguration](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/connectionhealthpolicy/ConnectionHealthPolicyConfiguration.ts).

- Also, I'm not sure this class is needed since only the `MeetingSessionPOSTLogger` is using this model internally. We do not need to expose unnecessary APIs for builders. (Note that all classes, types, and interfaces in individual files are exposed.)

test/logger/MeetingSessionPOSTLogger.test.ts
```diff
@@ -0,0 +1,351 @@
...
+        const startLogPublishSchedulerSpy = sinon.spy(logger, 'startLogPublishScheduler');
+        logger.startLogPublishScheduler(batchSize);
+        expect(startLogPublishSchedulerSpy.calledOnce).to.be.true;
+        logger.stop();
+        done();
+      });
+    });
+
+    it('handles when the fetch call fails', done => {
+      new TimeoutScheduler(20).start(() => {
```
simmkyu: You can remove outer `TimeoutScheduler` from all of your tests but just need an inner one (line 219) for testing the interval scheduler.

CHANGELOG.md
```diff
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add new ContentShareController APIs
 - Getting Started guide
 - Doc guide generator
-- Add basic component library setup
+- Add POSTLogger for meeting sessions
```
simmkyu: `Add basic component library setup` is still being removed in your file.

src/logger/MeetingLog.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import LogLevel from './LogLevel';
+
+export default class MeetingLog {
```
simmkyu: Update `MeetingLog` to `Log` as Andrew suggested.

test/dommock/DOMMockBuilder.ts
```diff
@@ -394,6 +394,9 @@ export default class DOMMockBuilder {
     GlobalAny.navigator = {
       mediaDevices: new mediaDevicesMaker(),
       userAgent: USER_AGENTS.get(mockBehavior.browserName),
+      sendBeacon(url: string, data?: Blob | Int8Array | Int16Array | Int32Array | Uint8Array | Uint16Array | Uint32Array | Uint8ClampedArray | Float32Array | Float64Array | DataView | ArrayBuffer | FormData | string | null): boolean {
```
simmkyu: Prepend ```_``` to unused arguments

PR #144, opened 2020-02-29
Add validation session Id and disable extendedDebugging flag for SauceLabs

ltrung:
*Description of changes*
Add validation for session Id and disable extendedDebugging flag for SauceLabs.

### DIFF
--- a/integration/js/utils/SdkBaseTest.js
+++ b/integration/js/utils/SdkBaseTest.js
@@ -1,7 +1,5 @@
-const {KiteBaseTest} = require('../node_modules/kite-common');
-const {AppPage} = require('../pages/AppPage');
+const {KiteBaseTest, TestUtils} = require('../node_modules/kite-common');
 const {SaucelabsSession} = require('./WebdriverSauceLabs');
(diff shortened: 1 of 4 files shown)


integration/js/utils/SdkBaseTest.js
```diff
@@ -109,11 +120,33 @@ class SdkBaseTest extends KiteBaseTest {
...
+        await this.quitSeleniumSessions();
+        return;
+      }
+    }
+    console.log('I am sending that I am ready');
+    this.io.emit('session_ready');
+
+    //Wait for other to be ready
+    if (this.numberOfParticipant > 1) {
+      await this.waitForTestReady();
```
katayal: What would happen in this case:
Thread 1 acquired a session on saucelabs and failed because the other thread was not ready. At this point the `numberAttendeeSessionReady` in server.js is 1.

Now the Thread 2 is able to acquire a session on saucelabs so the `numberAttendeeSessionReady` is now 2. The server.js will then send the message to Thread 1 (already ended) and Thread 2 (alive) that the tests are ready so Thread 2 will continue with the test.
We need a need a way to prevent this from happening.

PR #145, opened 2020-02-29
Added sleep timeout condition to reconnect logic

RidipDe:
*Issue #:* 474
Reconnected to the meeting even after putting laptop for sleep/closing lid for a longtime

*Description of changes*
Fixed reconnect logic to stop trying to reconnect to a meeting after the laptop is put to sleep.
Changed the hasPastReconnectDeadline logic. Adds DefaultReconnectController as a pingPongObserver and updates the time it received the last Pong in response for a ping.


src/reconnectcontroller/DefaultReconnectController.ts
```diff
@@ -25,24 +30,47 @@ export default class DefaultReconnectController implements ReconnectController {
...
 
-  startedConnectionAttempt(isFirstConnection: boolean): void {
+  startedConnectionAttempt(isFirstConnection: boolean, pingPong: PingPong): void {
     this._isFirstConnection = isFirstConnection;
     if (!this.firstConnectionAttempted) {
       this.firstConnectionAttempted = true;
       this.firstConnectionAttemptTimestamp = Date.now();
+      this.pingPong = pingPong;
+      this.pingPong.addObserver(this);
+      this.pingPong.start();
```
aburnson: The `PingPong` is already started and stopped in the connection monitor: https://github.com/aws/amazon-chime-sdk-js/blob/master/src/connectionmonitor/SignalingAndMetricsConnectionMonitor.ts#L43

Why do we need to start and stop it here? We should avoid having two different objects controlling the lifecycle of `PingPong`.

src/reconnectcontroller/DefaultReconnectController.ts
```diff
@@ -25,24 +30,47 @@ export default class DefaultReconnectController implements ReconnectController {
     return Date.now() - this.firstConnectionAttemptTimestamp;
   }
 
+  private timeSinceLastActiveMs(): number {
```
aburnson: Rather than implement all of this logic here and making the `ReconnectController` more complicated, I would suggest making a new monitor, e.g. `idlemonitor/IdleMonitor.ts` to handle the *detection* of the idle state. Then `ReconnectController` could then ask `IdleMonitor.isIdle()` in `hasPastReconnectDeadline` to determine whether it is allowed to reconnect.

src/reconnectcontroller/DefaultReconnectController.ts
```diff
@@ -25,24 +30,47 @@ export default class DefaultReconnectController implements ReconnectController {
...
   private hasPastReconnectDeadline(): boolean {
-    return this.timeSpentReconnectingMs() >= this.reconnectTimeoutMs;
+    const hasSpentTooMuchTimeReconnecting =
+      this.timeSpentReconnectingMs() >= this.reconnectTimeoutMs;
+    const hasBeenTooLongSinceConnected = this.timeSinceLastActiveMs() >= this.reconnectTimeoutMs;
+    return hasSpentTooMuchTimeReconnecting || hasBeenTooLongSinceConnected;
+  }
+
+  didReceivePong(_latencyMs: number, _clockSkewMs: number): void {
+    this.lastConnectionAttemptTimestamp = Date.now();
```
aburnson: This isn't the timestamp of the last connection attempt, it's the timestamp of when the connection was recently known to be active.

src/idlemonitor/DefaultIdleMonitor.ts
```diff
@@ -0,0 +1,41 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import PingPong from '../pingpong/PingPong';
+import PingPongObserver from '../pingpongobserver/PingPongObserver';
+import IdleMonitor from './IdleMonitor';
+
+export default class DefaultIdleMonitor implements IdleMonitor, PingPongObserver {
+  private canReceivePongs: boolean = false;
+  private lastPongResponseSeen: number = Date.now();
```
simmkyu: - Rename it to `lastActiveTimestampMs` because, technically, you don't get a pong when instantiating this class.

  `__TimestampMS` postfix examples in SDK
  ```
  connectionStartTimestampMs
  currentTimestampMs
  lastGoodSignalTimestampMs
  lastNoSignalTimestampMs
  lastPacketLossInboundTimestampMs
  lastWarnTimestampMs
  lastWeakSignalTimestampMs
  previousTimestampMs
  setLastGoodSignalTimestampMs
  setLastNoSignalTimestampMs
  setLastPacketLossInboundTimestampMs
  setLastWeakSignalTimestampMs
  startTimestampMs
  videoInputAttachedTimestampMs
  ```

CHANGELOG.md
```diff
@@ -37,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix crash in demo app when click on screen share view
 - Fix integration test completion time writer
 - Fix the ping pong reconnection issue
+- Added sleep timeout condition to reconnect logic
```
ltrung: Duplicate

src/idlemonitor/IdleMonitor.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  /**
+   * Add IdleMonitor to the list of PingPong observers
+   */
+  addToPingPongObserver(_pingPong: PingPong): void;
+
+  /**
+   * Remove IdleMonitor from the list of PingPong observers
+   */
+  removeIdleMonitorObserver(): void;
```
ltrung: Should the name be RemoveFromPingPongObserver to match the other method?

test/idlemonitor/DefaultIdleMonitor.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    forEachObserver(_observerFunc: (_observer: PingPongObserver) => void): void {}
+
+    start(): void {}
+
+    stop(): void {}
+  }
+
+  beforeEach(() => {
+    expect = chai.expect;
```
ltrung: Could you move this and maxIdleTimeMs up? We dont need to reset this for every test.

src/idlemonitor/DefaultIdleMonitor.ts
```diff
@@ -0,0 +1,41 @@
...
+    return Date.now() - this.lastPongResponseSeen;
+  }
+
+  isIdle(): boolean {
+    return this.timeSinceLastActiveMs() >= this.maxIdleTimeMs;
+  }
+
+  didReceivePong(_id: number, _latencyMs: number, _clockSkewMs: number): void {
+    this.lastPongResponseSeen = Date.now();
+    this.canReceivePongs = true;
```
simmkyu: The variable name says whether or not it can receive pongs. Then, should not we set this to `false` in `PingPongObserver.didMissPongs()`?

test/reconnectcontroller/DefaultReconnectController.test.ts
```diff
@@ -14,10 +17,32 @@ describe('DefaultReconnectController', () => {
     return new DefaultReconnectController(50, new FullJitterBackoff(10, 0, 0));
   };
 
+  class TestPingPong implements PingPong {
```
ltrung: Nit: We usually name this type of class NoOp...

test/reconnectcontroller/DefaultReconnectController.test.ts
```diff
@@ -41,9 +66,11 @@ describe('DefaultReconnectController', () => {
   });
 
   describe('retryWithBackoff', () => {
+    const controller = defaultController();
```
ltrung: Nit: Could we rename this to reconnectController?

test/reconnectcontroller/DefaultReconnectController.test.ts
```diff
@@ -82,9 +110,45 @@ describe('DefaultReconnectController', () => {
...
+        controller.retryWithBackoff(
+          () => {
+            done();
+          },
+          () => {}
+        )
+      ).to.equal(true);
+    });
+
+    it('stops calling the retry func if it is past the deadline', done => {
```
ltrung: test has the same name as the one above?

test/reconnectcontroller/DefaultReconnectController.test.ts
```diff
@@ -14,10 +17,32 @@ describe('DefaultReconnectController', () => {
...
+    isIdle(): boolean {
+      return false;
+    }
+    addToPingPongObserver(_pingPong: PingPong): void {}
+    didReceivePong(_id: number, _latencyMs: number, _clockSkewMs: number): void {}
+    removeIdleMonitorObserver(): void {}
+  }
+
   beforeEach(() => {
     expect = chai.expect;
```
ltrung: This can be moved up as well. No need for beforeEach here.

test/reconnectcontroller/DefaultReconnectController.test.ts
```diff
@@ -54,6 +81,7 @@ describe('DefaultReconnectController', () => {
 
     it('calls the cancel func if canceled after starting retry', done => {
       const controller = defaultController();
+      controller.startedConnectionAttempt(false, new TestIdleMonitor(new TestPingPong(), 50));
```
ltrung: Can we create a IdleMonitor object up top and pass it around?

test/reconnectcontroller/DefaultReconnectController.test.ts
```diff
@@ -82,9 +110,45 @@ describe('DefaultReconnectController', () => {
       });
     });
 
+    it('does not call  the retry func', done => {
```
ltrung: Can you add more to this description? When do we not call the retry func?

src/idlemonitor/DefaultIdleMonitor.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  removeIdleMonitorObserver(): void {
+    this.canReceivePongs = false;
+    this.pingPong.removeObserver(this);
+  }
+
+  private timeSinceLastActiveMs(): number {
+    if (!this.canReceivePongs) {
+      return 0;
```
simmkyu: Technically, returning `0` means that the session is active if this component cannot receive a pong. This does not make sense. Recommend making the following changes:

- Rename `this.canReceivePongs` to `this.enabled` (or other generic name).
- Set `this.enabled` to `true` in `start()`.
- Set `this.enabled` to `false` in `stop()`. `this.enabled` just tells whether `DefaultIdleMonitor` is enabled or not.
- In `isIdle()`, return `false` for the case you cannot determine whether it's idle or not.
  ```
   isIdle(): boolean {
    if (!this.enabled) {
      return false;
    }
    return this.timeSinceLastActiveMs() >= this.maxIdleTimeMs;
  }
  ```

src/reconnectcontroller/ReconnectController.ts
```diff
@@ -16,7 +18,7 @@ export default interface ReconnectController {
    * can be set.
    * @param{boolean} isFirstConnection whether this is the first attempt to connect for this session
    */
-  startedConnectionAttempt(isFirstConnection: boolean): void;
+  startedConnectionAttempt(isFirstConnection: boolean, idelMonitor: IdleMonitor): void;
```
simmkyu: I don't think it's a good idea to depend on `IdleMonitor` in the interface. If a builder implements `ReconnectController` or already has implemented it, they need to pass an `IdleMonitor` instance to `startedConnectionAttempt()`. From the API design perspective, what does `IdleMonitor` mean to `startedConnectionAttempt()`?

I think you can implement this feature without modifying `ReconnectController` and `DefaultReconnectController`. Instead, you can just use the reconnect controller to turn on and off from the outside. (I remember we discussed this before.)

I will comment in `DefaultReconnectController`.

src/task/CleanStoppedSessionTask.ts
```diff
@@ -41,6 +41,8 @@ export default class CleanStoppedSessionTask extends BaseTask {
       this.context.statsCollector = null;
       this.context.connectionMonitor.stop();
       this.context.connectionMonitor = null;
+      this.context.idleMonitor.removeIdleMonitorObserver();
```
simmkyu: Here, once you changed them to `start()` and `stop()`, this will be `stop()` that matches other monitor components.

src/reconnectcontroller/DefaultReconnectController.ts
```diff
@@ -26,7 +28,9 @@ export default class DefaultReconnectController implements ReconnectController {
   }
 
   private hasPastReconnectDeadline(): boolean {
```
simmkyu: - `hasPastReconnectDeadline()` is being used only by `retryWithBackoff()`.
- Let's take a look at `retryWithBackoff()`. In addition to `hasPastReconnectDeadline()`, it's using `this.shouldReconnect` to determine whether or not to retry, right?

```
  retryWithBackoff(retryFunc: () => void, cancelFunc: () => void): boolean {
    const willRetry = this.shouldReconnect && !this.hasPastReconnectDeadline();
    if (willRetry) {
   ...
```

- Look at `disableReconnect()`. This is a public method.
```
  disableReconnect(): void {
    this.shouldReconnect = false;
  }
```

In summary, you can disable the reconnect operation without modifying the `ReconnectController` interface and the `DefaultReconnectController` implementation.

Note that `reconnectController` is already available in `AudioVideoControllerState` so you can disable the reconnect operation literally anywhere in the audio-video controller, in tasks, or even in your component.

simmkyu:
https://github.com/aws/amazon-chime-sdk-js/pull/169

I made small changes from this PR. Thanks!
- Use the existing `SignalingAndMetricsConnectionMonitor` over a new `IdleMonitor`.
- Add a method to `ReconnectController`.

PR #147, opened 2020-03-04
Fixing infinite loop when retrying in audio and video integ tests

ltrung:
*Description of changes*
Fixing infinite loop when retrying in audio and video integ tests

### DIFF
--- a/integration/js/server.js
+++ b/integration/js/server.js
@@ -15,7 +15,7 @@ const port = process.argv[2];
 const numberOfParticipant = parseInt(process.argv[3], 10);
 const interval = 1000;
 let createMeetingPromise = undefined;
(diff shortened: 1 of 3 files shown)


integration/js/server.js
```diff
@@ -56,10 +56,17 @@ io.on('connection', function (socket) {
...
-      numberAttendeeReady -= 1;
+      attendeeReadyMap.set(socket, false);
     }
+
+    let numberAttendeeReady = 0;
+    attendeeReadyMap.forEach(function(value, socket) {
+      if (value === true) {
+        numberAttendeeReady += 1;
+      }
+    })
```
simmkyu: missing semicolon

PR #148, opened 2020-03-04
Make sure both participants in audio and video tests reach finish state before retrying

ltrung:
*Description of changes*
Make sure both participants in audio and video tests reach finish state before retrying


integration/js/utils/SdkBaseTest.js
```diff
@@ -12,6 +12,7 @@ class SdkBaseTest extends KiteBaseTest {
     this.meetingTitle = kiteConfig.uuid;
     this.testName = testName;
     this.testReady = false;
+    this.testFinish = false;
```
simmkyu: optional: I'd rename it to ```testFinished``` to follow conventions

PR #151, opened 2020-03-05
Add basic component library setup

davejlane:
*Issue #: n/a* 

*Description of changes*
This change introduces the basic elements of a UI component library that  is intended to be utilized alongside the Chime SDK. It is build using React, Typescript, and Styled Components,  and utilizes Storybook for developing and displaying components within the library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add new ContentShareController APIs
 - Getting Started guide
 - Doc guide generator
+- Add basic component library setup
 
 ### Changed
 - Do not bypass ice gathering based on sdp connection attributes for Safari on iOS
--- a/demos/components/README.md
+++ b/demos/components/README.md
@@ -1,0 +1,41 @@
+# Chime SDK UI Component Library
+
+This contains reusable components written with React, TypeScript and styled components to be used to create UIs with Chime SDK JS.
+
+## To genereate dependencies
+
+```
+npm install
+```
+
+## To run Storybook locally
+
+```
+npm start
+```
+
+## Build
+
+```
+npm run build
+```
+
+# Usage
+
+After publishing your lib (ie `my-react-lib`) you can use it by importing the styles in your App's entry point:
+
+
+
+Then you can import and use your components:
+
+```tsx
+//
+import React from 'react'
+import { Example } from 'my-react-lib'
+
+export class App extends React.Component {
+  render() {
+    return <Example />
+  }
+}
+```
--- a/demos/components/src/components/Button/PrimaryButton.tsx
+++ b/demos/components/src/components/Button/PrimaryButton.tsx
@@ -1,0 +1,8 @@
+import React from 'react';
+
+import { ButtonProps, Button } from './';
+import { StyledPrimaryButton } from './Styled';
+
+export const PrimaryButton: React.FC<ButtonProps> = (props) => <Button addStyles={StyledPrimaryButton} {...props} />;
+
+export default PrimaryButton;
--- a/demos/components/src/components/Button/SecondaryButton.tsx
+++ b/demos/components/src/components/Button/SecondaryButton.tsx
@@ -1,0 +1,8 @@
+import React from 'react';
+import { StyledSecondaryButton } from './Styled';
+
+import { Button, ButtonProps } from './';
+
+export const SecondaryButton: React.FC<ButtonProps> = props => <Button addStyles={StyledSecondaryButton} {...props} />;
+
+export default SecondaryButton;
--- a/demos/components/src/components/VideoTile/Styled.tsx
+++ b/demos/components/src/components/VideoTile/Styled.tsx
@@ -1,0 +1,81 @@
+import styled from 'styled-components';
+
+import { VideoTileProps } from './';
+
+export const StyledVideoTile = styled.div<VideoTileProps>`
+  width: 100%;
+  background: #1b1c20;
+  position: relative;
+  display: inline-block;
+  margin: 1%;
+  vertical-align: top;
+
+  &:before {
+    content: '';
+    display: block;
+    padding-top: 56.25%;
+  }
+
+  &:after {
+    content: '';
+    position: absolute;
+    box-shadow: 0 1rem 2.5rem 0 black;
+    opacity: 0;
+    transition: opacity 0.2s ease-in-out;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+  }
+
+  &:hover:after {
+    opacity: 0.3;
+  }
+
+  video {
+    position: absolute;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+    width: 100%;
+    height: 100%;
+  }
+
+  .icon {
+    width: 1.5rem;
+    height: 1.5rem;
+    display: inline-block;
+    background-color: papayawhip;
+    margin-right: 0.5rem;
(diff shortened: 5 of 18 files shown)


demos/components/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Build
+
+```
+npm run build
+```
+
+# Usage
+
+After publishing your lib (ie `my-react-lib`) you can use it by importing the styles in your App's entry point:
```
aburnson: Wonder if we could get a small demo app to go along with the PR to demonstrate this with? Maybe that demo could then live in `demos/react` and the `components` folder could live in the root at `/react` or `/components`?

davejlane: I was planning that as a next step. Initially I wanted to land the basic setup for a component library so that others could start utilizing and contributing to it. What are your thoughts?

demos/components/src/components/Button/SecondaryButton.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { StyledSecondaryButton } from './Styled';
+
+import { Button, ButtonProps } from './';
+
+export const SecondaryButton: React.FC<ButtonProps> = props => {
```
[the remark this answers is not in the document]
davejlane: That's somewhat intentional. In some cases, we're passing the entire props object as is to a child component, in others de-structuring is used for easier reference.

demos/components/src/components/VideoTile/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+  bindVideoTile: (arg0: any) => void;
+}
+
+export const VideoTile: React.SFC<VideoTileProps> = props => {
+  const videoEl = useRef(null);
+  const { bindVideoTile } = props;
+
+  useEffect(() => {
+    !!videoEl && bindVideoTile(videoEl.current);
+  }, [bindVideoTile, videoEl]);
```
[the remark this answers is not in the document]
davejlane: Storybook is being used to document and present UI components. In this case, we won't be binding a videoTile to the SDK, so I'm passing an empty function so that no action is taken.

Good catch on `useEffect`, I'll update  so that it is only called once. (The equivalent of `componentDidMount`.

demos/components/lib/components/Button/PrimaryButton.d.ts
```diff
@@ -0,0 +1,4 @@
+import React from 'react';
```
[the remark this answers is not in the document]
davejlane: That's a good question. Eventually, this package should be it's own `npm` module, in which case we'll need to define a strategy for preparing it's contents for consumption. I can remove this until we're at that point.

demos/components/webpack.config.js
```diff
@@ -0,0 +1,71 @@
...
+      commonjs: 'react',
+      commonjs2: 'react',
+      amd: 'react',
+      root: 'React'
+    }
+  },
+  module: {
+    rules: [
+      {
+        test: /\.css$/,
```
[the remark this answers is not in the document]
davejlane: They don't, but again I was thinking ahead to when this is a full module. I'll remove until we need it.

demos/components/src/components/Button/PrimaryButton.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { ButtonProps, Button } from './';
+import { StyledPrimaryButton } from './Styled';
+
+export const PrimaryButton: React.FC<ButtonProps> = (props) => {
+  const { icon, label } = props;
+  return (
+    <Button
+      {...props}
+      addStyles={StyledPrimaryButton}
+      icon={icon}
```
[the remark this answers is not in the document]
davejlane: That's a good callout. The props for this component are a combination of some custom props and native button props as defined by TypeScript. I can pass the entire props object here and disseminate in the root `Button` component.

demos/components/README.md
aburnson: Ok, that makes sense.

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add ping/pong to screen sharing start code path to ensure socket is viable
 - Enable integration tests for safari 12
 - Write timestamp for latest canary completion time
+- Add basic component library setup
```
simmkyu: Move this to `## [Unreleased]` section (line 14)

demos/components/src/components/VideoTile/Styled.tsx
```diff
@@ -0,0 +1,84 @@
...
+    bottom: 0;
+    width: 100%;
+    height: 100%;
+  }
+
+  .icon {
+    width: 1.5rem;
+    height: 1.5rem;
+    display: inline-block;
+<<<<<<< HEAD
```
simmkyu: Resolve a conflict in this file

demos/components/src/components/VideoTile/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  nameplate: string;
+  bindVideoTile: (arg0: any) => void;
+}
+
+export const VideoTile: React.SFC<VideoTileProps> = props => {
+  const videoEl = useRef(null);
+  const { bindVideoTile } = props;
+
+  useEffect(() => {
+    !!videoEl && bindVideoTile(videoEl.current);
```
simmkyu: Sorry for the confusion. If a builder passes a new function to ```props.bindVideoTile``` in the middle of a meeting, a component cannot bind a tile anymore. The previous code worked, but the caller (storybook) just needed to pass the same object.

e.g., Using the `useCallback` hook in a storybook (but we may not need this since a storybook is for a documentation.)
```
<VideoTile bindVideoTile={useCallback(() => {}, [])} nameplate="Test name" />
```

davejlane: I think that I understand your question better now, and I'm fairly unfamiliar with how the SDK will connect with this; you're saying that it's possible that the SDK might bass a new function to the videoTile, so we should make sure that this function gets called when it changes? I'll update it to make sure that it does that.

CHANGELOG.md
davejlane: Will do.

demos/components/src/components/VideoTile/index.tsx
simmkyu: It will depend on the library design: How will these React components use `amazon-chime-sdk-js`?

Option 1) Does a builder need to set up `amazon-chime-sdk-js` separately and pass a SDK object into these UI component as props?

Option 2) The UI library use `amazon-chime-sdk-js` internally. For example, in a typical React 3rd party library, a provider <ChimeProvider> passes context variables down to <VideoTile> and keeps business logic internal.

This PR is for adding basic UI setup, so let's not worry about the approach yet. Sorry, please revert it back to the previous revision.

```
useEffect(() => {
    !!videoEl && bindVideoTile(videoEl.current);
  }, [bindVideoTile, videoEl]);
```

davejlane: Thanks for the explanation, and I think that the implementation is not decided yet, although I'm leaning towards option 1, with some HOC that will manage the business logic of the SDK objects and pass information down to this more presentational component for rendering.

I should have given more context around Storybook: It's simply a tool for developing and viewing UI components. I agree that once we start building applications that integrate these components with the SDK, we'll get a much better picture of the best implementation approach.

demos/components/webpack.config.js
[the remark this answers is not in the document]
davejlane: I was wrong about this. I added some global style for Storybook, and this is required to process those styles while running storybook, so I'll leave this as is for now.

PR #153, opened 2020-03-07
Fix ping pong restart issue after reconnect

RidipDe:
*Issue #:* 
Bug discovered while solving the issue #474

*Description of changes*
The ping pong failed to start after the laptop was disconnected from the internet and then reconnected. The issue has been fixed.


CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix audio-video session stop to return Left status code
 - Fix crash in demo app when click on screen share view
 - Fix integration test completion time writer
+- - Fix ping pong restart issue after reconnect
```
aburnson: Extra dash

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -74,15 +75,14 @@ export default class DefaultAudioVideoController implements AudioVideoController
...
-
   private connectionHealthData = new ConnectionHealthData();
   private observerQueue: Set<AudioVideoObserver> = new Set<AudioVideoObserver>();
   private meetingSessionContext = new AudioVideoControllerState();
   private sessionStateController: SessionStateController;
 
   private static MIN_VOLUME_DECIBELS = -42;
   private static MAX_VOLUME_DECIBELS = -14;
-  private static PING_PONG_INTERVAL_MS = 10000;
+  private static PING_PONG_INTERVAL_MS = 5000;
```
aburnson: Why is the ping pong interval changing?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -63,6 +63,7 @@ import NScaleVideoUplinkBandwidthPolicy from '../videouplinkbandwidthpolicy/NSca
 import DefaultVolumeIndicatorAdapter from '../volumeindicatoradapter/DefaultVolumeIndicatorAdapter';
 import WebSocketAdapter from '../websocketadapter/WebSocketAdapter';
 import AudioVideoControllerState from './AudioVideoControllerState';
+//import WebSocketReadyState from "../websocketadapter/WebSocketReadyState";
```
aburnson: If this is no longer necessary, then remove it instead of commenting it out.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -216,16 +216,18 @@ export default class DefaultAudioVideoController implements AudioVideoController
     this.meetingSessionContext.videosToReceive = new DefaultVideoStreamIdSet();
     this.meetingSessionContext.videosPaused = new DefaultVideoStreamIdSet();
     this.meetingSessionContext.statsCollector = new DefaultStatsCollector(this, this.logger);
+    this.meetingSessionContext.pingPong = new DefaultPingPong(
+      this.meetingSessionContext.signalingClient,
+      DefaultAudioVideoController.PING_PONG_INTERVAL_MS,
+      this.logger
+    );
+    this.meetingSessionContext.pingPong.start();
```
aburnson: Connection monitor is currently [responsible for starting and stopping](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/connectionmonitor/SignalingAndMetricsConnectionMonitor.ts#L43) PingPong. Doing this splits responsibility over PingPong lifecycle.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -537,6 +539,7 @@ export default class DefaultAudioVideoController implements AudioVideoController
 
       this.sessionStateController.perform(SessionStateControllerAction.FinishConnecting, () => {
         this.actionFinishConnecting();
+        this.meetingSessionContext.pingPong.start();
```
aburnson: Why is this not in actionFinishConnecting?

RidipDe:
> 3\. actionDisconnect
Thank you Kyu for the explanation, it is of great help. 
As discussed before, the this.stop() in the [DefaultPingPong.handleSignalingClientEvent() ](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/pingpong/DefaultPingPong.ts#L117) is where the intervalScheduler is stopping, which in turn stops the  pingPong mechanism. So basically, we need to explicitly re-start this intervalScheduler. 
Whenever there is a reconnect after the reconnect, the DefaultAudioVideoController.actionReconnect() is called which does not have the MonitorTask. There were three methods that I thought could re-start the pingPong mechanism.
1) Add a new MonitorTask() as part of the array tasksToRunSerially in  [SerialGroupTask()](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/audiovideocontroller/DefaultAudioVideoController.ts#L506) call in the [DefaultAudioVideoController.actionReconnect()](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/audiovideocontroller/DefaultAudioVideoController.ts#L496). However, the exiting observer will still be active.
2) Use the this.meetingSessionContext.signalingClient.start() in the [DefaultAudioVideoController.actionFinishConnecting()](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/audiovideocontroller/DefaultAudioVideoController.ts#L302). Even in tis case, the exiting observer will still be active.
3) Add a new property "pingPong" to the AudioVideoControllerState which will save the pingPong state when the object of SignalingAndMetricsConnectionMonitor is created for the first time in [DefaultAudioVideoController.actionConnect()](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/audiovideocontroller/DefaultAudioVideoController.ts#L219), and then call the  this.meetingSessionContext.pingPong.start() in the DefaultAudioVideoController.actionFinishConnecting(). I found this method more appropriate because [DefaultPingPong.stop()](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/pingpong/DefaultPingPong.ts#L60) is called as part of the [DefaultPingPong.start()](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/pingpong/DefaultPingPong.ts#59). Which means we are removing the existing observer and then adding a new observer and then starting the intervalScheduler.

src/audiovideocontroller/DefaultAudioVideoController.ts
RidipDe: will change back to 10000

src/audiovideocontroller/DefaultAudioVideoController.ts
RidipDe: can be moved to be a part of the actionFinishConnecting()

src/audiovideocontroller/DefaultAudioVideoController.ts
RidipDe: Here is what I understand:
https://github.com/aws/amazon-chime-sdk-js/pull/153#issuecomment-596404883
I might be wrong, please let me know...

simmkyu:
> As discussed before, the this.stop() in the DefaultPingPong.handleSignalingClientEvent() is where the intervalScheduler is stopping, which in turn stops the pingPong mechanism. So basically, we need to explicitly re-start this intervalScheduler.

Since the ping pong stops in `DefaultPingPong.handleSignalingClientEvent()`, you should be able to restart that intervalScheduler within `DefaultPingPong` itself but not in `DefaultAudioVideoController`.

To simplify the problem, you can consider that the `DefaultPingPong` class has private start and stop methods, say `DefaultPingPong.internalStart()` and `DefaultPingPong.internalStop()`, in addition to the existing public `start()` and `stop()`.

When a laptop is disconnected from the Internet, `DefaultPingPong.internalStop()` is called, right? Then, you need to call `DefaultPingPong.internalStart()` when the laptop becomes available online.

simmkyu:
Closing this PR as https://github.com/aws/amazon-chime-sdk-js/pull/160 fixes `DefaultPingPong`

PR #156, opened 2020-03-10
Traps video sending SSRC change in consecutive negotiaion as error

XHatan:
*Issue #:* 

Currently, SDK lacked the mechanism to detect video sending section SSRC change during two consecutive negotiations. Such change will break a meeting by removing multiple attendees who are viewing the video.

*Description of changes*

1.  Detect SSRC change at CreateOffer and throws Error in `CreateSDPTask.ts`
2. Only the `update` function in DefaultAudioVideoController catches the Error and calls `restartLocalVideo` to have a correct inter-operation with server.
3. Unit test
4. Change Screen Sharing logging level from WARN to INFO.


src/sdp/DefaultSDP.ts
```diff
@@ -149,4 +214,105 @@ export default class DefaultSDP implements SDP {
 
     return new DefaultSDP(originalSdp);
   }
+
+  withPlanBSimulcast(videoSimulcastLayerCount: number): DefaultSDP {
```
XHatan: Simply revert some changes made in https://github.com/aws/amazon-chime-sdk-js/commit/abc3bf7b27a95152910d2faf8cbbb0a51a42f731#diff-56abeb90f57cd6acbb5beaf736d6ae2f

aburnson: Side note: in that change I also notice we have some tests that still consider Opera to be an unsupported browser:

```
    it('can be constructed with an unsupported browser', () => {
      const mockBuilder = new DOMMockBuilder();
      setUserAgent(SAFARI_USERAGENT);
      setUserAgent(OPERA_USERAGENT);
```

We should use something that will always be unsupported like IE.

PR #159, opened 2020-03-10
Add observer events for content sharing

ltrung:
*Description of changes*
Add some observer events for content sharing.


src/audiovideofacade/DefaultAudioVideoFacade.ts
```diff
@@ -386,6 +387,16 @@ export default class DefaultAudioVideoFacade implements AudioVideoFacade {
...
   }
 
+  addContentShareObserver(observer: ContentShareObserver): void {
+    this.contentShareController.addContentShareObserver(observer);
+    this.trace('addContentShareObserver');
+  }
+
+  removeContentShareObserver(observer: ContentShareObserver): void {
+    this.contentShareController.removeContentShareObserver(observer);
+    this.trace('addContentShareObserver');
```
simmkyu: => `removeContentShareObserver`

src/contentshareobserver/ContentShareObserver.ts
```diff
@@ -0,0 +1,21 @@
...
+   */
+  contentShareDidStop?(): void;
+  /**
+   * Called when a content share session is paused.
+   */
+  contentShareDidPause?(): void;
+  /**
+   * Called when a content share session is unpaused.
+   */
+  contentShareDidUnPause?(): void;
```
simmkyu: nit: lowercase `P` to match `ContentShareControllerFacade.unpauseContentShare()`

src/audiovideofacade/DefaultAudioVideoFacade.ts
ltrung: Fixed.

src/contentshareobserver/ContentShareObserver.ts
ltrung: Updated.

src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -53,14 +61,41 @@ export default class DefaultContentShareController implements ContentShareContro
 
   pauseContentShare(): void {
     this.mediaStreamBroker.toggleMediaStream(false);
+    this.forEachContentShareObserver(observer => {
```
aburnson: Ideally, these observer events would only fire if the state actually transitioned. For example, if you call pauseContentShare twice, this will generate two pause events, but the second one is redundant. If I call pauseContentShare, then I know that because I called it that I'm going to get a callback saying I paused... it's not telling me anything I didn't already know.

src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -52,15 +63,55 @@ export default class DefaultContentShareController implements ContentShareContro
...
+  }
+
+  audioVideoDidStart(): void {
+    this.forEachContentShareObserver(observer => {
+      Maybe.of(observer.contentShareDidStart).map(f => f.bind(observer)());
+    });
+  }
+
+  audioVideoDidStop(sessionStatus: MeetingSessionStatus): void {
+    if (sessionStatus.statusCode() === MeetingSessionStatusCode.Left) {
```
aburnson: I would think we would want to still broadcast events even if the status code was something other than Left

PR #162, opened 2020-03-12
Inject an environment variable to check if the demo app is running on…

simmkyu:
… localhost

*Issue #:* 

*Description of changes*
- Take `process.env.npm_config_is_local` from the NPM script in which you can pass using the `--is_local` option.
- Inject `process.env.npm_config_is_local` into the demo app using the Webpack EnvironmentPlugin.
- In the serverless script, always build a demo app even if there exists assets in the `/dist` folder. I think building a demo app per script execution is better than grabbing possibly old assets anyway.

### DIFF
--- a/demos/browser/app/meeting/meeting.ts
+++ b/demos/browser/app/meeting/meeting.ts
@@ -17,8 +17,10 @@ import {
   Device,
   DeviceChangeObserver,
   LogLevel,
+  Logger,
   MeetingSession,
   MeetingSessionConfiguration,
+  MeetingSessionPOSTLogger,
   MeetingSessionStatus,
   MeetingSessionStatusCode,
   MeetingSessionVideoAvailability,
@@ -96,6 +98,11 @@ class TestSound {
 }
 
 export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver {
+  static readonly DID: string = '+17035550122';
+  static readonly BASE_URL: string = [location.protocol, '//', location.host, location.pathname.replace(/\/*$/, '/')].join('');
+  static readonly LOGGER_BATCH_SIZE: number = 85;
+  static readonly LOGGER_INTERVAL_MS: number = 1150;
+
   showActiveSpeakerScores = false;
   activeSpeakerLayout = true;
   meeting: string | null = null;
@@ -103,9 +110,6 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver 
   voiceConnectorId: string | null = null;
   sipURI: string | null = null;
   region: string | null = null;
-  static readonly DID: string = '+17035550122';
-  static readonly BASE_URL: string = [location.protocol, '//', location.host, location.pathname.replace(/\/*$/, '/')].join('');
-
   meetingSession: MeetingSession | null = null;
   audioVideo: AudioVideoFacade | null = null;
   tileOrganizer: DemoTileOrganizer = new DemoTileOrganizer();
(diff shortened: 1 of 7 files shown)


demos/browser/webpack.config.js
```diff
@@ -18,6 +19,9 @@ module.exports = env => {
         inject: 'head',
       }),
       new HtmlWebpackInlineSourcePlugin(),
+      new webpack.EnvironmentPlugin({
+        IS_LOCAL: process.env.npm_config_is_local === 'true' ? 'true' : 'false'
```
ltrung: Does that mean we have to specify another argument --is_local=true when running npm run start?

simmkyu:
> Can we not just use a simple test from within browser JS like: `location.hostname === 'localhost' || location.hostname === '127.0.0.1'`?

If a customer maps a custom domain name to 127.0.0.1 (e.g., `/etc/hosts` in macOS), this will be broken. Minor issue though - they will just not see any logs in a browser console.

demos/browser/webpack.config.js
simmkyu: `start:fast` in `demos/browser/package.json` specifies `--is_local=true` so you don't need to pass any option.

demos/browser/app/meeting/meeting.ts
```diff
@@ -551,11 +554,11 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
...
   async initializeMeetingSession(configuration: MeetingSessionConfiguration): Promise<void> {
-    let logger :MeetingSessionPOSTLogger | ConsoleLogger = null;
-    if (!(DemoMeetingApp.BASE_URL.includes('localhost') || DemoMeetingApp.BASE_URL.includes("127.0.0.1"))) {
-      logger = new MeetingSessionPOSTLogger('SDK', configuration, 85, 1150, `${DemoMeetingApp.BASE_URL}logs`, LogLevel.INFO);
-    } else {
+    let logger: Logger;
+    if (isLocal) {
       logger = new ConsoleLogger('SDK', LogLevel.DEBUG);
+    } else {
+      logger = new MeetingSessionPOSTLogger('SDK', configuration, 85, 1150, `${DemoMeetingApp.BASE_URL}logs`, LogLevel.INFO);
```
asnyatkov: Can add constants for 85 and 1150?

simmkyu: - Added `readonly` constants.
- Config dictionary vs. individual parameters - Chime SDK has been following the latter pattern, passing individual parameters. e.g., [SignalingClientSubscribe](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/signalingclient/SignalingClientSubscribe.ts#L21)

aburnson:
>> Can we not just use a simple test from within browser JS like: `location.hostname === 'localhost' || location.hostname === '127.0.0.1'`?

> If a customer maps a custom domain name to 127.0.0.1 (e.g., `/etc/hosts` in macOS), this will be broken. Minor issue though - they will just not see any logs in a browser console.

Still seems like a lot of overhead to support this use case; and it's not clear why you would want to test the demos with a custom domain name. Could we not just create a utility function to detect local which could be overridden when doing testing?

simmkyu:
> > > Can we not just use a simple test from within browser JS like: `location.hostname === 'localhost' || location.hostname === '127.0.0.1'`?
> 
> > If a customer maps a custom domain name to 127.0.0.1 (e.g., `/etc/hosts` in macOS), this will be broken. Minor issue though - they will just not see any logs in a browser console.
> 
> Still seems like a lot of overhead to support this use case; and it's not clear why you would want to test the demos with a custom domain name. Could we not just create a utility function to detect local which could be overridden when doing testing?

Agree, adding an env variable in two layers is over-complicating demos. I'll update PR with a simple check against `localhost` and `127.0.0.1`.

PR #176, opened 2020-03-19
Use max-bundle RTCRtpPolicy for firefox

XHatan:
*Issue #:* 

*Description of changes*

`max-bundle` is the policy that is best compatible with our server. 
Safari is using this policy right now. 
Chrome still uses Plan B so it's left untouched.


aburnson:
Rebased and :+1:

PR #179, opened 2020-03-19
Add max-content-share query parameter

ltrung:
*Description of changes*
Add max-content-share query parameter to allow 2 content share at the same time


aburnson:
Needs `npm run build:release` for version bumps

PR #180, opened 2020-03-20
Add an integration test that checks only 2 content share are allowed

ltrung:
*Description of changes*
Add an integration test that checks only 2 content share are allowed

### DIFF
--- a/integration/configs/content_share_screen_capture_test.config.json
+++ b/integration/configs/content_share_screen_capture_test.config.json
@@ -18,11 +18,6 @@
   ],
   "clients": [
     {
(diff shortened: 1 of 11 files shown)


integration/configs/content_share_screen_capture_test.config.json
```diff
@@ -11,17 +11,12 @@
       "noOfThreads": 1,
       "testImpl": "ContentShareScreenCapture.js",
       "payload": {
-        "url": "http://localhost:8080/",
```
RidipDe: should be localhost

PR #184, opened 2020-03-25
Add additional observer to indicate downlink pressure

XHatan:
*Issue #:* 

*Description of changes*
Add two additional callback in `AudioVideoObserver` to pass information about downlink pressure.
1.  `videoNotReceivingEnoughData` callback:
`SdkBitrateFrame` is sent from server that contains the active list of video streams available Exact definition can be found in signaling protocol.
By cross-referencing server measured bitrates of streams with what's actually received from WebRTC stats, we can approximately determine the performance of the client downlink network for each individual stream.
A constant `DEFAULT_DOWNLINK_CALLRATE_UNDERSHOOT_FACTOR` is defined so we only fire callback when the value is dropped below this percentage. 

2. `estimatedDownlinkBandwidthLessThanRequired`
This callback will be fired when WebRTC exposed estimated downlink bandwidth is less than required bitrate (calculated from SdkBitrateFrame). This callback will only available on platform where WebRTC stats expose bandwidth estimation.  WebRTC estimated downlink bandwidth follows a state machine design. At transition state of the estimation process, the callback may be mis-fired so this callback should only be used as side information. 


demos/browser/app/meeting/meeting.ts
```diff
@@ -528,6 +528,10 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
     this.populateAudioOutputList();
   }
 
+  estimatedDownlinkBandwidthLessThanRequired(videoBandwidthBudgetKbps: number, requiredBandwidth: number): void {
```
aburnson: Also need to add this to meetingV2.ts

PR #189, opened 2020-03-26
Recreate meeting for demo meeting ID if ended

aburnson:
*Issue #:* 

*Description of changes*

Recreate meeting for demo meeting ID if ended

### DIFF
--- a/demos/serverless/src/handlers.js
+++ b/demos/serverless/src/handlers.js
@@ -32,6 +32,13 @@ const getMeeting = async(meetingTitle) => {
     return null;
   }
   const meetingData = JSON.parse(result.Item.Data.S);
+  try {
+    await chime.getMeeting({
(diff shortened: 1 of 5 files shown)


demos/serverless/src/handlers.js
```diff
@@ -32,6 +32,13 @@ const getMeeting = async(meetingTitle) => {
     return null;
   }
   const meetingData = JSON.parse(result.Item.Data.S);
+  try {
```
ltrung: So from what I understand, chime.getMeeting with an existing ID of a meeting that already ended will return an error that we will catch and return null instead. Is that correct?
We have a integ test to verify this behavior where you cannot rejoin an ended meeting. Will this break that test?

aburnson:
Closing for now to address @ltrung 's comment

PR #192, opened 2020-03-27
Add notification and icons to component lib

p-foucht:
*Issue #:* n/a

*Description of changes*
Adds icons and notification to component library. Also sets up knobs for storybook usage.


demos/components/src/components/Notification/index.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+
+import { StyledNotification, StyledButton } from './Styled';
+import { Caution, Check, Information } from '../icons';
+import { Button } from '../Button';
+
+export type Severity = 'success' | 'warning' | 'error' | 'info';
+export type Size = 'sm' | 'md' | 'lg' | 'fill';
```
simmkyu: How about using enums for string types in this PR? e.g., `Severity`, `Direction`

```
# nit: Do not need to full this rule in the demo apps, but Chime SDK prefers full words over abbreviations.
enum Size {
  Small,
  Medium,
   ...
}
```

For example, using an enum value as a key is less error-prone.
```
const sizes = {
  [Size.Small]: '30rem',
  [Size.Medium]: '50rem',
  ...
};
```

demos/components/src/components/Notification/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+      size={size}
+      {...props}
+    >
+      <div className="icon">{icon}</div>
+      <div className="message">{children}</div>
+      {onClose && (
+        <Button
+          className="button"
+          addStyles={StyledButton}
+          onClick={() => onClose()}
```
simmkyu: Passing a new function inline will keep force `<Button>` to re-render even if a parent component passes a memoized callback. Either pass `onClose` directly or created a memorized callback.

demos/components/src/components/Notification/index.tsx
p-foucht: I chose types here mainly because it felt like the better developer experience, since you still get the type checking/autocomplete without the little extra boilerplate that comes with enums

Happy to switch to enums though if you think they have other advantages over types

demos/components/src/components/Notification/index.tsx
p-foucht: Good point, will update. Originally I felt that re-rendering would be fairly inexpensive here, but for best practice I should have done it that way anyways

demos/components/src/components/Notification/index.tsx
simmkyu: Right, enum props will require both JS and TS to import all enums. String types looks good to me.

PR #199, opened 2020-04-01
Fix issue with retry

ltrung:
*Description of changes*
- Fix an issue that causes retry to not happen. 
- Allow to select Browserstack as an alternate selenium provider for debugging.
- Fix Kite installation issue to checkout a specific commit before the break changes that cause Travis tests to fail for the last 2 weeks.

### DIFF
--- a/integration/js/utils/SdkBaseTest.js
+++ b/integration/js/utils/SdkBaseTest.js
@@ -1,5 +1,6 @@
 const {KiteBaseTest, TestUtils} = require('../node_modules/kite-common');
 const {SaucelabsSession} = require('./WebdriverSauceLabs');
+const {BrowserStackSession} = require('./WebdriverBrowserStack');
 const {emitMetric} = require('./CloudWatch');
 const uuidv4 = require('uuid/v4');
 const fs = require('fs');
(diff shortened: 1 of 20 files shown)


integration/js/utils/SdkBaseTest.js
```diff
@@ -90,19 +91,24 @@ class SdkBaseTest extends KiteBaseTest {
...
-        await new Promise(r => setTimeout(r, 1000));
-      } else {
-        console.log(`Successfully create a Saucelabs session: ${sessionId}`);
-        return session;
+    if (process.env.SELENIUM_GRID_PROVIDER === "browserstack") {
+      const session = await BrowserStackSession.createSession(capabilities);
+      return session;
+    } else {
+      const invalidSessionIdRegEx = new RegExp(/^new_request:/);
+      for (let i =0; i< 3; i++) {
```
RidipDe, line 99: nit: spaces

RidipDe: should the value 3 be not based of the retry parameter?

aburnson: > nit: spaces

We should enable linting on the `/integration` folder.

aburnson, line 99: We should enable linting on the `/integration` folder

PR #204, opened 2020-04-02
Test fail stop travis

ltrung:
*Description of changes*
Stop travis job if integration tests fail

### DIFF
--- a/integration/js/script/failure-check
+++ b/integration/js/script/failure-check
@@ -59,11 +59,17 @@ end
 
 puts "\nGenerating failure report\n"
 puts "========================================="
+isFailed = false
 resultFiles = Dir["./kite-allure-reports/*result.json"]
 resultFiles.each do |resultFile|
   file = File.read(resultFile)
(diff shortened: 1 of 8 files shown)


integration/js/script/failure-check
```diff
@@ -59,11 +59,17 @@ end
...
 puts "\nGenerating failure report\n"
 puts "========================================="
+isFailed = false
 resultFiles = Dir["./kite-allure-reports/*result.json"]
 resultFiles.each do |resultFile|
   file = File.read(resultFile)
   result = JSON.parse(file)
   if (result["status"]) == "FAILED"
+    isFailed = true
     explain_failure(result)
```
RidipDe, line 69: optional: why do we need a new variable 'isFailed'? we can do exit(1) after explain_failure(result)

ltrung: I dont want to exit(1) right way. I would like to make sure we print out all the failure instead of quit right away.

Issue #232, opened 2020-04-02
Make content share frame rate configurable


aburnson:

ltrung:
PR merged. Closed

PR #239, opened 2020-04-03
Allow content share frame rate to be configurable

ltrung:
*Issue #232 :* 

*Description of changes*
- Change the default max frame rate for content share to 15 fps (from 3fps).
- Add an API to allow to set the max frame rate.


src/contentsharecontroller/ContentShareControllerFacade.ts
```diff
@@ -14,6 +14,11 @@ export default interface ContentShareControllerFacade {
    */
   startContentShareFromScreenCapture(sourceId?: string): Promise<void>;
 
+  /**
+   * Set the max frame rate for screen capture. By default, the max frame rate is 15.
+   */
+  setScreenCaptureMaxFrameRate(frameRate: number): void;
```
aburnson, line 20: Instead, could we add a parameter to `startContentShareFromScreenCapture` like this
`startContentShareFromScreenCapture(sourceId?: string | null, frameRate?: number): Promise<void>;` ?

src/contentsharecontroller/ContentShareMediaStreamBroker.ts
```diff
@@ -56,24 +57,32 @@ export default class ContentShareMediaStreamBroker implements MediaStreamBroker
...
+    sourceId?: string,
+    frameRate?: number
+  ): MediaStreamConstraints {
     return {
       audio: false,
       video: {
         ...(!sourceId && {
           frameRate: {
-            max: 3,
+            max: frameRate ? frameRate : ContentShareMediaStreamBroker.defaultFrameRate,
```
simmkyu, line 78: Don't need to update in this PR but later you can just use ```||```.
``frameRate || ContentShareMediaStreamBroker.defaultFrameRate``

test/contentsharecontroller/ContentShareMediaStreamBroker.test.ts
```diff
@@ -96,43 +115,64 @@ describe('ContentShareMediaStreamBroker', () => {
...
+      const streamConstraints: MediaStreamConstraints = {
+        audio: false,
+        video: {
+          frameRate: {
+            max: 30,
+          },
+        },
+      };
+      const spy = sinon.spy(contentShareMediaStreamBroker, 'acquireDisplayInputStream');
+      await contentShareMediaStreamBroker.acquireScreenCaptureDisplayInputStream(null, 30);
```
simmkyu, line 157:
```
async acquireScreenCaptureDisplayInputStream(sourceId?: string, frameRate?: number)
```

The type of `sourceId` is implicitly `string | undefined` but not `null`. You can pass `undefined` instead.

simmkyu: Andrew suggested the `string | null` type for `startContentShareFromScreenCapture()`. You can update that signature instead of changing the test file.

Issue #216, opened 2020-04-02
Add builder-facing info logs and lower internal log level to debug


aburnson:
From #198

vidya-mohan:
Please see https://aws.github.io/amazon-chime-sdk-js/modules/meetingevents.html for an improved debugging procedure.

PR #260, opened 2020-04-11
Do not log DefaultAudioVideoFacade TRACE events at INFO level.

rnewman:
This commit introduces a file-scoped constant to control this kind of
logging; it seems oriented at SDK developers, not at builders using the
SDK.

It is not appropriate to log high-volume events at INFO level, so this
also switches to use `Logger.debug` instead of `Logger.info` for these
events.

*Issue #:* 

*Description of changes*

### DIFF
--- a/src/audiovideofacade/DefaultAudioVideoFacade.ts
+++ b/src/audiovideofacade/DefaultAudioVideoFacade.ts
@@ -16,6 +16,9 @@ import RealtimeController from '../realtimecontroller/RealtimeController';
 import VideoTile from '../videotile/VideoTile';
 import VideoTileController from '../videotilecontroller/VideoTileController';
 
+// Flip this to `true` if you want lots of TRACE-level output logged at DEBUG.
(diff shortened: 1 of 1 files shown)


aburnson:
Closing as these logs are currently necessary in order to troubleshoot SDK integrations. We [ask for INFO-level logs](https://github.com/aws/amazon-chime-sdk-js/blob/master/.github/ISSUE_TEMPLATE/bug-report.md) in the bug report issue template.

In the future we may consider adding an additional TRACE level, in which case we could move INFO->DEBUG and DEBUG->TRACE. Then INFO could be reserved for higher-level information #216. However, this change cannot be made without a major version bump, so will need to save it for v2.

What we could potentially do in a backwards-compatible way is to add a log tagging mechanism where builders integrating with logging could subscribe/unsubscribe from a set of logging tags. For example, logs could be tagged as `API`, `AudioVideoController`, `SessionLifecycle`, etc.

PR #265, opened 2020-04-14
Add flag to enable WebRTC Unified Plan for Chromium-based browsers

XHatan:
*Issue #:* 
 
*Description of changes*

- Add feature flag to enable WebRTC Unified Plan for Chromium-based browsers (default false)
- Change demo to use Unified Plan


src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -29,6 +29,7 @@ export default class DefaultBrowserBehavior implements BrowserBehavior {
   };
 
   private chromeLike: string[] = ['chrome', 'edge-chromium', 'chromium-webview', 'opera'];
+  constructor(private enableUnifiedPlanForChromiumBasedBrowsers: boolean = false) {}
```
simmkyu, line 32: Accepting a JS object is more scalable if parameters will have default values. For example, if the constructor takes in one more parameter later, we need to explicitly pass the value for the first parameter. (builders too)

You can change a signature to the following:

```
private enableUnifiedPlanForChromiumBasedBrowsers: boolean;

constructor({ enableUnifiedPlanForChromiumBasedBrowsers = false }: { enableUnifiedPlanForChromiumBasedBrowsers?: boolean } = {}) {
    this.enableUnifiedPlanForChromiumBasedBrowsers = enableUnifiedPlanForChromiumBasedBrowsers;
}
```

And pass an object (no change is required in this file):
```
new DefaultBrowserBehavior({
  enableUnifiedPlanForChromiumBasedBrowsers: true
})
```

XHatan: Good idea. I will make the change.

PR #268, opened 2020-04-14
Expose ExternalUserID on videoTileDidUpdate

RidipDe:
*Description of changes*
Expose ExternalUserId for this API:
https://github.com/aws/amazon-chime-sdk-js/blob/9720363625a16fb702b8edd53b8efaa3a63a5f03/test/videotilecontroller/DefaultVideoTileController.test.ts#L47


src/videotile/VideoTileState.ts
```diff
@@ -13,6 +13,7 @@ export default class VideoTileState {
   paused: boolean = false;
   poorConnection: boolean = false;
   boundAttendeeId: string | null = null;
+  boundExternalUserId: string | null = null;
```
simmkyu, line 16: Where do we set `boundExternalUserId`? I can't find changes in the previous PR.

PR #269, opened 2020-04-14
Ensure attendee presence leave/join is ordered correctly

aburnson:
*Issue #:* 

*Description of changes*

Ensure attendee presence leave/join is ordered correctly. The server may send old stream ids that are leaving after new stream ids that joining. When mapped to attendee presence this can appear as though the same attendee is joining and then leaving, when really they were rejoining.

### DIFF
--- a/test/volumeindicatoradapter/DefaultVolumeIndicatorAdapter.test.ts
+++ b/test/volumeindicatoradapter/DefaultVolumeIndicatorAdapter.test.ts
@@ -119,6 +119,53 @@ describe('DefaultVolumeIndicatorAdapter', () => {
       expect(volumeUpdate).to.equal(2);
       expect(attendeeIdUpdate).to.equal(2);
     });
+
+    it('handles out of order stream ids', () => {
+      const streamInfo1Join = SdkAudioStreamIdInfo.create();
+      const streamInfo1Frame = SdkAudioStreamIdInfoFrame.create();
+      const streamInfo2Join = SdkAudioStreamIdInfo.create();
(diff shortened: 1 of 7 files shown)


test/volumeindicatoradapter/DefaultVolumeIndicatorAdapter.test.ts
```diff
@@ -119,6 +119,53 @@ describe('DefaultVolumeIndicatorAdapter', () => {
...
+      const streamInfo2Leave = SdkAudioStreamIdInfo.create();
+      const streamInfo2Frame = SdkAudioStreamIdInfoFrame.create();
+      streamInfo1Join.audioStreamId = 1;
+      streamInfo1Join.attendeeId = fooAttendee;
+      streamInfo1Join.externalUserId = fooExternal;
+      streamInfo1Frame.streams = [streamInfo1Join];
+      streamInfo2Join.audioStreamId = 2;
+      streamInfo2Join.attendeeId = fooAttendee;
+      streamInfo2Join.externalUserId = fooExternal;
+      streamInfo2Leave.audioStreamId = 1;
```
ltrung, line 136: Minor: For the leave frame, do we only send audioStreamId without attendee Id and externalUserId? If not, could we set it here so it is easy to understand that this is the same user from streamInfo1.

aburnson: Leave frames are indicated by the absence of an attendee ID and external user ID, so that is intentional.

PR #275, opened 2020-04-16
Add an observer to retrieve raw stats from RTCPeerConnection

ltrung:
*Description of changes*
Allow retrieving raw stats from RTCPeerConnection


src/statscollector/DefaultStatsCollector.ts
```diff
@@ -420,6 +420,11 @@ export default class DefaultStatsCollector implements StatsCollector {
   }
 
   private handleRawMetricReports(rawMetricReports: RawMetricReport[]): void {
```
simmkyu, line 422: Although we are calling this param `rawMetricReports`, it's not untouched data from `rtcPeerConnection.getStats`.

From 453-460, SDK is re-assigning raw data to a specific object shape.
https://github.com/aws/amazon-chime-sdk-js/blob/f8c345b95472cc2aea26e7e399ca72f3523900d2/src/statscollector/DefaultStatsCollector.ts#L453

I wonder if a builder wants raw data directly from `rtcPeerConnection.getStats()`.

simmkyu:
> I think it's useful to simply expose `RTCPeerConnection` so developers can manage when and how they are going to call `getStats`.

In that case, we may need to add these observer methods to tell whether the peer connection object exists.
- peerConnectionDidBecomeAvailable
- peerConnectionDidBecomeUnavailable

ltrung:
> I think it's useful to simply expose `RTCPeerConnection` so developers can manage when and how they are going to call `getStats`.

@XHatan  That's how I originally did it. However, Andrew was against exposing internal object.

src/statscollector/DefaultStatsCollector.ts
ltrung: Good point. Updated.

PR #286, opened 2020-04-19
Expose ExternalUserID on videoTileDidUpdate

RidipDe:
*Description of changes*
- Expose ExternalUserId for this API:
https://github.com/aws/amazon-chime-sdk-js/blob/9720363625a16fb702b8edd53b8efaa3a63a5f03/test/videotilecontroller/DefaultVideoTileController.test.ts#L47
- Updated the demo app to reflect this change.

Testing Methodology:
- Written unit tests - success
- When the video is turned on, we are able to see the external user id associated to the attendee.


src/task/CleanStoppedSessionTask.ts
```diff
@@ -63,7 +63,7 @@ export default class CleanStoppedSessionTask extends BaseTask {
 
       const tile = this.context.videoTileController.getLocalVideoTile();
       if (tile) {
-        tile.bindVideoStream('', true, null, null, null, null);
+        tile.bindVideoStream('', true, null, null, null, null, '');
```
simmkyu, line 66: I think passing `null` for external user ID is better. For example, some customers might be using an empty string as a valid external user ID.

Note that the attendee ID (first parameter) can be an empty string since SDK owns it.

src/task/CreatePeerConnectionTask.ts
```diff
@@ -187,7 +194,23 @@ export default class CreatePeerConnectionTask extends BaseTask implements Remova
       width = cap.width as number;
       height = cap.height as number;
     }
-    tile.bindVideoStream(attendeeId, false, stream, width, height, streamId);
+    let externalUserId = this.context.realtimeController.realtimeExternalUserIdFromAttendeeId(
+      attendeeId
+    );
+    tile.bindVideoStream(attendeeId, false, stream, width, height, streamId, externalUserId);
+    if (!externalUserId) {
```
simmkyu, line 201: @aburnson @XHatan I suggested we handle a case when the tile is added without `externalUserId`.

1. For customers using external user ID, it's a race condition. A video track is added ***before*** the `DefaultVolumeIndicatorAdapter` retrieves the associated external user ID from an audio stream.
2. For customers not using external user ID, SDK will still subscribe to presence changes. We don't know whether a customer is using external user ID for remote attendees.

Two questions:
- First of all, does this implementation make sense?
- For case (2), can we just check if an attendee is not present yet? i.e. If an attendee becomes present without external ID, this attendee will never have external ID in the future.
  ```
  if (!isAttendeePresentAlready) {
  ```

src/realtimecontroller/DefaultRealtimeController.ts
```diff
@@ -317,6 +317,12 @@ export default class DefaultRealtimeController implements RealtimeController {
     });
   }
 
+  realtimeExternalUserIdFromAttendeeId(attendeeId: string): string | null {
+    if (attendeeId in this.state.attendeeIdToExternalUserId) {
```
aburnson, line 321: What does this function return if `attendeeId` is not `attendeeIdToExternalUserId`?

src/realtimecontroller/RealtimeController.ts
```diff
@@ -58,6 +58,11 @@ export default interface RealtimeController {
     callback: (attendeeId: string, present: boolean, externalUserId?: string | null) => void
   ): void;
 
+  /**
+   * Retrieve the externalUserId from the attendeeId-externalUserId map
+   */
+  realtimeExternalUserIdFromAttendeeId(attendeeId: string): string;
```
aburnson, line 64: Return type does not match implementation.

src/task/CreatePeerConnectionTask.ts
```diff
@@ -187,7 +194,26 @@ export default class CreatePeerConnectionTask extends BaseTask implements Remova
...
+      const handler = (
+        presentAttendeeId: string,
+        present: boolean,
+        externalUserId: string
+      ): void => {
+        if (attendeeId === presentAttendeeId && present && externalUserId !== null) {
+          tile.bindVideoStream(attendeeId, false, stream, width, height, streamId, externalUserId);
+          this.context.realtimeController.realtimeUnsubscribeToAttendeeIdPresence(handler);
+          this.presenceHandlerSet.delete(handler);
```
aburnson, line 210: This implementation assumes that the presence handler is going to be called. What happens to the attendee presence event does not arrive (e.g. connection is lost)? How will this handler be cleaned up in that case?

I would suggest setting a timer that will automatically delete the handler if it is not called after a certain amount of time (say 10 seconds).

src/task/CreatePeerConnectionTask.ts
aburnson: Attendees will always have an external user ID since that is a required field. I'm OK with the workaround below (with one comment).

src/task/CreatePeerConnectionTask.ts
RidipDe: Have added a TimeoutScheduler to removeHandlers after every 10 seconds.

demos/browser/app/meeting/meeting.ts
```diff
@@ -1128,7 +1128,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
     // TODO: enforce roster names
     new TimeoutScheduler(200).start(() => {
       const rosterName = this.roster[tileState.boundAttendeeId]
-        ? this.roster[tileState.boundAttendeeId].name
+        ? `${this.roster[tileState.boundAttendeeId].name}${tileState.boundExternalUserId ? ` [${tileState.boundExternalUserId}]` : ``}`
```
aburnson, line 1131: Why does the external user ID need to be visible in the roster? Won't that cause the roster width to be very large?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1185,7 +1185,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
     // TODO: enforce roster names
     new TimeoutScheduler(2000).start(() => {
       const rosterName = this.roster[tileState.boundAttendeeId]
-        ? this.roster[tileState.boundAttendeeId].name
+        ? `${this.roster[tileState.boundAttendeeId].name}${tileState.boundExternalUserId ? ` [${tileState.boundExternalUserId}]` : ``}`
```
aburnson, line 1188: Same question as above.

src/realtimecontroller/DefaultRealtimeController.ts
```diff
@@ -317,6 +317,13 @@ export default class DefaultRealtimeController implements RealtimeController {
...
```
aburnson, line 321: Is this `if` necessary? Won't the `|| null` below handle this case?

src/task/CreatePeerConnectionTask.ts
```diff
@@ -187,7 +199,31 @@ export default class CreatePeerConnectionTask extends BaseTask implements Remova
       width = cap.width as number;
       height = cap.height as number;
     }
-    tile.bindVideoStream(attendeeId, false, stream, width, height, streamId);
+    const externalUserId = this.context.realtimeController.realtimeExternalUserIdFromAttendeeId(
+      attendeeId
+    );
+    tile.bindVideoStream(attendeeId, false, stream, width, height, streamId, externalUserId);
+    if (!externalUserId) {
+      const timeoutScheduler = new TimeoutScheduler(10000);
```
aburnson, line 207: Value of timer needs to go into a class constant.

src/task/CreatePeerConnectionTask.ts
```diff
@@ -187,7 +199,31 @@ export default class CreatePeerConnectionTask extends BaseTask implements Remova
       width = cap.width as number;
       height = cap.height as number;
     }
-    tile.bindVideoStream(attendeeId, false, stream, width, height, streamId);
+    const externalUserId = this.context.realtimeController.realtimeExternalUserIdFromAttendeeId(
+      attendeeId
+    );
+    tile.bindVideoStream(attendeeId, false, stream, width, height, streamId, externalUserId);
+    if (!externalUserId) {
```
aburnson, line 206: Suggest moving the contents of this if-body into a separate function with a name that indicates what it is doing.

src/task/CreatePeerConnectionTask.ts
```diff
@@ -187,7 +199,31 @@ export default class CreatePeerConnectionTask extends BaseTask implements Remova
...
+          tile.bindVideoStream(attendeeId, false, stream, width, height, streamId, externalUserId);
+          this.context.realtimeController.realtimeUnsubscribeToAttendeeIdPresence(handler);
+          this.presenceHandlerSet.delete(handler);
+          timeoutScheduler.stop();
+        }
+      };
+      this.context.realtimeController.realtimeSubscribeToAttendeeIdPresence(handler);
+      this.presenceHandlerSet.add(handler);
+      timeoutScheduler.start(() => {
+        this.removeHandlers();
```
aburnson, line 223: Asymmetric lifecycle: before the timer begins we add one handler, once the timer fires, it removes all handlers. Shouldn't it just remove the one handler that was added.

demos/browser/app/meeting/meeting.ts
RidipDe: We decided in a standup to have the external id in the video tile as trial for few days.

src/task/CreatePeerConnectionTask.ts
```diff
@@ -187,7 +223,13 @@ export default class CreatePeerConnectionTask extends BaseTask implements Remova
       width = cap.width as number;
       height = cap.height as number;
     }
-    tile.bindVideoStream(attendeeId, false, stream, width, height, streamId);
+    const externalUserId = this.context.realtimeController.realtimeExternalUserIdFromAttendeeId(
+      attendeeId
+    );
+    tile.bindVideoStream(attendeeId, false, stream, width, height, streamId, externalUserId);
+    if (!externalUserId) {
```
simmkyu, line 230: Missed in the previous PR: You need to check `!== null`. Customers might use an empty string as a valid external user ID.

```
if (externalUserId !== null) {
```

Advanced: `realtimeExternalUserIdFromAttendeeId` can return `string | null | undefined`.
- `string`: both attendee ID and external user ID exists
- `null`: attendee ID exists but this attendee does not have external user ID
- `undefined`: attendee ID does not exist <-- Only in this case, we can get into this if statement.

```
if (typeof externalUserId === 'undefined') {
   ...
}
```

aburnson: @simmkyu I believe ExternalUserId is a required field that must be between 2-64 characters:
https://docs.aws.amazon.com/chime/latest/APIReference/API_CreateAttendee.html#API_CreateAttendee_RequestSyntax

src/realtimecontroller/RealtimeController.ts
```diff
@@ -58,6 +58,11 @@ export default interface RealtimeController {
     callback: (attendeeId: string, present: boolean, externalUserId?: string | null) => void
   ): void;
 
+  /**
+   * Retrieve the externalUserId from the attendeeId-externalUserId map
```
aburnson, line 62: The fact that it is store in a map is an internal detail that does not need to be mentioned in the API method. Suggest: "Returns the external user ID for a given attendee ID"

src/task/CreatePeerConnectionTask.ts
```diff
@@ -24,7 +29,12 @@ export default class CreatePeerConnectionTask extends BaseTask implements Remova
   ];
   private removeVideoTrackEventListeners: { [trackId: string]: (() => void)[] } = {};
 
-  constructor(private context: AudioVideoControllerState) {
+  static readonly removeHandlerIntervalMs: number = 10000;
```
aburnson, line 32: We may not be perfectly consistent about this, but in general we use ALL_CAPS_CASE when defining constants to make it clear that they are constants, e.g.

https://github.com/aws/amazon-chime-sdk-js/blob/master/src/activespeakerdetector/DefaultActiveSpeakerDetector.ts#L16

src/task/CreatePeerConnectionTask.ts
aburnson: Also, when the external user id is read off the wire, it is converted into either a non-empty string or null here:
https://github.com/aws/amazon-chime-sdk-js/blob/efa5f360acc20793cf3993be5099264206e3327b/src/volumeindicatoradapter/DefaultVolumeIndicatorAdapter.ts#L39

src/videotile/DefaultVideoTile.ts
```diff
@@ -131,7 +131,8 @@ export default class DefaultVideoTile implements DevicePixelRatioObserver, Video
     mediaStream: MediaStream | null,
     contentWidth: number | null,
     contentHeight: number | null,
-    streamId: number | null
+    streamId: number | null,
+    externalUserId?: string | null
```
aburnson, line 135: Why is this parameter optional?

src/task/CreatePeerConnectionTask.ts
```diff
@@ -187,7 +226,13 @@ export default class CreatePeerConnectionTask extends BaseTask implements Remova
       width = cap.width as number;
       height = cap.height as number;
     }
-    tile.bindVideoStream(attendeeId, false, stream, width, height, streamId);
+    const externalUserId = this.context.realtimeController.realtimeExternalUserIdFromAttendeeId(
+      attendeeId
+    );
+    tile.bindVideoStream(attendeeId, false, stream, width, height, streamId, externalUserId);
```
aburnson, line 232: If `externalUserId` is `null` won't this cause the video tile update to not include a non-empty externalUserId? Isn't the goal to ensure that there is one when the tile is bound so that callbacks always have the externalUserId?

src/realtimecontroller/RealtimeController.ts
RidipDe: changed the comment

src/task/CreatePeerConnectionTask.ts
RidipDe: done

src/videotile/DefaultVideoTile.ts
RidipDe: To make sure that the API already being used by customers do not break.

src/task/CreatePeerConnectionTask.ts
RidipDe: if ecternalUserId is null, binding will take place twice

simmkyu:
```
if (externalUserId) {
  tile.bindVideoStream(attendeeId, false, stream, width, height, streamId, externalUserId);
} else {
  this.bindExternalIdToRemoteTile(tile, attendeeId, stream, width, height, streamId);
}
```

aburnson: @RidipDe still need to implement above logic ^

PR #288, opened 2020-04-20
Expose an API for GetStats from RTCPeerConnection

ltrung:
*Description of changes*
Expose an API for GetStats from RTCPeerConnection


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add post publish script
 - Add feature flag to enable WebRTC Unified Plan for Chromium-based browsers
 - Add link to Amazon Chime SDK Security in README
+- Expose an API for GetStats from RTCPeerConnection
```
simmkyu, line 29: Move it to a new "Unreleased" section

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -148,6 +148,13 @@ export default class DefaultAudioVideoController implements AudioVideoController
     return this._mediaStreamBroker;
   }
 
+  async getRTCPeerConnectionStats(selector?: MediaStreamTrack): Promise<RTCStatsReport> {
```
simmkyu, line 151: nit: `async` is not required since it doesn't `await` for any function but just return a promise.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -148,6 +148,13 @@ export default class DefaultAudioVideoController implements AudioVideoController
...
+    if (!this.rtcPeerConnection) {
+      return null;
+    }
+    return this.rtcPeerConnection.getStats(selector);
```
simmkyu, line 155: In DefaultStatsCollector, `getStats` takes in success and failure callbacks for Chromium WebRTC instead of using a Promise.
https://github.com/aws/amazon-chime-sdk-js/blob/master/src/statscollector/DefaultStatsCollector.ts#L447

Could you make sure supported browser versions have a promise-based `getStats`?

Here, we just return

CHANGELOG.md
ltrung: Fixed.

src/audiovideocontroller/DefaultAudioVideoController.ts
ltrung: Removed.

src/audiovideocontroller/DefaultAudioVideoController.ts
ltrung: That is what I thought at first but after talking to consumers, they do not need the legacy stats from Chrome. Chome currently also supports the new getStats that return Promise.

simmkyu: Confirmed with Trung that Chrome 58 started to support a promise-based callback

Issue #287, opened 2020-04-20
getFloatTimeDomainData function undefined on ios Devices for Audio preview 


JannikSt:
**Describe the bug**
As demonstrated in the browser demo app, we utilize the getFloatTimeDomainData function to show the audio input preview in our current webapp when the user selects their devices. This works fine on Android and on the supported desktop browsers. Unfortunately, the function remains undefined on ios devices, thus no inputaudiopreview can be shown. 

Is there something I'm missing here? 

I appreciate your help. 

**Logs** 
[...] 
[Info] 2020-04-20T13:50:01.427Z [INFO] SDK - API/DefaultAudioVideoFacade/232d5f5c-3539-4859-8c33-fbab2443caca/43c9d898-7795-4cba-890b-26b26b3d016a/createAnalyserNodeForAudioInput

[Info] 2020-04-20T13:50:01.429Z [INFO] SDK - requesting new audio device with constraint [object Object] 

[Info] 2020-04-20T13:50:01.430Z [INFO] SDK - API/DefaultAudioVideoFacade/232d5f5c-3539-4859-8c33-fbab2443caca/43c9d898-7795-4cba-890b-26b26b3d016a/chooseAudioInputDevice "523A254EE29807796C70E38C5F678C1340494703"
[Info] 2020-04-20T13:50:02.510Z [INFO] SDK - got audio device for constraints {"audio":{"deviceId":{"exact":"523A254EE29807796C70E38C5F678C1340494703"},"echoCancellation":true,"googEchoCancellation":true,"googAutoGainControl":true,"googNoiseSuppression":true,"googHighpassFilter":true,"googNoiseSuppression2":true,"googEchoCancellation2":true,"googAutoGainControl2":true}} 

[Info] 2020-04-20T13:50:02.511Z [INFO] SDK - checking whether existing audio device can be reused 

[Info] 2020-04-20T13:50:02.511Z [INFO] SDK - reusing existing audio device 

[Info] 2020-04-20T13:50:02.512Z [INFO] SDK - cannot replace audio track due to: no active meeting and peer connection 

[Info] 2020-04-20T13:50:02.512Z [INFO] SDK - API/DefaultDeviceController/chooseAudioInputDevice "523A254EE29807796C70E38C5F678C1340494703" -> "PermissionGrantedByUser" 

[Info] 2020-04-20T13:50:02.873Z [INFO] SDK - API/DefaultDeviceController/createAnalyserNodeForAudioInput 

[...]

PR #294, opened 2020-04-21
Use getByteTimeDomainData to support iOS Safari in meeting demo

aburnson:
*Issue #:* #287 

*Description of changes*

Use getByteTimeDomainData to support iOS Safari in meeting demo

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -958,19 +958,19 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
     if (!analyserNode) {
       return;
     }
-    if (!analyserNode.getFloatTimeDomainData) {
+    if (!analyserNode.getByteTimeDomainData) {
       document.getElementById('audio-preview').parentElement.style.visibility = 'hidden';
       return;
     }
-    const data = new Float32Array(analyserNode.fftSize);
+    const data = new Uint8Array(analyserNode.fftSize);
     let frameIndex = 0;
     this.analyserNodeCallback = () => {
       if (frameIndex === 0) {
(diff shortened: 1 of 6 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -962,15 +962,15 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
       document.getElementById('audio-preview').parentElement.style.visibility = 'hidden';
```
ltrung, line 962: Do we need to update this to getByteTimeDomainData too? Also does this check work?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -962,15 +962,15 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
...
     let frameIndex = 0;
     this.analyserNodeCallback = () => {
       if (frameIndex === 0) {
-        analyserNode.getFloatTimeDomainData(data);
+        analyserNode.getByteTimeDomainData(data);
         const lowest = 0.01;
         let max = lowest;
         for (const f of data) {
-          max = Math.max(max, Math.abs(f));
+          max = Math.max(max, (f - 128) / 128);
```
ltrung, line 973: Could you explain more about this logic?

demos/browser/app/meetingV2/meetingV2.ts
aburnson: Good call. Thanks.

demos/browser/app/meetingV2/meetingV2.ts
aburnson: This math normalizes the values. Byte time domain data is stored 0-255 where 0...128...256 would be equivalent to -1...0...1.

Issue #277, opened 2020-04-16
demos/serverless: example is incompatible with Windows 10 Command Prompt; not documented


swoldemi:
Example in question lives here: https://github.com/aws/amazon-chime-sdk-js/tree/master/demos/serverless

**Important:** Please attach INFO-level logs with timestamps to the report from *before* the session started up until the point at which the unexpected behavior occured. For multi-participant calls, attach a log file for each participant if possible.
```
C:\Users\me\Projects\amazon-chime-sdk-js\demos\serverless>  node ./deploy.js -r us-east-1 -b my-bucket -s testing-stack -a meeting
,aws-cli/1.16.198 Python/3.6.0 Windows/10 botocore/1.12.188
,
,SAM CLI, version 0.47.0
,
Verifying application meeting
Application ../browser/dist/meeting.html does not exist. Rebuilding demo apps
Command npm failed with ENOENT <---- This shouldn't happen
C:\Users\me\Projects\amazon-chime-sdk-js\demos\serverless> where npm
C:\Program Files\nodejs\npm
C:\Program Files\nodejs\npm.cmd
```

**Meeting and attendee IDs:**
N/A

**Timeline**
Outline the main actions the user performed and when:
See Command Prompt output above.

**Describe the bug**
Describe the bug and the timestamp at which it happened relative to the timestamp in the log file.
See Command Prompt output above.
`deploy.js` is not considerate of Windows Command Prompt users. For [`child_process.spawnSync`](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options) to work on Windows, it seems some additional parameters should be passed. This happens transiently across different commands (that DO exist in PATH), but is sometimes solved by appending the `.cmd` suffix on the command. (e.g. `npm.cmd` will not result in ENOENT). In any case, running `node ./deploy.js -r us-east-1 -b mybucket -s stackname -a appname` to deploy the application will still fail because `cp` is not a native Command Prompt resource: https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/serverless/deploy.js#L137-L142

**Screenshots**
If applicable, add screenshots to help explain the problem.
N/A

**Platform**
 - OS :Microsoft Windows 10 [Version 10.0.18362.720]
 - Browser:  N/A
 - Browser Version: N/A 
 - Amazon Chime SDK GitHub Commit: f8c345b95472cc2aea26e7e399ca72f3523900d2

[...]

PR #307, opened 2020-04-27
Fix serverless deploy script to work on Windows

ltrung:
*Issue #277 :* 

*Description of changes*
Fix the current serverless deploy script to work on Windows:
- Add options shell= true in SpawnSync to make sure shell commands are run in a shell. This fix the issue where commands are not found in Windows.
- Use copy file from fs-extra package instead of os-dependent copy command. Previously, we use cp which does not exist in Windows.

Test by manually deploying in both Mac and Windows.

### DIFF
--- a/demos/serverless/README.md
+++ b/demos/serverless/README.md
@@ -23,7 +23,7 @@ API Gateway deployment that runs the `meeting` demo.
 
 ```
 cd demos/serverless
-node ./deploy.js -r us-east-1 -b <my-bucket> -s <my-stack-name> -a meeting
+npm run deploy -- -r us-east-1 -b <my-bucket> -s <my-stack-name> -a meeting
 ```
 
 The script will create an S3 bucket and CloudFormation stack
--- a/demos/serverless/deploy.js
+++ b/demos/serverless/deploy.js
@@ -1,5 +1,5 @@
 const { spawnSync } = require('child_process');
-const fs = require("fs");
+const fs = require('fs-extra');
 const path = require("path");
 
 // Parameters
@@ -80,6 +80,10 @@ function parseArgs() {
 }
 
 function spawnOrFail(command, args, options) {
+  options = {
+    ...options,
+    shell: true
+  };
   const cmd = spawnSync(command, args, options);
   if (cmd.error) {
     console.log(`Command ${command} failed with ${cmd.error.code}`);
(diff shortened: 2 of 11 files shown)


demos/serverless/README.md
```diff
@@ -23,7 +23,7 @@ API Gateway deployment that runs the `meeting` demo.
 
 ```
 cd demos/serverless
-node ./deploy.js -r us-east-1 -b <my-bucket> -s <my-stack-name> -a meeting
+npm run deploy -r us-east-1 -b <my-bucket> -s <my-stack-name> -a meeting
```
simmkyu, line 26: Make sure to change other files running `npm ./deploy.js`

https://github.com/aws/amazon-chime-sdk-js/blob/a8eac1cc2ac7fe66f3a72ef2c9aa04da761beb41/script/deploy-canary-demo#L11

https://github.com/aws/amazon-chime-sdk-js/blob/a8eac1cc2ac7fe66f3a72ef2c9aa04da761beb41/script/deploy-canary-demo#L18

https://github.com/aws/amazon-chime-sdk-js/blob/a8eac1cc2ac7fe66f3a72ef2c9aa04da761beb41/script/postpublish#L51

demos/serverless/deploy.js
```diff
@@ -1,5 +1,5 @@
 const { spawnSync } = require('child_process');
-const fs = require("fs");
+const fs = require('fs-extra')
```
simmkyu, line 2: missing `;`

demos/serverless/deploy.js
```diff
@@ -80,6 +80,11 @@ function parseArgs() {
 }
 
 function spawnOrFail(command, args, options) {
+  if (options) {
```
simmkyu, line 83: How about assigning since `shell` will always be `true`?

```
options = {
  ...options,
  shell: true
};
```

demos/serverless/README.md
ltrung: Good catch. Updated.

demos/serverless/deploy.js
ltrung: Fixed.

demos/serverless/deploy.js
ltrung: Nice. Updated

Issue #319, opened 2020-04-30
Local video playback freezes when stopping/starting video on Safari


jeffmerten:
**Meeting and attendee IDs:**
N/A

**Timeline**
Outline the main actions the user performed and when:
• Open meetingV2 example.
• Go through settings and into video conference and enable the video so you see your own playback.
• Disable the video feed
• Re-enable the video feed

You'll see the video tile for your own playback, but now it's frozen.

**Describe the bug**
Our application has a video toggle button that lets our users turn off and on their video. When this is toggled off we call `stopLocalVideoTile()` and `chooseVideoInputDevice(null)`. When it is toggled on we call `chooseVideoInputDevice(deviceId)` and then `startLocalVideoTile()` after the promise has been resolved.

The latter appears to resume the video playback for us but causes the video playback to freeze. This only happens on safari and mobile safari. We also haven't seen any warn or error logs logged from inside amazon-chime-sdk-js.

It appears this can also be reproduced in the meetingV2 example.

**Screenshots**
N/A

**Platform**
 - OS: MacOS
 - Browser: Safari
 - Browser Version: 13.0.5
 - Amazon Chime SDK GitHub Commit: d658830a1f1d151c12a9fb89e371984bea3f9ebf (April 29th, 2020)

bizready:
I have a similar issue. I can toggle on/off video and LED provides confirmation of on/off but I do not see Video display on the tile.

dinesh-cm:
We do have the same issue for our users opening the meeting URL in iOS browser.

ltrung:
Thank you for reporting. The issue has been fixed in this commit https://github.com/aws/amazon-chime-sdk-js/commit/d3964b8631baed7b23167a686b1063648c764995
@bizready and @dinesh-cm, I think they maybe different issues. Could you open a new issue?

Issue #157, opened 2020-03-10
Fix the video tile unbinding issue


simmkyu:
### Problem

From [DefaultVideoTile.disconnectVideoStreamFromVideoElement()](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/videotile/DefaultVideoTile.ts#L75):

```
// Need to yield the message loop before clearing `srcObject` to
// prevent Safari from crashing.
new AsyncScheduler().start(() => {
  videoElement.srcObject = null;
});
```

`DefaultVideoTile.disconnectVideoStreamFromVideoElement()` is setting `videoElement.srcObject` to `null` after a slight delay. However, `disconnectVideoStreamFromVideoElement()` and all callers, such as `VideoTileController.unbindVideoElement()`-are not async functions.

See this example:

```
// Assume that a <video> element "element" is bound to oldTileId.
// Unbind "element" asynchronously.
VideoTileController.unbindVideoElement(oldTileId);

// Bind "element" synchronously.
VideoTileController.bindVideoElement(newTileId, element);

// After binding "element," "element.srcObject" is set to null.
```

#### Option 1 (preferred):
Set `videoElement.srcObject` to `null` synchronously. We need to investigate why this can cause Safari to crash.

#### Option 2:
Make all callers to `async`. This requires significant changes that should be related to `2.0.0`.

aburnson:
Fixed in #325

PR #324, opened 2020-05-01
Fix video freeze in Safari after toggling off and on

ltrung:
**Issue #319 and #157 :** 

**Description of changes:**
- Switch back to use requestAnimationFrame instead of AsyncScheduler to fix the local video freeze in Safari.
- Check the video element's srcObject before clearing it out to make sure the video element's srcObject has not been overriden.

**Testing**

Yes
- Add unit tests
- Manually testing in Safari and verified that the local video is not frozen after toggling off then on again.
- Retest the issue #3 and verify there is no crash when remote video is added or toggled.
- Retest the fix in #63 and verify the following case is working:
  
    - In browser window 1, join A and enable video
    - In browser window 2, join B and enable video
    - Confirm both windows show local/remote videos
    - In browser window 1, open a new tab (e.g. example.com) and leave focused
    - In browser window 2, toggle B video off and on
    - Switch back to tab in browser window 1 with participant A
    - Video appears (and not empty)

### DIFF
--- a/src/videotile/DefaultVideoTile.ts
+++ b/src/videotile/DefaultVideoTile.ts
@@ -4,7 +4,6 @@
 import DevicePixelRatioMonitor from '../devicepixelratiomonitor/DevicePixelRatioMonitor';
 import DevicePixelRatioObserver from '../devicepixelratioobserver/DevicePixelRatioObserver';
 import DefaultModality from '../modality/DefaultModality';
-import AsyncScheduler from '../scheduler/AsyncScheduler';
 import VideoTileController from '../videotilecontroller/VideoTileController';
 import VideoTile from './VideoTile';
 import VideoTileState from './VideoTileState';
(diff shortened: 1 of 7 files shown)


src/videotile/DefaultVideoTile.ts
```diff
@@ -78,10 +78,15 @@ export default class DefaultVideoTile implements DevicePixelRatioObserver, Video
...
-      new AsyncScheduler().start(() => {
-        videoElement.srcObject = null;
+      const clearSrcCallBack = ((originalSource: MediaStream | MediaSource | Blob | null) => {
+        return () => {
+          if (videoElement.srcObject === originalSource) {
+            videoElement.srcObject = null;
+          }
+        };
+      })(videoElement.srcObject);
+      requestAnimationFrame(() => {
```
justindarc, line 88: Have you tested what happens when the browser tab is backgrounded? For power savings reasons, `requestAnimationFrame` does not tick when the page is hidden (like when you switch tabs in the browser).”

edwardbc:
This is to be happening on Firefox as well. Video start, stop and start from another user will freeze video on Firefox. 
```videoElement.srcObject = null;```
Doesn't get executed in this case.

ltrung:
> This is to be happening on Firefox as well. Video start, stop and start from another user will freeze video on Firefox.
> `videoElement.srcObject = null;`
> Doesn't get executed in this case.

@edwardbc Could you please open a new issue with steps to reproduce in Firefox so we could track it?

PR #331, opened 2020-05-03
Add more content sharing integration tests

ltrung:
**Description of changes:**
- When retrying integration tests, end the current Sauce Labs session and start a new one. This ensures we have a clean state every time.
- Adding 3 new test cases for Content Share:
  - If an attendee joins later, verify that they can see content share.
  - If the sharer leaves the meeting, verify that the content share also ends.
  - If the sharer closes the app, verify that the content share also ends.
**Testing**

- Manually run the integration tests and verify that tests pass.

### DIFF
--- a/integration/js/utils/SdkBaseTest.js
+++ b/integration/js/utils/SdkBaseTest.js
@@ -118,12 +118,30 @@ class SdkBaseTest extends KiteBaseTest {
     }
   }
 
+  async initializeSeleniumSession(numberOfSeleniumSessions) {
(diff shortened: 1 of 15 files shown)


integration/js/utils/SdkBaseTest.js
```diff
@@ -241,6 +230,24 @@ class SdkBaseTest extends KiteBaseTest {
       await this.seleniumSessions[i].quit();
     }
   }
+
+  async printRunDetails(testResult) {
```
RidipDe, line 234: nit: can we change the method name to something like "sessionRunDetails" to differentiate from the printRunDetails() in the WebdriverSauceLabs.js which is actually logging the session details ?

PR #334, opened 2020-05-04
Add banners to read me file to show latest npm version and downloads

RidipDe:
**Description of changes:**
Add gifs to read me file to show latest npm version and downloads

**Testing**

Yes
Change to the readme file, reflected in the read me window.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 # Amazon Chime SDK for JavaScript
 
+<a href="https://www.npmjs.com/package/amazon-chime-sdk-js"><img src="https://img.shields.io/npm/v/amazon-chime-sdk-js?style=flat-square"></a>
+<a href="https://travis-ci.org/github/aws/amazon-chime-sdk-jsr"><img src="https://travis-ci.org/aws/amazon-chime-sdk-js.svg?branch=master"></a>
+
 ### Build video calling, audio calling, and screen sharing applications powered by Amazon Chime.
 
 The Amazon Chime SDK makes it easy to add collaborative audio calling,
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -1,5 +1,9 @@
 # Amazon Chime SDK for JavaScript
 
+<a href="https://www.npmjs.com/package/amazon-chime-sdk-js"><img src="https://img.shields.io/npm/v/amazon-chime-sdk-js?style=flat-square"></a>
+<a href="https://travis-ci.org/github/aws/amazon-chime-sdk-jsr"><img src="https://travis-ci.org/aws/amazon-chime-sdk-js.svg?branch=master"></a>
+<a href="https://www.npmjs.com/package/amazon-chime-sdk-js"><img src="https://img.shields.io/npm/dm/amazon-chime-sdk-js?style=flat-square"></a>
```
ltrung, line 5: Thought we decided not to have the download info anymore

README.md
```diff
@@ -1,5 +1,8 @@
 # Amazon Chime SDK for JavaScript
 
+<a href="https://www.npmjs.com/package/amazon-chime-sdk-js"><img src="https://img.shields.io/npm/v/amazon-chime-sdk-js?style=flat-square"></a>
+<a href="https://travis-ci.org/github/aws/amazon-chime-sdk-jsr"><img src="https://travis-ci.org/aws/amazon-chime-sdk-js.svg?branch=master"></a>
+
```
ltrung, line 5: Isn't this Travis just info from the build when a PR is merged to master? If so it's not very useful as we only do deployment. Also if there is no demo change, the deployment script will return an error and Travis will show as failure. This will be confusing to show

aburnson: @ltrung 

> Isn't this Travis just info from the build when a PR is merged to master? If so it's not very useful as we only do deployment.

I think this is OK. In order for the PR to merge it must successfully pass the unit and integration tests. It might be worth having the merge to master also re-run `npm run build:release` since that is relatively quick.

> Also if there is no demo change, the deployment script will return an error and Travis will show as failure. This will be confusing to show

Could you clarify why this happens? If that's the case we should try to fix this in the Travis config.

PR #338, opened 2020-05-05
Add method to get the nearest media region

RidipDe:
**Description of changes:**
https://nearest-media-region.l.chime.aws returns the nearest AWS region that can host a meeting. We use this publicly available API, to select the region returned as the default media region in the demo app.

**Testing**

Yes
Selects us-west-2 as the closest media region
https://ciavhkpz27.execute-api.us-east-1.amazonaws.com/Prod/v2/

### DIFF
--- a/demos/browser/app/meeting/meeting.ts
+++ b/demos/browser/app/meeting/meeting.ts
@@ -144,6 +144,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver 
       "amazon-chime-sdk-js@" + Versioning.sdkVersion;
     this.initEventListeners();
     this.initParameters();
+    this.getNearestMediaRegion();
   }
 
   initParameters(): void {
@@ -472,6 +473,30 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver 
         window.location = window.location.pathname;
       });
     });
+  }
+
+  getNearestMediaRegion(): void {
+    const supportedMediaRegions: Array<string> = ['ap-northeast-1', 'ap-southeast-1', 'ap-southeast-2', 'ca-central-1', 'eu-central-1', 'eu-north-1', 'eu-west-1', 'eu-west-2', 'eu-west-3', 'sa-east-1', 'us-east-1', 'us-east-2', 'us-west-1', 'us-west-2'];
+    new AsyncScheduler().start(
+      async (): Promise<void> => {
+        try {
+          const nearestMediaRegionResponse = await fetch(
+            `https://nearest-media-region.l.chime.aws`,
+            {
+              method: 'GET',
+            }
+          );
+          const nearestMediaRegionJSON = await nearestMediaRegionResponse.json();
+          const nearestMediaRegion = nearestMediaRegionJSON.region;
+          if (supportedMediaRegions.indexOf(nearestMediaRegion) !== -1) {
+            (document.getElementById('inputRegion') as HTMLInputElement).value = nearestMediaRegion;
+          } else {
+            throw new Error(`${nearestMediaRegion} doesn't exist`);
+          }
+        } catch (error) {
+          this.log('Default media region selected: ' + error.message);
+        }
+      });
   }
 
   toggleButton(button: string, state?: 'on' | 'off'): boolean {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add bandwidth policy to meeting session configuration to allow overriding default policies
 - Add more content sharing integration tests
 - Add gifs to read me file to show latest npm version and downloads
+- Add method to get the nearest media region
 
 ### Changed
 - Simplify meeting demos to leverage externalUserId in roster
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -150,6 +150,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
       "amazon-chime-sdk-js@" + Versioning.sdkVersion;
     this.initEventListeners();
     this.initParameters();
+    this.getNearestMediaRegion();
     if (this.isRecorder()) {
       new AsyncScheduler().start(async () => {
         this.meeting = new URL(window.location.href).searchParams.get('m');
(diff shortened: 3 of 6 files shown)


demos/browser/app/meeting/meeting.ts
```diff
@@ -474,6 +475,32 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
...
+          const nearestMediaRegionJSON = await nearestMediaRegionResponse.json();
+          const nearestMediaRegion = nearestMediaRegionJSON.region;
+          if(nearestMediaRegion !== null || nearestMediaRegion !== '') {
+            (document.getElementById('inputRegion') as HTMLInputElement).value = nearestMediaRegion;
+          }
+          else {
+            (document.getElementById('inputRegion') as HTMLInputElement).value = defaultNearestMediaRegion;
+          }
+        } catch (error) {
+          (document.getElementById('inputRegion') as HTMLInputElement).value = defaultNearestMediaRegion;
```
mikeantonelli, line 499: It feels a bit odd to inform the client that `us-east-1` is the closest region and silently swallow an error. Could the client set a `defaultNearestMediaRegion` on setup? Could you at least log an error/warning here, and on line 496, that informs the client a issue occurred and a default will be returned?

RidipDe: Thank you @mikeantonelli for your comment.
The getNearestMediaRegion() will GET the nearest media region as a fetch call to https://nearest-media-region.l.chime.aws and set the default option in the inputRegion drop down menu to the region being returned by this call. If the call to the API fails by any chance, the "us-east-1" will be selected as the default option as the closest region in the dropdown. Till this point, we are just 'selecting' the nearest media region. Meetings are not created/joined until the join button is clicked. So the inputRegion option can be changed by the client manually before clicking the join button to create/join the meeting.

mikeantonelli: Thanks for the clarification

demos/browser/app/meeting/meeting.ts
```diff
@@ -474,6 +475,32 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
...
+      async (): Promise<void> => {
+        this.showProgress('progress-authenticate');
+        try {
+          const nearestMediaRegionResponse = await fetch(
+            `https://nearest-media-region.l.chime.aws`,
+            {
+              method: 'GET',
+            }
+          );
+          const nearestMediaRegionJSON = await nearestMediaRegionResponse.json();
```
ltrung, line 490: the join button event handler created in initEventListeners is also an aysnc job which reads from inputRegion. So if this job for some reason completes after the users click Join, it would not work. Maybe its fine and we can just use the default region in that case. What do you think @simmkyu

demos/browser/app/meeting/meeting.ts
```diff
@@ -474,6 +475,32 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
...
+          const nearestMediaRegion = nearestMediaRegionJSON.region;
+          if(nearestMediaRegion !== null || nearestMediaRegion !== '') {
+            (document.getElementById('inputRegion') as HTMLInputElement).value = nearestMediaRegion;
```
ltrung, line 493: We hard-coded the list of available region so you may want to check whether this nearest media region is in that list.

RidipDe: If the list of media regions change in the future, the demo app might not reflect the change unless we hard-code and update the list within the demo app.

ltrung: Sure but alternatively, you set a value that is not in the list (UI bug). A better way is to fetch the list of media regions and the nearest media region and populate the list instead of hard coding it.

demos/browser/app/meeting/meeting.ts
ltrung: We already set default region to use-east-1 so this is unnecessary. Also we should log error if the fetch failed.

demos/browser/app/meeting/meeting.ts
```diff
@@ -474,6 +475,32 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
     });
   }
 
+  getNearestMediaRegion(): void {
+    const supportedMediaRegions: Array<string> = ['ap-northeast-1', 'ap-southeast-1', 'ap-southeast-2', 'ca-central-1', 'eu-central-1', 'eu-north-1', 'eu-west-1', 'eu-west-2', 'eu-west-3', 'sa-east-1', 'us-east-1', 'us-east-2', 'us-west-1', 'us-west-2'];
+    const defaultNearestMediaRegion = "us-east-1";
```
simmkyu, line 480: nit: Use single quotes here and in line 485 to match other strings

demos/browser/app/meeting/meeting.ts
```diff
@@ -474,6 +475,32 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
...
+    new AsyncScheduler().start(
+      async (): Promise<void> => {
+      try {
```
simmkyu, line 483: Fix indentations in this file and `meetingV2.ts`

demos/browser/app/meeting/meeting.ts
```diff
@@ -474,6 +475,32 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
...
+        const nearestMediaRegionResponse = await fetch(
+          `https://nearest-media-region.l.chime.aws`,
+          {
+            method: 'GET',
+          }
+        );
+        const nearestMediaRegionJSON = await nearestMediaRegionResponse.json();
+        const nearestMediaRegion = nearestMediaRegionJSON.region;
+        if(nearestMediaRegion !== null || nearestMediaRegion !== '' || supportedMediaRegions.indexOf(nearestMediaRegion) != -1) {
```
simmkyu, line 492: - Use a strict inequality `!== -1`
- Add a space `if (`
- You can just do the following since `[].indexOf(null)` and `[].indefOf('')` will be `-1`.

How about this?
   ```
   if (supportedMediaRegions.indexOf(nearestMediaRegion) !== -1) {
      ...
   } else {
      throw new Error(`${nearestMediaRegion} doesn't exist`);
   }
   ```

demos/browser/app/meeting/meeting.ts
```diff
@@ -474,6 +475,32 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
...
+          (document.getElementById('inputRegion') as HTMLInputElement).value = nearestMediaRegion;
+        }
```
simmkyu, line 494: I understand that the demo is not auto-formatted with ESLint and Prettier. Please follow the convention in this file.
`} else {`

demos/browser/app/meeting/meeting.ts
```diff
@@ -474,6 +475,34 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
     });
   }
 
+  getNearestMediaRegion(): void {
+    const supportedMediaRegions: Array<string> = ['ap-northeast-1', 'ap-southeast-1', 'ap-southeast-2', 'ca-central-1', 'eu-central-1', 'eu-north-1', 'eu-west-1', 'eu-west-2', 'eu-west-3', 'sa-east-1', 'us-east-1', 'us-east-2', 'us-west-1', 'us-west-2'];
+    const defaultNearestMediaRegion = "us-east-1";
+    new AsyncScheduler().start(
+      async (): Promise<void> => {
+      try {
+        const nearestMediaRegionResponse = await fetch(
```
simmkyu, line 484: From @ltrung:
> the join button event handler created in initEventListeners is also an aysnc job which reads from inputRegion. So if this job for some reason completes after the users click Join, it would not work. Maybe its fine and we can just use the default region in that case. What do you think @simmkyu

I think it should be fine to use the default region since it's a demo app. In production, we'll enable the "Join" button only after regions are selected.

demos/browser/app/meeting/meeting.ts
```diff
@@ -474,6 +475,30 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
     });
   }
 
+  getNearestMediaRegion(): void {
+    const supportedMediaRegions: Array<string> = ['ap-northeast-1', 'ap-southeast-1', 'ap-southeast-2', 'ca-central-1', 'eu-central-1', 'eu-north-1', 'eu-west-1', 'eu-west-2', 'eu-west-3', 'sa-east-1', 'us-east-1', 'us-east-2', 'us-west-1', 'us-west-2'];
```
tenge-amzn, line 479: Is there a reason for this extra check on support regions? If there's new meeting region launched in the future and client code copied this snippet - they'll got errors when new regions is returned from nearest-media-region

PR #345, opened 2020-05-06
Increase the size of content share video tile for the demo app in small screen

ltrung:
**Description of changes:**

**Testing**

- Manually test the demo app in a Pixel 3 XL and verify that content share tile is bigger (full width, and twice the height of the video tile).

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -33,7 +33,7 @@ import {
 } from '../../../../src/index';
 
 class DemoTileOrganizer {
-  private static MAX_TILES = 17;
+  static MAX_TILES = 17;
   private tiles: { [id: number]: number } = {};
   public tileStates: {[id: number]: boolean } = {};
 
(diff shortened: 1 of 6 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1317,6 +1330,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
 
   updateTilePlacement(tileIndex: number, x: number, y: number, w: number, h: number): void {
     const tile = document.getElementById(`tile-${tileIndex}`) as HTMLDivElement;
+    tile.classList.add(this.isContentTile(tileIndex) ? 'content-share-tile' : 'video-tile');
```
simmkyu, line 1333: When the tile element changes from the content share to the normal video, does the demo app replace this class with a new one?

ltrung: Good catch. I have updated the code to remove the previous class before adding a new one.

PR #378, opened 2020-05-14
Add data message APIs

ltrung:
**Description of changes:**
Add new data message APIs to allow sending message between attendees.
**Testing**

2. How did you test these changes? Manual testing.


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -594,7 +625,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
   async initializeMeetingSession(configuration: MeetingSessionConfiguration): Promise<void> {
     let logger: Logger;
     if (location.hostname === 'localhost' || location.hostname === '127.0.0.1') {
-      logger = new ConsoleLogger('SDK', LogLevel.INFO);
+      logger = new ConsoleLogger('SDK', LogLevel.DEBUG);
```
aburnson, line 628: Let's revert this change.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -611,12 +642,16 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
     this.meetingSession = new DefaultMeetingSession(configuration, logger, deviceController);
     this.audioVideo = this.meetingSession.audioVideo;
 
+    // XXX
+    (global as any).audioVideo = this.audioVideo;
```
aburnson, line 646: Let's revert this change.

demos/browser/app/meetingV2/meetingV2.ts
ltrung: Fixed.

demos/browser/app/meetingV2/meetingV2.ts
ltrung: Fixed.

simmkyu:
I pushed a fix for the CSS issue in Safari.

aburnson:
> I pushed a fix for the CSS issue in Safari.

Thanks @simmkyu

src/datamessage/DataMessage.ts
```diff
@@ -0,0 +1,67 @@
...
+    senderAttendeeId: string,
+    senderExternalUserId: string,
+    throttled?: boolean | null
+  ) {
+    this.timestampMs = timestampMs;
+    this.topic = topic;
+    this.data = data;
+    this.senderAttendeeId = senderAttendeeId;
+    this.senderExternalUserId = senderExternalUserId;
+    this.throttled = !!throttled;
```
simmkyu, line 51: `!!throttled` will always be a boolean, but the comment says,
> null if this DataMessage is not a sender receipt

src/datamessage/DataMessage.ts
```diff
@@ -0,0 +1,67 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/* eslint @typescript-eslint/no-explicit-any: 0 */
```
simmkyu, line 4: nit: you can put this line before `json(): any {` since `any` is being used only once

src/realtimecontroller/RealtimeState.ts
```diff
@@ -83,4 +84,18 @@ export default class RealtimeState {
...
+  sendDataMessageCallbacks: ((
+    topic: string, // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    data: Uint8Array | string | any,
+    lifetimeMs?: number
+  ) => void)[] = [];
+
+  /**
+   * Callbacks to listen for receiving message from data channel based on given topic
+   */
+  didReceiveDataMessageCallbacks: Map<string, ((dataMessage: DataMessage) => void)[]> = new Map();
```
simmkyu, line 100: nit: only this method starts with `did`. Exising properties do not start with a verb.

src/task/SendAndReceiveDataMessagesTask.ts
```diff
@@ -0,0 +1,100 @@
...
+      this.validateDataMessage(topic, uint8Data, lifetimeMs);
+      const message = SdkDataMessagePayload.create();
+      message.topic = topic;
+      message.lifetimeMs = lifetimeMs;
+      message.data = uint8Data;
+      const messageFrame = SdkDataMessageFrame.create();
+      messageFrame.dataMessagePayloads = [message];
+      this.context.signalingClient.sendDataMessage(messageFrame);
+    } else {
+      throw new Error('Signaling client is not ready');
```
simmkyu, line 83: Who is catching these errors?

`realtimeSendDataMessage` will call an array of callbacks including this one. e.g. Assume that a builder has added their callbacks. If `sendDataMessageHandler` is the second element in an array, a builder's 3rd, 4th, and subsequent callbacks will not be invoked.

```
  realtimeSendDataMessage(
    topic: string,
    data: Uint8Array | string | any, // eslint-disable-line @typescript-eslint/no-explicit-any
    lifetimeMs?: number
  ): void {
    this.wrap(() => {
      for (const fn of this.state.sendDataMessageCallbacks) {
        fn(topic, data, lifetimeMs);
      }
    });
  }
```

aburnson: Error-handling here could be improved; we probably shouldn't throw here. We may want to buffer up the message to send later or return the message back to the sender indicating it wasn't delivered (similar to `throttled`).

PR #384, opened 2020-05-17
Remove ExternalMeetingId param from createMeeting function

nakshathru:
`ERROR	UnexpectedParameter: Unexpected key 'ExternalMeetingId' found in params`
Getting this error while trying to create a new meeting using lambda

Upon removing the ExternalMeetingId param, meeting created successfully.
Is this parameter got removed in the latest sdk ??


asnyatkov:
Hi @nakshathru, ExternalMeetingId is a new parameter added to the API on 04/07. It takes some time for the updated AWS SDK to propagate through Lambda infrastructure. In order to get the up-to-date version, you can pull the most recent NPM and install it with your application, like we do in the serverless demo: https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/serverless/deploy.js#L14

Issue #347, opened 2020-05-06
Firefox Video tiles freeze after turning on/off


edwardbc:
**Meeting and attendee IDs:**
Meeting ID: 608e301f-7825-4977-8e40-af53b1a5655

**Timeline**
1. Chrome user joins the meeting and starts Video
2. Chrome user starts Video
3. Firefox user joins the meeting
4. Firefox user starts video (Warning 1)
5. Firefox user stops video, 
6. Firefox user starts video again (Warning 2)
7. Remote Video tile on Firefox freezes

*Warning 1:* ```2020-05-06T21:12:22.445Z [WARN] SDK - removing existing tiles 1 with same attendee id e3a8fe33-eec2-4b92-8bb9-b14a2a237bfc```

*Warning 2:* ```2020-05-06T21:13:13.897Z [WARN] SDK - removing existing tiles 3 with same attendee id e3a8fe33-eec2-4b92-8bb9-b14a2a237bfc```

**Describe the bug**
On Firefox, when joining a meeting with an existing video tile, then turning on video a warning is displayed (first sign), then when the other user turns off and on their video it freezes and `videoTileWasRemoved` never gets called. Subsequent attempts to display the remote video tile fail and it's never restored on Firefox

NOTE: Doesn't always happen when Firefox joins the meeting first. There needs to be an active video tile in place when Firefox joins to replicate this easily.
Additionally, on a separate app, calling `.audioVideo.getAllVideoTiles()` on the console after the Firefox user joins with Video, before toggling any tiles on the other side, it returns three tiles instead of two on Firefox, the third having a `tileState` full of `null` values. I couldn't replicate this specific case on the Browser demo, but might be related.

**Logs**
From browser demo. Initially replicated in a different app and flow.
[ChromeUser.log](https://github.com/aws/amazon-chime-sdk-js/files/4589558/ChromeUser.log)
[FirefoxUser.log](https://github.com/aws/amazon-chime-sdk-js/files/4589559/FirefoxUser.log)

**Platform**
 - OS: 
 - Browser:  Firefox
 - Browser Version:  76.0
 - Amazon Chime SDK GitHub Commit: 1.4.0

edwardbc:
There's a third warning that is sometimes displayed when enabling video again on the non-Firefox attendee
```2020-05-06T21:49:30.851Z [WARN] ChimeMeetingLogs - no stream found for tile=null```

lhristov:
[...]

PR #389, opened 2020-05-18
Fix Firefox 76 missing tile issue

alochbau-amzn:
**Issue #:** 
https://github.com/aws/amazon-chime-sdk-js/issues/347

**Description of changes:**
With Firefox version 76, the webrtc library issues a
trackEvent for any SDP m line in the initial offer/answer
and for any m line that had any change at all for follow
on offer/answers.  Add logic to ignore inactive m lines
or m lines that we already have a remote video tile for.

**Testing**

Yes
Using the demo app, added different versions of Firefox, Chrome and Safari into a meeting.  Enabled/disabled video and verified all tiles were visible.


edwardbc:
Seems the fixes from pull https://github.com/aws/amazon-chime-sdk-js/pull/389 fix the issue with the additional track on FF 76 and tiles are restored correctly so far. 

There's an error that it's triggered now every now and then after testing the above scenario and turning on and off the video on both clients , eventually the following error can be seen: 
Log sample 
```
INFO: videoAvailabilityDidChange  : { remoteVideoAvailable: false, canStartLocalVideo: true }

ERROR: TypeError: this.removeTrackRemovedEventListeners[track.id] is not a function

INFO: videoAvailabilityDidChange  ()  { remoteVideoAvailable: true, canStartLocalVideo: true }
```
On:
https://github.com/aws/amazon-chime-sdk-js/blob/5a9d8d0a6b6c0809c7a856a1e3da89b1706dfdb0/src/task/CreatePeerConnectionTask.ts#L263

PR #415, opened 2020-05-26
Fix FinishGatheringICECandidatesTask when there no turn creds

XHatan:
**Issue #:** 

**Description of changes:**
When there are no turn credentials, ice gathering promise might never get resolved because ice gathering state change is fired by a different event. Only affects use case where there is no turn creds.

**Testing**

Yes

Local demo testing


src/task/FinishGatheringICECandidatesTask.ts
```diff
@@ -86,6 +86,14 @@ export default class FinishGatheringICECandidatesTask extends BaseTask {
         this.removeEventListener();
         reject(error);
       };
+      if (!this.context.turnCredentials) {
+        this.context.peer.addEventListener('icegatheringstatechange', () => {
```
simmkyu, line 90: - How can `this.context.turnCredentials` be `null` or `undefined`? `ReceiveTURNCredentialsTask` in the start operation is always setting this to an `MeetingSessionTURNCredentials` instance.
- How about removing this event listener in `this.removeEventListener` too?

XHatan: * This is only for a debug mode testing with local server so there won't be any turn relay. 
* "How about removing this event listener in this.removeEventListener too", Sure I could do that.

Issue #278, opened 2020-04-16
The Camera LED can't be turned off


simmkyu:
**Important:** Please attach INFO-level logs with timestamps to the report from *before* the session started up until the point at which the unexpected behavior occured. For multi-participant calls, attach a log file for each participant if possible.

**Meeting and attendee IDs:**
List the meeting and attendee IDs that the attached logs apply to.
N/A

**Timeline**
Outline the main actions the user performed and when:
N/A (See below)

**Describe the bug**
Describe the bug and the timestamp at which it happened relative to the timestamp in the log file.

Repro step 1:
1. Open the browser demo app.
  https://github.com/aws/amazon-chime-sdk-js/tree/master/demos/browser
2. Type meeting ID and name. Click "Continue."
3. Immediately click "Join" in the device page before seeing your preview video.
4. LED light is turned on. You can't turn it off by calling stop methods.
   ```
   stopVideoPreviewForVideoInput()
   chooseVideoInputDevice(null)
   ```

Repro step 2:
1. Open meetingV2.ts
   https://github.com/aws/amazon-chime-sdk-js/blob/ae80650701b8691d1ad490ef1cb6a94bc0d195ac/demos/browser/app/meetingV2/meetingV2.ts#L1012
2.  Remove `await` from the first `chooseVideoInputDevice` call. Then, call `chooseVideoInputDevice()` after starting the video preview.
    ```
        this.audioVideo.chooseVideoInputDevice(device);
        if (showPreview) {
          this.audioVideo.startVideoPreviewForVideoInput(document.getElementById(
            'video-preview'
          ) as HTMLVideoElement);
        }
        await this.audioVideo.chooseVideoInputDevice(device);
    ```
3. Open the demo app and join the meeting. LED will be on permanently until closing a browser tab or window.

**Screenshots**
If applicable, add screenshots to help explain the problem.

**Platform**
 - OS: macOS 10.14.6
 - Browser: Chrome
 - Browser Version: 80.0.3987.163
 - Amazon Chime SDK GitHub Commit: f8c345b95472cc2aea26e7e399ca72f3523900d2

RossWilliams:
Could I summarise the issue as saying if you have multiple media streams for a track, the tracks cannot be stopped? Is this due to internal state tracking within the sdk, or a web platform issue?

simmkyu:
[...]

PR #435, opened 2020-06-01
Fix the LED light issue and add the device controller test

simmkyu:
**Issue #:** 
https://github.com/aws/amazon-chime-sdk-js/issues/278

**Description of changes:**

- Add device controller tests.
- Fix and improve device APIs.

### Problem 1: Releasing a previously acquired device

In the following diagram, say you have called `chooseVideoInputDevice` API three times. Before this PR, all three will release the active device `A`. After this PR, each API call will release devices `X` and `Y` acquired by other API calls in progress or completed.

Screen Shot 2020-06-01 at 9 42 05 AM

<br/><br/>

### Problem 2: Selecting no device (`null`) while one or more APIs are running

Assume that you have called `chooseVideoInputDevice` API with the device `X`. Before you allow or deny the permission, you called `chooseVideoInputDevice` with `null` to unselect all devices. Since you explicitly indicated "no device," SDK will release a device X acquired before passing `null`.

Screen Shot 2020-06-01 at 9 55 02 AM

**Testing**

Yes

Tested in a browser demo


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -476,15 +479,19 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     device: Device,
     fromAcquire: boolean
   ): Promise<DevicePermission> {
+    this.videoInputDeviceCount += 1;
```
ltrung, line 482: Should this only increase if kind === 'video'

simmkyu: A good catch - This counter might be used for audio later. I'll rename it to `this.inputDeviceCount`.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -496,30 +503,36 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
         this.logger.info(`using media stream ${stream.id} for ${kind} device`);
-        newDevice = new DeviceSelection();
         newDevice.stream = stream;
         newDevice.constraints = proposedConstraints;
       } else {
-        newDevice = await this.getDeviceFromBrowser(proposedConstraints);
+        newDevice.stream = await navigator.mediaDevices.getUserMedia(proposedConstraints);
+        newDevice.constraints = proposedConstraints;
+
+        if (kind === 'video' && this.lastNoVideoInputDeviceCount > callCount) {
```
ltrung, line 525: When can this.lastNoVideoInputDeviceCount > callCount? I saw they all is set to this.videoInputDeviceCount.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -548,36 +561,44 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
         : DevicePermission.PermissionDeniedByUser;
     }
     this.logger.info(`got ${kind} device for constraints ${JSON.stringify(proposedConstraints)}`);
-    const restartVideo =
-      kind === 'video' &&
-      !fromAcquire &&
-      this.boundAudioVideoController &&
-      this.boundAudioVideoController.videoTileController.hasStartedLocalVideoTile();
+
+    const oldStream: MediaStream | null = this.activeDevices[kind]
```
ltrung, line 565: So if I understand this correctly, the main fix is to release the media stream after get the new stream from GetUserMedia correct? Before we do it before getting the new stream, so all ChooseVideoInputDevice calls will release stream A (as in your example).

src/devicecontroller/DefaultDeviceController.ts
simmkyu: 1. Call `chooseVideoInputDevice()` and wait for `getUserMedia()` to resolve. A user needs to click Allow or Deny, or it will take a few milliseconds if previously allowed.
2. While waiting for `getUserMedia()`, call `chooseVideoInputDevice(null)` to unselect all devices.
3. Now, the previous call at (1) is resolved.
   At this point, `this.lastNoVideoInputDeviceCount > callCount` will be `true`.

src/devicecontroller/DefaultDeviceController.ts
ltrung: Yes, I was about to ask whether we have the same issue for ChooseAudio<Input/Output>Device?

simmkyu: - Both APIs have the same issue of not releasing a previous device if the API is getting multiple times.

- But calling `chooseAudioInputDevice(null)` creates an empty audio device, while calling `chooseVideoInputDevice(null)` unselects devices. I haven't looked into details for the audio flow.

src/devicecontroller/DefaultDeviceController.ts
simmkyu: Yes, in this PR, SDK is releasing an old stream ***after*** `getUserMedia()` is resolved. Before this PR, SDK released it before.

Issue #263, opened 2020-04-14
Return specific errors for TaskFailed errors


simmkyu:
Chime SDK throws a general error `TaskFailed` when [MeetingSessionStatusCode](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/meetingsession/MeetingSessionStatusCode.ts) doesn't have a corresponding status. Since the release, we observed that SDK had stopped the session with `TaskFailed` in the following cases:
- A task in [src/task](https://github.com/aws/amazon-chime-sdk-js/tree/master/src/task) has an uncaught error
  e.g. `Cannot read property '__property_name__' of undefined`.
- The connection is timed out after 15 seconds (default).
- SDK fails to reconnect the session after multiple retries.

We should add new status codes for these common cases.

kaaaren:
Fix https://github.com/aws/amazon-chime-sdk-js/pull/581 is merged, closing this issue

PR #464, opened 2020-06-08
Restart the session if the attendee is not present 

simmkyu:
**Issue #:** 
N/A

**Description of changes:**
- If a non-zero `MeetingSessionConfiguration.attendeePresenceTimeoutMs` is specified, SDK will wait for the current attendee to appear. If not present, SDK will reconnect the session.
- In Firefox and Safari, SDK will ask for the microphone permission before reconnect.

**Problem to fix**
If you haven't called `chooseAudioInputDevice` with a non-null device before and start a meeting, SDK creates an empty audio device and joins a meeting. In Firefox and Safari, if a user hasn't allowed microphone before and then join a meeting with an empty audio device, SDK fails to send an audio packet to the backend. Since SDK uses an audio info frame to notify an attendee presence, an attendee doesn't show up.

**Testing**

   Yes
   Manually intercepting the attendee presence event and checking whether SDK is reconnecting or not


src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -286,6 +287,7 @@ export default class DefaultAudioVideoController implements AudioVideoController
               new FinishGatheringICECandidatesTask(this.meetingSessionContext),
               new SubscribeAndReceiveSubscribeAckTask(this.meetingSessionContext),
               new SetRemoteDescriptionTask(this.meetingSessionContext),
+              new ValidateAttendeePresenceTask(this.meetingSessionContext),
```
XHatan, line 290: Should we make the time out a configurable value in `MeetingSessionConfiguration` ?
I would like this feature to have a less blast radius, e.g the default `ValidateAttendeePresenceTask` is a no-op, if user configure the time out then this task will then wait for attendee presence

simmkyu: That's a good suggestion. I can make it configurable. Why do you want to have a smaller blast radius?
- `null` (default) : no-op. SDK will not restart the session.
- `number`: SDK will wait the given amount of time and restart if not present.

One more question - do you see any issue programmatically choosing audio device when SDK restarts for the no-presence event?

```
const devices = await app.audioVideo.listAudioInputDevices();
if (!devices.length) {
  await app.audioVideo.chooseAudioInputDevice(devices[0].deviceId);  
}
```

XHatan: > Why do you want to have a smaller blast radius?
null (default) : no-op. SDK will not restart the session.
number: SDK will wait the given amount of time and restart if not present.

I think I am being super cautious here about developers having an option to disable this safety check. But on second thought, "attendee presence" is pretty vital and we should validate this. So it's up to you :)


>One more question - do you see any issue programmatically choosing audio device when SDK restarts for the no-presence event?
const devices = await app.audioVideo.listAudioInputDevices();
if (!devices.length) {
  await app.audioVideo.chooseAudioInputDevice(devices[0].deviceId);  
}

Where is the code ? Is it demo application ?

simmkyu: We don't have that code yet. If you don't call `chooseAudioInputDevice`, SDK joins a meeting with an empty audio device.

In Safari, if you join a meeting with an empty audio device, an attendee is not present. To mitigate this issue, SDK can explicitly call `chooseAudioInputDevice` before restarting the session between line 302 and 303 here.

https://github.com/aws/amazon-chime-sdk-js/blob/master/src/audiovideocontroller/DefaultAudioVideoController.ts#L302-L303

XHatan: might need some input from @aburnson 

My first impression is that `chooseAudioInputDevice` on a real device should only be triggered by user action, if SDK calls `chooseAudioInputDevice` in this case

aburnson: Device selection should only occur by a user via the developer's application. The SDK should not assume a device since that could lead to leaking data from an unintended device. I think it could be reasonable to call `chooseAudioInputDevice` for the previously selected device (or a null device) if the intention is to try to kick the session join through again.

simmkyu: Thanks for the suggestions. In this PR, I'll not call `chooseAudioInputDevice` during reconnect. I'll push a new revision with a configurable timeout.

simmkyu:
This PR was using `RealTimeController.realtimeExternalUserIdFromAttendeeId` API to check whether the current attendee exists. If not, the task subsribes to the present event.

The recent commit "Parse External User ID from Index Frame Message" (87be5956d1b64bba9b61473b6a1a8d5523f15c4d) has removed this API. Removing it makes sense. We should expose the least information possible and suggest using the callback. I'll need to refactor this PR to incorporate this change.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -290,16 +292,33 @@ export default class DefaultAudioVideoController implements AudioVideoController
...
+        // TODO: Remove this workaround once the root cause is identified.
+        // Assume you haven't allowed microphone before and joined the session with an empty audio device.
+        // In Safari, SDK doesn't receive an attendee presence event. In Firefox, no ice candidates were gathered.
+        if (
+          this.meetingSessionContext.browserBehavior.requiresGetUserMediaForNoAttendeePresence() &&
+          (status.statusCode() === MeetingSessionStatusCode.NoAttendeePresent ||
+            (error && error.message && error.message.includes('no ice candidates were gathered')))
+        ) {
+          try {
+            await navigator.mediaDevices.getUserMedia({ audio: true });
```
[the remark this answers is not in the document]
simmkyu: I just tried using the media stream broker. It doesn't work because `acquireAudioInputStream` also creates and returns an empty audio device. The `getUserMedia` doesn't get called ([here](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/devicecontroller/DefaultDeviceController.ts#L522)).

simmkyu:
@aburnson Thanks for the suggestions.

- Removed unnecessary logic from `MonitorTask` and `WaitForAttendeePresenceTask` (renamed from `ValidateAttendeePresenceTask`).
- Haven't modified `TimeoutScheduler` for taking a callback to thorw an error. I want to imporve error handling with https://github.com/aws/amazon-chime-sdk-js/issues/263.
- Using `acquireAudioInputStream` doesn't work because it's also creating an empty audio device and not prompting the permission dialog. How about adding an API to `MediaStreamBroker` for calling the audio-only `getUserMedia`?
  ```
  await navigator.mediaDevices.getUserMedia({ audio: true });
  ```

simmkyu:
I just tested that this PR didn't work in Safari if the bluetooth headphone was connected. I needed to choose the MacBook microphone explicitly instead of `getUserMedia({ audio: true })`.

```
await navigator.mediaDevices.getUserMedia({ audio: true });
```

simmkyu:
@aburnson @dylonChime I updated PR. I removed the `getUserMedia` workaround from this PR and created https://github.com/aws/amazon-chime-sdk-js/issues/474 for an empty audio issue. SDK will just restart the session if an attendee is not present for given amount of time.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -285,7 +286,19 @@ export default class DefaultAudioVideoController implements AudioVideoController
               new SetLocalDescriptionTask(this.meetingSessionContext),
               new FinishGatheringICECandidatesTask(this.meetingSessionContext),
               new SubscribeAndReceiveSubscribeAckTask(this.meetingSessionContext),
-              new SetRemoteDescriptionTask(this.meetingSessionContext),
+              ...(this.meetingSessionContext.meetingSessionConfiguration.attendeePresenceTimeoutMs >
```
aburnson, line 289: Is splat/array necessary here? Each array in the ternary expression has one thing in it...

simmkyu: Right, `...[]` is unnecessary. I udpated PR

PR #490, opened 2020-06-17
Fixing the ability to choose default in chrome browsers

RidipDe:
**Issue #:** 
When a meeting is active and we connect a bluetooth device to our system and then select the default (which is the bluetooth device) option from the audio input drop down, the switch does not take place. The app continues to use the audio input device that it has been using before the bluetooth was connected.

**Description of changes:**
The constraint check for chooseInputDevice() now checks with the help of groupId in addition to deviceId.
Calling the MediaStreamTrack.stop() once before making the switch to the new device. This call tells the user agent that the track's source is no longer needed by the MediaStreamTrack

**Testing**
Yes
Tested it manually. We can now select the default option in the drop down audio input menu or perform await app.audioVideo.chooseAudioInputDevice('default');
Inter browser tests have also been performed between Chrome and Firefox


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Updated cloudwatch log stream ID to have attendee_id
+- Ability to choose default in chrome browsers
```
ltrung, line 22: This is a bit misleading. We support choosing default in Chrome before so could you add more details as well as the PR title?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -520,7 +542,12 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
         newDevice.stream = stream;
         newDevice.constraints = proposedConstraints;
       } else {
-        newDevice.stream = await navigator.mediaDevices.getUserMedia(proposedConstraints);
+        if (kind === 'audio' && this.activeDevices[kind]) {
```
ltrung, line 545: Why do we need to do this here? This should be done in releaseMediaStream already

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -492,9 +508,15 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
       return DevicePermission.PermissionGrantedByBrowser;
     }
 
-    const proposedConstraints: MediaStreamConstraints | null = this.calculateMediaStreamConstraints(
+    let proposedConstraints: MediaStreamConstraints | null = this.calculateMediaStreamConstraints(
```
ltrung, line 511: Cant we modify DeviceSelection and add groupId there? Then we do not need to maintain another mapping and need to do this either.

src/devicecontroller/DefaultDeviceController.ts
RidipDe: We need to stop the active MediaTrack at least once before we call the getUserMedia(constraint)
This call tells the user agent that the track's source is no longer needed by the MediaStreamTrack

CHANGELOG.md
RidipDe: Selecting the 'default' in chrome never actually chose the 'default' in Chrome. Please let me know what should we have here instead

simmkyu: Not only Chrome, but also Opera also lists the same device twice

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -41,6 +41,11 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
   private videoMaxBandwidthKbps: number = DefaultDeviceController.defaultVideoMaxBandwidthKbps;
 
   private useWebAudio: boolean = false;
+  private deviceIdGroupIdMap: { [mediaDeviceKind: string]: Map<string, string> | null } = {
```
simmkyu, line 44: From `Map<string, string> | null`, `| null` is not required since it's always be `new Map()`.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -408,6 +413,17 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
       }
     }
     this.deviceInfoCache = devices;
+    this.setDeviceIdGroupIdMap();
+  }
+
+  private setDeviceIdGroupIdMap(): void {
+    for (const device of this.deviceInfoCache) {
+      if (device.groupId) {
```
simmkyu, line 421: - nit: It can be in just one line without this `if` statement.
    ```
    this.deviceIdGroupIdMap[device.kind].set(device.deviceId, device.groupId || '');
    ```
- nit: Since you can simplify it, you can just put it into `updateDeviceInfoCacheFromBrowser()`.

src/devicecontroller/DefaultDeviceController.ts
simmkyu: - A good point - SDK is already releasing an old audio stream at line 618.
- Assume you are using Microphone1, and the browser doesn't store a previously-selected permisison, e.g. FF. Now, you are choosing Microphone2. If you release an active input here, you might be muted until you allow the microphone permission opened from `getUserMedia`.

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -240,8 +251,8 @@ describe('DefaultDeviceController', () => {
         await new Promise(resolve => new TimeoutScheduler(10).start(resolve));
         deviceController.chooseAudioInputDevice(stringDeviceIds[3]);
       });
-      new TimeoutScheduler(500).start(() => {
-        expect(index).to.equal(3);
+      new TimeoutScheduler(100).start(() => {
+        expect(index).to.equal(1);
```
simmkyu, line 255: I don't think this test is related to your fix for choosing the "default."

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -464,6 +475,17 @@ describe('DefaultDeviceController', () => {
       await deviceController.chooseAudioOutputDevice(stringDeviceId);
       expect(spy.called).to.be.true;
     });
+
+    it('binds the audio device if the audio-video is bound with groupId', async () => {
```
simmkyu, line 479: This new `chooseAudioOutputDevice` test is also not related to your fix for `chooseAudioInputDevice`. I added an overall comment for testing strategies.

RidipDe: This is for just testing the mapping.

src/devicecontroller/DefaultDeviceController.ts
RidipDe: done

src/devicecontroller/DefaultDeviceController.ts
RidipDe: For using 'default', once the MediaStreamTrack reference is removed, only then the mapping between the ‘default’ and the original device is dissociated from the browser.
I can add another check like:
JSON.parse(JSON.stringify(proposedConstraints.audio)).deviceId === 'default' before the stop is called

CHANGELOG.md
simmkyu: > Selecting the 'default' in chrome never actually chose the 'default' in Chrome. Please let me know what should we have here instead

Before this PR, choosing `default` works if you have not selected any device or have seleted a non-`default` device before.

src/devicecontroller/DefaultDeviceController.ts
simmkyu: Could you elaborate this a bit more? Is this for fixing the Chrome issue where calling `getUserMedia` with `default` doesn't work when another `default` is already selected?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -630,6 +659,10 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
       return null;
     } else if (typeof device === 'string') {
       trackConstraints.deviceId = { exact: device };
+      if (kind === 'audio' && setGroupId === true) {
+        groupId = this.deviceIdGroupIdMap['audioinput'].get(device);
+        trackConstraints.groupId = groupId;
```
simmkyu, line 664: This PR assumes that the `Device` device is a string ID; however, `Device` can be `string | MediaTrackConstraints | MediaStream | null`.

For example, this flow won't work.
1. `chooseAudioInputDevice('default')` to choose the default.
2. Connect a new device "Bluetooth Headphone." A new default is "Bluetooth Headphone."
3. Choose the new default by passing the constraints.
    ```
    chooseAudioInputDevice({
      deviceId: { exact: 'default' };
    })
    ```

simmkyu:
You are going in the right direction. We need to take these two into account.

1. Handle the `Device` type:
https://github.com/aws/amazon-chime-sdk-js/blob/5edcc57511fd6617bfb9f376269b2054fbdf942b/src/devicecontroller/Device.ts#L4
2. Don't pass `groupId` to `getUserMedia` to avoid regressions

---

### Suggestions

How about just removing a group ID right before `getUserMedia`?
```
const { groupId, ...constraintsWithoutGroupId } = proposedConstraints;
newDevice.stream = await navigator.mediaDevices.getUserMedia(
  constraintsWithoutGroupId
);
```

Then, always accept `groupId` in `calculateMediaStreamConstraints` instead of a boolean argument only for the string device ID.
```
private calculateMediaStreamConstraints(
  kind: string,
  device: Device,
  groupId: string
```

simmkyu:
From https://github.com/aws/amazon-chime-sdk-js/pull/496:

We are overcomplicating this a bit. This PR still has a few issues.
- You are stopping the previous `default` and then call `getUserMedia`. In some browser such as FF, `getUserMedia` might prompt the permission dialog and wait for the user's answer. If you stop the previous `default`, then the user is getting muted until allowing the microphone permission. Also, the backend might drop this user since it doesn't receive audio for a while.
- Comparing using `this.isGroudpIdSame` won't work if a user has never listed devices before. `this.deviceIdGroupIdMap` will be empty. It's working in the SDK demo since we list devices in the device selection page.

---

Sorry for all the back and forth. In addition to my suggestions https://github.com/aws/amazon-chime-sdk-js/pull/490#issuecomment-645754243, how about simplifying this PR like the following?

### Suggestions

1. Instead of maintaining the `this.isGroudpIdSame` map, just get the group ID in `chooseInputDevice` and pass it to `calculateMediaStreamConstraints`.

    ```
    const groupId = groupIdFromDevice(device);

    const proposedConstraints: MediaStreamConstraints | null = this.calculateMediaStreamConstraints(
      kind,
      device,
      groupId
    );
    ```
2. Accept `groupId` in `calculateMediaStreamConstraints` and assign it to `trackConstraints` if exists.
    ```
    private calculateMediaStreamConstraints(
      kind: string,
      device: Device,
      groupId: string
      
      if (groupId) {
        trackConstraints.groupId = groupId;
      }
    ```
3. Right before calling `getUserMedia`, remove a group ID to avoid regression.
    ```
    const { groupId, ...constraintsWithoutGroupId } = proposedConstraints;
    newDevice.stream = await navigator.mediaDevices.getUserMedia(
      constraintsWithoutGroupId
    );
    ```

PR #495, opened 2020-06-17
Add resize listener on HTMLVideoElement in Demo

XHatan:
**Issue #:** 

**Description of changes:**

**Testing**


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1445,6 +1446,21 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
     return tiles;
   }
 
+  setUpVideoTileElementResizer(): void {
+    for (let i = 0; i <= 16; i++) {
+      const videoElem = document.getElementById(`video-${i}`) as HTMLVideoElement;
+      videoElem.onresize = () => {
+        this.log(`video-${i} intrinsic resolution ${videoElem.videoWidth}x${videoElem.videoHeight}`);
```
davejlane, line 1453: Do you want to log every resize? My concern is that any time the viewport is resized or when another video is added or removed and affects the layout, then this function will fire for up to 16 video elements. It might be worth only logging if the video dimensions matches the portrait dimensions you're looking for?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1445,6 +1446,21 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
     return tiles;
   }
 
+  setUpVideoTileElementResizer(): void {
+    for (let i = 0; i <= 16; i++) {
```
ltrung, line 1450: We have 17 video tiles right?

RidipDe: yup 17 tiles -> [0,16]

demos/browser/app/meeting/meeting.ts
```diff
@@ -491,6 +492,21 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver
...
+  setUpVideoTileElementResizer(): void {
+    for (let i = 0; i <= DemoTileOrganizer.MAX_TILES; i++) {
+      const videoElem = document.getElementById(`video-${i}`) as HTMLVideoElement;
+      videoElem.onresize = () => {
+        if (videoElem.videoHeight > videoElem.videoWidth) {
+          // portrait mode
+          this.log(`video-${i} changed to portrait mode resolution ${videoElem.videoWidth}x${videoElem.videoHeight}`);
+          videoElem.style.objectFit = 'contain';
+        } else {
+          videoElem.style.objectFit = 'cover';
```
mikeantonelli, line 504: Shouldn't we also compare to the aspect ratio of the `parentElement`? We don't want to apply `cover` when rendering a 16:9 video in a 9:16 (or 4:3) container, because the visible area of the video will pixelate due to stretching and video will focus in the absolute center commonly missing the user's intended focus area.

XHatan: The PR is mainly to show one way of obtaining intrinsic video resolution from HTMLVideoElement. And I agree with your comments, depending on the layout, developers need to handle cropping differently based on various intrinsic video resolutions.

PR #496, opened 2020-06-18
Fixing the ability to choose default input in browsers when default changes

RidipDe:
**Issue #:** 
When a new audio input device is connected to a system while the 'default' system microphone is in use such that the new input device becomes the system default, the call to the method getUserMedia() with audio constraint deviceId as 'default' in chromium based browsers continues to use the previously active media stream (system microphone) instead of the the expected new stream (new audio input device). However if we call the getUserMedia() with audio constraint deviceId with alpha-numeric deviceId of the new device, it returns the stream of the new device. 
When a meeting is active and we connect an external device (lets say bluetooth device) to our system and then select the default which is the bluetooth device option from the audio input drop down, the switch does not take place. The app continues to use the audio input device that it has been using before the bluetooth was connected.

**Description of changes:**
Change 1:
In chromium browsers running on desktops and mobile devices, the system speaker is the 'default' device id.
With an external audio device connected (like a bluetooth audio device), the external audio device becomes the 'default' but the bowser browser don't automatically switch the device. The user has to manually reselect the default device. 
When choosing the new device with deviceId 'default', while the the previous device with deviceId 'default' is active, the current logic in the SDK checks the constraints of the activeDevice and the new requesting device based on just the deviceId. So when requesting the new 'default' device, the current logic returns the already active media stream as it thinks that a request to the already active 'default' device is made. 
Each device also has a groupId associated. While the deviceId is unique to each device, the groupId may/may not group multiple devices into single group. So instead of just using the deviceId as a constraint check for the condition, deviceId and the corresponding groupId should be used as a unique pair to validate the check. So when a new device id added the deviceId and groupId pair will be unique.

Change 2:
Following the above problem we have another issue that needs to be addressed to solve the issue. If we try to call the getUserMedia() on the 'default' device, since the browser is already using the 'default', it is going to create a new stream but will point to the old 'default' device. So we need to ensure that the old media stream is stopped/released before requesting the new 'default' device in the getUserMedia(). 
The understanding here is, that the mapping between the deviceId and the device is not updated until the track is stopped. 
If we call the getUserMedia() with 'default' as deviceId in the constraint without stopping the active track, it returns the previously acquired MediaStream associated to the 'default' deviceId. But once we stop the active media track and then call the getUserMedia(), it will tell the user agent that the track's source is no longer needed by the MediaStreamTrack. This way, the browser will know to clean up the existing references and update the mapping.

**Repro Steps using the demo meeting app:**
1) Start a meeting with no external device in a Chromium based browser. The 'default' audio input is the system microphone.
2) Connect an external device (bluetooth for example). Under the hood, the bluetooth becomes the new default.
3) From the audio input drop down, select the 'default'.
 The switch in the microphone does not take place. The SDK continues to use the old 'default' microphone

**Testing**
Yes
Tested it manually. We can now select the default option in the drop down audio input menu or perform await app.audioVideo.chooseAudioInputDevice('default');
Inter browser tests have also been performed between Chrome and Firefox


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -492,21 +507,24 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
       return DevicePermission.PermissionGrantedByBrowser;
     }
 
-    const proposedConstraints: MediaStreamConstraints | null = this.calculateMediaStreamConstraints(
+    let proposedConstraints: MediaStreamConstraints | null = this.calculateMediaStreamConstraints(
```
ltrung, line 510: Should be const here?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -475,6 +481,15 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     return device && device.id ? device : null;
   }
 
+  private isGroudpIdSame(groupId: string, kind: string, device: Device): boolean {
```
ltrung, line 484: nit: rename it to hasSameGroupId?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -475,6 +481,15 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+    if (device === null) return false;
+    if (typeof device !== 'string')
+      device = JSON.parse(JSON.stringify(device)).deviceId
```
ltrung, line 487: Could we move JSON.stringify to a variable?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -475,6 +481,15 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+        ? JSON.parse(JSON.stringify(device)).deviceId.exact
+        : JSON.parse(JSON.stringify(device)).streamId;
+    return groupId === this.deviceIdGroupIdMap[kind + 'input'].get(device);
```
ltrung, line 490: this does not work for audiooutput right?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -520,9 +538,16 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
         newDevice.stream = stream;
         newDevice.constraints = proposedConstraints;
       } else {
+        //Check can be removed after we confirm that getUserMedia bug has been fixed by chromium
```
ltrung, line 541: Need a link to the actual Chrome bug. And please put it in a format TODO: remove when this bug is fixed.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -520,9 +538,16 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+        if (
+          kind === 'audio' &&
+          this.activeDevices[kind] &&
```
ltrung, line 544: can we just check if the parameter device === default here?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -550,6 +575,13 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
           }
         });
       }
+
+      let groupId = this.deviceIdGroupIdMap[kind + 'input'].get(device);
+      if (groupId === null) {
```
ltrung, line 580: why cant groupid be null? I dont think we need this if stmt.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -492,21 +507,24 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
       kind,
       device
     );
+
     if (
       this.activeDevices[kind] &&
       this.activeDevices[kind].matchesConstraints(proposedConstraints) &&
-      this.activeDevices[kind].stream.active
+      this.activeDevices[kind].stream.active &&
+      this.activeDevices[kind].groupId !== null &&
```
ltrung, line 519: This is not a correct logic since we want to skip groupId check if groupId is null. Also since this only happens to Chrome default, should we rework this to its own method and only check groupId for Chrome and default to reduce the change of bug?

src/devicecontroller/DefaultDeviceController.ts
RidipDe: done

src/devicecontroller/DefaultDeviceController.ts
RidipDe: done

src/devicecontroller/DefaultDeviceController.ts
RidipDe: done

src/devicecontroller/DefaultDeviceController.ts
RidipDe: We dont know how the change in the video input will behave, if lets say we connect a usb webcam in the middle of the meeting session this device will become the default.
Also since we are targeting AudioTrack only, we are checking on 'audio' kind

ltrung: I meant we should check if device === default here too because the bug only happens if we switch from default to default right?

simmkyu:
We are overcomplicating this a bit. This PR still has a few issues.
- You are stopping the previous `default` and then call `getUserMedia`. In some browser such as FF, `getUserMedia` might prompt the permission dialog and wait for the user's answer. If you stop the previous `default`, then the user is getting muted until allowing the microphone permission. Also, the backend might drop this user since it doesn't receive audio for a while.
- Comparing using `this.isGroudpIdSame` won't work if a user has never listed devices before. `this.deviceIdGroupIdMap` will be empty. It's working in the SDK demo since we list devices in the device selection page.

---

Sorry for all the back and forth. In addition to my suggestions https://github.com/aws/amazon-chime-sdk-js/pull/490#issuecomment-645754243, how about simplifying this PR like the following?

### Suggestions

1. Instead of maintaining the `this.isGroudpIdSame` map, just get the group ID in `chooseInputDevice` and pass it to `calculateMediaStreamConstraints`.

    ```
    const groupId = groupIdFromDevice(device);

    const proposedConstraints: MediaStreamConstraints | null = this.calculateMediaStreamConstraints(
      kind,
      device,
      groupId
    );
    ```
2. Accept `groupId` in `calculateMediaStreamConstraints` and assign it to `trackConstraints` if exists.
    ```
    private calculateMediaStreamConstraints(
      kind: string,
      device: Device,
      groupId: string
      
      if (groupId) {
        trackConstraints.groupId = groupId;
      }
    ```
3. Right before calling `getUserMedia`, remove a group ID to avoid regression.
    ```
    const { groupId, ...constraintsWithoutGroupId } = proposedConstraints;
    newDevice.stream = await navigator.mediaDevices.getUserMedia(
      constraintsWithoutGroupId
    );
    ```

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -520,9 +548,24 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
         newDevice.stream = stream;
         newDevice.constraints = proposedConstraints;
       } else {
+        // TODO: remove check when the bug is fixed.
```
aburnson, line 551: When what bug is fixed?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -520,9 +548,24 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+        if (
+          kind === 'audio' &&
+          this.activeDevices[kind] &&
+          device &&
+          this.getDeviceIdStr(device) === 'default'
```
aburnson, line 556: Why does it matter what the device label of the device is (e.g. `default`)? Shouldn't this logic apply any time there is a request for a device that has the same device id as existing but a different group id?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -520,9 +548,24 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+        ) {
+          if (
+            this.activeDevices[kind].constraints &&
+            this.activeDevices[kind].constraints.audio &&
+            JSON.parse(JSON.stringify(this.activeDevices[kind].constraints.audio)).deviceId
```
aburnson, line 561: Why roundtrip through JSON?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -520,9 +548,24 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+              .exact === 'default'
+          ) {
+            this.activeDevices[kind].stream.getAudioTracks()[0].stop();
```
aburnson, line 564: Why not use [this.releaseMediaStream](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/devicecontroller/DefaultDeviceController.ts#L231)?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -520,9 +548,24 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
         newDevice.constraints = proposedConstraints;
       } else {
+        // TODO: remove check when the bug is fixed.
+        if (
+          kind === 'audio' &&
+          this.activeDevices[kind] &&
+          device &&
+          this.getDeviceIdStr(device) === 'default'
+        ) {
+          if (
```
aburnson, line 558: What is the reason for having two `ifs`? Can they be combined with `&&`?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -641,8 +685,6 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
       trackConstraints.width = trackConstraints.width || this.videoWidth;
       trackConstraints.height = trackConstraints.height || this.videoHeight;
       trackConstraints.frameRate = trackConstraints.frameRate || this.videoFrameRate;
-      // TODO: try to replace hard-code value related to videos into quality-level presets
```
aburnson, line 644: Why is this comment being removed?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -41,6 +41,11 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
   private videoMaxBandwidthKbps: number = DefaultDeviceController.defaultVideoMaxBandwidthKbps;
 
   private useWebAudio: boolean = false;
+  private deviceIdGroupIdMap: { [mediaDeviceKind: string]: Map<Device, string> } = {
```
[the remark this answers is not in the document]
RidipDe: Few browsers like Safari does not have groupId as part of the device information. To make the solution oblivious of the browser I added this state so that it can have a mapping between the deviceId and the groupId (or '' if group id is missing)

src/devicecontroller/DefaultDeviceController.ts
RidipDe: deviceId is unique but the groupId is not. With the 'default' in this  condition check, we are making sure that the releaseMediaStream() or stop() is called only for the the device which has the deviceId 'default'

RidipDe: However if we do a check 
this.getDeviceIdStr(device) === JSON.stringify(this.activeDevices[kind].constraints.audio).deviceId.exact 
instead, the 'default' check can be avoided

src/devicecontroller/DefaultDeviceController.ts
RidipDe: updated

RidipDe:
> Questions on the PR overview:
> 
> > When a meeting is active and we connect a bluetooth device to our system
> 
> * Is this specific to Bluetooth devices?
> * What if the device is already connected, can the issue be reproduced?
> * What browser does this impact?
> 
> > then select the default (which is the bluetooth device) option from the audio input drop down
> 
> * Does this impact the SDK or just the meeting demo?
> 
> > The constraint check for chooseInputDevice() now checks with the help of groupId in addition to deviceId.
> 
> * What does this mean specifically? How is the groupId used?
> 
> > This call tells the user agent that the track's source is no longer needed by the MediaStreamTrack.
> 
> * Why does this resolve the issue?
> 
> > How did you test these changes?
> 
> * Are there unit tests to test behavior with the workaround and without the workaround?


> Is this specific to Bluetooth devices?
The issue can occur whenever an external device is attached to the system in the middle of the meeting. Bluetooth was the external device used for testing here

> What if the device is already connected, can the issue be reproduced?
If the device is already connected, and the default changes while in the middle of meeting, the issue can be reproduced.

> What browser does this impact?
Any browser that uses 'default' as the deviceId. Chrome and Opera are couple of examples

> Does this impact the SDK or just the meeting demo?
It affects both the SDK and the meeting demo

> What does this mean specifically? How is the groupId used?
groupId is being used as a condition check along with the constraint parameter in this PR.
If activeDevice (deviceId and groupId) = requestingDevice (deviceId and groupId) continue using the same device, else new create DeviceSelection() with the same deviceId and the groupId
Also, when in some browsers the groupId is undefined (like Safari), we assign a default groupId to make the solution oblivious of the browser.

> Why does this resolve the issue?
When the getUserMedia() is called with 'default' as the constraint deviceId, while the stream is active, the 'default' refers to the old 'default' even after the new device was connected in the middle of the meeting and it is chosen by the OS to become the 'default'

src/devicecontroller/DefaultDeviceController.ts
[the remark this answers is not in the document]
aburnson: Do the browsers that have an empty group ID require the workaround logic?

src/devicecontroller/DefaultDeviceController.ts
aburnson: My question is really what about this workaround needs to be specific to the device that is labeled `default`? Do we need to know that the device is the default one? Can we not trigger the workaround on any existing device whose group id is now different?

src/devicecontroller/DefaultDeviceController.ts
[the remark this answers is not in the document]
RidipDe: No browsers such as Safari does not have groupId. It has unique deviceId which is never reassigned like in the case of chromium based browser.

simmkyu:
@aburnson @RidipDe Based on PR #490 (closed), please take a look at my minimal solution for choosing the default: https://github.com/aws/amazon-chime-sdk-js/commit/81f546234a2f75aa0f2fc4b974e7d8450677e1f4
- Include `groupId` in [`DeviceSelection.constraints`](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/devicecontroller/DeviceSelection.ts#L5) if the given device is a string ID.
- To avoid any regression, call `getUserMedia()` without a group ID as before.

Issue #445, opened 2020-06-03
DeviceChangeObserver method is not triggered in iOS and Android


simmkyu:
**Important:** Please attach INFO-level logs with timestamps to the report from *before* the session started up until the point at which the unexpected behavior occured. For multi-participant calls, attach a log file for each participant if possible.

**Meeting and attendee IDs:**
List the meeting and attendee IDs that the attached logs apply to.
N/A

**Timeline**
Outline the main actions the user performed and when:
N/A

**Describe the bug**
Describe the bug and the timestamp at which it happened relative to the timestamp in the log file.

[`DeviceChangeObserver`](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/devicechangeobserver/DeviceChangeObserver.ts) methods are not getting called in iOS and Android.

Repro steps
1. Implement `audioInputsChanged`, `audioOutputsChanged`, and `videoInputsChanged` to detect if the current device list has changed in iOS or Android.
2 Connect your bluetooth audio device
3. `audioInputsChanged` and `audioOutputsChanged` are not triggered.

**Screenshots**
If applicable, add screenshots to help explain the problem.

**Platform**
 - OS: iOS and Android
 - Browser: Safari (iOS) and Chrome (Android)
 - Browser Version: All versions
 - Amazon Chime SDK GitHub Commit: 7f7e612a001b918a20d8b170900b4705a2c86ed9

simmkyu:
I confirmed that in iOS Safari, `devicechange` is triggered.
- `DeviceChangeObserver.audioInputsChanged` is not called because the device list doesn't include my bluetooth device AirPods. Thus, there's no change between old and new lists.
- `DeviceChangeObserver.audioOutputsChanged` is not called because macOS and iOS Safari don't list audio output devices. Safari is using the device selected by a computer or an iOS device.

Try the following steps to validate the `devicechange` event in the demo app.
1. Open the SDK demo in iOS Safari.
2. In the Safari console, run this code to listen to the `devicechange` event.
    ```
    navigator.mediaDevices.addEventListener('devicechange', async () => {
      console.warn(await navigator.mediaDevices.enumerateDevices());
    });
    ```
3. Type a meeting title and name. Tap "Continue."
4. The device permisison dialog appears. Tap "Allow."
5. In the Safari console, you can see that `devicechange` is triggerd.

PR #503, opened 2020-06-22
Add a MediaDevices proxy to support the "devicechange" in all supported browsers

simmkyu:
**Issue #:** 
https://github.com/aws/amazon-chime-sdk-js/issues/445

**Description of changes:**
- Add a factory that creates the MediaDevices proxy. In a browser without the `devicechange` support, the proxy will poll a list of devices every one second and call listeners if the list changes. It will start polling only if there are one or more listeners.
- Why not using [the existing logic `DefaultDeviceController.handleDeviceChange()`](https://github.com/aws/amazon-chime-sdk-js/blob/8e5802c04031ac62a2c5c5fe75af5bb4ffbf12d0/src/devicecontroller/DefaultDeviceController.ts#L63)? (I could make it work by adding more conditions, but I prefer not to modify it.)
  - `handleDeviceChange()` calls `getUserMedia()` when a label doesn't exist.
  - `handleDeviceChange()` calls itself recursively to defer a task.
  - `handleDeviceChange()` uses all device info fields for comparison.

**Testing**

  Yes
  Tested by connecting and disconnecting devices.


simmkyu:
Thanks Trung and Andrew for reviewing. We decided to polyfill `MediaDevices.addEventListener` instead of adding a new component.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -456,6 +462,7 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
   private forEachObserver(observerFunc: (observer: DeviceChangeObserver) => void): void {
     for (const observer of this.deviceChangeObservers) {
       new AsyncScheduler().start(() => {
+        /* istanbul ignore else */
```
XHatan, line 465: what is "istanbul ignore else " ?

simmkyu: Istanbul will skip the `else` case for coverage. Unless we make `forEachObserver` public, it's impossible to cover the `else` case.

SDK is iterating listeners in `this.deviceChangeObservers`. In the next tick (`setTimeout` with zero delay), SDK checks whether the listener exists in `this.deviceChangeObservers`.

simmkyu:
Rebased

PR #522, opened 2020-06-28
Fixed setSinkId() from throwing DOMException in Chromium browsers

RidipDe:
**Issue #:** 
The audio output fails to switch after an active external audio device is disconnected from the system in the middle of a meeting. The chooseAudioOutputDevice() is used to select the external device as well as the fall back device after the external device disconnects using the corresponding hex device id in a Chromium based browser and not the using the ‘default’ deviceId.

chooseAudioOutputDevice(‘default’). → disconnect B/T → chooseAudioOutputDevice(<hex-id>) works fine.
chooseAudioOutputDevice(<bt_hex-id>). → disconnect B/T → chooseAudioOutputDevice(<any_hex-id>) throws DOMException
chooseAudioOutputDevice(<bt_hex-id>). → disconnect B/T → chooseAudioOutputDevice(‘default’) throws DOMException

Uncaught (in promise) DOMException: The operation could not be performed and was aborted

**Repro Steps:**
We were able to reproduce the issue. Steps involved were as follows: 

1. Connect any external audio output device (bluetooth headphones for eg) and start a meeting
2. Explicitly select bluetooth audio output device using the hex deviceId [not ‘default’]
3. Disconnect bluetooth from the system
4. Explicitly select MAC’s audio output device using the deviceId

**Description of changes:**
An external device is connected to an active meeting and is chosen using the chooseAudioOutputDevice() using its hex device id. The external device gets disconnected in the middle of the meeting. As a fall back mechanism, another audio output device is chosen from the list of available audio output devices using the hex device id as a parameter to chooseAudioOutputDevice(). The request fails to fulfill. The meeting is left with no active audio output stream.

In chromium based browsers, under the hood the audio element keeps pointing to the device associated to the active stream and continues to refer to the last active device even after the device gets disconnected. When the device disconnects, the audio element should release the stream holding the device and then update the sinkId to point to any of the available devices. After the sinkId is updated, the stream should be re-attached to the audio element.

In this PR, for a chromium based browser, the active audio output stream is nullified before setting up the sinkId using the setSinkId(deviceId). The same stream which was active before the setSinkId() call is restored after the call setSinkId() operation is fulfilled and the sinkId for the audio element is updated.

**Testing**

Yes
Manually, the setSinkId() does not throw the DOMException error with this change. 
Joined two meetings and performed the above mentioned repro steps and verified that the sinkId attribute of the audio element is updated. I was also able to listen to the sound from the desired audio output device. 


src/audiomixcontroller/DefaultAudioMixController.ts
```diff
@@ -51,8 +53,19 @@ export default class DefaultAudioMixController implements AudioMixController {
         // @ts-ignore
         const oldSinkId: string = this.audioElement.sinkId;
         if (newSinkId !== oldSinkId) {
-          // @ts-ignore
-          this.audioElement.setSinkId(newSinkId);
+          if (new DefaultBrowserBehavior().hasChromiumWebRTC()) {
```
asnyatkov, line 56: Instead of creating new instance every time the API is used, I suggest to create member variable and initialize it only once.

RidipDe: Fixed

RidipDe: Done, created an instance variable to use instead

jens-team-ai (commented):
I only tested this briefly, but at least on Chrome/Brave I cannot play a test sound on any non-default device. It errors with "deviceId not found". Digging into this a bit it appears that the Jitsi-reported deviceIds are different from what `navigator.mediaDevices.enumerateDevices()` returns. Not sure why and what to do about it (besides name matching instead of device Id matching).

PR #523, opened 2020-06-28
Remove "multiple" attribute from browser demo optional feature HTMLSelectElement 

XHatan:
**Issue #:** 

HTMLSelectElement causes some confusion on Android Chrome, even when user selects an element, on Android Chrome it showed "0 selected".

**Description of changes:**

**Testing**

yes
with demo on android chrome, smoke tested functionality with chrome/safari/firefox.

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.html
+++ b/demos/browser/app/meetingV2/meetingV2.html
@@ -45,10 +45,11 @@
       <div class="row mt-3">
         <div class="col-12">
           <label for="optional-features">Optional Features</label>
-          <select id="optional-features" class="custom-select" style="width:100%" multiple>
-            <option value="">--Please choose an option--</option>
+          <select id="optional-features" class="custom-select" style="width:100%">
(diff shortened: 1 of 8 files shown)


demos/browser/app/meetingV2/meetingV2.html
```diff
@@ -45,10 +45,11 @@ <h1 class="h3 mb-3 font-weight-normal">Join a meeting</h1>
       <div class="row mt-3">
         <div class="col-12">
           <label for="optional-features">Optional Features</label>
-          <select id="optional-features" class="custom-select" style="width:100%" multiple>
```
ltrung, line 48: Does that mean I cannot enable simulcast and unified plan option at the same time?

XHatan: That's right, it can not be enabled at the same time. The reason I added this is to allow easy testing for QA and the multiple selection confuses them.

Issue #479, opened 2020-06-15
 Cannot read property 'id' of undefined In angular 


mangalagiri:
In angular ----------> ERROR TypeError: Cannot read property 'id' of undefined at CreatePeerConnectionTask.push../node_modules/amazon-chime-sdk-js/build/task/CreatePeerConnectionTask.js.CreatePeerConnectionTask.addRemoteVideoTrack (CreatePeerConnectionTask.js:228)

![MadhuLatest](https://user-images.githubusercontent.com/18577663/84617401-6f3cec80-aeec-11ea-92cd-f585a1ee85f3.png)

**Meeting and attendee IDs:**
List the meeting and attendee IDs that the attached logs apply to.

**Timeline**
Outline the main actions the user performed and when:

* <timestamp> <action>
* <timestamp> <action>

**Describe the bug**
Describe the bug and the timestamp at which it happened relative to the timestamp in the log file.

**Screenshots**
If applicable, add screenshots to help explain the problem.

**Platform**
 - OS: Windows 
 - Browser: Chrome 
 - Browser Version:  Version 74.0.3729.108 (Official Build) (64-bit)
 - Amazon Chime SDK GitHub Commit: amazon-chime-sdk-js@1.7.5

simmkyu:
In https://github.com/aws/amazon-chime-sdk-js/issues/482, the attached screenshot shows that the session was connected okay. Are you still getting this `Cannot read property 'id' of undefined` error in your Angular application?

Two previous lines show that both audio and video tracks have valid IDs.

```
[INFO] SDK - received track event: kind=audio id=6af63725-9dbb-4cc0-8453-f332b24c837f label=6af63725-9dbb-4cc0-8453-f332b24c837f
[INFO] SDK - received track event: kind=video id=f033ede5-5848-4a26-b59f-548ce74a8828 label=f033ede5-5848-4a26-b59f-548ce74a8828
```

In the stack trace, the error originated from one of Angular dependencies: `zone-evergreen.js`. `zone-evergreen.js` from *Zone.js* patches web APIs in modern browsers. Have you included *Zone.js* patch for RTCPeerConnection APIs (`webapis-rtc-peer-connection.js`)?

dashawk:
I have the same issue above. I use webpack to bundle my app using `commonjs2` as libraryTarget.

![image](https://user-images.githubusercontent.com/1437205/85377017-149a3500-b56b-11ea-96a3-dc7a8aad50d4.png)

[...]

PR #538, opened 2020-07-06
Ensure video input is connected before configuring in SDP

alochbau-amzn:
**Issue #:** 
https://github.com/aws/amazon-chime-sdk-js/issues/479

**Description of changes:**
The SDP Offer option offerToReceiveVideo was being set to true if
the local video tile existed.  There is a small race condition where
the local video tile could exist but the associated transceiver input
hasn't been created yet.  The resulting behavior was a track event would
be received without a MediaStream which was throwing an error.

Change the setting of the offerToReceiveVideo flag based on the state
of the transceiver instead of the local tile.  Also add a check in the
track event to verify a stream exists and print a warning otherwise.

**Testing**
Enabled unified plan and enabled the local camera immediately upon receiving the canStartLocalVideoEvent.
Verified the track event was not received without a stream.  Tested with and without Unified Plan on Chrome and on the other supported browsers.

Yes
Modified Browser demo


alochbau-amzn:
I Squashed the merge commit.

XHatan:
Do we know what causes the race between attaching track to transceiver and creating offer ? The track attachment should be finished before creating SDP, I am wondering if we are ‘await’ on the right condition ?

alochbau-amzn:
The race condition is occurring just as we are getting ready to send the first SDP.
- Just as the client is joining the meeting the application calls startLocalVideoTile before any SDP is sent.
- The videoTileController adds the local video tile.
- The audioVideoController.update is called which triggers multiple tasks including AttachMediaInputTask and CreateSDPTask.  
- CreateSDPTask is querying the videoTileController to see if there is a local video tile, it is returning true, but in this case the AttachMediaInputTask hasn't gotten to the point of adding the video input to the transceiver.  The setting of that option then triggers a trackEvent without a stream

PR #548, opened 2020-07-14
Fix content share video to be displayed below the data message text box in the demo app

RidipDe:
**Issue #:** 
When the content share test video is turned on in the mobile browser, it appears behind the roaster in the landscape mode and does not display in the portrait mode, even though the content share actually keeps playing.

**Description of changes:**
 The content share test video is brought below the data message text box. Changed the meetingV2.html file and the display property of the content share video

Repro Steps:
    User joins in the meeting with PIN and name in landscape mode in mobile device.
    User clicks on the share content->test video

**Testing**
Yes
Manually testing and verifying it

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -1267,6 +1267,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
         if (videoUrl) {
           videoFile.src = videoUrl;
         }
+        videoFile.style.display = 'block';
         await videoFile.play();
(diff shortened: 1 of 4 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1267,6 +1267,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
         if (videoUrl) {
           videoFile.src = videoUrl;
         }
+        videoFile.style.display = 'block';
```
ltrung, line 1270: This breaks video integration tests. This is unrelated right?

PR #553, opened 2020-07-15
Add log to list the set of constraints supported by the browser

RidipDe:
**Issue #:** 
Currently. we are not logging the set of constraints that are supported by the browser. We need to log this to make sure that while choosing an input device, the builder can verify that the attributes passed in the proposedConstrains is an attribute in the constraints list that is supported by the browser.

**Description of changes:**
We use the [getSupportedConstraints](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getSupportedConstraints) to list all the supported browser constraints.

**Testing**

Yes
Manually

### DIFF
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -65,6 +65,10 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
       mediaDeviceWrapper.addEventListener('devicechange', () => {
         this.handleDeviceChange();
       });
+      const supportedConstraints = navigator.mediaDevices.getSupportedConstraints();
+      this.logger.info(
+        `Supported Constraints in this browser ${JSON.stringify(supportedConstraints)}`
+      );
     } catch (error) {
       logger.error(error.message);
     }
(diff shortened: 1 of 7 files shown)


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -676,6 +676,10 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     kind: string,
     device: Device
   ): MediaStreamConstraints | null {
+    let supportedConstraints = navigator.mediaDevices.getSupportedConstraints();
+    this.logger.warn(
```
simmkyu, line 680: - Log only once in the constructor since the list won't change
- Use `info()` as it's not a warning or error
- nit: Use `const`
- nit: Use template strings (``) over the plus sign (+) for concatenation

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -67,6 +67,10 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     } catch (error) {
       logger.error(error.message);
     }
+    const supportedConstraints = navigator.mediaDevices.getSupportedConstraints();
```
simmkyu, line 70: Sorry, I missed it in the previous revision. Could you put your line into the above `try/catch` block between line 66 and 67?
In a non-secured website, `navigator.mediaDevices` will be `undefined`.

PR #556, opened 2020-07-15
Add device change observer events when the current audio/video input stream ended

ltrung:
**Description of changes:**
Raise a device change observer event when the current audio/video input stream ended.
**Testing**

- Unit tests
- Test in Chrome and Firefox by manually calling stream track stop to trigger the ended event.
- Test in Chrome and Firefox by disconnecting bluetooth headphone and the ended event fires and the new observer events are triggered.
- Test in Chrome and Firefox when switching devices and the ended event does not fire and the new observer events are not triggered.
- Test in Chrome and Firefox when turning on/off camera that the ended event does not fire and the new observer events are not triggered.


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -454,6 +462,26 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     this.alreadyHandlingDeviceChange = false;
   }
 
+  private async handleDeviceStreamEnded(kind: string, deviceId: string): Promise<void> {
+    if (
+      kind === 'video' &&
+      this.boundAudioVideoController &&
+      this.boundAudioVideoController.videoTileController.hasStartedLocalVideoTile()
+    ) {
+      this.boundAudioVideoController.videoTileController.stopLocalVideoTile();
```
RidipDe, line 471: can we have an additional method for this functionality to stop the LocalVideoTile so it can be reused in releaseMediaStream()

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -454,6 +462,26 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     this.alreadyHandlingDeviceChange = false;
   }
 
+  private async handleDeviceStreamEnded(kind: string, deviceId: string): Promise<void> {
+    if (
```
simmkyu, line 466: I don't think this `if` statement is required. Calling `chooseInputDevice(video, null, false)` below will eventually stop local video tile.

Stack trace
1. `chooseInputDevice(video, null, false)`
2. `this.releaseMediaStream(this.activeDevices[kind].stream)`
3. In `releaseMediaStream`, you call stop `localVideoTile()` of matching active devices

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -522,6 +550,22 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
   }
 
+  private getActiveDeviceId(kind: string): string | null {
+    /* istanbul ignore else */
+    if (this.activeDevices[kind] && this.activeDevices[kind].constraints) {
+      const activeDeviceMediaTrackConstraints =
+        this.activeDevices[kind].constraints.audio || this.activeDevices[kind].constraints.video;
+      const activeDeviceConstrainDOMStringParameters = (activeDeviceMediaTrackConstraints as MediaTrackConstraints)
+        .deviceId;
+      const activeDeviceId = (activeDeviceConstrainDOMStringParameters as ConstrainDOMStringParameters)
```
simmkyu, line 560: - Can we reuse `getDeviceIdStr()`?
- @RidipDe @ltrung  In `getDeviceIdStr()` too, it's not a good idea to extract the `exact` property only.

   Public APIs `chooseAudioInputDevice` and `chooseVideoInputDevice` take the `Device` type as an argument. The `Device` type can be `string | MediaTrackConstraints | MediaStream | null`. Customers can call it with a non-exact consraint, e.g.

  ```
  // This will fail
  chooseAudioInputDevice({
     audio: {
        deviceId: { ideal: 'my-device-id }
     }
  })
  ```

  You can skip it in this PR but please create an issue for tracking. This code will also be broken by our change soon. We'll remove the `exact` keyword conditionally for the Safari 12.0.0 and 12.1.0 support.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -249,15 +249,23 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+          kind === 'video' &&
+          this.boundAudioVideoController &&
+          this.boundAudioVideoController.videoTileController.hasStartedLocalVideoTile()
+        ) {
+          this.boundAudioVideoController.videoTileController.stopLocalVideoTile();
+        }
       }
     }
+
+    for (const track of tracksToStop) {
```
simmkyu, line 265: Not a fan of code comments, but in this case, I think it's worth explaining why we need to set the active device `null` first and then stop tracks.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -454,6 +462,26 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
 
+  private async handleDeviceStreamEnded(kind: string, deviceId: string): Promise<void> {
+    if (
+      kind === 'video' &&
+      this.boundAudioVideoController &&
+      this.boundAudioVideoController.videoTileController.hasStartedLocalVideoTile()
+    ) {
+      this.boundAudioVideoController.videoTileController.stopLocalVideoTile();
+    }
+    await this.chooseInputDevice(kind, null, false);
```
simmkyu, line 473: `chooseInputDevice` is a long method. (Hope we can break it down soon.) I wonder if we want to put it into `try`/`catch` block.

RidipDe, line 473: Isn't this statement a repetition of https://github.com/aws/amazon-chime-sdk-js/blob/master/src/devicecontroller/DefaultDeviceController.ts#L593

We have an 'ended' event which is called when the device is disconnected which sets the device to null

src/devicecontroller/DefaultDeviceController.ts
ltrung: Yes, I confirmed it because I could never trigger that code to call if we put it after chooseInputDevice(null) but it depends on the value of this.activeDevices[kind].stream so I chose to keep it just in case. Let me know if you think it is totally unnecessary and I will remove it.

src/devicecontroller/DefaultDeviceController.ts
ltrung: I would prefer to leave it as is as it is how it was before and we usually use it like that in other places.

simmkyu: Before this PR, we are calling it asynchronously without `await`, so the next line is guaranteed to be called. In this PR, the observer lines won't be reached.

```
this.chooseInputDevice(kind, null, false);
if (
  kind === 'video' &&
  this.boundAudioVideoController &&
  this.boundAudioVideoController.videoTileController.hasStartedLocalVideoTile()
) {
  this.boundAudioVideoController.videoTileController.stopLocalVideoTile();
}
```

src/devicecontroller/DefaultDeviceController.ts
ltrung: We removed this in another place so it shows up in one place right now.

src/devicecontroller/DefaultDeviceController.ts
ltrung: Removed since handleDeviceStreamEnded only called when activeDevices[kind].stream exists so stopLocalVideoTile should always be called.

src/devicecontroller/DefaultDeviceController.ts
ltrung: Good catch. I will write up a GitHub issue and work on it after this PR since this is an existing code (getActiveDeviceId is in Ridip's previous commit) and we want to get this in our next NPM push.

src/devicecontroller/DefaultDeviceController.ts
ltrung: Added comment

src/devicecontroller/DefaultDeviceController.ts
simmkyu: Trung and I decided to skip the `try/catch` block here as `chooseInputDevice` is catching thrown errors from browser API calls.

src/devicecontroller/DefaultDeviceController.ts
RidipDe: Since our SDK was using the 'exact' keyword convention we decided to extract only the 'exact' property

RidipDe: Since exact and ideal have the same union type, we can add additional if condition to check the case where we have 'ideal' property instead of the 'exact'
    exact?: string | string[];
    ideal?: string | string[];

RidipDe: `if ((deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters).ideal) {
        /* istanbul ignore else */
        if (
          (deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters).ideal as string
        ) {
          return (deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters)
            .ideal as string;
        }
      }`

PR #566, opened 2020-07-19
Fix content share test video in Firefox

RidipDe:
**Issue #:** 
When the content share test video is started on a Firefox browser, we are not able to see the test video being played at the receivers end on a chrome/firefox browser. The captureStream() used is not supported by firefox browsers. Instead it needs to use mozCaptureStream()

**Issue Repro steps#:** 
1) User A starts a meeting from firefox and user B joins the same meeting from either chrome / firefox
2) User A starts content share test video, and hear the audio of the content share test video
3) User B is not able to view the test video
4) Neither user A nor user B is able to see the <<content share>> in the roaster.

**Description of changes:**
In this PR we are calling the mozCaptureMediaStream() on a video element with a MediaStream while the video element is playing.

**Testing**

Yes
Tested manually on Firefox v77.0.1 on MAC

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -18,6 +18,7 @@ import {
   DefaultMeetingSession,
   DefaultModality,
   Device,
+  DefaultBrowserBehavior,
   DeviceChangeObserver,
   LogLevel,
   Logger,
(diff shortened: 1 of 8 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1276,8 +1279,15 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
           videoFile.src = videoUrl;
         }
         await videoFile.play();
-        // @ts-ignore
-        const mediaStream: MediaStream = videoFile.captureStream();
+        let mediaStream: MediaStream;
+        if(this.defaultBrowserBehaviour.hasFirefoxWebRTC()) {
+          videoFile.play();
```
ltrung, line 1284: Do you need this? We call await videoFile.play() above?

PR #581, opened 2020-07-23
Add error message for TaskFailed errors

kaaaren:
**Issue #:** 

**Description of changes:**
Add error message for TaskFailed errors in DefaultAudioVideoController, so that a detailed error message will show after unknown TaskFailed warning 

**Testing**

Yes
By manually throw errors in tasks and checking console error messages in browser demo app

### DIFF
--- a/src/audiovideocontroller/AudioVideoController.ts
+++ b/src/audiovideocontroller/AudioVideoController.ts
@@ -47,7 +47,7 @@ export default interface AudioVideoController extends AudioVideoControllerFacade
   /**
    * Handles the meeting session status and returns true if it will restart the session.
    */
-  handleMeetingSessionStatus(status: MeetingSessionStatus): boolean;
+  handleMeetingSessionStatus(status: MeetingSessionStatus, error: Error | null): boolean;
 
   /**
    * Sets the max bandwidth for video publishing
--- a/src/audiovideocontroller/DefaultAudioVideoController.ts
+++ b/src/audiovideocontroller/DefaultAudioVideoController.ts
@@ -359,7 +359,7 @@ export default class DefaultAudioVideoController implements AudioVideoController
           this.getMeetingStatusCode(error) || MeetingSessionStatusCode.TaskFailed
         );
         await this.actionDisconnect(status, true);
-        if (!this.handleMeetingSessionStatus(status)) {
+        if (!this.handleMeetingSessionStatus(status, error)) {
           this.forEachObserver(observer => {
             Maybe.of(observer.audioVideoDidStop).map(f => f.bind(observer)(status));
           });
@@ -528,7 +528,7 @@ export default class DefaultAudioVideoController implements AudioVideoController
         if (status.statusCode() !== MeetingSessionStatusCode.IncompatibleSDP) {
           this.logger.info('failed to update audio-video session');
         }
-        this.handleMeetingSessionStatus(status);
+        this.handleMeetingSessionStatus(status, error);
       });
     }
   }
@@ -559,7 +559,6 @@ export default class DefaultAudioVideoController implements AudioVideoController
         this.logger.info('canceled retry');
       }
     );
-
     if (!willRetry) {
       this.sessionStateController.perform(SessionStateControllerAction.Fail, () => {
         this.actionDisconnect(status, false);
@@ -618,11 +617,10 @@ export default class DefaultAudioVideoController implements AudioVideoController
       // To perform the "Reconnect" action again, the session should be in the "Connected" state.
       this.sessionStateController.perform(SessionStateControllerAction.FinishConnecting, () => {
         this.logger.info('failed to reconnect audio-video session');
-        this.handleMeetingSessionStatus(
-          new MeetingSessionStatus(
-            this.getMeetingStatusCode(error) || MeetingSessionStatusCode.TaskFailed
-          )
+        const status = new MeetingSessionStatus(
+          this.getMeetingStatusCode(error) || MeetingSessionStatusCode.TaskFailed
         );
+        this.handleMeetingSessionStatus(status, error);
       });
     }
     this.connectionHealthData.setConnectionStartTime();
@@ -655,7 +653,7 @@ export default class DefaultAudioVideoController implements AudioVideoController
     }
   }
 
-  handleMeetingSessionStatus(status: MeetingSessionStatus): boolean {
+  handleMeetingSessionStatus(status: MeetingSessionStatus, error: Error | null): boolean {
     this.logger.info(`handling status: ${MeetingSessionStatusCode[status.statusCode()]}`);
     if (!status.isTerminal()) {
       if (this.meetingSessionContext.statsCollector) {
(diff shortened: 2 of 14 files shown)


src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -618,11 +621,10 @@ export default class DefaultAudioVideoController implements AudioVideoController
...
       this.sessionStateController.perform(SessionStateControllerAction.FinishConnecting, () => {
         this.logger.info('failed to reconnect audio-video session');
-        this.handleMeetingSessionStatus(
-          new MeetingSessionStatus(
-            this.getMeetingStatusCode(error) || MeetingSessionStatusCode.TaskFailed
-          )
+        const status = new MeetingSessionStatus(
+          this.getMeetingStatusCode(error) || MeetingSessionStatusCode.TaskFailed
         );
+        this.handleMeetingSessionStatus(status, error.message);
```
ltrung, line 627: Do we have error.name here? If so, should we pass it in?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -559,8 +559,11 @@ export default class DefaultAudioVideoController implements AudioVideoController
         this.logger.info('canceled retry');
       }
     );
-
     if (!willRetry) {
+      errorMessage &&
+        this.logger.error(
+          `failure message of ${MeetingSessionStatusCode[status.statusCode()]}: ${errorMessage}`
```
anuranduttaroy, line 565: Can we rephrase the start of this error message? Something like 

`reconnection failed with status code ${MeetingSessionStatusCode[status.statusCode()]}: ${errorMessage}`

src/audiovideocontroller/DefaultAudioVideoController.ts
kaaaren: That's a good call, I tested with manually throw errors, so error.name is just "Error" plus the error status code is added. I can add error.name to provide more info.

src/audiovideocontroller/DefaultAudioVideoController.ts
kaaaren: Yeah sure, we can rephrase the error message, but I'm not sure if `reconnection failed..` is accurate here as the error message could be from actionConnect or actionUpdate failure

src/audiovideocontroller/AudioVideoController.ts
```diff
@@ -42,12 +42,12 @@ export default interface AudioVideoController extends AudioVideoControllerFacade
...
    * Restarts the peer connection and/or the session.
    */
-  reconnect(status: MeetingSessionStatus): boolean;
+  reconnect(status: MeetingSessionStatus, errorMessage?: string | null): boolean;
 
   /**
    * Handles the meeting session status and returns true if it will restart the session.
    */
-  handleMeetingSessionStatus(status: MeetingSessionStatus): boolean;
+  handleMeetingSessionStatus(status: MeetingSessionStatus, errorMessage?: string | null): boolean;
```
simmkyu, line 50: - Nice idea. Instead of a string, how about taking an Error object? In a place where you log, you can access `error.message` (and more properties such as `error.name` and `error.stack` later)

- Recommend removing `?`. Although we are using an optional argument (`?`) in several places, we prefer a required argument over `?`.  In case of no error from the caller, they can pass `null`.

  There will just be three more calls in `MonitorTask`, one of which has an error object too!

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -559,8 +559,11 @@ export default class DefaultAudioVideoController implements AudioVideoController
         this.logger.info('canceled retry');
       }
     );
-
     if (!willRetry) {
+      errorMessage &&
```
simmkyu, line 563: How about logging this error message in `handleMeetingSessionStatus` where we log several messages? You can do the following:

```
const willRetry = this.reconnect(status, errorMessage);
if (!willRetry) {
  // Log here
}
return willRetry;
```

src/audiovideocontroller/DefaultAudioVideoController.ts
simmkyu: Recommend passing an Error object instead of a string. See a comment above.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -688,7 +690,7 @@ export default class DefaultAudioVideoController implements AudioVideoController
     }
     if (status.isFailure() || status.isTerminal()) {
       if (this.meetingSessionContext.reconnectController) {
-        return this.reconnect(status);
+        return this.reconnect(status, errorMessage);
```
simmkyu, line 693: https://github.com/aws/amazon-chime-sdk-js/blob/a4ee269d18b2f796494b38b436e467ebf9ca5fd3/src/audiovideocontroller/DefaultAudioVideoController.ts#L678-L682

- How about removing this warning above and logging it like the following?

```
if (status.isFailure() || status.isTerminal()) {
  if (this.meetingSessionContext.reconnectController) {
    const willRetry = this.reconnect(status);
    if (willRetry) {
      this.logger.warn(`will retry due to status code ${MeetingSessionStatusCode[status.statusCode()]}${error ? ` and error: ${error.message}` : ``}`);
    } else {
      this.logger.error(`failed with status code ${MeetingSessionStatusCode[status.statusCode()]}${error ? ` and error: ${error.message}` : ``}`);
    }
    return willRetry;
  }
}
```

src/audiovideocontroller/DefaultAudioVideoController.ts
kaaaren: Yep, sounds good, I'll update in the next revision

src/audiovideocontroller/AudioVideoController.ts
kaaaren: 1) Yep, passing Error is more flexible, will change in the next revision. 2) Will remove optional argument and add error in MonitorTask

src/audiovideocontroller/DefaultAudioVideoController.ts
kaaaren: Yep, removed the status.isFailure() warning and moved there in the new revision

src/audiovideocontroller/DefaultAudioVideoController.ts
kaaaren: Yeah, updated, it's more cleaner this way.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -618,11 +617,10 @@ export default class DefaultAudioVideoController implements AudioVideoController
       // To perform the "Reconnect" action again, the session should be in the "Connected" state.
       this.sessionStateController.perform(SessionStateControllerAction.FinishConnecting, () => {
         this.logger.info('failed to reconnect audio-video session');
-        this.handleMeetingSessionStatus(
-          new MeetingSessionStatus(
-            this.getMeetingStatusCode(error) || MeetingSessionStatusCode.TaskFailed
-          )
+        const status = new MeetingSessionStatus(
```
simmkyu, line 620: nit: This change is not required

PR #585, opened 2020-07-24
Avoid pixel3(XL) encoder issue with size not 16's multiples

yixial-amzn:
Ensure uplink resolution dimensions are aligned

**Issue #:** 
When simulcast on Chrome is enabled, Pixel3(XL) encoder has issue with width that are not 16's multiples. Remote sites will see stripped video from Pixel 3(XL)

**Description of changes:**
Make sure chosen resolution in simulcast uplink have height aligned.

**Testing**

1. Have you successfully run `npm run build:release` locally?
Yes

2. How did you test these changes?
With local js-sdk demo server, add pixel3XL and a few other mac clients (chrome, 2 Firefox and Safari) into a same chime meeting. Verified on webrtc-internals that all remote sites see live pixel3XL video without artifacts.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #587, opened 2020-07-27
Revert "Avoid pixel3(XL) encoder issue with size not 16's multiples"

yixial-amzn:
Reverts aws/amazon-chime-sdk-js#585


yixial-amzn:
Revert this change as js-sdk plans to do a NPM today.

PR #589, opened 2020-07-29
Fix race condition in Chromium browsers when consecutive audio bind operations take place 

RidipDe:
**Issue #:** 
The sound test button in the 'Select Device' page does not respond in chromium based browsers. In the demo app's [TestSound class](https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/browser/app/meetingV2/meetingV2.ts#L70) we see that the audio output bind operations are called consecutively. Upon investigation its found that for chromium browsers, there needs to be a slight delay between binding the audio device and the audio element stream. The order to call these 3 methods should be independent to the SDK

We recently made a modification in our SDK to set the sinkId a bit differently for chromium based browsers to handle an edge case which resulted in DOMException. The bindAudioMix() is a non async function and as per the [publicly available setSinkId documentation](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/setSinkId) setSinkId() returns a Promise. We had to call the setSinkId() in the bindAudioMix in a AsyncScheduler block. This allows async execution of the block of code.

*Root cause*
When a call to the setSinkId() is made, under the hood chromium based browsers expects the audioElement srcObject to be null while setting up the sinkId as discussed in the PR #522 
With this AsyncScheduler call in the SDK and the 3 consecutive calls to the bind methods in the demo app, a race condition would set up the audio stream (srcObject) of the audio element before the sinkId is set. Thus we cannot listen to the TestSound when the sound test button in the demo app is clicked on any chromium based browser.

**Description of changes:**
Two changes have been made in this PR. 
1) In order to avoid the unexpected race condition, we removed the AsyncScheduler from the the non async method bindAudioMix.
2) For chromium based browsers, in order to accommodate some time for the setSinkId() operation to complete before the stream is attached back to the srcObject of the audio element, then() was introduced over the setSinkId()

**Testing**

Yes

I have tested the following three major scenarios on the browser:
1) Being able listen to the beep upon clicking the Test speaker button on the device selection page of the demo app. Thus testing that the consecutive calls to the bind operations does not result in race condition.
2) The DOMException does not occur as tested for PR #522.
3) We are able to switch between audio output devices while in an active meeting.

### DIFF
--- a/src/audiomixcontroller/DefaultAudioMixController.ts
+++ b/src/audiomixcontroller/DefaultAudioMixController.ts
@@ -3,7 +3,6 @@
 
 import BrowserBehavior from '../browserbehavior/BrowserBehavior';
 import DefaultBrowserBehavior from '../browserbehavior/DefaultBrowserBehavior';
-import AsyncScheduler from '../scheduler/AsyncScheduler';
 import AudioMixController from './AudioMixController';
 
 export default class DefaultAudioMixController implements AudioMixController {
@@ -56,13 +55,14 @@ export default class DefaultAudioMixController implements AudioMixController {
         const oldSinkId: string = this.audioElement.sinkId;
         if (newSinkId !== oldSinkId) {
           if (this.browserBehavior.hasChromiumWebRTC()) {
-            new AsyncScheduler().start(async () => {
-              const existingStream = await this.audioElement.srcObject;
-              this.audioElement.srcObject = null;
-              // @ts-ignore
-              await this.audioElement.setSinkId(newSinkId);
-              this.audioElement.srcObject = existingStream;
-              this.audioStream = existingStream as MediaStream;
+            const existingAudioElement = this.audioElement;
+            const existingstream = this.audioStream;
+            existingAudioElement.srcObject = null;
+            // @ts-ignore
+            existingAudioElement.setSinkId(newSinkId).then(() => {
+              if (this.audioElement === existingAudioElement) {
+                existingAudioElement.srcObject = existingstream;
+              }
             });
           } else {
             // @ts-ignore
--- a/test/audiomixcontroller/DefaultAudioMixController.test.ts
+++ b/test/audiomixcontroller/DefaultAudioMixController.test.ts
@@ -4,6 +4,7 @@
 import * as chai from 'chai';
 
 import DefaultAudioMixController from '../../src/audiomixcontroller/DefaultAudioMixController';
+import TimeoutScheduler from '../../src/scheduler/TimeoutScheduler';
 import DOMMockBehavior from '../dommock/DOMMockBehavior';
 import DOMMockBuilder from '../dommock/DOMMockBuilder';
 
@@ -72,13 +73,15 @@ describe('DefaultAudioMixController', () => {
     expect(defaultAudioMixController.bindAudioStream(stream)).to.equal(true);
   });
 
-  it('can successfully bind and sink an audio stream in chromium based browser', () => {
+  it('can successfully bind and sink an audio stream in chromium based browser', async () => {
     behavior.browserName = 'chrome';
     domMockBuilder = new DOMMockBuilder(behavior);
     defaultAudioMixController = new DefaultAudioMixController();
     defaultAudioMixController.bindAudioDevice(device);
     defaultAudioMixController.bindAudioElement(element);
+    await new Promise(resolve => new TimeoutScheduler(50).start(resolve));
     expect(defaultAudioMixController.bindAudioStream(stream)).to.equal(true);
+    expect(element.srcObject).to.equal(stream);
   });
 
   it('can bind an audio device, but not yet sink', () => {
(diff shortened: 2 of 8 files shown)


src/audiomixcontroller/DefaultAudioMixController.ts
```diff
@@ -56,13 +56,12 @@ export default class DefaultAudioMixController implements AudioMixController {
...
-            new AsyncScheduler().start(async () => {
-              const existingStream = await this.audioElement.srcObject;
-              this.audioElement.srcObject = null;
-              // @ts-ignore
-              await this.audioElement.setSinkId(newSinkId);
+            const existingStream = this.audioElement.srcObject;
+            this.audioElement.srcObject = null;
+            // @ts-ignore
+            this.audioElement.setSinkId(newSinkId);
+            new TimeoutScheduler(50).start(() => {
```
simmkyu, line 63: Using any timeout-related API within a non-async function is error-prone. These are some examples.

This code will throw an error:
```
mixController.bindAudioElement(elementA);
mixController.unbindAudioElement();
```

`elementA` will have `elementB`'s stream, while the user just want to swich the element.
```
mixController.bindAudioElement(elementA);
mixController.bindAudioElement(elementB);
```

---

The `setSink` failure on Chrome should be a common error. I wonder if there's another solution rather than nullifying it. Do you have the Chromium bug link that says `srcObject` should be `null`?

RidipDe: It was based out of experimentation and having the following minimal repro as follows:

//create audio
audio = document.createElement('audio');
audio.autoplay = true

//set sinkId to the bluetooth device
await audio.setSinkId(<bt_hex-id>)
console.log('Audio is being played on ' + audio.sinkId);

//set new media stream
 audio.srcObject= new MediaStream()
 
//confirm current sinkId
audio.sinkId
 
//disconnect bluetooth and set the sinkId to the mac device
await audio.setSinkId(<mac_hex-id>)
console.log('Audio is being played on ' + audio.sinkId);

while the audio.srcObject is not null, and when we try to set up the sinkId, we will see a DOMException

ltrung: Agree that we should avoid setTimeout if possible. It seems hacky.

test/audiomixcontroller/DefaultAudioMixController.test.ts
```diff
@@ -95,6 +95,14 @@ describe('DefaultAudioMixController', () => {
     expect(defaultAudioMixController.bindAudioDevice(device)).to.equal(true);
   });
 
+  it('can successfully set audio element and unbind element', () => {
```
simmkyu, line 98: Nice try, but it won't test anything for two reasons.

#### Reason 1: DOMMockBuilder

In DOMMockBuilder, `setSinkId` just sets but returns immediately.

https://github.com/aws/amazon-chime-sdk-js/blob/218330d4046fd18b30cd488a12b0babee00a31ae/test/dommock/DOMMockBuilder.ts#L856-L862
  
Try adding the artificial delay. As you discovered, it will set `sinkId` and then take some time to finish. (Optional: You can even disable setting `srcObject` while this function is running to simulate the exact behavior of Chrome.)

```
setSinkId(deviceId: string): Promise<undefined> {
  this.sinkId = deviceId;
  await new Promise(resolve => setTimeout(resolve, mockBehavior.asyncWaitMs));
  return undefined;
}
```

#### Reason 2: Async

This `it` test will immediately finish after running `defaultAudioMixController.unbindAudioElement();`. Note that `bindAudioElement` and `unbindAudioElement` have the side effect unfortunately.

Convert your `it` to the `async` function, then add some delay, and then use `expect` to verify.

```
defaultAudioMixController.bindAudioElement(element);
defaultAudioMixController.unbindAudioElement();
await new Promise(resolve => new TimeoutScheduler(1000).start(resolve));

expect(element.srcObject).to.equal(null);
```

src/audiomixcontroller/DefaultAudioMixController.ts
```diff
@@ -56,13 +55,13 @@ export default class DefaultAudioMixController implements AudioMixController {
...
         if (newSinkId !== oldSinkId) {
           if (this.browserBehavior.hasChromiumWebRTC()) {
-            new AsyncScheduler().start(async () => {
-              const existingStream = await this.audioElement.srcObject;
-              this.audioElement.srcObject = null;
-              // @ts-ignore
-              await this.audioElement.setSinkId(newSinkId);
-              this.audioElement.srcObject = existingStream;
-              this.audioStream = existingStream as MediaStream;
+            const existingStream = this.audioElement.srcObject;
```
simmkyu, line 58: What will happen if the user changes mind and calls it with another element?
```
bindAudioStream(streamA);
bindAudioElement(elementA);
// changed mind
bindAudioStream(streamB);
bindAudioElement(elementB);

// elementA.srcObject will reference streamB!
```

Instead of the stream, how about saving the element itself temporarily? In the callback, you can do the following:
```
const existingAudioElement = this.audioElement;
this.audioElement.srcObject = null;
// @ts-ignore
this.audioElement.setSinkId(newSinkId).then(() => {
  if (this.audioElement === existingAudioElement) {
     this.audioElement.srcObject = this.audioStream;
  }
});
```

src/audiomixcontroller/DefaultAudioMixController.ts
RidipDe: Yeah, we had AsyncScheduler before for setting sinkId. Removed that and now in order to achieve a slight delay before the stream is reattached to the srcObject I had to use TimeoutScheduler.
Trying to figure out an alternative approach.

src/audiomixcontroller/DefaultAudioMixController.ts
```diff
@@ -56,13 +55,13 @@ export default class DefaultAudioMixController implements AudioMixController {
...
+            this.audioElement.srcObject = null;
+            // @ts-ignore
+            this.audioElement.setSinkId(newSinkId).then(() => {
```
simmkyu, line 61: What will happen when `setSinkId` fails?

In the current PR, `audioElement.srcObject` will be set to `null`. You can catch the error and restore `srcObject` but it'll be difficult to figure out the error as this compoent doens't have any logger.

src/audiomixcontroller/DefaultAudioMixController.ts
RidipDe: Great Catch

src/audiomixcontroller/DefaultAudioMixController.ts
RidipDe: Yeah if we catch the exception thrown by setSinkId(), we might miss on errors generated from the setSinkId() eg the DOMException

RidipDe:
> Nice! Please make sure to test possible race conditions and basic scenarios in a real browser too. Probably, using `app.audioVideo.audioMixController` in the demo app will be helpful.

I have tested the following three major scenarios on the browser:
1) Being able listen to the beep upon clicking the Test speaker button on the device selection page of the demo app. Thus testing that the consecutive calls to the bind operations does not result in race condition.
2) The DOMException does not occur as tested for PR #522 
3) We are able to switch between audio output devices while in an active meeting

PR #595, opened 2020-07-30
Fix invalid constraints and disable Unified Plan in safari 12.0

kaaaren:
**Issue #:** 
This is fix the issue that user is unable to enable video in iPad Safari 12.1 version. 
The first `failed to get audio device for constraints { __SDK_CONSTRAINTS__ } Invalid constraint` is due to the WebKit bug: https://bugs.webkit.org/show_bug.cgi?id=187933

The second error `AttachMediaInputTask error: Attempted to assign to readonly property.` is caused by that unified plan is not supported in Safari 12, according to https://webkit.org/blog/8672/on-the-road-to-webrtc-1-0-including-vp8/, SDP Unified Plan is turned off by default in iOS 12.2.

**Description of changes:**
Fix invalid constraints and disable Unified Plan in safari 12.0

**Testing**

Yes
Tested iPad 12.1.4 (Safari 12.0.0) and MacOS 12.0.0 using SauceLabs:
1. Verified in ios safari 12.0.0, that there is no `Invalid constraint` error and plan-b is used. User is able to turn on video and browser doesn't crash and no error is found.
2. Verified in macOS safari 12.0.0, that there is no `Invalid constraint` error and plan-b is used. But no camera is available for the destop device with saucelab


test/browserbehavior/DefaultBrowserBehavior.test.ts
```diff
@@ -122,41 +127,30 @@ describe('DefaultBrowserBehavior', () => {
...
-      expect(
-        new DefaultBrowserBehavior({
-          enableUnifiedPlanForChromiumBasedBrowsers: !enableUnifiedPlan,
-        }).requiresUnifiedPlan()
-      ).to.eq(true);
-      expect(
-        new DefaultBrowserBehavior({
-          enableUnifiedPlanForChromiumBasedBrowsers: !enableUnifiedPlan,
-        }).requiresUnifiedPlanMunging()
-      ).to.eq(true);
```
kaaaren, line 145: Unit test for requiresUnifiedPlan() is removed here as `RTCRtpTransceiver.prototype.hasOwnProperty('currentDirection');` will always be false here

src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -188,4 +192,9 @@ export default class DefaultBrowserBehavior implements BrowserBehavior {
   private isPixel3(): boolean {
     return /( pixel 3)/i.test(navigator.userAgent);
   }
+
+  private isUnifiedPlanSupported(): boolean {
+    const res = RTCRtpTransceiver.prototype.hasOwnProperty('currentDirection');
```
anuranduttaroy, line 197: Can we have just
```
return RTCRtpTransceiver.prototype.hasOwnProperty('currentDirection');
```

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -555,6 +555,10 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
           return (deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters)
             .exact as string;
         }
+      } else {
```
ltrung, line 558: Do we need to do the same in getActiveDeviceId?

src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -117,6 +117,10 @@ export default class DefaultBrowserBehavior implements BrowserBehavior {
     return this.isSafari();
   }
 
+  requiresNoExactMediaStreamConstraints(): boolean {
+    return this.isIOSSafari() && (this.version() === '12.0.0' || this.version() === '12.1.0');
```
ltrung, line 121: Is this only for iOS Safari or can it be for desktop Safari?
Also is this for any 12.0+ and 12.1+ or just 12.0.0 and 12.1.0?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -555,6 +555,10 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
           return (deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters)
```
simmkyu, line 555: https://github.com/aws/amazon-chime-sdk-js/blob/03a195883a5bd306bdbd1e87126e4070dc7bdc4e/src/devicecontroller/DefaultDeviceController.ts#L550

This line will fail if `deviceId` is a string because it's trying to access `exact` on the `string`.  Please test `this.getDeviceIdStr` with different types of inputs.

```
if (typeof deviceMediaTrackConstraints.deviceId === 'string') {
  return deviceMediaTrackConstraints.deviceId;
} else if ('exact' in deviceMediaTrackConstraints.deviceId) {
  return deviceMediaTrackConstraints.deviceId.exact;
}
```

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -555,6 +555,10 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
             .exact as string;
         }
+      } else {
+        if (deviceMediaTrackConstraints.deviceId as string) {
```
simmkyu, line 559: I understand you followed the pattern in this function. If I'm not missing TypeScript features, using `as` in `if` doesn't have any effect. @RidipDe 

[Click here to try in the TS playgrond](https://www.typescriptlang.org/play/?target=5&ssl=1&ssc=1&pln=7&pc=2#code/JYMwBAFA5FYIYGcwIC4CdgDsDmBKMA3gFBilgDGA9pgpQDYCmAdHZdtAHKUphoNzkAFgwAmUXAG4iAXyJFQkFAE8ADg0rgYYALy6wUVBhzjCJMlRr1mrdlAAqg4EifwKlNH3I8A7nCXiZIA):
```
if ('' as string) {
    console.log('Not reached');
}

if (typeof '' === 'string') {
    console.log('This is a correct way')
}
```

src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -67,7 +67,7 @@ export default class DefaultBrowserBehavior implements BrowserBehavior {
   }
 
   requiresUnifiedPlan(): boolean {
-    let shouldEnable = this.isSafari() || this.isFirefox();
+    let shouldEnable = (this.isSafari() && this.isUnifiedPlanSupported()) || this.isFirefox();
```
XHatan, line 70: it's safer to check isUnifiedPlanSupported for all browsers and not just for safari.

src/devicecontroller/DefaultDeviceController.ts
simmkyu: @kaaaren I'm looking into two methods `getDeviceIdStr` and `getActiveDeviceId` now and will share you with some suggestions today.

simmkyu: The custom type ```Device``` can be `string | MediaTrackConstraints | MediaStream | null`. In case the `device` argument is `MediaTrackConstraints`, `device.deviceId` can be:
|Type|Handled by SDK?|
|--|--|
| a string | ✅ Added and handled by this PR |
| a string array | |
| an object with a string `exact` | ✅ |
| an object with a string array `exact` | |
| an object with a string `ideal` | |
| an object with a string array `ideal` | |

We are missing 4 other cases. In this PR, let's handle your string change. I'll create an issue for this.

```
private getDeviceIdStr(device: Device): string | null {
  if (device === null) {
    return null;
  } else if (typeof device === 'string') {
    return device;
  } else if ((device as MediaStream).id) {
    return (device as MediaStream).id;
  }

  const constraints: MediaTrackConstraints = device as MediaTrackConstraints;
  if (!constraints.deviceId) {
    return '';
  } else if (typeof constraints.deviceId === 'string') {
    return constraints.deviceId;
  }

  const deviceIdConstraint: ConstrainDOMStringParameters = constraints.deviceId as ConstrainDOMStringParameters;
  if (typeof deviceIdConstraint.exact === 'string') {
    return deviceIdConstraint.exact;
  }

  return '';
}
```


Need to update  `getActiveDeviceId` too

```
private getActiveDeviceId(kind: string): string | null {
  /* istanbul ignore else */
  if (this.activeDevices[kind] && this.activeDevices[kind].constraints) {
    const activeDeviceMediaTrackConstraints =
      this.activeDevices[kind].constraints.audio || this.activeDevices[kind].constraints.video;
    const activeDeviceConstrainDOMStringParameters = (activeDeviceMediaTrackConstraints as MediaTrackConstraints)
      .deviceId;

    let activeDeviceId: string;
    if (typeof activeDeviceConstrainDOMStringParameters === 'string') {
      activeDeviceId = activeDeviceConstrainDOMStringParameters as string;
    } else {
      activeDeviceId = (activeDeviceConstrainDOMStringParameters as ConstrainDOMStringParameters).exact as string;
    }

    return activeDeviceId;
  }
  /* istanbul ignore next */
  return null;
}
```

ltrung: Nice job Kyu. Very through comment 👍

src/browserbehavior/DefaultBrowserBehavior.ts
kaaaren: Yep, will update

src/browserbehavior/DefaultBrowserBehavior.ts
kaaaren: Yeah, I tested with macOS safari 12.0 and it has a exact for deviceId.

src/devicecontroller/DefaultDeviceController.ts
kaaaren: Good catch, and I see Kyu's comments below, I'll update accodingly

src/browserbehavior/DefaultBrowserBehavior.ts
kaaaren: Will add the check there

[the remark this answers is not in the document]
kaaaren: Ok, will do

src/devicecontroller/DefaultDeviceController.ts
kaaaren: Thanks for the detailed explanation!! Will update

RidipDe: Yeah while writing this snippet we took into consideration only the constrains types being used in our DefaultDeviceController in our SDK like the string, null and the one with the exact keyword. 
We use them in the proposed constraints for the parameter in getUserMedia()

src/devicecontroller/DefaultDeviceController.ts
RidipDe: I read about union type in typescript here: https://www.typescriptlang.org/docs/handbook/advanced-types.html
and implemented accordingly. Its not just checking the type in the above implementation. Union types are useful for modeling situations when values can overlap in the types they can take on.

deviceMediaTrackConstraints.deviceId can be either
``` string | string[] | ConstrainDOMStringParameters```

The scenario already handled is if deviceMediaTrackConstraints.deviceId  is referred as ConstrainDOMStringParameters then get the string `exact` attribute. 

We have just handled this condition as of now because in our SDK, we are currently assigning the device id as a `string` as an `exact` property in this line https://github.com/aws/amazon-chime-sdk-js/blob/master/src/devicecontroller/DefaultDeviceController.ts#L731

We cannot do 
```
if (typeof deviceMediaTrackConstraints.deviceId === ConstrainDOMStringParameters)

as 'ConstrainDOMStringParameters' only refers to a type and not value
```

As per need we need to keep adding the checks
example against a string and string[] and ideal
For checking if its type of 'string' or not we need to use this
```
if (typeof ' ' === 'string') {
    console.log('This is a correct way')
}
```


If we want the `ideal` string property of the device, we need to add
```
      /* istanbul ignore else */
      if ((deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters).ideal) {
        /* istanbul ignore else */
        if (
          (deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters).ideal as string
        ) {
          return (deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters)
            .ideal as string;
        }
      }
```

Similarly, as per the need, we need to handle string[] for properties `exact` and `ideal`

Comment (simmkyu) [reply src/devicecontroller/DefaultDeviceController.ts:559]: > If we want the ideal string property of the device, we need to add

You might be confused of the type conversion (`as string`) vs. the type check. Assume that `deviceId` is a string array. The following code will still return a string array.

```
/* istanbul ignore else */
if ((deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters).ideal) {
  /* istanbul ignore else */
  if (
    (deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters).ideal as string
  ) {
    return (deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters)
      .ideal as string;
  }
}
```

The 2nd `if` statement doesn't check whether the condition is a string or not, but it will be `true` if the condition is a non-falsy item. i.e. It's no different from having only one `if` statement.

```
if ((deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters).ideal) {
  return (deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters)
    .ideal as string;
}
```

To check if the `ideal` type is a string, you can use `typeof`.
```
if (typeof (deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters).ideal === 'string') {
  return (deviceMediaTrackConstraints.deviceId as ConstrainDOMStringParameters)
    .ideal as string;
}
```

File: test/dommock/DOMMockBuilder.ts [hunk 8]
```diff
@@ -736,6 +742,39 @@ export default class DOMMockBuilder {
       }
     };
 
+    const mockRTCRtpTransceiverWithCurrentDirection = class MockRTCRtpTransceiverWithCurrentDirection {
```
simmkyu, line 745: I don't think you need to create the full clone of `MockRTCRtpTransceiver` but how about just setting it once?

```
if (mockBehavior.isUnifiedPlanSupported) {
  MockRTCRtpTransceiver.prototype.currentDirection = 'sendrecv';
}
```

Note that `RTCRtpTransceiver.prototype.hasOwnProperty('currentDirection');` will be `true` only when you explicitly set the prototype like the above.

src/browserbehavior/DefaultBrowserBehavior.ts
simmkyu: @XHatan I think @kaaaren wrote the Safari Unified Plan checker based on this WebKit article:
https://webkit.org/blog/8672/on-the-road-to-webrtc-1-0-including-vp8/

We might need another Unified Plan checker for Chrome using `addTransceiver` but only if we plan to support lower versions: https://chromestatus.com/feature/4749621767176192

src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -188,4 +192,9 @@ export default class DefaultBrowserBehavior implements BrowserBehavior {
   private isPixel3(): boolean {
     return /( pixel 3)/i.test(navigator.userAgent);
   }
+
+  //To detect whether Safari uses Unified Plan
+  private usesUnifiedPlan(): boolean {
```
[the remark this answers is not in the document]
kaaaren: yep, updated

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -176,6 +176,14 @@ describe('DefaultDeviceController', () => {
   });
 
   describe('chooseAudioInputDevice', () => {
+    const IOS_SAFARI12_0_USERAGENT =
```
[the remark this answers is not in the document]
kaaaren: Thanks, updated to remove `setUserAgent` and use `DOMMockBehavior`

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -224,7 +232,10 @@ describe('DefaultDeviceController', () => {
 
     it('attaches the audio input stream to the audio context', async () => {
       deviceController.enableWebAudio(true);
-      let permission = await deviceController.chooseAudioInputDevice(stringDeviceId);
+      const device = {
```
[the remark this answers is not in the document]
kaaaren: Great catch, you are right it's not testing the code. I've reverted the change and added a new test case for string `deviceId`

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -302,6 +313,46 @@ describe('DefaultDeviceController', () => {
         done();
       });
     });
+
+    it('releases all previously-acquired audio streams with iOS 12', done => {
+      setUserAgent(IOS_SAFARI12_0_USERAGENT);
```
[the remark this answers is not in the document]
kaaaren: Gotcha, updated to use `DOMMockBehavior` and remove `setUserAgent`

test/browserbehavior/DefaultBrowserBehavior.test.ts
```diff
@@ -115,48 +124,76 @@ describe('DefaultBrowserBehavior', () => {
 
     it('can detect Safari', () => {
       setUserAgent(SAFARI_USERAGENT);
+      domMockBehavior = new DOMMockBehavior();
+      domMockBehavior.usesUnifiedPlan = true;
+      domMockBehavior.browserName = 'safari';
```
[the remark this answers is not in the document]
kaaaren: Gotcha, I updated to use either `setUserAgent` or `domMockBehavior` in DefaultDeviceController.test. But here using `setUserAgent` only won't work because we need to set `domMockBehavior.isUnifiedPlanSupported = false;` for safari 12 & 11 and ios safari 12.0 & 12.1 in domMockBehavior, so that we can turn off unified plan in DOMMockBuilder. So I use `DOMMockBehavior` instead here. 
Let me know if there is a cleaner way to do that :)

test/browserbehavior/DefaultBrowserBehavior.test.ts
```diff
@@ -115,48 +124,76 @@ describe('DefaultBrowserBehavior', () => {
...
       expect(new DefaultBrowserBehavior().requiresBundlePolicy()).to.eq('max-bundle');
+      domMockBehavior = new DOMMockBehavior();
+      domMockBehavior.usesUnifiedPlan = false;
+      domMockBehavior.browserName = 'safari';
+      mockBuilder = new DOMMockBuilder(domMockBehavior);
+      expect(new DefaultBrowserBehavior().requiresNoExactMediaStreamConstraints()).to.eq(false);
+      expect(new DefaultBrowserBehavior().requiresUnifiedPlan()).to.eq(false);
+      expect(new DefaultBrowserBehavior().requiresUnifiedPlanMunging()).to.eq(false);
+
+      setUserAgent(IOS_SAFARI12_0_USERAGENT);
```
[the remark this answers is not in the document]
kaaaren: Good call, I've put each safari version into it's own test block.

test/dommock/DOMMockBuilder.ts
kaaaren: Good point, updated.

test/browserbehavior/DefaultBrowserBehavior.test.ts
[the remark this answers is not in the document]
simmkyu: Using `DOMMockBehavior` makes sense here. I think we should keep replacing `setUserAgent` with `DOMMockBuilder` later.

PR #604, opened 2020-08-04
Change content share video check integration tests to use attendee name

ltrung:
**Description of changes:**
This is to update the video check for content share integration tests to use attendee name instead of video element index so that we do not need to rely on knowing how demo app assign video elements and reduce the breaking whenever we add or modify video elements in the demo app.

**Testing**

Manually run all content share video tests.

### DIFF
--- a/integration/js/pages/AppPage.js
+++ b/integration/js/pages/AppPage.js
@@ -309,34 +309,32 @@ class AppPage {
     return checked.result;
   }
 
-  async videoCheckLong(stepInfo, index, expectedState) {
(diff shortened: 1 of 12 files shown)


integration/js/pages/AppPage.js
```diff
@@ -309,34 +309,32 @@ class AppPage {
     return checked.result;
   }
 
-  async videoCheckLong(stepInfo, index, expectedState) {
```
anuranduttaroy, line 312: Was this not being used anywhere?

PR #605, opened 2020-08-06
Fix isSupported API in return true for Firefox on Android

ltrung:
**Description of changes:**
Fix isSupported API in return true for Firefox on Android when chime JS SDK currently does not support that platform.

**Testing**

2. How did you test these changes? Manually run a serverless demo in Firefox Android and verify that this warning message is logged: 
>this browser is not currently supported. Stability may suffer. Supported browsers are: Google Chrome 78+.

### DIFF
--- a/src/browserbehavior/DefaultBrowserBehavior.ts
+++ b/src/browserbehavior/DefaultBrowserBehavior.ts
@@ -133,10 +133,22 @@ export default class DefaultBrowserBehavior implements BrowserBehavior {
   }
 
   isSupported(): boolean {
-    return this.majorVersion() >= this.browserSupport[this.browser.name];
+    if (
+      !this.browserSupport[this.browser.name] ||
+      this.majorVersion() < this.browserSupport[this.browser.name]
+    ) {
+      return false;
+    }
+    if (this.browser.name === 'firefox' && this.isAndroid()) {
+      return false;
+    }
+    return true;
   }
 
   supportString(): string {
+    if (this.isAndroid()) {
+      return `${this.browserName['chrome']} ${this.browserSupport['chrome']}+`;
+    }
     const s: string[] = [];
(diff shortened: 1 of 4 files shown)


src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -133,10 +133,22 @@ export default class DefaultBrowserBehavior implements BrowserBehavior {
...
+    ) {
+      return false;
+    }
+    if (this.browser.name === 'firefox' && this.isAndroid()) {
+      return false;
+    }
+    return true;
   }
 
   supportString(): string {
```
simmkyu, line 148: I see that we are just using  `supportString` to log a list of supported browsers. I wonder if we can just put the whole string in https://docs.aws.amazon.com/chime/latest/dg/meetings-sdk.html#mtg-browsers because even in Android, it should return a list of supported browsers for desktop and others

```
- Mozilla Firefox (version 60 and later), for macOS and Windows
- Google Chrome (version 78 and later), for macOS, Windows, and Ubuntu LTS 16.04 and later
- Google Chrome for Android also supported for audio and video only (no content sharing)
- Chromium-based Edge (version 79 and later), for Windows
- Chromium-based Electron (Electron 7 and later, with Chromium version 78 and later)
- Safari (version 12, audio and video only, no content sharing), for macOS
- Safari (version 12.1.1 and later, audio and video only, no content sharing), for iOS
- Safari (version 13 and later, content sharing with screen capture requires turning on the Develop, Experimental Features, Screen Capture feature in the browser), for macOS
- Opera (version 66 and later), for macOS and Windows
```

ltrung: Discussed with Kyu, since supportedString is a public API, we will not modify it to return the above string at this time

PR #612, opened 2020-08-10
Makes AudioVideoController update to take VideoSubscribeContext

XHatan:
**Issue #:** 
A race condition can occur when a new IndexFrame is sent to SDK client in the middle of negotiation which overrides the VideoStreamIndex.
Another race condition can occur during two consecutive negotiations, the first round of track events are dispatched during the second negotiation.
Untitled Diagram (1)

**Description of changes:**

1. Abstract video signaling parameters in VideoSubscribeContext
2. Make AudioVideoController update takes VideoSubscribeContext instead of sharing data in AudioVideoControllerState

**Testing**

Yes
* unit tests

* on demo app
  * normal meeting with up to 16 tiles of videos
  * pausing / unpausing video
  * content sharing with multiple videos
  * trigger reconnection


src/videosubscribecontext/DefaultVideoSubscribeContext.ts
```diff
@@ -0,0 +1,68 @@
...
+import VideoStreamIndex from '../videostreamindex/VideoStreamIndex';
+import VideoSubscribeContext from './VideoSubscribeContext';
+
+export default class DefaultVideoSubscribeContext implements VideoSubscribeContext {
+  private _videoSubscriptions: number[] = [];
+  private _videoStreamIndex: VideoStreamIndex | null = null;
+  private _videosToReceive: VideoStreamIdSet = new DefaultVideoStreamIdSet();
+  private _videosPaused: VideoStreamIdSet = new DefaultVideoStreamIdSet();
+
+  constructor() {}
```
simmkyu, line 15: Not required

src/videosubscribecontext/VideoSubscribeContext.ts
```diff
@@ -0,0 +1,64 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import VideoStreamIdSet from '../videostreamidset/VideoStreamIdSet';
+import VideoStreamIndex from '../videostreamindex/VideoStreamIndex';
+
+export default interface VideoSubscribeContext {
```
simmkyu, line 7: - nit: Recommend matching the method order bewteen interface and implementation
  e.g.
  1. clone get functions
  2. ref get functions
  3. set functions
  4. wantsReceiveVideo()
  5. clone()

- If I were to rewrite this class in V2, I would state the "cloned" version more explicitly. Don't worry about this yet.
```
getClonedVideoStreamIndex();

// Typical get and set reference
getVideoStreamIndex();
setVideoStreamIndex(videoStreamIndex: VideoStreamIndex)
```

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -425,8 +438,10 @@ export default class DefaultAudioVideoController implements AudioVideoController
     });
   }
 
-  update(): boolean {
+  update(videoSubscribeContext: VideoSubscribeContext): boolean {
+    const nextVideoSubscribeContext = videoSubscribeContext.clone();
```
simmkyu, line 442: nit: how about removing this line and just assigning it to `this.meetingSessionContext`? Creating the *next* and immediately assigning it to *current* don't explain much.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -234,7 +243,8 @@ export default class DefaultAudioVideoController implements AudioVideoController
     this.meetingSessionContext.videoTileController = this._videoTileController;
     this.meetingSessionContext.videoDownlinkBandwidthPolicy = this.configuration.videoDownlinkBandwidthPolicy;
     this.meetingSessionContext.videoUplinkBandwidthPolicy = this.configuration.videoUplinkBandwidthPolicy;
-
+    this.meetingSessionContext.currentVideoSubscribeContext = new DefaultVideoSubscribeContext();
+    this.meetingSessionContext.previousVideoSubscribeContext = new DefaultVideoSubscribeContext();
```
simmkyu, line 247: - I'm probably missing something. You are initializing `this.meetingSessionContext.previousVideoSubscribeContext` here and using it in `CreatePeerConnectionTask`. But you don't set `previousVideoSubscribeContext` or call any update APIs on it. What is `previousVideoSubscribeContext` for?

- Assume that you keep setting `previousVideoSubscribeContext` to `currentVideoSubscribeContext` per update completion. What happens if more than one new IndexFrame arrives SDK during the negotiation? i.e. `previousVideoSubscribeContext` stores just one history snapshot.
 
   > Make AudioVideoController update takes VideoSubscribeContext instead of sharing data in AudioVideoControllerState
   
   In the CR descipriton, a brief explanation about `currentVideoSubscribeContext`, `previousVideoSubscribeContext`, and the full lifcycle would be useful.

   e.g. I'm understanding in this way. Per update operation, SDK will be using an immutable set of the VideoSubscribeContext object until it's complete. What happen when a new IndexFrame is coming?

XHatan: > * I'm probably missing something. You are initializing `this.meetingSessionContext.previousVideoSubscribeContext` here and using it in `CreatePeerConnectionTask`. But you don't set `previousVideoSubscribeContext` or call any update APIs on it. What is `previousVideoSubscribeContext` for?

looked like I messed it up while rebasing

> * Assume that you keep setting `previousVideoSubscribeContext` to `currentVideoSubscribeContext` per update completion. What happens if more than one new IndexFrame arrives SDK during the negotiation? i.e. `previousVideoSubscribeContext` stores just one history snapshot.
>   > Make AudioVideoController update takes VideoSubscribeContext instead of sharing data in AudioVideoControllerState
>   

If track event is two negotiation behind, the issue would still exist. What's your suggestion ? I don't think there are timing guarantees from webrtc

>   
>   In the CR descipriton, a brief explanation about `currentVideoSubscribeContext`, `previousVideoSubscribeContext`, and the full lifcycle would be useful.
>   e.g. I'm understanding in this way. Per update operation, SDK will be using an immutable set of the VideoSubscribeContext object until it's complete. What happen when a new IndexFrame is coming?



Untitled Diagram

PR #617, opened 2020-08-12
Do not subscribe to self content share video

ltrung:
**Description of changes:**
- Update downlink policies to not consider self content share video when subscribing for video stream index.
- Manually raise videoTileDidUpdate in startContentShare to be backward-compatible with the previous behavior.

**Testing**

2. How did you test these changes? Test with local demo app and verify that content share video is still shown to sharer but addVideoTile is not called from createPeerConnectionTask.


src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -48,10 +50,21 @@ export default class DefaultContentShareController
...
-      this.audioVideo.videoTileController.startLocalVideoTile();
+      this.contentAudioVideo.videoTileController.startLocalVideoTile();
     }
+
+    const tile = this.attendeeAudioVideo.videoTileController.addVideoTile();
+    tile.bindVideoStream(
+      this.contentAudioVideo.configuration.credentials.attendeeId,
+      false,
+      stream,
+      null,
```
XHatan, line 63: can we fill the width and height ?

src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -83,7 +96,7 @@ export default class DefaultContentShareController
   }
 
   stopContentShare(): void {
-    this.audioVideo.stop();
+    this.contentAudioVideo.stop();
```
XHatan, line 99: A general question, does videoTileWasRemoved get fired when stopContentShare is called ?  should we call another bindStream here?

src/contentsharecontroller/DefaultContentShareController.ts
ltrung: We do not set width and height for content share currently. I checked the current behavior and it does not contain width and height either.

PR #619, opened 2020-08-13
Exclude self content share stream index

ltrung:
**Description of changes:**
-  Exclude self content share video stream index.
-  Manually raise videoTileDidUpdate in startContentShare to be backward-compatible with the previous behavior.

**Testing**

2. How did you test these changes? Test with local demo app and verify that content share video is still shown to sharer but addVideoTile is not called from createPeerConnectionTask.


src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -106,6 +110,16 @@ export default class DefaultContentShareController
   }
 
   audioVideoDidStart(): void {
+    this.contentShareTile = this.attendeeAudioVideo.videoTileController.addVideoTile();
+    this.contentShareTile.bindVideoStream(
+      this.contentAudioVideo.configuration.credentials.attendeeId,
+      false,
+      this.mediaStreamBroker.mediaStream,
+      null,
+      null,
```
aburnson, line 119: Investigate whether width and height can be derived from MediaStream like it is done here: https://github.com/aws/amazon-chime-sdk-js/blob/0e03fdce362efea84da16af9b28735a72661654a/src/task/CreatePeerConnectionTask.ts#L216

ltrung: Added dimension.

src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -105,18 +111,66 @@ export default class DefaultContentShareController
...
+        const isContentAttendee = new DefaultModality(attendeeId).hasModality(
+          DefaultModality.MODALITY_CONTENT
+        );
+        const isSelfAttendee =
+          new DefaultModality(attendeeId).base() ===
+          this.attendeeAudioVideo.configuration.credentials.attendeeId;
+        if (isContentAttendee && isSelfAttendee) {
+          if (present && !this.contentShareTile) {
+            const stream = this.mediaStreamBroker.mediaStream;
+            if (stream.getVideoTracks().length > 0) {
```
aburnson, line 145: For readability, to reduce the level of nesting here, you could return early if the conditionals are false.

src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -105,18 +111,66 @@ export default class DefaultContentShareController
...
     });
   }
+
+  private setupContentShareEvents(): void {
+    /*
+     *  We use realtimeSubscribeToAttendeeIdPresence instead of audioVideoDidStart because audioVideoDidStart fires
+     *  before the capacity check in Tincan while when realtimeSubscribeToAttendeeIdPresence fires, we know the
+     *  content attendee has been able to pass the capacity check and join the call so we can start the local
+     *  content share video
+     */
```
aburnson, line 133: Prefer single-line `//` over doc-gen style `/* */` for inline comments.

ltrung: Good to know. Fixed.

src/contentsharecontroller/DefaultContentShareController.ts
ltrung: Return early for the two outer loops.

PR #636, opened 2020-08-20
Prevent error accessing browser.name or browser.version when using custom user agent

rlebosse:
Right now, if your browser has a custom user agent set, there will be an error accessing `this.browser.name` when you try to start the Chime Meeting.

Capture d’écran 2020-08-20 à 11 18 30

This PR adds a fallback if `detect-browser`does not succeed detecting current browser.


RidipDe:
Will have a NPM Push. Will close it for now.

PR #640, opened 2020-08-25
Add npm login and logout as part of publish script

RidipDe:
**Issue #:** 
The `git clean` command [as part of the post publish script](https://github.com/aws/amazon-chime-sdk-js/blob/master/script/postpublish#L27) deletes all the untracked files. So running `npm run logout` after the `npm publish` threw an error saying that the `.npmrc` file that stored the auth token is already deleted.

**Description of changes:**
We moved the `npm run login` at the end of the pre-publish command and `npm run logout` at the beginning of the post-publish script, so that the npm login / logout can be incorporated as part of the `npm publish` script

**Testing**

Yes
Tested locally - npm login and logout after publish was successful

### DIFF
--- a/script/postpublish
+++ b/script/postpublish
@@ -1,6 +1,12 @@
 #!/usr/bin/env ruby
 require 'json'
 
+def verbose command
+  puts("--> #{command}") || system(command) || fail("Failed: #{command}")
+end
+
+verbose("npm run logout")
+
 Dir.chdir(File.expand_path(File.join(File.dirname(__FILE__), '..')))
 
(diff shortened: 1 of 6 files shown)


script/postpublish
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env ruby
 require 'json'
 
+verbose("npm run logout")
```
ltrung, line 4: Should the definition of verbose go above this?

RidipDe: fixed

PR #644, opened 2020-08-31
Handle Index  message during resubscribe

alochbau-amzn:
**Description of changes:**

Receiving an index message during resubscription process
can result in track to attendee or stream to not return valid data.

Change VideoStreamIndex to keep and only update track based data
after a Subscribe Ack.  Other fields in the module can safely
be updated after every index message.

**Testing**
Created unit tests to simulate errors reported by customers and verified
fixes resolved the problem.
Made test calls using the serverless demos and verified video reception

Yes

Unit tests and manual calling


src/videostreamindex/DefaultVideoStreamIndex.ts
```diff
@@ -67,12 +74,23 @@ export default class DefaultVideoStreamIndex implements VideoStreamIndex {
     this.currentIndex = indexFrame;
     this.streamToAttendeeMap = null;
     this.streamToExternalUserIdMap = null;
-    this.ssrcToStreamMap = null;
+  }
+
+  integrateSubscribeFrame(): void {
+    // This is called just as a Subscribe is being sent.  Save corresponding Index
+    this.indexForSubscribe = this.currentIndex;
```
XHatan, line 81: is it possible that a track event comes late (or browser event loop is running late) and track event happens at the next negotiation, index frame is already overwritten ?

alochbau-amzn: I think it should be ok.  Track events only occur when a track is added or removed and occurs right after receiving the subscribeAck.  The mapping from stream to attendee will happen just as subscribeAck happens using the indexForSubscribe.  If for some reason the track event never comes, then a tile won't be added and I don't see an issue.  If it comes late, then it is using that mapping that was created when the subscribeAck happened.  I don't think an old one would ever come after a second subscribeAck.

src/videostreamindex/DefaultVideoStreamIndex.ts
```diff
@@ -277,24 +281,16 @@ export default class DefaultVideoStreamIndex implements VideoStreamIndex {
 
   streamIdForTrack(trackId: string): number {
```
simmkyu, line 282: All other getter APIs, modified in this PR, are called after `SubscribeAndReceiveSubscribeAckTask` completes. For example, after the "update" operation, the RTCPeerConnection track handler is calling `attendeeIdForTrack` to get the attendee ID. PR makes this attendee ID valid until the next subscribe ack.

But `streamIdForTrack` can be called before `SubscribeAndReceiveSubscribeAckTask`. Here's a stack trace:

1. The "update" operation always executes `AttachMediaInputTask` before `SubscribeAndReceiveSubscribeAckTask`.
2. `AttachMediaInputTask` calls `transceiverController.updateVideoTransceivers`.
3. `transceiverController.updateVideoTransceivers` calls `transceiverController.unsubscribeTransceivers`.
3. `transceiverController.unsubscribeTransceivers` calls this `streamIdForTrack` API.
4. After `AttachMediaInputTask` completes, `SubscribeAndReceiveSubscribeAckTask`

Could you review this flow too when SDK receives multiple video streams?

simmkyu: Please ignore this comment. The consecutive "update" operation will be deferred, so each "udpate" operation unsubscribes transceivers using the previously-built subscribe map. I can't think of race conditions as the map will be updated per `SubscribeAndReceiveSubscribeAckTask`.

image

src/videostreamindex/DefaultVideoStreamIndex.ts
```diff
@@ -67,12 +74,23 @@ export default class DefaultVideoStreamIndex implements VideoStreamIndex {
     this.currentIndex = indexFrame;
     this.streamToAttendeeMap = null;
     this.streamToExternalUserIdMap = null;
-    this.ssrcToStreamMap = null;
+  }
+
+  integrateSubscribeFrame(): void {
```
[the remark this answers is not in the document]
alochbau-amzn: I went back and forth on the naming of this.   I landed on trying to keep it consistent with the concept of "integrate..." since it is called as a Protocol Message is sent or received.  No problem renaming it though.

src/videostreamindex/DefaultVideoStreamIndex.ts
```diff
@@ -19,11 +19,18 @@ import VideoStreamDescription from './VideoStreamDescription';
  */
 export default class DefaultVideoStreamIndex implements VideoStreamIndex {
   protected currentIndex: SdkIndexFrame | null = null;
+  protected indexForSubscribe: SdkIndexFrame | null = null;
   protected currentSubscribeAck: SdkSubscribeAckFrame | null = null;
+
+  // These are based on the index at the time of the last Subscribe Ack
   protected trackToStreamMap: Map<string, number> | null = null;
```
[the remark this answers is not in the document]
alochbau-amzn: Will do.

src/videostreamindex/DefaultVideoStreamIndex.ts
XHatan: > If for some reason the track event never comes, then a tile won't be added and I don't see an issue

in this case, on unified plan, next time negotiation happens, won't track event never get fired and we are missing a tile ? 

> I don't think an old one would ever come after a second subscribeAck.

I guess it's a rare case but want to be sure on the model here. I couldn't find any documentation on the timing of track event. With early media, I used to expect event track come around sdp answer, but from various logs, track event can be really late.

Issue #627, opened 2020-08-15
UnbindVideoElement


76kennyboss:
Hello, 

Am trying to get the following effect.
1.  Video streams are already "bind" successfully to small thumb nails
2. When user press on the thumb nails, video stream is first "unbind" from the smaller thumb nail and then "bind" to a larger sized element.

Am unable to get both `unbindVideoElement` and `bindVideoElement` working in stage (2).

RidipDe:
Hello,
can you please let us know what you are trying to achieve and what is actually happening? 
Can you please provide us with SDK based API minimal repro that you are performing at your end.

vidya-mohan:
@76kennyboss please help the team by keeping the original bug template and provide the details with sufficient information to make progress

simmkyu:
`audioVideo.unbindVideoElement()` API seems to be broken.

Repro steps in the Chime SDK demo
1. Join a meeting and turn on your camera. Open the browser console and take a note of the video tile ID, e.g. `1`.
2. In the browser console, run the following command with the tile ID.
   ```
   app.audioVideo.unbindVideoElement(__tileId__)
   ```
3. The video element keeps displaying the video.

As a workaround, you might resize the same `<video>` element, e.g. using CSS or HTML attributes, when the user clicks on it.

michhyun1:
A fix has been pushed for this, the video stream should be removed from the video element when you unbind the video tile from the video element now.

michhyun1:
This commit was recently reverted, we need to push another fix for this issue

ileccese:
Hi, have you pushed another fix for this issue? I am also facing problems from this bug

michhyun1:
@ileccese could you please elaborate? The unbindVideoElement should alter the tileState of the videoTile, which you can then reference in videoTileDidUpdate.

ileccese:
@michhyun1 I'm seeing the same behavior that Kyu described above - the video tile still displays the video stream after unbindVideoElement has been called

michhyun1:
[...]

PR #680, opened 2020-09-10
Fix unbinding video tile bug

michhyun1:
**Issue #:**
https://github.com/aws/amazon-chime-sdk-js/issues/627
**Description of changes:**
Fix bug related to unbinding a video tile from a video element

**Testing**

2. How did you test these changes? Tested locally on a local instance of the meeting demo. 

### DIFF
--- a/src/videotile/DefaultVideoTile.ts
+++ b/src/videotile/DefaultVideoTile.ts
@@ -181,7 +181,14 @@ export default class DefaultVideoTile implements DevicePixelRatioObserver, Video
 
   bindVideoElement(videoElement: HTMLVideoElement | null): void {
     let tileUpdated = false;
+    // If videoElement is null, we are doing an unbinding
     if (this.tileState.boundVideoElement !== videoElement) {
+      if (videoElement === null) {
+        DefaultVideoTile.disconnectVideoStreamFromVideoElement(
+          this.tileState.boundVideoElement,
+          false
+        );
+      }
       this.tileState.boundVideoElement = videoElement;
       tileUpdated = true;
     }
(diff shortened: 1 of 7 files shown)


src/videotile/DefaultVideoTile.ts
```diff
@@ -181,7 +181,16 @@ export default class DefaultVideoTile implements DevicePixelRatioObserver, Video
...
   bindVideoElement(videoElement: HTMLVideoElement | null): void {
     let tileUpdated = false;
+    // If videoElement is null, we are doing an unbinding
     if (this.tileState.boundVideoElement !== videoElement) {
+      if (videoElement === null) {
+        DefaultVideoTile.disconnectVideoStreamFromVideoElement(
+          this.tileState.boundVideoElement,
+          false
+        );
+        this.tileState.videoElementCSSWidthPixels = null;
```
ltrung, line 191: Could we keep the logic below as before? Not sure why we want to move it here. We will no longer handle the case where boundVideoElement is null and videoElement is null but the height and width are not.

michhyun1: youre right, I moved it back and just added this condition

src/videotile/DefaultVideoTile.ts
```diff
@@ -181,7 +181,14 @@ export default class DefaultVideoTile implements DevicePixelRatioObserver, Video
 
   bindVideoElement(videoElement: HTMLVideoElement | null): void {
     let tileUpdated = false;
+    // If videoElement is null, we are doing an unbinding
```
simmkyu, line 184: nit: The comment is not needed as the code is self-describing.

simmkyu (approved):

PR #681, opened 2020-09-11
Fix tone does not stop when calling MeetingReadinessChecker.checkAudioOutput multiple times

ltrung:
**Issue:** 
Calling MeetingReadinessChecker multiple times will cause the test tone to not stop.

**Description of changes:**
Before playing the test tone, make sure to stop any tone played previously.

**Testing**

2. How did you test these changes? Manually run the meeting readiness checker demo and call the checkAudioOutput API multiple times and verify that the test tone stops.

### DIFF
--- a/src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
+++ b/src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
@@ -94,6 +94,9 @@ export default class DefaultMeetingReadinessChecker implements MeetingReadinessC
     const rampSec = 0.1;
     const maxGainValue = 0.1;
 
+    if (this.oscillatorNode) {
+      this.stopTone();
+    }
     this.audioContext = DefaultDeviceController.getAudioContext();
     this.gainNode = this.audioContext.createGain();
     this.gainNode.gain.value = 0;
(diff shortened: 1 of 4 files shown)


anuranduttaroy (approved):

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
```diff
@@ -94,6 +94,9 @@ export default class DefaultMeetingReadinessChecker implements MeetingReadinessC
     const rampSec = 0.1;
     const maxGainValue = 0.1;
 
+    if (this.oscillatorNode) {
```
simmkyu, line 97: Stopping the previous tone is one approach, but a better way would be to stop the corresponding tone only.

```
private playTone(...) {
  // Create AudioContext nodes as local variables

  return () => {
     // Stop and disconnect nodes
  };
}
```

// In the caller
```
let stopTone;
try {
  stopTone = this.playTone();
} catch (error) {
} finally {
  stopTone();
}
```

simmkyu: NVM.. since it takes a `HTMLAudioElement` parameter, my suggestion can be error prone too if they pass the same audio element.

simmkyu (approved):

nicklera1:
Here is what I'm getting.
```
Uncaught (in promise) DOMException: Failed to execute 'disconnect' on 'AudioNode': the given destination is not connected.DefaultMeetingReadinessChecker.js:182 Uncaught (in promise) DOMException: Failed to execute 'disconnect' on 'AudioNode': the given destination is not connected.
    at DefaultMeetingReadinessChecker.app_meeting.../../node_modules/amazon-chime-sdk-js/build/meetingreadinesschecker/DefaultMeetingReadinessChecker.js.DefaultMeetingReadinessChecker.stopTone (https://localhost:9000/:12246:29)
    at DefaultMeetingReadinessChecker.<anonymous> (https://localhost:9000/:12206:30)
    at step (https://localhost:9000/:12112:23)
    at Object.next (https://localhost:9000/:12093:53)
    at fulfilled (https://localhost:9000/:12084:58)
```

PR #684, opened 2020-09-11
Change error to warn for logging Cloudwatch errors

RidipDe:
**Issue #:**
We were logging errors generated from the cloudwatch logs if any as error. We just need it to warn us instead.

**Description of changes:**
Changing error logging to warn logging on the handler

**Testing**

Yes
Yes

No

No

### DIFF
--- a/demos/serverless/src/handlers.js
+++ b/demos/serverless/src/handlers.js
@@ -135,7 +135,12 @@ exports.logs = async (event, context) => {
   try {
     await cloudWatchClient.putLogEvents(putLogEventsInput).promise();
   } catch (error) {
-    console.error(`Failed to put CloudWatch log events with error ${error.message} and params ${JSON.stringify(putLogEventsInput)}`);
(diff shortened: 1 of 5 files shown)


demos/serverless/src/handlers.js
```diff
@@ -135,7 +135,7 @@ exports.logs = async (event, context) => {
   try {
     await cloudWatchClient.putLogEvents(putLogEventsInput).promise();
   } catch (error) {
-    console.error(`Failed to put CloudWatch log events with error ${error.message} and params ${JSON.stringify(putLogEventsInput)}`);
+    console.warn(`Failed to put CloudWatch log events with error ${error.message} and params ${JSON.stringify(putLogEventsInput)}`);
```
ltrung, line 138: We need to log error for everything except the DataAlreadyAcceptedException since we decided to skip that one.

anuranduttaroy: Why are we doing this anyways?

ltrung: We decide to not handle DataAlreadyAcceptedException so just want to catch and skip that error.

ltrung (approved):

Issue #456, opened 2020-06-05
"Cannot read property 'addEventListener' of undefined" error on HTTP


simmkyu:
**Important:** Please attach INFO-level logs with timestamps to the report from *before* the session started up until the point at which the unexpected behavior occured. For multi-participant calls, attach a log file for each participant if possible.

**Meeting and attendee IDs:**
List the meeting and attendee IDs that the attached logs apply to.
N/A

**Timeline**
Outline the main actions the user performed and when:
N/A

**Describe the bug**
Describe the bug and the timestamp at which it happened relative to the timestamp in the log file.

After deploying the browser demo to a non-HTTPS host, SDK displays this error message when attempting to join a meeting.

<img src="https://user-images.githubusercontent.com/36240708/83927697-8dfbf000-a742-11ea-836d-a9cb4135babc.png" width="300"></img>

The cause is that media device APIs are only available in secure contexts, e.g. HTTPS/TLS in a browser. The `navigator.mediaDevices` is `undefined` unless a customer explicitly overrides the option.
https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia

The SDK demo needs to display a better error message.

**Screenshots**
If applicable, add screenshots to help explain the problem.

**Platform**
 - OS: macOS/Windows
 - Browser: Chrome/FF
 - Browser Version: Tested in Chrome 83.0.4103.61 but recent versions of all modern browsers don't allow the media API access.
 - Amazon Chime SDK GitHub Commit: c610e1222af3c60cdadac193d7aba3e6afdc0e5a

PR #692, opened 2020-09-17
Add messaging for getUserMedia error on HTTP site

bradmanAWS:
**Issue #:**
#456
**Description of changes:**
Added improved messaging for getUserMedia error when on HTTP sites. 
**Testing**
yes
I loaded the app via `http:` and `https:` in Chrome, Safari, and Firefox. The new error message is only shown when joining a meeting fails and the client is using `http:`. 

No
No

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -211,11 +211,12 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
           try {
             chimeMeetingId = await this.authenticate();
           } catch (error) {
+            const httpErrorMessage = 'UserMedia is not allowed in HTTP sites. Either use HTTPS or enable media capture on insecure sites.';
             (document.getElementById(
               'failed-meeting'
             ) as HTMLDivElement).innerText = `Meeting ID: ${this.meeting}`;
             (document.getElementById('failed-meeting-error') as HTMLDivElement).innerText =
-              error.message;
+              window.location.protocol === 'http:' ? httpErrorMessage : error.message;
             this.switchToFlow('flow-failed-meeting');
             return;
           }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update integration-watchlist to include demos/browser with no exception for package.json
 - Change error to warn for logging Cloudwatch errors
 - Update README with use case to handle `realtimeSubscribeToVolumeIndicator` updates efficiently
+- Change error messaging for getUserMedia error
 
 ### Removed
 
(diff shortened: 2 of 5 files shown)


anuranduttaroy:
Do we not want to bump up the version number in [package.json](https://github.com/aws/amazon-chime-sdk-js/blob/getusermedia-error/package.json#L3) and [Versioning.ts](https://github.com/aws/amazon-chime-sdk-js/blob/getusermedia-error/src/versioning/Versioning.ts#L21) for this change?

ltrung (commented):
How did you test these changes? Can you specify it in more details other than "locally"?

Also the error message in the issue is "Cannot get addEventListener of undefined" but in here you check for MediaDeviceInfo is not defined in the error message so will this work?

simmkyu (dismissed):

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -211,11 +211,12 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
           try {
             chimeMeetingId = await this.authenticate();
           } catch (error) {
+            const mediaDeviceInfoErrorMessage = 'getUserMedia is not allowed in HTTP sites. Either use HTTPS or enable media capture on insecure sites.';
             (document.getElementById(
               'failed-meeting'
             ) as HTMLDivElement).innerText = `Meeting ID: ${this.meeting}`;
             (document.getElementById('failed-meeting-error') as HTMLDivElement).innerText =
-              error.message;
+              error.message.includes('MediaDeviceInfo is not defined') ? mediaDeviceInfoErrorMessage : error.message;
```
RidipDe, line 219: Its possible that a few browsers do not support the MediaDeviceInfo API like the Opera, which is a Chromium based browser. Ref. https://developer.mozilla.org/en-US/docs/Web/API/MediaDeviceInfo

The error might still be 'MediaDeviceInfo is not defined'. So with the current 'mediaDeviceInfoErrorMessage' it might confuse the user even after the user redirected to HTTPS in a non supported browser as the user will continue to see the same error.

How about using the following instead 
```
const mediaDeviceInfoErrorMessage = 'MediaDeviceInfo is not defined. Please redirect to HTTPS or use one of the officially supported browsers';
 ```

Comment (bradmanAWS) [reply demos/browser/app/meetingV2/meetingV2.ts]: I'm changed the implementation to only show the new error message when error occurs and the user is using `http:` protocol.

File: demos/browser/package.json [hunk 2]
```diff
@@ -33,7 +33,11 @@
   "dependencies": {
     "amazon-chime-sdk-js": "file:../..",
     "aws-sdk": "^2.755.0",
+<<<<<<< HEAD
```
ltrung, line 36: Merge conflict.

ltrung (changes requested):

ltrung (approved):

PR #693, opened 2020-09-17
Fix css bug to account for 4.5.1 bootstrap update

michhyun1:
**Issue #:**

**Description of changes:**
Bootstrap introduced a code change  in 4.5.1. (https://github.com/twbs/bootstrap/pull/30940) that breaks the css format in our demo app.  We just need to override the flex property to make sure the flex-basis property isnt 100%.

Also, bump bootstrap to 4.5.1.

**Testing**
Visual testing in the demo app.

2. How did you test these changes? visually verified on the demo app
3. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? n/a
4. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved? n/a

### DIFF
--- a/demos/browser/package.json
+++ b/demos/browser/package.json
@@ -32,8 +32,8 @@
   },
   "dependencies": {
     "amazon-chime-sdk-js": "file:../..",
-    "aws-sdk": "^2.751.0",
-    "bootstrap": "4.5.0",
+    "aws-sdk": "^2.755.0",
+    "bootstrap": "^4.5.1",
     "compression": "^1.7.4",
     "jquery": "^3.5.1",
     "lodash": "^4.17.20",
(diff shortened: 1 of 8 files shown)


demos/browser/package.json
```diff
@@ -32,8 +32,8 @@
   },
   "dependencies": {
     "amazon-chime-sdk-js": "file:../..",
-    "aws-sdk": "^2.751.0",
-    "bootstrap": "4.5.0",
+    "aws-sdk": "^2.755.0",
+    "bootstrap": "4.5.1",
```
simmkyu, line 36: We have removed `^` to use `4.5.0`. Please use `^` to update to all minior/patch versions, e.g. any version < 5.0.0.

simmkyu (commented):
Please test your change in Safari too.

simmkyu (dismissed):
LGTM! I have a minor comment on using `^` in `package.json

demos/browser/package.json
michhyun1: sure, will do

simmkyu (approved):

PR #699, opened 2020-09-18
Do not choose the "null" device for initial reconnection

simmkyu:
**Issue #:**

**Description of changes:**
SDK retriees to connect the session in two cases:
1. When an initial connection fails, or
2. When an ongoing meeting has a network issue or other "retryable" errors

In case 1, SDK fully disconnects the session and performs the "start" operation again. Before this PR, SDK deselects both audio and video inputs. After this PR, SDK keeps using the same devices during reconnection.

**Testing**

   By injecting a failed task during initial connection, make sure that SDK doesn't deselect audio and video inputs. After multiple retry attempts, let SDK connect the session. Ensure that the audio and video are working fine.


test/task/ReleaseMediaInputTask.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  it('releases audio and video device', async () => {
+    let called = 0;
+    class MockMediaStreamBroker extends NoOpMediaStreamBroker {
+      releaseMediaStream(mediaStream: MediaStream): void {
+        const track = mediaStream.getTracks()[0];
+        if (track.kind === 'audio') {
+          called += 1;
+        } else if (track.kind === 'video') {
+          called += 1;
```
michhyun1, line 60: nit, track.kind === 'audio' || track.kind === 'video'

michhyun1 (approved):

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -416,6 +417,21 @@ export default class DefaultAudioVideoController implements AudioVideoController
     } catch (error) {
       this.logger.info('fail to clean');
     }
+
+    if (!reconnecting) {
+      try {
+        await new SerialGroupTask(this.logger, this.wrapTaskName('AudioVideoReleaseMedia'), [
+          new TimeoutTask(
```
ltrung, line 424: There is only one task here to release media input task. Do we need to put it in SerialGroupTask or just TimeoutTask is fine?

src/task/CleanStoppedSessionTask.ts
```diff
@@ -55,11 +55,6 @@ export default class CleanStoppedSessionTask extends BaseTask {
       this.context.turnCredentials = null;
       this.context.videoSubscriptions = null;
       this.context.transceiverController.reset();
-      this.context.mediaStreamBroker.releaseMediaStream(this.context.activeAudioInput);
```
ltrung, line 58: I feel like it should be part of the CleanStoppedSessionTask to clean up the media stream input and that should be the default behavior.
Could we instead pass in a flag to control whether we clean up the media stream input then pass reconnecting to it?

src/audiovideocontroller/DefaultAudioVideoController.ts
simmkyu: It's for logging purpose. `__serialTaskName__/Timeout15000ms/__taskName__` will be printed.
e.g. `AudioVideoReleaseMedia/Timeout15000ms/ReleaseMediaInputTask`

Without SerialGroupTask, the message starts with `Timeout15000ms/`.

src/task/CleanStoppedSessionTask.ts
simmkyu: Thanks for the suggestion.

- All child tasks accept only the `AudioVideoControllerState` context as a single parameter. Thus, I didn't add an extra flag in the constructor but abstract them into an explicit task.
- We can add a flag into the context, but it will then require to maintain the flag state. We wanted to avoid child tasks to know the session state.

ltrung: It makes sense to me. The only issue I have with this is now we need to know/remember cleanStoppedSesssionTask does not release media stream input and has to remember to call it seperately.

simmkyu: The name `CleanStoppedSessionTask` is a general cleaning-up task, while `LeaveAndReceiveLeaveAckTask` and `ReleaseMediaInputTask` are specific tasks.

`LeaveAndReceiveLeaveAckTask` was also a part of `CleanStoppedSessionTask` but separated for having its own timeout.

simmkyu:
I'll reopen this PR after 1.18 release

PR #702, opened 2020-09-20
Create log stream before logging begins

RidipDe:
**Issue #:**
We recently observed ResourceAlreadyExistsException on Log stream - we already have a mechanism to check if a log stream exists - and if it exists we re-use that or else we create a new log stream. 
A race condition is taking place, two or more processes reach the createLogStream() in ensureLogStream at the same time. While one gets the priority and creates a logStream, the other process fails to create a LogStream with the same logStream name. So this process fails for the first time and is retried at the CW end. Upon retry the previously failed process executes without error on the LogStream returned by ensureLogStream - thus there is no data loss but we see ERROR entry due to the failed execution.

**Description of changes:**
We are creating a log stream at the beginning of the meeting itself. So there should be no race condition when multiple process start writing to a log.

**Testing**

Yes

Tested by removing the endpoint logs and running the serverless demo app to see that the application is creating a new Log Stream

No

No

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -674,6 +674,24 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
     }
   }
 
+  async createLogStream(): Promise<void> {
+    const body = JSON.stringify({
+      meetingId: this.meetingSession.configuration.meetingId,
+      attendeeId: this.meetingSession.configuration.credentials.attendeeId,
+    });
+    try {
+      const response = await fetch(`${DemoMeetingApp.BASE_URL}create_log_stream`, {
+        method: 'POST',
+        body
+      });
+      if (response.status === 200) {
+        console.log('Log stream created');
+      }
+    } catch (error) {
+      console.error(error.message);
+    }
+  }
+
   async initializeMeetingSession(configuration: MeetingSessionConfiguration): Promise<void> {
     let logger: Logger;
     const logLevel = LogLevel.INFO;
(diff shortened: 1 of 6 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -248,6 +248,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
           }
           await this.openAudioOutputFromSelection();
           this.hideProgress('progress-authenticate');
+          await this.createLogStream();
```
ltrung, line 251: Should this go to initializeMeetingSession when we intialize MeetingPostSessionLogger. There is no reason to create log stream for local host.
```
      logger = consoleLogger;
    } else {
      logger = new MultiLogger(
        consoleLogger,
        new MeetingSessionPOSTLogger(
          'SDK',
          configuration,
          DemoMeetingApp.LOGGER_BATCH_SIZE,
          DemoMeetingApp.LOGGER_INTERVAL_MS,
          `${DemoMeetingApp.BASE_URL}logs`,
          logLevel
        ),
      );
    }
```

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -673,6 +674,24 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
...
+      meetingId: this.meetingSession.configuration.meetingId,
+      attendeeId: this.meetingSession.configuration.credentials.attendeeId,
+    });
+    try {
+      const response = await fetch(`${DemoMeetingApp.BASE_URL}create_log_stream`, {
+        method: 'POST',
+        body
+      });
+      if (response.status === 200) {
+        console.error('Log stream created');
```
ltrung, line 688: Why do we log this as error?

ltrung (changes requested):
Should we also add code to give warn for InvalidSequenceToken?

PR #709, opened 2020-09-23
Fix bootstrap bug in V1 meeting

michhyun1:
**Issue #:**

**Description of changes:**
Adjust for bootstrap 4.5.1 addition to row class. Need to unset the flex property that was introduced in 4.5.1.

**Testing**
Ran meeting V1 and meeting V2 and visually verified.

yes

 Ran meeting V1 and meeting V2 and visually verified.

n/a
n/a

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.html
+++ b/demos/browser/app/meetingV2/meetingV2.html
@@ -226,7 +226,7 @@
   </div>
   <audio id="meeting-audio" style="display:none"></audio>
   <div id="meeting-container" class="container-fluid h-100" style="display:flex; flex-flow:column">
-    <div class="row mb-3 mb-lg-0" style="flex-basis: auto;">
+    <div class="row mb-3 mb-lg-0" style="flex: unset;">
       <div class="col-12 col-lg-3 order-1 order-lg-1 text-center text-lg-left">
         <div id="meeting-id" class="navbar-brand text-muted m-0 m-lg-2"></div>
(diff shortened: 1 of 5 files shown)


demos/browser/app/meetingV2/meetingV2.html
```diff
@@ -226,7 +226,7 @@ <h1 class="h3 mb-3 font-weight-normal text-center">Select devices</h1>
   </div>
   <audio id="meeting-audio" style="display:none"></audio>
   <div id="meeting-container" class="container-fluid h-100" style="display:flex; flex-flow:column">
-    <div class="row mb-3 mb-lg-0" style="flex-basis: auto;">
+    <div class="row mb-3 mb-lg-0" style="flex: unset;">
```
RidipDe, line 229: Why are we changing the styling in meeting V2 ?

RidipDe: Is it to be consistent with the meeting V1 ?

michhyun1: yeah, flex sets three different properties: flex-grow, flex-shrink, and flex-basis. 

Initially I thought we could just reset flex-basis, but I think it would be better to reset all the flex properties to default by doing unset. (this is how it was prior to the 4.5.1 bootstrap change)

RidipDe (dismissed):

anuranduttaroy (dismissed):

ltrung (dismissed):

anuranduttaroy (approved):

PR #712, opened 2020-09-23
Use credentials sent via signaling connection JOIN_ACK

aburnson:
**Description of changes:**

Use credentials sent via signaling connection JOIN_ACK

**Testing**

Yes.

Tested in the meeting demo:
- Verified that when credentials are received via signaling the TURN credentials fetch is bypassed
- Verified that if credentials are not received via signaling, TURN credentials `fetch` fallback occurs
- Verified that when meeting has already ended, signaling channel returns WebSocket close message and terminates session start attempt

Yes. Multiple enum values that were used to describe "meeting ended" (`AudioCallEnded`, `TURNMeetingEnded`, `MeetingEnded`) have been collapsed into a single enum value (`6`). This change allows builders to simply check for the `MeetingEnded` status rather than multiple separate statuses. The original enum names still exist and are backwards compatible. These changes have been reviewed and approved separately by @ltrung and @simmkyu .

Yes. The wire protocol and service has been updated to include TURN credentials in the JOIN_ACK frame. If the client sees that they are present, then the regular TURN credentials `fetch` is skipped and the peer connection can be constructed immediately using the received credentials. This protocol change has been reviewed and approved separately by @codeslinger.


protocol/SignalingProtocol.proto
```diff
@@ -111,6 +113,7 @@ message SdkIndexFrame {
     optional bool at_capacity = 1;
     repeated SdkStreamDescriptor sources = 2;
     repeated uint32 paused_at_source_ids = 3;
+    optional uint32 num_participants = 4;
```
ltrung, line 116: What is this used for?

simmkyu (dismissed):

protocol/SignalingProtocol.proto
aburnson: This field was generated from the most current service protocol spec. It is appearing now since it had not been synced to the JS SDK protocol before now.

simmkyu (dismissed):

ltrung (dismissed):

src/meetingsession/MeetingSessionStatusCode.ts
```diff
@@ -24,20 +24,32 @@ export enum MeetingSessionStatusCode {
...
+  /**
+   * Deprecated. The meeting has ended. This is a legacy alias for MeetingEnded and will
+   * be removed in v2.0.0.
+   */
+  TURNMeetingEnded = 6,
+
+  /**
+   * The meeting has ended.
+   */
+  MeetingEnded = 6,
```
anuranduttaroy, line 51: We changed the `TURNMeetingEnded = 24` to `TURNMeetingEnded = 6` here. Isn’t that backward incompatible?

src/task/ReceiveTURNCredentialsTask.ts
```diff
@@ -32,8 +32,15 @@ export default class ReceiveTURNCredentialsTask extends BaseTask {
   }
 
   async run(): Promise<void> {
+    if (this.context.turnCredentials) {
+      this.context.logger.info('TURN credentials available, skipping credentials fetch');
+      return;
+    } else {
```
anuranduttaroy, line 38: This else seems to be not required.

src/meetingsession/MeetingSessionStatusCode.ts
aburnson: It is ok to change the underlying internal value since these numbers are only used to differentiate the enum values. As long as the client application uses the names, the behavior is the same. This change has been verified in the meeting demo.

simmkyu (dismissed):

src/task/ReceiveTURNCredentialsTask.ts
aburnson: Fixed in latest rev

simmkyu (approved):

src/task/ReceiveTURNCredentialsTask.ts
```diff
@@ -94,7 +101,7 @@ export default class ReceiveTURNCredentialsTask extends BaseTask {
         return this.context.meetingSessionConfiguration.urls.urlRewriter(uri);
       })
       .filter((uri: string) => {
-        return uri !== '';
+        return !!uri;
```
ltrung, line 104: Is this what cause the meeting readiness tests to fail earlier?

simmkyu (approved):

ltrung (approved):

PR #716, opened 2020-09-24
Upgrade dependencies

rnewman:
**Issue #:**

None.

**Description of changes:**

This sequence of commits updates our dependencies to relatively current versions, updating and fixing our lint configuration to match. It also switches to `ts-loader`, away from the unmaintained `awesome-typescript-loader`. See individual commit messages for details.

I split this set of changes across multiple commits to make review of the large number of automated lint fixes a little less painful.

The `detect-browser` upgrade includes no changes that should affect our platform-detection behavior.

**Testing**

`npm run build:release` passes locally. These changes are almost entirely TypeScript-oriented (and the rest are trivial), and thus do not affect behavior.

No public APIs or request methods have changed apart from the introduction of a `ConnectionMetrics` type in place of the use of `{}`. This should not be a breaking change for any consumers.


package.json
```diff
@@ -27,6 +27,7 @@
     "doc": "node node_modules/typedoc/bin/typedoc --options typedoc.json",
     "postdoc": "node script/update-typedoc-link.js",
     "test": "npm run test:fast",
+    "test:richnew": "cross-env TS_NODE_CACHE=false nyc mocha -rv ts-node/register \"test/videouplinkbandwidthpolicy/**/*.test.ts\"",
```
ltrung, line 30: Do we want to remove this :)

tsconfig.test.json
```diff
@@ -0,0 +1,18 @@
...
+    "noUnusedLocals": true,
+    "downlevelIteration": true,
+    "experimentalDecorators": true,
+    "module": "commonjs",
+    "target": "es5",
+    "lib": ["es6", "dom"],
+    "declaration": true
+  },
+  "include": [
+    "./test/**/*",
```
ltrung, line 16: Why cant we include this in tsconfig.json? The current one include the src and guides folder:
```
  "include": [
    "./src/**/*",
    "./guides/docs.ts"
  ]
```

package.json
rnewman: `git add -u` strikes again!

tsconfig.test.json
rnewman: Including files here means they are included in the transpiled output.

The right fix for this is to use project references:

https://www.typescriptlang.org/docs/handbook/project-references.html

to make a test project that depends on the app project. We can do this in a follow-up. It's not a big deal for now, because the second definition is only used for the lint config.

rnewman: Updated commit with project references pushed.

rnewman:
Travis seems to disagree about what the content of the docs should be.

```
> script/postbuild

Error: there are uncommitted changes:

 M docs/classes/allhighestvideobandwidthpolicy.html
 M docs/classes/asyncscheduler.html
 M docs/classes/attachmediainputtask.html
 M docs/classes/audiovideocontrollerstate.html
 M docs/classes/baseconnectionhealthpolicy.html
…
```

rnewman:
Filed #720 for the Travis bug.

script/update-typedoc-link.js
```diff
@@ -32,12 +32,12 @@ walk('docs')
...
   });
 
-fs.readFile('docs/assets/js/search.js', 'utf8', (err, data) => {
+fs.readFile('docs/assets/js/search.json', 'utf8', (err, data) => {
   if (err) return console.error(err);
 
   let result = data.replace(/["][,]["]/g, '",\n"');
   result = result.replace(/[}][,][{]/g, '},\n{');
-  fs.writeFile('docs/assets/js/search.js', result, 'utf8', err => {
+  fs.writeFile('docs/assets/js/search.json', result, 'utf8', err => {
```
ltrung, line 40: Could you explain how does this work?

ltrung (approved):

script/update-typedoc-link.js
rnewman: [The fix is in `package.json`](https://github.com/aws/amazon-chime-sdk-js/blob/master/package.json#L27) — we add an explicit reference to the main `amazon-chime-sdk-js` repo as a command-line argument, so everywhere that builds now does so assuming a PR to the main repo.

```
    "doc": "node node_modules/typedoc/bin/typedoc --options typedoc.json --gitRemote https://github.com/aws/amazon-chime-sdk-js",
```

(I tried to add that as a config option in the typedoc config file and it complained, but the arg worked.)

This change just tracks a change in typedoc, which needed to upgrade to gain that command-line argument:

https://github.com/TypeStrong/typedoc/issues/1247#issuecomment-616291123

PR #723, opened 2020-09-25
Reintroduce defined-in links to GitHub.

rnewman:
**Issue #:**

None.

**Description of changes:**

This commit breaks out doc generation into a Ruby script so we can
determine the Git remote at runtime. This is necessary to support forks
(which need to use 'upstream') and Travis (which uses 'origin').

We also explicitly pass `master` as the revision against which to link, which makes our doc links more stable.

This change convinces TypeDoc that this is a GitHub repo, which brings
back the "Defined in" links that I broke in my last commit.

**Testing**

`npm run build:release` passes locally. Changes are doc-only.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "build:publish": "npm run build && npm run lint && npm run check && npm run doc && npm run test:retry",
     "postbuild:publish": "script/postbuild",
     "predoc": "rm -rf docs && node script/build-guides.js",
-    "doc": "node node_modules/typedoc/bin/typedoc --options typedoc.json --gitRemote https://github.com/aws/amazon-chime-sdk-js",
+    "doc": "script/generate-docs",
     "postdoc": "node script/update-typedoc-link.js",
     "test": "npm run test:fast",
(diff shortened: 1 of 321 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
     "build:publish": "npm run build && npm run lint && npm run check && npm run doc && npm run test:retry",
     "postbuild:publish": "script/postbuild",
     "predoc": "rm -rf docs && node script/build-guides.js",
-    "doc": "node node_modules/typedoc/bin/typedoc --options typedoc.json --gitRemote https://github.com/aws/amazon-chime-sdk-js",
+    "doc": "script/generate-docs",
```
ltrung, line 27: So when you run npm run build:release from a fork, will it use upstream/master?

rnewman: Correct. Locally:

```
> script/generate-docs

Generating docs against upstream/master.
```

and [in Travis it uses origin/master](https://travis-ci.org/github/aws/amazon-chime-sdk-js/jobs/730261169#L268).

ltrung (approved):

PR #724, opened 2020-09-25
Create log stream before logging begins

RidipDe:
**Issue #:**
We recently observed ResourceAlreadyExistsException on Log stream - we already have a mechanism to check if a log stream exists - and if it exists we re-use that or else we create a new log stream. 
A race condition is taking place, two or more processes reach the createLogStream() in ensureLogStream at the same time. While one gets the priority and creates a logStream, the other process fails to create a LogStream with the same logStream name. So this process fails for the first time and is retried at the CW end. Upon retry the previously failed process executes without error on the LogStream returned by ensureLogStream - thus there is no data loss but we see ERROR entry due to the failed execution.

**Description of changes:**
We are creating a log stream at the beginning of the meeting itself. So there should be no race condition when multiple process start writing to a log.

**Testing**

Yes

Verified by running the serverless demo app successfully that create_log_stream endpoint is hit before the logs endpoint in the network tab
Tested by removing the endpoint logs and running the serverless demo app to see that the application is creating a new Log Stream in the cloudwatch.

N/A

N/A

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -666,6 +666,24 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
     }
   }
 
+  async createLogStream(configuration: MeetingSessionConfiguration): Promise<void> {
+    const body = JSON.stringify({
+      meetingId: configuration.meetingId,
+      attendeeId: configuration.credentials.attendeeId,
+    });
+    try {
+      const response = await fetch(`${DemoMeetingApp.BASE_URL}create_log_stream`, {
+        method: 'POST',
+        body
+      });
+      if (response.status === 200) {
+        console.log('Log stream created');
+      }
+    } catch (error) {
+      console.error(error.message);
+    }
+  }
+
   async initializeMeetingSession(configuration: MeetingSessionConfiguration): Promise<void> {
     let logger: Logger;
     const logLevel = LogLevel.INFO;
@@ -673,6 +691,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
     if (location.hostname === 'localhost' || location.hostname === '127.0.0.1') {
       logger = consoleLogger;
     } else {
+      await this.createLogStream(configuration);
       logger = new MultiLogger(
         consoleLogger,
         new MeetingSessionPOSTLogger(
--- a/demos/serverless/src/handlers.js
+++ b/demos/serverless/src/handlers.js
@@ -136,7 +136,7 @@ exports.logs = async (event, context) => {
     await cloudWatchClient.putLogEvents(putLogEventsInput).promise();
   } catch (error) {
     const errorMessage = `Failed to put CloudWatch log events with error ${error} and params ${JSON.stringify(putLogEventsInput)}`;
-    if (error.code === 'DataAlreadyAcceptedException') {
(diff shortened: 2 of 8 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -666,13 +666,32 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
...
+      meetingId: configuration.meetingId,
+      attendeeId: configuration.credentials.attendeeId,
+    });
+    try {
+      const response = await fetch(`${DemoMeetingApp.BASE_URL}create_log_stream`, {
+        method: 'POST',
+        body
+      });
+      if (response.status === 200) {
+        console.log('Log stream created');
```
devalevenkatesh, line 680: Do we need to handle for any other statuses here? Like throw error when access related error comes in. Not sure about the working here, but if we get a security token exception for the fetch call is it catched in `response.status` or in catch block.

demos/serverless/src/handlers.js
```diff
@@ -203,6 +203,20 @@ async function ensureLogStream(cloudWatchClient, logStreamName) {
...
+  if (!body.meetingId || !body.attendeeId) {
+    return response(400, 'application/json', JSON.stringify({error: 'Need properties: meetingId, attendeeId'}));
+  }
+  const logStreamName = `ChimeSDKMeeting_${body.meetingId.toString()}_${body.attendeeId.toString()}`;
+  const cloudWatchClient = new AWS.CloudWatchLogs({ apiVersion: '2014-03-28' });
+  await cloudWatchClient.createLogStream({
+    logGroupName: logGroupName,
+    logStreamName: logStreamName,
+  }).promise();
+  return response(200, 'application/json', JSON.stringify({}));
```
devalevenkatesh, line 217: Correct me here if this is needed, will it be better to have some data in response, like if the stream successfully got created or for some reason errors out.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -666,13 +666,32 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
...
+  }
+
   async initializeMeetingSession(configuration: MeetingSessionConfiguration): Promise<void> {
     let logger: Logger;
     const logLevel = LogLevel.INFO;
     const consoleLogger = logger = new ConsoleLogger('SDK', logLevel);
     if (location.hostname === 'localhost' || location.hostname === '127.0.0.1') {
       logger = consoleLogger;
     } else {
+      await this.createLogStream(configuration);
```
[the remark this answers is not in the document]
RidipDe: console.error in createLogStream is sufficient as the handler has logic to create a stream if it does not exist.

demos/browser/app/meetingV2/meetingV2.ts
RidipDe: createLogStream is called as soon as the meeting session is initialized. So token exception for the fetch is unlikely to occur.
Moreover the handler has logic to create a stream if it does not exist.

demos/serverless/src/handlers.js
RidipDe: We just use the response status in the demo app to verify successful creation of the stream name.
The handler has logic to create a stream if it does not exist once the app stats emitting logs to cloudwatch.

ltrung (approved):

PR #729, opened 2020-09-28
Increase roster check time for sip test

katayal:
SIp test is failing because roster updates are not coming in time. Increasing the roster check duration.

### DIFF
--- a/integration/js/SipCallTest.js
+++ b/integration/js/SipCallTest.js
@@ -1,5 +1,5 @@
 const {OpenAppStep, JoinMeetingStep, AuthenticateUserStep, GetSipUriForCallStep} = require('./steps');
-const {UserJoinedMeetingCheck, UserAuthenticationCheck, RemoteAudioCheck, RosterCheck} = require('./checks');
+const {UserJoinedMeetingCheck, UserAuthenticationCheck, RemoteAudioCheck, RosterCheck, RosterCheckConfig} = require('./checks');
 const {AppPage} = require('./pages/AppPage');
 const {TestUtils} = require('./node_modules/kite-common');
 const SdkBaseTest = require('./utils/SdkBaseTest');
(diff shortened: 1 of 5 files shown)


integration/js/pages/AppPage.js
```diff
@@ -279,20 +279,23 @@ class AppPage {
...
+          console.log("Roster check completed successfully in : %d secs", (start - end)/1000);
           return true;
         }
       } catch (err) {
       }
-      await TestUtils.waitAround(10);
+      await TestUtils.waitAround(waitTimeMs);
       i++;
     }
+    console.log("Roster check failed in : %d secs", (start - end)/1000);
```
ltrung, line 298: So if roster check fail, where do we populate end variable here?

katayal: oh yeah fixing it

integration/js/pages/AppPage.js
```diff
@@ -279,20 +279,23 @@ class AppPage {
     return await this.driver.findElement(elements.failedMeetingFlow).isDisplayed();
   }
 
-  async rosterCheck(numberOfParticipants) {
+  async rosterCheck(numberOfParticipants, checkCount = 10, waitTimeMs = 100) {
```
ltrung, line 282: nit: Why dont we just pass RosterCheckConfig here?

integration/js/pages/AppPage.js
```diff
@@ -279,20 +279,24 @@ class AppPage {
...
+          console.log("Roster check completed successfully in : %d secs", (start - end)/1000);
           return true;
         }
       } catch (err) {
       }
-      await TestUtils.waitAround(10);
+      await TestUtils.waitAround(waitTimeMs);
       i++;
     }
+    let end = performance.now();
```
ltrung, line 298: Variable end declared twice in this function.

integration/js/SipCallTest.js
```diff
@@ -34,7 +34,7 @@ class SipCallTest extends SdkBaseTest {
     const sipCallClient = new SipCallClient(this.payload.sipTestAssetPath, this.payload.resultPath);
     const sipCall = sipCallClient.call(this.payload.voiceConnectorId, this.sipUri);
 
-    await test_window.runCommands(async () => await await RosterCheck.executeStep(this, session, 2));
+    await test_window.runCommands(async () => await await RosterCheck.executeStep(this, session, 2, new RosterCheckConfig(60, 500)));
```
ltrung, line 37: Will 30 second be enough? I thought this could take 35s before the SIP attendee shows up.

PR #741, opened 2020-09-30
Make browser demo audio input selection asynchronous.

rnewman:
This simply alters the signature of `audioInputSelectionToDevice` to
make future extension more straightforward.

**Issue #:**

None.

**Testing**

This is a demo-only change. The demo runs correctly locally, with input device selection tested across two devices.


anuranduttaroy:
The build failed as the version number in not bumped up in the PR.

rnewman:
Why do we need to bounce the version number when not making a change to the library? Should I file a separate issue to track that bug?

ltrung (approved):

PR #742, opened 2020-09-30
Show video WebRTC stats and attendeeId on video tile hover

devalevenkatesh:
**Issue #:**
NA

**Description of changes:**
- Additions to demo browser app to show WebRTC video stats and attendeeId for upstream and downstream which will help QA team while testing simulcast changes.
- For the upstream, I could not found a way to calculate `packet loss` since I could only find `packetsSent` but not `packetsLost`.
- Similarly for downstream, I could not find `framesPerSecond` will check for other metrics and add later.
- Stats are shown only when using Chrome.

**Testing**

2. How did you test these changes? Locally in chrome, checked other browsers no errors but the simulcast resolutions will be empty 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add a Travis check to make sure version update
 - Add metrics for Selenium initialization metrics for integration tests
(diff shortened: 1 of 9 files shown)


devalevenkatesh:
Will re-open once simulcast changes PR is in, so that one PR at a time can be focused and reviewed.

PR #749, opened 2020-10-01
Extend Logger.debug to support strings, not just functions.

rnewman:
**Issue #:**

None.

**Description of changes:**

I ran into this sharp edge, which manifested as some of my code bailing
out with `debugFunction is not a function`.

This change extends the interface to accept strings as well as
functions.

Additionally, this change adds trivial memoization in `MultiLogger`,
ensuring that debug functions are called exactly once, no matter how
many loggers will consume the resultant string.

**Testing**

Yes.

Unit tests are included. Tested locally with browser demo with additional `logger.debug` calls added for both branches.

I'm not sure: are these logging interfaces intended to be implemented by customers? If so, this will be a minor API signature change which has not been explicitly reviewed or approved.

No.

### DIFF
--- a/src/logger/MultiLogger.ts
+++ b/src/logger/MultiLogger.ts
@@ -32,9 +32,20 @@ export default class MultiLogger implements Logger {
     }
   }
 
-  debug(debugFunction: () => string): void {
+  debug(debugFunction: string | (() => string)): void {
+    let message: string;
+    const memoized =
+      typeof debugFunction === 'string'
+        ? debugFunction
+        : () => {
+            if (!message) {
+              message = debugFunction();
+            }
+            return message;
+          };
+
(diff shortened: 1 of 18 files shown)


src/logger/MultiLogger.ts
```diff
@@ -32,9 +32,20 @@ export default class MultiLogger implements Logger {
     }
   }
 
-  debug(debugFunction: () => string): void {
+  debug(debugFunction: string | (() => string)): void {
+    let v: string;
```
ltrung, line 36: nit: Could we change this v to better variable name?

rnewman: Changed to `message`. Good call.

ltrung (approved):

PR #763, opened 2020-10-07
Fall back to `null` device if there is any error while acquiring the audio device

anuranduttaroy:
**Issue #:** NA

**Description of changes:** 
If there is any error while acquiring the audio device, we fall back to null device.
Reason: If audio device selection fails (e.g. NotReadableError), the peer connection is left hanging with no active audio track since we release the previously attached track. If no audio packet has yet been sent to the server, the server will not emit the joined event.

**Testing**

Yes.

Tested by running the serverless demo and meeting readiness checker. No failures.

NA

NA


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -705,6 +705,17 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+       * If there is any error while acquiring the audio device, we fall back to null device.
+       * Reason: If device selection fails (e.g. NotReadableError), the peer connection is left hanging
+       * with no active audio track since we release the previously attached track.
+       * If no audio packet has yet been sent to the server, the server will not emit the joined event.
+       */
+      if (kind === 'audio') {
+        this.logger.info(`choosing null ${kind} device instead`);
+        newDevice.stream = DefaultDeviceController.createEmptyAudioDevice() as MediaStream;
+        newDevice.constraints = null;
+      }
```
aburnson, line 718: If `createEmptyAudioDevice` throws, this will prevent `deniedForDuration` from running and returning the correct error code. Could we wrap this in a try / catch block and make this a best-effort fallback? The catch can just log out.

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -212,6 +212,13 @@ describe('DefaultDeviceController', () => {
       expect(permission).to.equal(DevicePermission.PermissionGrantedByBrowser);
     });
 
+    it('denies the permission by browser', async () => {
```
simmkyu, line 215: This test doesn't check if the selected device is the `null` device.

simmkyu (changes requested):

simmkyu (approved):

simmkyu (approved):
Please make sure to test your change in supported browsers

anuranduttaroy:
> Please make sure to test your change in supported browsers

Yes. Tested by running the serverless demo app on Chrome, Firefox & Safari.

Issue #719, opened 2020-09-24
Remove the restriction on single commits in a PR


rnewman:
It would be a great deal more humane to configure this repo to require a squash-and-rebase merge for PRs rather than to require in `build:release` that a PR consists of a single commit away from a branch named `origin/master`.

See https://docs.github.com/en/github/administering-a-repository/configuring-pull-request-merges for how to do this.

This would be an experience improvement for the contributor (can write meaningful commits, can use ordinary GitHub development workflows for iteration), the reviewer (can review meaningful commits, can see incremental fixes in the PR), and future spelunkers (can see the PR history without squashed commits). 

This would also provide a small improvement in the fallibility of those custom build scripts. Any developer who doesn't have `origin` as a remote in their repo will have to go through multiple rounds of pushes to submit a PR, because the prebuild script will bomb out without failing the build if `origin/master` doesn't exist.

rnewman:
Fixed by #772.

PR #772, opened 2020-10-10
Read versioning from tags, include SHA-1 of commit.

rnewman:
**Issue #:**

None.

**Description of changes:**

This change uses semver tags stored in Git as the canonical record for the current version, using `git describe` to compute the version number built into `Versioning.ts`, and recording untagged commits using the standard semver delta syntax. A release build will be `v1.19.14`, and the next untagged local commit beyond it will be `v1.19.14+1-gaa46ace`.

That commit hash is also accessible via `Versioning`.

Developers now do not need to modify `Versioning.ts` or `package.json` on each commit: every commit is uniquely identified by a delta beyond the last tagged version, and the version string is baked into the build as part of `npm run build`.

The act of running `scripts/publish` will bump the package version, create the right Git tag to match, then do a build. The build will result in a new version being built into `Versioning` via the generated version file.

Additionally, this change removes the restriction that each build must consist of a single commit on top of `origin/master`, but preserves the restriction that `CHANGELOG.md` must have been modified since the base revision.

The `publish` script has not been tested, for obvious reasons.

The purpose of this change is to allow us to converge on a more normal development process, whereby every change is uniquely identifiable by hash and base version, but an entire sequence of commits can land between assigned version numbers. As such, a version number now corresponds to a release, not to a commit.

With this change every release is tagged twice; I don't know why we use a custom tag format (`amazon-chime-sdk-js@1.19.14`), but I figured I wouldn't break that process.

This commit also rewrites the versioning tests to be simpler by using a regex, as well as cleaning up the `check-code-style.js` script to respect ignored files rather than special-casing `.DS_Store`. Please forgive me.

Feedback welcome; I'm sure the scripting is incomplete.


aburnson (approved):

package.json
```diff
@@ -56,6 +56,7 @@
     "eslint-plugin-simple-import-sort": "^4.0.0",
     "esm": "^3.2.25",
     "fs-extra": "^8.1.0",
+    "git-describe": "^4.0.4",
```
ltrung, line 59: Do we need this dependency? Or can we use git describe directly?

src/versioning/Versioning.ts
```diff
@@ -18,7 +19,14 @@ export default class Versioning {
...
    */
   static get sdkVersion(): string {
-    return '1.19.14';
+    return semverString;
+  }
+
+  /**
+   * Return the SHA-1 of the Git commit from which this build was created.
+   */
+  static get buildSHA(): string {
```
ltrung, line 28: I assume this will fail the unit test coverage :D

script/publish
```diff
@@ -74,14 +78,14 @@ verbose('git fetch origin')
...
 
-File.write(version_file, File.read(version_file).gsub(/return '[.0-9]+';/, "return '#{version_string}';"))
 File.write(changelog_file, File.read(changelog_file).gsub(/\[Unreleased\]/, "[#{version_string}] - #{Date.today}"))
 verbose("npm version #{version_string} --no-git-tag-version")
 
 verbose('git add -A')
 verbose("git commit -m 'Publish #{tag}'")
 verbose("git tag -d #{tag} || true")
 verbose("git tag #{tag}")
+verbose("git tag #{semvertag} -m \"Release #{semvertag}.\"")
```
ltrung, line 88: Why do we need multiple tags?

src/versioning/Versioning.ts
rnewman: Hah, I believe you're right!

script/publish
rnewman: I don't know what the `amazon-chime-sdk-js@1.19.14` tags are for. The `v1.19.14`-style ones are what the rest of the world uses to indicate semver releases, including `git describe`. All of this magic functionality comes for free with Git if we just use the right tag format!

I would be happy to kill the custom one, but I figure it exists for *something*.

script/publish
```diff
@@ -74,14 +78,14 @@ verbose('git fetch origin')
 verbose('git reset --hard origin/master')
 verbose('git clean -ffxd .')
 
-File.write(version_file, File.read(version_file).gsub(/return '[.0-9]+';/, "return '#{version_string}';"))
```
ltrung, line 77: Does this mean the version file will not be updated to the NPM version? 
So if the current version is 1.19.0-20-g1a54723, after running NPM publish, we will have 1.20.0 but the version file will still be 1.19.0-20-g1a54723 right?

src/versioning/Versioning.ts
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import DefaultBrowserBehavior from '../browserbehavior/DefaultBrowserBehavior';
+import { hash, semverString } from './version.json';
```
ltrung, line 5: Do we need to check in version.json?

script/publish
rnewman: The `Versioning.ts` file no longer mentions a version at all.

`src/versioning/version.json` will be updated during `build`, right after `barrelize.js` is run.

That JSON file will be copied into the build along with the `tsc`-compiled `Versioning.js`.

`version.json` is not checked into version control at all: it is materialized at build time by looking at the current Git tags.

src/versioning/Versioning.ts
rnewman: No. It's specifically ignored in `.gitignore` so that nobody accidentally checks it in. It's just how we get the contents of `git describe` into the build output.

package.json
rnewman: It's an exceptionally small module and it's only in `devDependencies`, so I'm happy to save work where I can.

rnewman:
I tested the `publish` script locally by commenting out the destructive operations, and it correctly tags, updates package.json, and so on.

There's now a small hack in `Versioning.ts` to accommodate the chicken-and-egg absence of tagging. It can be removed as soon as someone tags a point release.

rnewman:
Well, that took way too many iterations, but I think this is good to go now.

rnewman:
@aburnson: if you run

```
git checkout master
git tag v1.19.14 -m "Release v1.19.14."
git push origin v1.19.14
```

then we can pre-emptively kill the test workaround before or after this lands.

aburnson (approved):

script/publish
ltrung: @aburnson Should we stop use this tag amazon-chime-sdk-js@#{version_string}?

ltrung (commented):
So if we allow multiple commits in a PR now, do we need a process/mechanism to enforce squashing when merging PR to master?

rnewman:
> So if we allow multiple commits in a PR now, do we need a process/mechanism to enforce squashing when merging PR to master?

Yes, in the general case. (I expect long-running branches like a v2 to result in a fast-forward merge, and to not merge via PR.)

There is a link to how to configure a repo to require squashing in #719.

Once you require squashing, the problem shifts to getting good squash commit messages for non-unit PRs.

My preference is to use a good PR 'description' section as the commit message body, and the PR title as the commit message. There is tooling for this: https://github.com/zachwhaley/squashed-merge-message.

PR #778, opened 2020-10-13
Make maxVideos in JoinFrame public configurable

kai-amzn:
**Issue #:**
None.

**Description of changes:**
Add a field maxVideoStreamsSubscribeTo in MeetingSessionConfiguration for builders to customize the max number of video streams an attendee want to subscribe. JoinFrame sent from SDK client to back end will be updated accordingly.
Remove the hardcode of private maxVideos.

**Testing**
Yes
Local test with demo browser:
Test the case of using media session default configuration, got maxVideoStreamsSubscribeTo equals to 16, local Tincan also get same value.
Test the case of set media session configuration maxVideoStreamsSubscribeTo to 25, got 25 in media session creation, local Tincan also get same value.

Yes, add a field maxVideoStreamsSubscribeTo in MeetingSessionConfiguration, which has been discussed in meeting.

No.


alochbau-amzn:
I would really prefer that we move away from setting a max value in the JOIN frame.  Limiting the list of remote videos returned in the INDEX message is not the best way to achieve this functionality.   The clients then have no indication of how many remote sources are available in the meeting and also no control over which remote sources to view.

The better way to achieve this functionality would be to have the INDEX message contain the full list of remote sources, then have a configuration variable in the clients for maximum number of videos to display.  The client downlink policy would then be modified to pick which remote sources to view based on local logic.

This is also a more scalable solution where we don't have to keep modifying clients and backend logic to increase the maximum number of active video sources allowed.

kai-amzn:
> I would really prefer that we move away from setting a max value in the JOIN frame. Limiting the list of remote videos returned in the INDEX message is not the best way to achieve this functionality. The clients then have no indication of how many remote sources are available in the meeting and also no control over which remote sources to view.
> 
> The better way to achieve this functionality would be to have the INDEX message contain the full list of remote sources, then have a configuration variable in the clients for maximum number of videos to display. The client downlink policy would then be modified to pick which remote sources to view based on local logic.
> 
> This is also a more scalable solution where we don't have to keep modifying clients and backend logic to increase the maximum number of active video sources allowed.

Yes I agree with "have the INDEX message contain the full list of remote sources", I think some of these already covered by @alochbau-amzn 's ongoing work.

The purpose of this PR is to remove the current hard code on max video tiles to display, with the backward compatibility to not impacting any current builder; as my opinion it doesn't conflict with INDEX message/downlink policy change as this only limit max number of video streams client receive.

aburnson (approved):

p-foucht:
Temporarily closing until release

PR #779, opened 2020-10-13
Add a tsconfig.json to browser demo to specify a custom root.

rnewman:
**Issue #:**

Fixing the build output broke the demo build.

**Description of changes:**

This adds a root to a new `tsconfig.json` in the demo repo, allowing all of the source to be loaded.

**Testing**

No. Fingers crossed.

Browser demo now runs.

No.

No.

### DIFF
--- a/demos/browser/package-lock.json
+++ b/demos/browser/package-lock.json
@@ -274,6 +274,7 @@
       "requires": {
         "@types/dom-mediacapture-record": "^1.0.4",
         "detect-browser": "^5.1.0",
+        "long": "^4.0.0",
         "protobufjs": "~6.8.8",
         "resize-observer": "^1.0.0"
       },
--- a/demos/browser/tsconfig.json
+++ b/demos/browser/tsconfig.json
@@ -1,0 +1,10 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "./build",
+    "rootDir": "../.."
+  },
+  "include": [
+    "./app/**/*"
+  ]
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update meeting readiness checker demo app with new regions CPT, MXP, BOM and ICN
 - Update meeting readiness checker demo app to create meeting after the checker starts
 - Alter the versioning script to require less ritual
-- Correct TypeScript build to generate correct artifacts in `build/`
 
 ### Removed
 
@@ -36,6 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix video track sometimes being removed and added on simulcast receive stream switch
 - Enabled termination protection for serverless demo cloudformation stack
 - Simulcast optimizations
+- Correct TypeScript build to generate correct artifacts in `build/`
+- Correct TypeScript configuration for demo app
 
 ## [1.19.0] - 2020-09-29
 ### Added


demos/browser/package-lock.json
```diff
@@ -274,6 +274,7 @@
       "requires": {
         "@types/dom-mediacapture-record": "^1.0.4",
         "detect-browser": "^5.1.0",
+        "long": "^4.0.0",
```
ltrung, line 277: Is this intentional?

demos/browser/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "./build",
```
devalevenkatesh, line 4: This build directory specified by `outDir` needs to be same as the webpack output directory right?

demos/browser/tsconfig.json
```diff
@@ -0,0 +1,10 @@
...
+    "rootDir": "../.."
```
devalevenkatesh, line 5: Also due to this, does it need to have `composite` set to true in the root `tsconfig.json`? I checked TS compiler is giving a red alert for this. Demo works though without adding that composite true with this config.

demos/browser/package-lock.json
rnewman: Generated by the install script, not by me.

demos/browser/tsconfig.json
rnewman: I don't think so; webpack's ts-loader will do its thing.

devalevenkatesh: So with this, two directories get generated first is `build` generated by compiler and the second is `dist` generated by webpack.

rnewman: That's probably ideal: we don't actually want `tsc`'s output written over Webpack's. Perhaps we want `noEmit` here, just using `tsc` for type checking?

devalevenkatesh: I think if webpack uses ts-loader for checking and compiling `ts` files, then we should not emit the output files using `tsconfig`, since webpack generates them.

rnewman:
This gives a correct build structure for the main lib, and:

```
[rnewman:~/sdk]$ ls -l demos/browser/dist 
total 12912
-rw-r--r--  1 rnewman  staff  3292206 Oct 13 14:55 meetingV2-bundle.js
-rw-r--r--  1 rnewman  staff  3316515 Oct 13 14:55 meetingV2.html
[rnewman:~/sdk]$ ls -l demos/browser/build
total 0
drwxr-xr-x   3 rnewman  staff    96 Oct 13 14:55 demos
drwxr-xr-x  74 rnewman  staff  2368 Oct 13 14:55 src
[rnewman:~/sdk]$ npm pack -q
amazon-chime-sdk-js-1.19.16.tgz
[rnewman:~/sdk]$ tar tzf amazon-chime-sdk-js-1.19.16.tgz | fgrep index
package/build/videostreamindex/DefaultVideoStreamIndex.js
package/build/index.js
package/build/videostreamindex/SimulcastVideoStreamIndex.js
package/build/videostreamindex/VideoStreamDescription.js
package/build/videostreamindex/VideoStreamIndex.js
package/build/videostreamindex/DefaultVideoStreamIndex.js.map
package/build/index.js.map
package/build/videostreamindex/SimulcastVideoStreamIndex.js.map
package/build/videostreamindex/VideoStreamDescription.js.map
package/build/videostreamindex/VideoStreamIndex.js.map
package/build/videostreamindex/DefaultVideoStreamIndex.d.ts
package/build/index.d.ts
package/build/videostreamindex/SimulcastVideoStreamIndex.d.ts
package/build/videostreamindex/VideoStreamDescription.d.ts
package/build/videostreamindex/VideoStreamIndex.d.tsi
```

devalevenkatesh (approved):

PR #798, opened 2020-10-20
Handle case where meeting or attendee response property can accept null

RidipDe:
**Issue #:**
In typescript, null is also treated as an object. So the check [this line](https://github.com/aws/amazon-chime-sdk-js/commit/cfcf723552d949a506237465598f7c1fdd824689#diff-1571e62a3f2170289ce6edb2c4853ae210629ff52e8d7f6e28e8640b5aa611a4R180) would turn false and the execution continued.
So, in the current scenario, we do not handle case where meeting or attendee response property is null.

**Description of changes:**
1) We are adding an explicit check to see if a meeting or attendee response attribute is null or not.

**Testing**

Yes
Manually, was able to login to the meeting app
Yes
No

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -206,6 +206,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
           try {
             chimeMeetingId = await this.authenticate();
           } catch (error) {
+            console.error(error);
             const httpErrorMessage = 'UserMedia is not allowed in HTTP sites. Either use HTTPS or enable media capture on insecure sites.';
             (document.getElementById(
               'failed-meeting'
--- a/src/meetingsession/MeetingSessionConfiguration.ts
+++ b/src/meetingsession/MeetingSessionConfiguration.ts
@@ -177,7 +177,9 @@ export default class MeetingSessionConfiguration {
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private toLowerCasePropertyNames(input: any): any {
-    if (typeof input !== 'object') {
+    if (input === null) {
+      return null;
+    } else if (typeof input !== 'object') {
       return input;
     } else if (Array.isArray(input)) {
       return input.map(this.toLowerCasePropertyNames);
--- a/test/meetingsession/MeetingSessionConfiguration.test.ts
+++ b/test/meetingsession/MeetingSessionConfiguration.test.ts
@@ -17,6 +17,104 @@ describe('MeetingSessionConfiguration', () => {
       expect(configuration.urls).to.be.null;
     });
 
+    it('any attribute except mediaplacement can be null', () => {
+      const configuration = new MeetingSessionConfiguration(
+        {
+          MeetingId: 'meeting-id',
+          ExternalMeetingId: null,
+          MediaPlacement: {
+            AudioHostUrl: 'audio-host-url',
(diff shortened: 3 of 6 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -206,6 +206,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
           try {
             chimeMeetingId = await this.authenticate();
           } catch (error) {
+            console.error(error);
```
simmkyu, line 209: Remove the `console.error`

test/meetingsession/MeetingSessionConfiguration.test.ts
```diff
@@ -17,6 +17,38 @@ describe('MeetingSessionConfiguration', () => {
       expect(configuration.urls).to.be.null;
     });
 
+    it('any attribute can be null', () => {
```
simmkyu, line 20: Please add unit tests for these cases too:
- `undefined` property
- An empty string property `''`
- `null` property in the child object, e.g. `MediaPlacement.SignalingUrl` is `null`
- an array of `null` and `undefined` elements, e.g. `[null, undefined, { ... }]`

You can add one more test, including all those cases

simmkyu (changes requested):

demos/browser/app/meetingV2/meetingV2.ts
simmkyu: I understand now. The following code shows the same message "UserMedia is not allowed in HTTP sites" for other erros too. Let's keep this `console.error` for now now to miss other errors.

src/meetingsession/MeetingSessionConfiguration.ts
```diff
@@ -145,12 +145,14 @@ export default class MeetingSessionConfiguration {
       }
       this.meetingId = createMeetingResponse.meetingid;
       this.urls = new MeetingSessionURLs();
-      this.urls.audioHostURL = createMeetingResponse.mediaplacement.audiohosturl;
-      this.urls.screenDataURL = createMeetingResponse.mediaplacement.screendataurl;
-      this.urls.screenSharingURL = createMeetingResponse.mediaplacement.screensharingurl;
-      this.urls.screenViewingURL = createMeetingResponse.mediaplacement.screenviewingurl;
-      this.urls.signalingURL = createMeetingResponse.mediaplacement.signalingurl;
-      this.urls.turnControlURL = createMeetingResponse.mediaplacement.turncontrolurl;
+      if (createMeetingResponse.mediaplacement !== null) {
```
simmkyu, line 148: Note that this `if` condition can be `true` when `createMeetingResponse.mediaplacement` is `undefined`, any number, and any string. Then, accessing properties on those primitives will fail.

src/meetingsession/MeetingSessionConfiguration.ts
```diff
@@ -151,6 +151,7 @@ export default class MeetingSessionConfiguration {
       this.urls.screenViewingURL = createMeetingResponse.mediaplacement.screenviewingurl;
       this.urls.signalingURL = createMeetingResponse.mediaplacement.signalingurl;
       this.urls.turnControlURL = createMeetingResponse.mediaplacement.turncontrolurl;
+
```
simmkyu, line 154: Remove an empty line here

simmkyu (approved):

simmkyu (dismissed):

simmkyu (approved):

Issue #799, opened 2020-10-21
Cannot run browser demo - "not a git repository" error


k-ahn:
I've downloaded the entire repository, navigated to `demos/browser`, initiated `npm run start` as advised then i get this error:

```
C:\Apps\chimeAppDemo\amazon-chime-sdk-js-master\node_modules\git-describe\lib\git-describe.js:81
        throw err;
        ^

Error: Git returned with status 128: fatal: not a git repository (or any of the parent directories): .git
    at handleProcessResult (C:\Apps\chimeAppDemo\amazon-chime-sdk-js-master\node_modules\git-describe\lib\git-describe.js:76:19)        
    at wrapper (C:\Apps\chimeAppDemo\amazon-chime-sdk-js-master\node_modules\lodash\lodash.js:5213:19)
    at gitDescribe (C:\Apps\chimeAppDemo\amazon-chime-sdk-js-master\node_modules\git-describe\lib\git-describe.js:46:16)
    at wrapper (C:\Apps\chimeAppDemo\amazon-chime-sdk-js-master\node_modules\lodash\lodash.js:5213:19)
    at Object.<anonymous> (C:\Apps\chimeAppDemo\amazon-chime-sdk-js-master\script\generate-version.js:7:36)
    at Module._compile (internal/modules/cjs/loader.js:1138:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1158:10)
    at Module.load (internal/modules/cjs/loader.js:986:32)
    at Function.Module._load (internal/modules/cjs/loader.js:879:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! amazon-chime-sdk-js@1.20.1 build: `npm install && node script/barrelize.js && node script/generate-version.js && npm run tsc:src && npm run tsc:guides && npm run copy`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the amazon-chime-sdk-js@1.20.1 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\kevin\AppData\Roaming\npm-cache\_logs\2020-10-20T23_59_42_547Z-debug.log
npm ERR! code ELIFECYCLE
npm ERR! errno 1
[...]

PR #801, opened 2020-10-21
Fall back to a synthetic version number if built with no .git directory. (#799)

rnewman:
**Issue #:**

#799.

**Description of changes:**

This commit makes it possible to build the `version.ts` file without the scaffolding of an actual Git checkout. The generated version number will be pulled from `package.json`. See code comments.

**Testing**

Yes.

Builds work.

No.

No.

### DIFF
--- a/script/generate-version.js
+++ b/script/generate-version.js
@@ -1,10 +1,31 @@
 #!/usr/bin/env node
 
-const { writeFileSync } = require('fs');
+const { readFileSync, writeFileSync } = require('fs');
 const { gitDescribeSync } = require('git-describe');
 const path = require('path');
 
-const description = JSON.stringify(gitDescribeSync(), null, 2);
+let description;
+try {
+  description = JSON.stringify(gitDescribeSync(), null, 2);
+} catch (e) {
+  console.log(`Got error ${e} getting description of current version.`);
+  console.log('For an accurate version number, or for distribution, please use a full Git checkout of this repository.');
+
+  // Probably not a checkout. This is what you'll get if you use
+  // "Download ZIP" in the GitHub UI.
+  // Synthesize a version file
+  // from the current package.json.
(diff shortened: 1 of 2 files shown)


script/generate-version.js
```diff
@@ -1,10 +1,31 @@
...
+  // Synthesize a version file
+  // from the current package.json.
+  const package = readFileSync(path.join(__dirname, '../package.json'));
+  const version = JSON.parse(package).version;
+  console.log(`Using synthetic build description with version ${version}.`)
+
+  description = JSON.stringify({
+    hash: 'gdeadbeef',
+    raw: 'v' + version,
+    semverString: version,
```
devalevenkatesh, line 25: I see `semverString` as just the version here where as in normal git repository `gitDescribeSync` will generate the `semverString` as something like `1.20.2+1.g42b39c4` (seems like the hash is getting attached to it). 

Also this is not adding all the keys like `semver`, `suffix` etc observed in the current generated `version.ts` will that affect the build as well as the publish process later? 

Also this is scoped down to error case so impact will be low if any I assume.

rnewman: By definition it's not possible for us to attach the commit hash of the current commit when we're not in a Git repo. The `gdeadbeef` is there as a sentinel value to support the hash accessor in `Versioning`.

The omitted keys are not used. All consumers of this data should go through `Versioning`.

The sole purpose for this commit is to allow someone to run `npm run build` on a downloaded copy of the repo and have it succeed rather than fail because there's no `.git` directory.

We expect real consumers to use a package manager (most) or to use a local clone (most of the rest), both of which will have correct, detailed version info.

devalevenkatesh: Cool got it thanks!

devalevenkatesh (approved):

devalevenkatesh (approved):

Issue #765, opened 2020-10-07
Allow combination  of optional feature in demo


XHatan:
UI enhancement: the optional feature box is a single selection, it should allow combinations

PR #815, opened 2020-10-26
Change optional features as list of input boxes in demo

XHatan:
**Issue #:**

#765  to allow selection of combination of optional features

**Description of changes:**

Change selection box to list of input boxes to allow selection of combination of optional features.
Also change the simulcast integration test based on the change.

**Testing**

2. How did you test these changes? demo 
3. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? no change to JS SDK source
4. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved? no

In firefox and safari, demo join page looks like

Screen Shot 2020-10-26 at 11 08 00 AM

In chrome, demo join page looks like 
Screen Shot 2020-10-26 at 11 12 44 AM

Checked it also works in android and iOS

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.html
+++ b/demos/browser/app/meetingV2/meetingV2.html
@@ -48,12 +48,21 @@
       </div>
       <div class="row mt-3">
         <div class="col-12">
-          <label for="optional-features">Optional Features</label>
-          <select id="optional-features" class="custom-select" style="width:100%">
-            <option value="" selected>None</option>
-            <option value="simulcast">Enable Simulcast For Chrome</option>
-            <option value="webaudio">Use WebAudio</option>
-          </select>
+          <fieldset>
+            <legend>Choose your optional features</legend>
+            <div class="custom-control custom-checkbox" style="text-align: left;">
+              <input type="checkbox" id="webaudio" class="custom-control-input">
+              <label for="webaudio" class="custom-control-label">Use WebAudio</label>
+            </div>
+            <div class="custom-control custom-checkbox" style="text-align: left;">
+              <input type="checkbox" id="simulcast" class="custom-control-input">
+              <label for="simulcast" class="custom-control-label">Enable Simulcast for Chrome</label>
+            </div>
+            <div class="custom-control custom-checkbox" style="text-align: left;">
+              <input type="checkbox" id="planB" class="custom-control-input">
+              <label for="planB" class="custom-control-label">Disable Unified Plan for Chrome</label>
+            </div>
+          </fieldset>
         </div>
       </div>
       <div class="row mt-3">
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -133,7 +133,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
   audioVideo: AudioVideoFacade | null = null;
   tileOrganizer: DemoTileOrganizer = new DemoTileOrganizer();
   canStartLocalVideo: boolean = true;
-  defaultBrowserBehaviour: DefaultBrowserBehavior;
+  defaultBrowserBehaviour: DefaultBrowserBehavior = new DefaultBrowserBehavior();;
   // eslint-disable-next-line
   roster: any = {};
   tileIndexToTileId: { [id: number]: number } = {};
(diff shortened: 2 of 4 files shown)


demos/browser/app/meetingV2/meetingV2.html
```diff
@@ -48,13 +48,25 @@ <h1 class="h3 mb-3 font-weight-normal">Join a meeting</h1>
...
+            <div class="custom-control custom-checkbox" style="text-align: left;">
+              <input type="checkbox" id="webaudio" class="custom-control-input">
+              <label for="webaudio" class="custom-control-label">Use WebAudio</label>
+            </div>
+
+            <div class="custom-control custom-checkbox" style="text-align: left;">
+              <input type="checkbox" id="planB" class="custom-control-input">
+              <label for="planB" class="custom-control-label">Disable Unified Plan for Chrome</label>
+            </div>
+            </select>
```
richnew10, line 66: I would suggest grouping the two Chrome options together, and perhaps putting Web Audio first.

demos/browser/app/meetingV2/meetingV2.html
```diff
@@ -48,13 +48,25 @@ <h1 class="h3 mb-3 font-weight-normal">Join a meeting</h1>
...
+          </div>
```
richnew10, line 67: I think you have a spurious pair of

```
</select>
</div>
```

at the end here, or maybe just the `select` is redundant. Please double check.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -190,15 +190,23 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
...
+    
     document.getElementById('form-authenticate').addEventListener('submit', e => {
       e.preventDefault();
       this.meeting = (document.getElementById('inputMeeting') as HTMLInputElement).value;
       this.name = (document.getElementById('inputName') as HTMLInputElement).value;
       this.region = (document.getElementById('inputRegion') as HTMLInputElement).value;
+      this.enableSimulcast = (document.getElementById('simulcast') as HTMLInputElement).checked;
+      this.enableWebAudio = (document.getElementById('webaudio') as HTMLInputElement).checked;
+      // js sdk default is true
+      this.enableUnifiedPlanForChromiumBasedBrowsers = !(document.getElementById('planB') as HTMLInputElement).checked;;
```
richnew10, line 209: Double `;;`.

If the default is true, you should probably default the input box as `checked`, no?

richnew10 (changes requested):

demos/browser/app/meetingV2/meetingV2.ts
XHatan: "Disable Unified Plan for Chrome" is default unchecked equivalent to  "MeetingSessionConfiguration" has "enableUnifiedPlanForChromiumBasedBrowsers" to be true.
The selection is "Disable Unified Plan for Chrome" since developer could still disable it by setting the field in MeetingSessionConfiguration to false.

demos/browser/app/meetingV2/meetingV2.html
XHatan: good catch. thanks

richnew10 (approved):

richnew10:
This failed `integ_video`:

```
Generating failure report

=========================================

Run configuration:

 Client 1 : MAC Firefox 76

 Client 2 : MAC Firefox 76

Failure details:

 [firefox] Check the remote videos

 [firefox] Waiting for other participants to complete their video checks
```

richnew10 (approved):

richnew10 (approved):

integration/js/pages/AppPage.js
```diff
@@ -78,8 +78,12 @@ class AppPage {
   }
 
   async chooseUseSimulcast() {
-    let featureSelection = await this.driver.findElement(elements.optionalFeatures);
-    await featureSelection.sendKeys('simulcast');
+    let simulcastFeature = await this.driver.findElement(elements.simulcastFeature);
+    if (simulcastFeature.isSelected()) {
```
richnew10, line 82: This needs to be `await feature.isSelected()`.

integration/js/pages/AppPage.js
```diff
@@ -78,8 +78,12 @@ class AppPage {
...
+      console.log('simulcast is selected');
+    } else {
+      await simulcastFeature.click();
```
richnew10, line 85: And this will raise an error because the label covers the checkbox. You should find the label with a CSS selector:

```
By.css('label[for="simulcast"]')
```

and click that instead.

richnew10 (changes requested):

richnew10:
The errors in this PR imply that we are never testing the simulcast mode, because we never think the checkbox is unchecked (you should add a test line for that!) and so we never click it. I want this change to land so I can use it, but how do we make sure we actually have coverage for simulcast both on and off?

XHatan:
temporarily close this

PR #816, opened 2020-10-27
Client analytics

simmkyu:
To reviewers: Recommend opening the "Meeting Events" guide in another tab and reviewing the code.
https://github.com/aws/amazon-chime-sdk-js/blob/c64631b9c93cdbf76492d53be8f6ffdcdd2c7fa1/guides/06_Meeting_Events.md

**Issue #:**
N/A

**Description of changes:**
- Add the event controller to record meeting history and notify an observer with attributes.
- Capture device and audio-video events.
- Add the guide.
- [Demo] Refactor and add a Lambda function for uploading meeting events in JSON string.
- [Demo] Add the CloudWatch dashboard to the CF template visualizing meeting events

**Testing**

2. How did you test these changes? Deployed the serverless demo and check the dashboard
3. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? Yes


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -513,8 +514,6 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
         (buttonMeetingLeave as HTMLButtonElement).disabled = true;
         this.leave();
         (buttonMeetingLeave as HTMLButtonElement).disabled = false;
-        // @ts-ignore
-        window.location = window.location.pathname;
```
RidipDe, line 517: Is this being removed to ensure the logs go in even after the meeting leave/end button is clicked ?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -704,6 +734,16 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
           logLevel
         ),
       );
+
+      await this.createLogStream(configuration, 'create_browser_event_log_stream');
+      this.meetingEventPOSTLogger = new MeetingSessionPOSTLogger(
+        'SDKEvent',
+        configuration,
+        DemoMeetingApp.LOGGER_BATCH_SIZE,
+        DemoMeetingApp.LOGGER_INTERVAL_MS,
```
RidipDe, line 743: Should we change the interval at which the SDKEvents will be pushed to Cloudwatch to avoid throttling issues.
LOGGER_INTERVAL_MS is used by for 'SDK'.
https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/cloudwatch_limits_cwl.html

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1395,6 +1435,11 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
     this.log(`session stopped from ${JSON.stringify(sessionStatus)}`);
     if (sessionStatus.statusCode() === MeetingSessionStatusCode.AudioCallEnded) {
       this.log(`meeting ended`);
+      // @ts-ignore
+      window.location = window.location.pathname;
```
RidipDe, line 1439: Oh i see, we are performing the redirection only after the 'AudioCallEnded' and not after the leave / end button is clicked.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -667,21 +666,52 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
...
+      case 'audioInputFailed':
+      case 'videoInputFailed':
+      case 'meetingStartFailed':
+      case 'meetingFailed': {
+        // Send the last 5 minutes of events.
+        this.meetingEventPOSTLogger?.info(JSON.stringify({
+          name,
+          attributes: {
+            ...otherAttributes,
+            meetingHistory: meetingHistory.filter(({ timestampMs }) => Date.now() - timestampMs < 5 * 60 * 1000)
```
anuranduttaroy, line 710: Should we store this `5 * 60 * 1000` in a variable which can be configured by the builders?

demos/browser/app/meetingV2/meetingV2.ts
simmkyu: Yes, we are handling redirection in the `audioVideoDidStop`.

demos/browser/app/meetingV2/meetingV2.ts
simmkyu: That's a good point, but it should be okay for two reasons.
- The previous line will add a small delay between two POST loggers.
- The event logger will send only small number of events, e.g. AudioSelected -> Requested -> Succeeded -> Ended

```
// 1. SDK logger starts here

// 2. Wait for the log stream.
await this.createLogStream(configuration, 'create_browser_event_log_stream');

// 3. After a delay, the event logger starts here
this.meetingEventPOSTLogger = new MeetingSessionPOSTLogger(
```

demos/browser/app/meetingV2/meetingV2.ts
simmkyu: Addeed it as a static variable: `MAX_MEETING_HISTORY_MS`

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -706,6 +737,16 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
           logLevel
         ),
       );
+
+      await this.createLogStream(configuration, 'create_browser_event_log_stream');
```
ltrung, line 741: Create 2 log streams after clicking on the Join button can create a noticable delay. Could we run both of them and wait for them at the end?

ltrung (approved):

demos/browser/app/meetingV2/meetingV2.ts
simmkyu: A new revision is creating two log streams simulatenously. Thanks for the suggestion.

```
await Promise.all([
  this.createLogStream(configuration, 'create_log_stream'),
  this.createLogStream(configuration, 'create_browser_event_log_stream')
]);
```

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -700,11 +716,33 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+      } else if (error?.name) {
+        errorMessage = error.name;
+      } else if (error?.message) {
+        errorMessage = error.message;
+      } else {
+        errorMessage = 'UnknownError';
+      }
+
+      if (kind === 'audio') {
+        this.boundAudioVideoController?.eventController.publishEvent('audioInputFailed', {
```
devalevenkatesh, line 731: Question: Would it be better if we have these `string` events attached to some enum since the `eventDidReceive` and `publishEvent` gets an event name based on `EventName` type? For example rather than 'audioInputFailed' we could have this governed by some string enum (not indexed) and use the same enum for `name` argument in the `eventDidReceive`?

simmkyu: `publishEvent` takes in the `EventName` type that is a string union of `audioInputFailed` and other event strings, so you can't call `publishEvent` with a non-event string.

Also, I chose a string union type over a enum for two reasons.
1. A builder doesn't need to import additonal enum files for comparsion.
2. A builder doesn't have to know the enum value.

Example using enums
```
// 1. A JavaScript builder needs to import enum files.
import EventName from './EventName';
import EventAttributes from './EventAttributes';

eventDidReceive(name, attributes): void {
  if (name === EventName.AudioInputFailed) {
     // 2. Before sending this event to the analytics tool, a builder needs to check the value of 
           "EventName.AudioInputFailed" to make sure that it indicates the event name.   
```

Example using a string union type
```
eventDidReceive(name, attributes): void {
   if (name === 'audioInputFailed') {
       // They know the "name" argument is an event-name string.
}
```

devalevenkatesh: Cool, got it thanks for the explanation!

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -383,12 +401,13 @@ export default class DefaultAudioVideoController implements AudioVideoController
...
   stop(): void {
     this.sessionStateController.perform(SessionStateControllerAction.Disconnect, () => {
-      this.actionDisconnect(new MeetingSessionStatus(MeetingSessionStatusCode.Left), false);
+      this.actionDisconnect(new MeetingSessionStatus(MeetingSessionStatusCode.Left), null, false);
     });
   }
 
   private async actionDisconnect(
     status: MeetingSessionStatus,
+    error: Error | null,
```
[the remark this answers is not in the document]
simmkyu: I moved the "error" parameter to 3rd but didn't make it optional. I think it's better for passing values explicitly. Optional parameters are useful for extending public APIs.

anuranduttaroy (approved):

aburnson (approved):

src/audiovideocontroller/AudioVideoController.ts
```diff
@@ -104,4 +105,9 @@ export default interface AudioVideoController extends AudioVideoControllerFacade
    * Returns the [[AudioMixController]] for this audio-video controller.
    */
   readonly audioMixController: AudioMixController;
+
+  /**
+   * Returns the [[EventController]] for this audio-video controller.
+   */
+  readonly eventController: EventController;
```
rnewman, line 112: This is a breaking API change.

PR #825, opened 2020-10-30
Update few VideoTileController and VideoTile APIs documentation

devalevenkatesh:
**Issue #:**
NA

**Description of changes:**
- Update few `VideoTileController` and `VideoTile` APIs documentation to better detail what the underlying code does.

**Testing**

2. How did you test these changes? Just documentation update.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? No need, since documentation update.
4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? Yes, documentation update is reviewed with @XHatan .

### DIFF
--- a/src/videotilecontroller/VideoTileController.ts
+++ b/src/videotilecontroller/VideoTileController.ts
@@ -12,25 +12,29 @@ import VideoTileState from '../videotile/VideoTileState';
  */
 export default interface VideoTileController {
   /**
-   * Binds the video element to the tile. This should also be called any time
-   * the layout of the video element changes, for example, when changing its
+   * Binds the video element to the tile if it exists for the provided tileId.
+   * This should also be called any time the layout of the video element changes, for example, when changing its
    * dimensions.
    */
   bindVideoElement(tileId: number, videoElement: HTMLVideoElement): void;
 
   /**
-   * Unbinds the video element from the tile.
+   * Unbinds the video element from the tile if it exists for the provided tileId.
+   * The video tile's bounded video element and that element's width and height are set to null.
+   * This does not remove the provided tileId mapping from the tile map in the [[DefaultVideoTileController]].
+   * To remove the mapping and destroy the tile for this tileId, you can use [[removeVideoTile]].
    */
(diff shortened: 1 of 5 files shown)


src/videotilecontroller/VideoTileController.ts
```diff
@@ -50,12 +52,14 @@ export default interface VideoTileController {
   getLocalVideoTile(): VideoTile | null;
 
   /**
-   * Pauses the video tile, if it exists.
+   * Pauses the video tile if it exists and sends the updated video tile state
```
alochbau-amzn, line 55: I would suggest adding some explanation here or in the API Overview doc about calling pause on local vs remote tiles.   This API is intended to be called on remote videos.  If called on a local video tile, then the tile will no longer be updated but the local source will continue to be sent into the meeting.

Also include something similar in unpauseVideoTile and the similar API's in VideoTile.ts

src/videotilecontroller/VideoTileController.ts
```diff
@@ -85,7 +89,9 @@ export default interface VideoTileController {
   addVideoTile(): VideoTile;
 
   /**
-   * Removes a tile by tile id.
+   * Disconnects video source from the video element bounded with the video tile,
```
alochbau-amzn, line 92: Seeing as this is part of the facade, should we document the purpose of removeVideoTile and what happens when called on local vs remote?
Maybe also discuss why you may want to unbind a video tile vs remove it?

devalevenkatesh: I have updated this to tell the difference in the API documentation to make it more clear. Thanks! For the local and remote removeVideoTile acts similar since the removeLocalVideoTile calls removeVideoTile internally with the existing local tileId.

src/videotilecontroller/VideoTileController.ts
devalevenkatesh: I have updated it now, thanks!

alochbau-amzn (approved):

anuranduttaroy (approved):

PR #826, opened 2020-10-30
Add verdaccio script for faster release process

michhyun1:
**Issue #:**
n/a

**Description of changes:**
Add verdaccio script that will automate the process of installing the verdaccio module, publishing the new version of the module to verdaccio, and then running the demo app with the newly published module in verdaccio.

**Testing**

2. How did you test these changes? ran the script, verified that it starts the demo app with the correctly installed version of the sdk.

### DIFF
--- a/script/verdaccio.js
+++ b/script/verdaccio.js
@@ -1,0 +1,277 @@
+/*
+    Verdaccio app script
+    This script will install and spin up a verdaccio server, which acts as a "dry-run" npm publishing tool.
+    We will publish our npm module to our locally hosted verdaccio server the same way we would publish to the npm registry.
+    Then we will modify the demo app to pull the sdk npm module from verdaccio, install it, and run the 
+    demo app with the module hosted in our local verdaccio server.
+    Once the demo app is running the script will hang - waiting for manual verification steps to be completed - and after testing
+    is finished the user can type 'yes' to allow the script to clean up and finish.
+*/
+const {
+    logger,
+    spawnOrFail,
+    process,
+    shouldContinuePrompt
+  } = require('./utilities');
+  const path = require('path');
+  const cp = require('child_process');
+  const fs = require('fs');
+  const prompt = require('prompt-sync')();
+  const PORT = 4873;
+  const verdaccioEndpoint = `http://localhost:${PORT}`;
+  const verdaccioAppDir = path.join(__dirname, '../../verdaccio_app');
+  const sdkTestDir = './test-chime-sdk-js';
+  const SDK_APP_NAME = 'amazon-chime-sdk-js';
+  const applicationProcesses = [];
+
+  const cleanupBeforeExit = () => {
+    logger.log('Terminating processes...');
+    applicationProcesses.map(process => {
+      logger.log(`Killing process: ${process.pid}`);
+      spawnOrFail('kill', [`kill -9`, `${process.pid}`]);
+    });
+    cleanup();
+    process.exit(0);
+  };
+
+  cleanup= () => {
+    spawnOrFail('ps', [
+      `-ef | grep -E '(webpack|verdaccio)' | grep -v grep | awk '{ print $2 }' | xargs kill -9`
+    ]);
+  }
+  
+  process.on('SIGTERM', () => {
+    logger.warn("Received SIGTERM, cleaning up.")
+    cleanupBeforeExit();
+  });  
+  process.on('SIGINT', () => {
+    logger.warn("Received SIGINT, cleaning up.")
+    cleanupBeforeExit();
+  });
+  process.on('error', (e) => {
+      logger.error(`Error received: ${e}`);
+      cleanupBeforeExit();
+  })
+  
+  // Create verdaccio testing directory and cd in to it
+  if (!fs.existsSync(verdaccioAppDir)) {
+    logger.log(`Creating verdaccio directory: ${verdaccioAppDir}`);
+    fs.mkdirSync(verdaccioAppDir);
+  }
+  
+  //  Go to verdaccio dir
+  logger.log(`Change directory to verdaccio directory: ${verdaccioAppDir}`);
+  process.chdir(verdaccioAppDir);
+
+  // Initiate basic node package
+  logger.log(`Creating basic node package`);
+  spawnOrFail('npm', ['init', '-f', '-y']);
+  
+  // Install verdaccio within new package
+  logger.log(`Installing Verdaccio`);
+  spawnOrFail('npm', ['i', 'verdaccio']);
+  
+  generateVerdaccioConfigFile();
+  
+  // Start verdaccio application
+  try {
+    startVerdaccioApp();
+  } catch (e) {
+    logger.error('Failed to start verdaccio', e);
+  }
+  
+  // Make SDK test directory
+  if (!fs.existsSync(sdkTestDir)) {
+    logger.log(`Creating SDK testing directory: ${sdkTestDir}`);
+    fs.mkdirSync(sdkTestDir);
+  } else {
+    logger.log(`SDK testing directory already exists: ${sdkTestDir}`);
+  }
+
+  logger.log(`Navigate to SDK testing directory: ${sdkTestDir}`);
+  process.chdir(sdkTestDir);
+  
+  // Clone SDK from github
+  logger.log(`Cloning ${SDK_APP_NAME} into ${sdkTestDir}`);
+  
+  //Clean up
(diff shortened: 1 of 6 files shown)


script/verdaccio.js
```diff
@@ -0,0 +1,220 @@
...
+  // navigate in to cloned SDK
+  process.chdir(path.join(`./${SDK_APP_NAME}`));
+  
+  // Update package.json version
+  let package = fs.readFileSync('package.json', 'utf-8');
+
+  const packageJson = JSON.parse(package);
+  const originalVersion = packageJson.version;
+  const versions = originalVersion.split('.').map(x => parseInt(x));
+  versions[2] += 1;
```
devalevenkatesh, line 107: Do we want to prompt for the MAJOR, MINOR or PATCH in the version here, or for testing its fine to just increase the patch?

script/verdaccio.js
```diff
@@ -0,0 +1,220 @@
...
+  logger.log(`Installing and Building...`);
+  cp.execSync('cd ../.. && npm run build');
+  cp.execSync('npm install');
+  cp.execSync('npm install aws-sdk');
+  cp.execSync('npm run build');
+  const demoAppClientProcess = cp.exec('node server.js');
+  applicationProcesses.push(demoAppClientProcess);
+  
+  logger.log(`
+      Starting V1 meeting demo app. Please wait for the server to start up.
```
devalevenkatesh, line 158: This should be `V2` right? Also are we not including `V1` testing run?

script/verdaccio.js
```diff
@@ -0,0 +1,220 @@
...
+      App will be running on: http://127.0.0.1:8080/
+  
+      Once application is loaded verify that its working properly.
```
devalevenkatesh, line 161: Nit: Indentation for this `logger.log` is a little off.

script/verdaccio.js
michhyun1: Yes, it should be V2, thanks

script/verdaccio.js
ltrung: We may not need to prompt for MAJOR MINOR or PATCH and instead just provide a prompt to enter the version to be used.

script/verdaccio.js
```diff
@@ -0,0 +1,220 @@
...
+  const originalVersion = packageJson.version;
+  const versions = originalVersion.split('.').map(x => parseInt(x));
+  versions[2] += 1;
+  const newVersion = versions.join('.');
+
+  logger.log(`Changing version number from ${originalVersion} to ${newVersion}`);
+
+  // Update version to new version
+  let versionFile = fs.readFileSync('./src/versioning/Versioning.ts', 'utf-8');
+  fs.writeFileSync('./src/versioning/Versioning.ts', versionFile.replace('return VERSION.semverString;', `return '${newVersion};'`));
```
ltrung, line 114: We should not need to do that anymore after Richard's change.
To update package version, we may want to use npm version --no-git-tag-version like in publish script.

script/verdaccio.js
```diff
@@ -0,0 +1,220 @@
...
+      Starting V1 meeting demo app. Please wait for the server to start up.
+      App will be running on: http://127.0.0.1:8080/
+  
+      Once application is loaded verify that its working properly.
+      `);
+  
+  logger.warn(
+    `Have you completed the demo app manual verification steps and are ready to continue with RELEASE?`
+  );
+  shouldContinuePrompt(cleanupBeforeExit);
```
ltrung, line 167: So if you enter Yes here we will proceed to cleanupBeforeExit, if you enter No you also call cleanupBeforeExit right?

ltrung, line 167: However the v1 meeting demo and backward compatibility test?

script/verdaccio.js
michhyun1: We do this for the components release, and IMO its easier to have patch, minor, major. Less room for error

script/verdaccio.js
michhyun1: Yes, thats right

michhyun1: I thought we wont be doing this for the meeting V1 anymore, and I still havent covered the backwards compatibility test yet

script/verdaccio.js
michhyun1: This is following the current steps in the MCM, we manually change this version before we push to verdaccio. Calling 'npm version' would only update hte package.json version, not this folder. I noticed we have a script called generate-version.js, but since this is just a dryrun essentially, I think we can jsut manually update this Versioning file, like we do in the MCM today.

script/verdaccio.js
devalevenkatesh: I think no update is made here right to get the explicit major, minor and patch or version directly as @ltrung said?

script/verdaccio.js
```diff
@@ -0,0 +1,226 @@
...
+    Starting meeting demo app. Please wait for the server to start up.
+    App will be running on: http://127.0.0.1:8080/
+
+    Once application is loaded verify that its working properly.`
+  );
+  
+  logger.warn(
+    `Have you completed the demo app manual verification steps and are ready to continue with RELEASE?`
+  );
+  shouldContinuePrompt(cleanupBeforeExit());
```
devalevenkatesh, line 173: `cleanupBeforeExit` twice? Since `shouldContinuePrompt` calls `cleanupBeforeExit` as a callback it can be either of the below two right?
 1. `shouldContinuePrompt(cleanupBeforeExit);`
2. ```shouldContinuePrompt(); cleanupBeforeExit();```

michhyun1: If we receive a 'yes' and there is no callback, this will call 'exit 1.' So instead, we call this callback to cleanup before exiting

devalevenkatesh: Okok, got it, also I think it should `shouldContinuePrompt(cleanupBeforeExit);` and not `shouldContinuePrompt(cleanupBeforeExit());` does it not give out error for the later?

script/verdaccio.js
ltrung: we are still doing v1 until Anuran remove it :D

script/verdaccio.js
ltrung: We should update the MCM to remove the manually update the versioning. That is not how the publish script works anymore.

script/verdaccio.js
michhyun1: yeah good catch, thanks.

script/verdaccio.js
ltrung: This is not correct though because MCM can do MAJOR, MINOR, PATCH or manual version.

script/verdaccio.js
```diff
@@ -0,0 +1,277 @@
...
+  spawnOrFail('npm', [`install ${SDK_APP_NAME} --registry ${verdaccioEndpoint}`]);
+
+  let meetingV2File = fs.readFileSync('app/meetingV2/meetingV2.ts', 'utf-8');
+  fs.writeFileSync('app/meetingV2/meetingV2.ts', meetingV2File.replace('../../../../src/index', 'amazon-chime-sdk-js'));
+
+  logger.log(`Installing...`);
+  cp.execSync('npm install');
+  cp.execSync('npm install aws-sdk');
+  logger.log(`Building...`);
+  cp.execSync('npm run build');
```
ltrung, line 156: Why dont we just exe 'npm run start' here?

script/verdaccio.js
```diff
@@ -0,0 +1,277 @@
...
+  demoAppClientProcess.kill();
+
+  // Start Backwards compatible testing
+  logger.log("Starting Backwards Compatibility Testing")
+  // Clean up
+  process.chdir('../..');
+  spawnOrFail('rm', ['-rf', `${SDK_APP_NAME}`]);
+
+  let previousVersion = prompt(
+    `Please type the previous NPM version to continue: `
```
ltrung, line 180: Could we just get the latest version from github (there should be a tag for that release)

script/verdaccio.js
michhyun1: Two things:
1. Npm run start runs `npm run deps`, which builds the local module, which we dont need to do since we're using the sdk module from verdaccio
2. There was some issues with spawning/closing the processes created with doing exec(npm run start) because npm creates child processes for each npm command inside of npm run start, so I was running into some issues cleaning it up and actually starting the server.

michhyun1:
Keeping this PR on pause until we decide if we should use verdaccio still

michhyun1:
We'll be replacing verdaccio with just npm pack, so closing this PR

PR #829, opened 2020-11-02
Update instructions for running tests locally.

rnewman:
This gets to the point of (failing) tests for me with a local Selenium grid.

### DIFF
--- a/integration/js/README.md
+++ b/integration/js/README.md
@@ -1,32 +1,139 @@
-## Integration Tests
+# Integration Tests
 
 This guide will help you setup and run integration tests for [Chime SDK](https://github.com/aws/amazon-chime-sdk-js) using [Chime SDK Browser Demo](https://github.com/aws/amazon-chime-sdk-js/tree/master/demos/browser).
 
+## Prerequisites
+
+To run integration tests you will need:
+
+- Node 10 or higher. Node 14 is supported.
+- npm 6.11 or higher. 6.14.8 is supported.
+- [KITE](https://github.com/webrtc/KITE).
+
+### Installing KITE on macOS
+
+Run the following.
+
+```shell
+# Install Maven.
+brew install maven
+
+# If you intend to run tests locally, install browser drivers.
+brew install maven geckodriver chromedriver
+
+# Wherever you like.
+export KITEDIR=~/repos
+cd $KITEDIR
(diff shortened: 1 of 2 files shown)


integration/js/README.md
```diff
@@ -2,13 +2,96 @@
...
+### Running the tests
+
+```shell
+export KITE_JAR="$KITE_HOME/KITE-Engine/target"
+cd integration
+export SELENIUM_GRID_PROVIDER=local
+```
+
+```shell
+$KITE_HOME/scripts/mac/path/r configs/meeting_end_test.config.json
```
ltrung, line 84: Could we change meeting_end_test.config.json to something like <test_name>.config.json or say for example to run meeting end test:

integration/js/README.md
```diff
@@ -2,13 +2,96 @@
...
 
+## Testing locally
+
+To run integration tests you will need:
+
+- Node 10 or higher. Node 14 is supported.
+- npm 6.11 or higher. 6.14.8 is supported.
+- [KITE](https://github.com/webrtc/KITE).
+
+### Installing KITE on macOS
```
ltrung, line 13: This is common for both locally and SauceLabs right. Could we move it outside to before test locally?

ltrung: Could we make it not specific to Mac OS? KITE has setup scripts for Linux and Windows too.

integration/js/README.md
```diff
@@ -2,13 +2,96 @@
...
+```shell
+cd integration/js
+npm install
+cd -
+```
+
+### Running the tests
+
+```shell
+export KITE_JAR="$KITE_HOME/KITE-Engine/target"
```
ltrung, line 78: We already did this step above.

integration/js/README.md
ltrung: We already add $KITE_HOME/scripts/mac/path/ to $PATH above so we could just do r configs/.. right?

integration/js/README.md
rnewman: We should, but I cannot write those docs.

integration/js/README.md
rnewman: No — `r` is a shell builtin, so even if it's in your path, it won't work.

ltrung (approved):

Issue #831, opened 2020-11-03
Video "flashing" off and on when another users joins


stopdropandrew:
**Describe the bug**

The simplest way to test is to join a meeting with three users. Alternate disabling/enabling sending video on two of the clients, while watching the video streams on the third. After you've cycled to get the `receiveStreamIds` array into the state `[0, <streamId>, 0, <streamId>]` when you enable the video you'll see both videos "flash" on the third user's screen.

It seems like the SDP subscribe ack is using the position in that array to set the label (ie `v_2`, `v_4`) and when it changes the trackLabel from `v_3` to `v_2` the client is removing the first video tile and resubscribing.

The issue was introduced in `v1.19.0` - this commit https://github.com/aws/amazon-chime-sdk-js/commit/34db7e4c7992ec865420d2dd2fb35778dd576b62

I've sliced up my logs to highlight the specific issue.

**Timeline**

##### Join a meeting with two other users
```
2020-11-02T22:23:38.673Z [INFO] VideoMeeting - 
adding transceiver mid: null subscription: 25 direction: recvonly
14:23:38.674 ConsoleLogger.ts:81 2020-11-02T22:23:38.674Z [INFO] VideoMeeting - 
adding transceiver mid: null subscription: 27 direction: recvonly
14:23:38.674 ConsoleLogger.ts:78 2020-11-02T22:23:38.674Z [DEBUG] VideoMeeting - 
transceiver index=0 mid=null subscription=0 direction=inactive
transceiver index=1 mid=null subscription=25 direction=recvonly
transceiver index=2 mid=null subscription=27 direction=recvonly
14:23:39.667 ConsoleLogger.ts:78 2020-11-02T22:23:39.667Z [DEBUG] VideoMeeting - 
transceiver index=0 mid=1 subscription=0 direction=sendrecv
transceiver index=1 mid=2 subscription=25 direction=recvonly
transceiver index=2 mid=3 subscription=27 direction=recvonly
```

##### Remote user A disables video
```
14:25:24.465 ConsoleLogger.ts:78 2020-11-02T22:25:24.465Z [DEBUG] VideoMeeting - 
transceiver index=0 mid=1 subscription=0 direction=sendrecv
transceiver index=1 mid=2 subscription=25 direction=recvonly
transceiver index=2 mid=3 subscription=0 direction=inactive
```

##### Remote user A enables video
```
[...]

PR #832, opened 2020-11-03
Don't shift existing streams to new transceivers

stopdropandrew:
**Issue #: https://github.com/aws/amazon-chime-sdk-js/issues/831**

**Description of changes:**

Rather than always shifting videos into available transceiver slots, only use previously available slots for *new* videos.

**Testing**

1. Have you successfully run `npm run build:release` locally? *Yes*
2. How did you test these changes? *Added a test for the specific use case*
3. Can these changes be tested using the demo application? *Should be testable via any app. Fixes the issue in our app*
4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? *No*
5. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved? *No*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Demo] Fix serverless deploy script to not print out logs
 - [Test] Make sure to error out if failed to grab Sauce Lab sessions
 - Fixed deploy github action with correct keys
+- Fix transceiver reordering bug
 
 ## [1.21.0] - 2020-10-29
 ### Added
--- a/docs/classes/defaulttransceivercontroller.html
+++ b/docs/classes/defaulttransceivercontroller.html
@@ -249,7 +249,7 @@
 						<li class="tsd-description">
 							<aside class="tsd-sources">
 								<ul>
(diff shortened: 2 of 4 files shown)


XHatan:
Hey @alochbau-amzn I think this reverted the fix you made, do you mind taking a look ?

alochbau-amzn:
This is a good change.  It is not reverting my change, but just making sure that the same transceiver is always used for existing remote videos.

stopdropandrew:
I pushed a version that fixes the changelog conflict. What's the timeline on a merge and release?

@alochbau-amzn's commit in 1.19.0 is great, it fixed a high impact but relatively uncommon issue - we we're really excited when we saw that in the changelog, thank you! Unfortunately the issue it introduced, that this pull fixes, is highly occurring but relatively low impact, we can't upgrade our SDK, and it is blocking a new feature in our app, until this is in.

alochbau-amzn:
@stopdropandrew We are doing some maintenance and changes to our CI process and checks.  Once that is done we will re-run the checks and if successful should be able to merge it in.  It would then go out in the subsequent release.

alochbau-amzn:
This has been fixed with a similar change made in https://github.com/aws/amazon-chime-sdk-js/pull/844

PR #834, opened 2020-11-03
Add optional features as list of input boxes

XHatan:
**Issue #:**
additional fix for 

https://github.com/aws/amazon-chime-sdk-js/pull/815 

**Description of changes:**

Change selection box to list of input boxes to allow selection of combination of optional features.
Also change the simulcast integration test based on the change.

**Testing**

1). manual testing with demo page
2) local grid and KITE auto testing - box is checked and log show simulcast is selected.
 - ```$KITE_HOME/scripts/mac/path/r configs/video_test_simulcast.config.json```

3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? yes

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -133,7 +133,7 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
   audioVideo: AudioVideoFacade | null = null;
   tileOrganizer: DemoTileOrganizer = new DemoTileOrganizer();
   canStartLocalVideo: boolean = true;
-  defaultBrowserBehaviour: DefaultBrowserBehavior;
+  defaultBrowserBehaviour: DefaultBrowserBehavior = new DefaultBrowserBehavior();;
   // eslint-disable-next-line
   roster: any = {};
   tileIndexToTileId: { [id: number]: number } = {};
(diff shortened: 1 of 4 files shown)


integration/js/pages/AppPage.js
```diff
@@ -77,9 +81,24 @@ class AppPage {
     await authenticateButton.click();
   }
 
+  async chooseUseWebAudio() {
+    let webAudioFeature = await this.driver.findElement(elements.webAudioFeature);
+    let webAudioFeatureLabel = await this.driver.findElement(elements.webAudioFeatureLabel);
+    if (await webAudioFeature.isSelected()) {
+         console.log('Web Audio is selected');
```
richnew10, line 88: Nit: indenting. I can fix this when I come by here again.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -318,28 +326,6 @@ export class DemoMeetingApp implements AudioVideoObserver, DeviceChangeObserver,
...
-      this.enableWebAudio = false;
-      this.enableUnifiedPlanForChromiumBasedBrowsers = true;
-      this.log("Feature lists:");
-      for (let i = 0; i < collections.length; i++) {
-        // hard code magic
-        if (collections[i].value === 'simulcast') {
-          this.enableSimulcast = true;
-          this.log('attempt to enable simulcast');
-          const videoInputQuality = document.getElementById('video-input-quality') as HTMLSelectElement;
-          videoInputQuality.value = '720p';
```
richnew10, line 334: Do we no longer need to do this?

integration/js/pages/AppPage.js
XHatan: updated

demos/browser/app/meetingV2/meetingV2.ts
XHatan: updated

richnew10 (approved):
So this is fine, but that nothing broke implies that there's no integration test coverage for simulcast itself — to test that when we try to enable simulcast we actually get it. Seems like we should address that.

XHatan:
> So this is fine, but that nothing broke implies that there's no integration test coverage for simulcast itself — to test that when we try to enable simulcast we actually get it. Seems like we should address that.

True. The feature testing relies on a button to be clicked. If demo page is changed, we might slip again and not get test coverage.

PR #850, opened 2020-11-08
Slightly rework DefaultEventController to create less garbage.

rnewman:
**Issue #:**

N/A.

**Description of changes:**

This uses slightly more efficient patterns when invoking observers, creates less garbage and does less work each time an event is processed, and additionally reuses the event object across multiple observers within the same event call. This is made safer by using `Object.freeze` to prevent observers from modifying the event. 

Future efficiency improvements can come from avoiding the need to create new event objects at all, by passing the 'state/context' attributes separately from the 'event' attributes.

Future type safety improvements can come from using a discriminated union pattern, rather than defining `EventAttributes` as an attribute soup.

Note that all of the private members appear in the generated API docs. This is partly fixed by adding an `@internal` annotation; the doc generator will respect that when some in-progress code brings that change into the master branch.

**Testing**

Yes.

Only unit tests; no manual testing.

I'm not sure.

No, although the event interface does not make it clear whether objects will be reused.

No.


rnewman:
Something is definitely broken with this CI configuration.

simmkyu (approved):

PR #856, opened 2020-11-12
Modify Github Actions Jobs

michhyun1:
**Issue #:**

**Description of changes:**
Don't use matrix strategy, because we conditionally run the integ test jobs, and the job wont report any results if they are skipped within the matrix strategy. We need to run all the jobs regardless of the contents of the PR, and then skip the integ tests that need to be skipped, so that we can require those tests for all PRs. A skipped test will still allow the workflow to pass. 

**Testing**
Needs to be submitted as a PR to test

2. How did you test these changes? needs to be submitted as a PR to test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Update the clicking sound answer in FAQs
+- Fix Github Actions CI workflow to include all integ tests
 
 ## [1.22.0] - 2020-11-10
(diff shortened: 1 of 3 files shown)


ltrung (approved):

CHANGELOG.md
```diff
@@ -14,7 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 ### Fixed
+<<<<<<< HEAD
```
XHatan, line 17: a merge conflict left over here

PR #881, opened 2020-11-19
Add installation instructions to migration guide. Clean up.

rnewman:
The deprecation notice points to the migration guide, so we should add upgrade/installation instructions here, too.

This commit also fixes a few lint complaints in the CHANGELOG in order to bypass that warning.

### DIFF
--- a/guides/08_Migration_to_2_0.md
+++ b/guides/08_Migration_to_2_0.md
@@ -1,18 +1,29 @@
 # Migration from SDK v1 to SDK v2
 
+## Installation
+
+Installation involves adjusting your `package.json` to depend on version `2.0.0`.
+
+```shell
+npm install --save amazon-chime-sdk-js@2
+```
+
+## Interface changes
+
 Version 2 of the Amazon Chime SDK for JavaScript makes a small number of interface
 changes, as well as removing some deprecated interfaces.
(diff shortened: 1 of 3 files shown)


rnewman:
This will fail because GitHub Actions is not set up for external contributors.

guides/08_Migration_to_2_0.md
```diff
@@ -174,23 +186,26 @@ class MyDeviceController implements DeviceController {
...
 
 Here is the list of new error classes introduced in version 2.0:
-- `GetUserMediaError`
-- `NotFoundError`
-- `NotReadableError`
-- `OverconstrainedError`
-- `PermissionDeniedError`
-- `TypeError`
+
+* `GetUserMediaError`
```
vidya-mohan, line 203: nit: formatting

vidya-mohan (approved):

guides/08_Migration_to_2_0.md
richnew10: Could you clarify, @vidya-mohan? The linter specifically asks for `*` rather than `-`, and I indented the subclasses underneath the parent `GetUserMediaError` class, because it shows the inheritance relationship.

PR #883, opened 2020-11-19
Updated migration document to add more information about `bindAudioDevice()` API behavior

anuranduttaroy:
**Issue #:** NA

**Description of changes:** Migration document update to add more information about `bindAudioDevice()` API.

**Testing**

Yes.

Changes can be seen here: https://github.com/aws/amazon-chime-sdk-js/blob/migration-doc-update/guides/08_Migration_to_2_0.md#update-bindaudioelement-bindaudiostream-and-bindaudiodevice-to-return-promisevoid-instead-of-boolean

NA

NA

NA


guides/08_Migration_to_2_0.md
```diff
@@ -126,6 +126,9 @@ try {
 await audioMixController.bindAudioStream(destinationStream.stream);
 ```
 
+Additionally, the `bindAudioDevice()` will throws an error with a message `'Cannot select audio output device. This browser does not support setSinkId.'`
+if the browser does not support [`setSinkId()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/setSinkId) API. Read more about browsers supporting `setSinkId`: https://caniuse.com/?search=setSinkId
```
RidipDe, line 130: 'if the browser does not support setSinkId API, then...'
We are missing the 'then' part

guides/08_Migration_to_2_0.md
```diff
@@ -126,6 +126,9 @@ try {
 await audioMixController.bindAudioStream(destinationStream.stream);
 ```
 
+Additionally, the `bindAudioDevice()` will throws an error with a message `'Cannot select audio output device. This browser does not support setSinkId.'`
```
devalevenkatesh, line 129: will throws -> will throw

guides/08_Migration_to_2_0.md
devalevenkatesh: I think we can reword this to:
Additionally, if the browser does not support [`setSinkId()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/setSinkId) API, the `bindAudioDevice()` will throw an error with a message `'Cannot select audio output device. This browser does not support setSinkId.'` Check [caniuse.com](https://caniuse.com/?search=setSinkId) for the `setSinkId` browser support.

RidipDe (approved):

devalevenkatesh (approved):

devalevenkatesh (approved):

PR #887, opened 2020-11-20
Generate less garbage in DefaultRealtimeController.

rnewman:
This eliminates the 'wrap' function, which was generating about 20 closures
per second.
    
Extensive testing is added to exercise each failure case. These failures
should be impossible in practice, but there we are.
    
Additionally, this commit removes the prohibition against the use of
stubs. Stubs are already prevalent in the codebase, and using sinon's
improves the understandability of tests.

**Issue #:**

None.

**Testing**

Yes.

Unit testing.

This should apply to all applications.

No.

No.


richnew10:
Merged in #889.

Issue #886, opened 2020-11-20
Tag doesn't match package.json amazon-chime-sdk-js field in singlejs demo


alhambra1:
**Important:** Please attach INFO-level logs with timestamps to the report from *before* the session started up until the point at which the unexpected behavior occured. For multi-participant calls, attach a log file for each participant if possible.

**Meeting and attendee IDs:**
List the meeting and attendee IDs that the attached logs apply to.

**Timeline**
Outline the main actions the user performed and when:

* <timestamp> <action>
* <timestamp> <action>

**Describe the bug**
Describe the bug and the timestamp at which it happened relative to the timestamp in the log file.

**Screenshots**
If applicable, add screenshots to help explain the problem.

**Platform**
 - OS: 
 - Browser: 
 - Browser Version: 
 - Amazon Chime SDK GitHub Commit:

Issue #851, opened 2020-11-09
[TypeError] e.calculateMediaStreamConstraints(src/devicecontroller/DefaultDeviceController) errorCannot set property 'echoCancellation' of undefined


lhong375:
This is reported by Sentry from our app (backyard.co)

**Timeline**
Outline the main actions the user performed and when:
dozen every day starting Nov 6.

* <timestamp> <action>
* <timestamp> <action>

**Describe the bug**

reported by Sentry.  When the error happens, looks like the user failed to grant permissions. I know the type of devices (where we call .echoCancellation on) is either null or string. still somehow we end up with undefined ? we should probably guard against it as well.

*stack trace*
TypeError: Cannot set property 'echoCancellation' of undefined
  at e.calculateMediaStreamConstraints(../../../src/devicecontroller/DefaultDeviceController.ts:868:24)
  at e.<anonymous>(../../../src/devicecontroller/DefaultDeviceController.ts:662:69)
  at ? (2.9e8bd285.chunk.js:8:1)
  at Object.next(2.9e8bd285.chunk.js:8:1)
  at ? (2.9e8bd285.chunk.js:8:1)
  at new Promise(<anonymous>)
  at r(2.9e8bd285.chunk.js:8:1)
  at e.chooseInputDevice(2.9e8bd285.chunk.js:636:5)
  at e.<anonymous>(../../../src/devicecontroller/DefaultDeviceController.ts:102:31)
  at ? (2.9e8bd285.chunk.js:8:1)
  at Object.next(2.9e8bd285.chunk.js:8:1)
  at ? (2.9e8bd285.chunk.js:8:1)
  at new Promise(<anonymous>)
  at r(2.9e8bd285.chunk.js:8:1)
  at e.chooseAudioInputDevice(2.9e8bd285.chunk.js:101:32)
  at chooseAudioInputDevice(../../../src/audiovideofacade/DefaultAudioVideoFacade.ts:329:42)
  at call(../../core/video/src/chime/ChimeRoom.ts:131:61)
  at tryCatch(../../node_modules/regenerator-runtime/runtime.js:63:40)
  at _invoke(../../node_modules/regenerator-runtime/runtime.js:293:22)
  at key(../../node_modules/regenerator-runtime/runtime.js:118:21)
  at asyncGeneratorStep(../../node_modules/babel-preset-react-app/node_modules/@babel/runtime/helpers/esm/asyncToGenerator.js:3:20)
[...]

Issue #674, opened 2020-09-10
interface MediaDeviceInfo is assigned to variable


MangoBird:
I'm getting error 'MediaDeviceInfo is not defined' from official browser demo app sporadically. I found typescript interface MediaDeviceInfo(which doesn't exists after compiled) is assigned to const variable doesNotHaveAccessToMediaDevices in private method updateDeviceInfoCacheFromBrowser of DefaultDeviceController class.

ltrung:
@MangoBird  Could you clarify on what browser do you encounter this issue or do you see it in all browsers? MediaDeviceInfo does not exist usually indicate a browser we do not support.

MangoBird:
@ltrung The browser was Chrome but I don't think the issue is browser related. The issue is with Typescript interface being assigned to a variable which is a grammatical mistake.

DefaultDeviceController.ts
line 411: const doesNotHaveAccessToMediaDevices = !MediaDeviceInfo;   <-- MediaDeviceInfo is not defined at runtime

ltrung:
Yes, `MediaDeviceInfo` is represetend as a TypeScript interface, but we don’t actually assign the interface to a variable. We convert it to a boolean and then assign.
Could you double check whether you use https and not http?

MangoBird:
It's a local environment so I use http for demo application.

@ltrung  Are you saying !MediaDivceInfo is typecasted by an exclamation mark or Is there some build process that swaps all typescript interface into boolean value? 

If you mean the former, it doesn't work because while exclamation can typecast the type undefined to a boolean value, but cannot typecast unrecognizable series of string which MediaDeviceInfo is after typescript code is compiled to javascript.

If you mean the latter, it would work but that's not what I get from demo application after compilation. MediaDeviceInfo remains in the compiled code base as it is not some boolean value.

MangoBird:
Oh my bad. I found that MediaDeviceInfo is defined in the web api. Sorry. I thought it was only typescript Interface that you guys defined. Thank you for your time.

MangoBird:
@ltrung I was thinking about this and I think typecasting MediaDeviceInfo into boolean value to handle the case where the browser doesn't support MediaDeviceInfo API is the wrong way to handle such case as it would throw an error instead of returning true(for the reason I mentioned as a former case above). I think the correct way would be using typeof keyword as demonstrated below.

[...]

PR #900, opened 2020-11-25
Log error if pass undefined to choose input API

ltrung:
**Issue #:**
#674: Prevent error exception when MediaDeviceInfo is not defined.
#851: Log error and return if device parameter is undefined.
#886: Update single.js package.json to use latest v2 version of amazon-chime-sdk-js.

**Testing**

- Unit tests
- Open a web browser and manually call chooseAudioInputDevice(undefined) and verify that an error is logged.
- Open a web browser and manually call delete MediaDeviceInfo and listAudioDevice and verify that there is no error thrown.
No

### DIFF
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -133,6 +133,10 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
   }
 
   async chooseAudioInputDevice(device: AudioInputDevice): Promise<void> {
+    if (device === undefined) {
+      this.logger.error('Audio input device cannot be undefined');
+      return;
+    }
     if (isAudioTransformDevice(device)) {
       // N.B., do not JSON.stringify here — for some kinds of devices this
       // will cause a cyclic object reference error.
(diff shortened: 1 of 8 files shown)


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -799,6 +799,11 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     device: Device,
     fromAcquire: boolean
   ): Promise<void> {
+    if (device === undefined) {
```
simmkyu, line 802: The 2.0 device controller runs more operations after calling `chooseInputIntrinsicDevice()`. For example, if the `device` argument is `undefined`, it'll call `pushVideoMeetingStateForPermissions()`.

https://github.com/aws/amazon-chime-sdk-js/blob/bfc4c600fb7e68f2d358ecb6c7fd096d30b2d430/src/devicecontroller/DefaultDeviceController.ts#L177-L185

Then, the `undefined` device will be considered as `videoInputSelected`.

https://github.com/aws/amazon-chime-sdk-js/blob/bfc4c600fb7e68f2d358ecb6c7fd096d30b2d430/src/devicecontroller/DefaultDeviceController.ts#L129-L134

---

- Please make sure that just returning void for the `undefined` device makes sense in other cases.

- How about throwing an error?

simmkyu (approved):

simmkyu (approved):

PR #901, opened 2020-11-26
Fix generate version script for forked repo PRs

michhyun1:
**Description of changes:**
If a forked repo submits a PR to this repo, doing just a checkout of this repo is not enough. It needs to add the upstream remote repo, fetch down the tags, and then gitDescribeSync will return the correct version. Or we can just create it manually, like we do here. 

**Testing**

2. How did you test these changes? Forked the repo and ran generate-version

### DIFF
--- a/script/generate-version.js
+++ b/script/generate-version.js
@@ -2,19 +2,15 @@
 
 const { readFileSync, writeFileSync } = require('fs');
 const { gitDescribeSync } = require('git-describe');
+const exec = require('child_process').execSync;
 const path = require('path');
 
-let description;
-try {
-  description = JSON.stringify(gitDescribeSync(), null, 2);
-} catch (e) {
-  console.log(`Got error ${e} getting description of current version.`);
-  console.log('For an accurate version number, or for distribution, please use a full Git checkout of this repository.');
-
+function manuallyCreateDescription() {
   // Probably not a checkout. This is what you'll get if you use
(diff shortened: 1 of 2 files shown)


richnew10:
Why not just fetch the tags directly from the remote?

```
git fetch --tags https://github.com/aws/amazon-chime-sdk-js
```

should work fine. Generating a `deadbeef` file should be the last resort.

michhyun1:
> Why not just fetch the tags directly from the remote?
> 
> ```
> git fetch --tags https://github.com/aws/amazon-chime-sdk-js
> ```
> 
> should work fine. Generating a `deadbeef` file should be the last resort.

Yeah, that should work. Thanks

script/generate-version.js
```diff
@@ -27,6 +23,16 @@ try {
   });
 }
 
+let description;
+try {
+  exec(`git fetch --tags https://github.com/aws/amazon-chime-sdk-js`);
```
richnew10, line 28: Move this to its own `try` block: the error message below will be confusing if fetching tags fails because _e.g._, the building machine is offline, and if this fetch fails you should still move on to try the next clause, because we might have the correct tags already.

richnew10 (changes requested):

script/generate-version.js
michhyun1: yup, thanks. Fixed

richnew10 (approved):

Issue #908, opened 2020-11-30
globalThis is not defined


rlebosse:
When importing code from `amazon-chime-sdk-js`, I have a blocking error when I am using a browser that does not support `globalThis`.

I run into the following error : `globalThis is not defined`

**Platform**
 - Browser: Edge 18 / IE 11 / Chrome 67
 - Amazon Chime SDK version : 2.1.0

rlebosse:
I submitted the following PR to fix the problem : https://github.com/aws/amazon-chime-sdk-js/pull/909

richnew10:
Fixed by #910; please reopen if needed.

PR #909, opened 2020-11-30
Hotfix/globalthis is not defined (critical)

rlebosse:
**Issue #:** https://github.com/aws/amazon-chime-sdk-js/issues/908

**Description of changes:** Check that `globalThis` is defined before trying to extract `globalThis['AudioWorkletNode']`

**Testing**


rnewman:
Thanks for the PR, @rlebosse.

`globalThis` is available in all browsers that the Amazon Chime SDK supports:

https://docs.aws.amazon.com/chime/latest/dg/meetings-sdk.html#mtg-browsers
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#Browser_compatibility

Can you tell us where you're seeing issues?

Additionally, the code in `libs` is imported from a separate repo; this change would need to be made upstream.

rlebosse:
I'm facing the error as soon as I import some code from `amazon-chime-sdk-js`.

No need to try to execute some code from the `amazon-chime-sdk-js` lib to get the error, the import is enough.

In my own opinion, that should not be the case due to the fact that we need to be able to call `new BrowserBehavior().isSupported()` to know if the current browser is supported by the JS SDK.

rnewman:
Which browser are you using?

rlebosse:
I ran into the issue on Edge 18 / IE 11 / Chrome 67

rnewman:
I would be quite surprised if SDK code loads at all in IE11, because it necessarily uses ES2015 syntax like `class`, and successfully transpiling classes that extend built-in classes (e.g., `AudioWorkletNode`) to ES5 is quite tricky: if you want to support IE, you will likely need to use code splitting in Webpack or some other mechanism to deliver different code to legacy browsers.

I am preparing an upstream change to introduce a guard similar to the one in this PR, and extend it to the support check, also.

richnew10:
#910 just landed; please test against `master` and reopen #908 if needed.

rlebosse:
@richnew10 , I tested against `master`, it works perfectly, the error is gone, thanks a lot for your reactivity !

richnew10:
Glad to hear it! This will go out with the next dot release; it's safe for you to track `master` in the interim.

PR #910, opened 2020-11-30
Adjust usage of `globalThis` in Amazon Voice Focus.

richnew10:
**Issue #:**

#908.

**Description of changes:**

This commit allows built SDK code to load in unsupported browsers that
don't include `globalThis`, and also adds an explicit support check to
reject those environments when evaluating support for Amazon Voice Focus.

**Testing**

No.

Running Voice Focus's own test suite. Integration tests will catch the rest.

"Test" is a weird word here.

No.

No.


richnew10:
N.B., the `export`/`const` style change is due to upgrading TypeScript in the upstream VF repo — this is all generated code.

XHatan (approved):

ltrung (approved):

PR #911, opened 2020-11-30
Add VideoProcessor API

XHatan:
**Issue #:**

**Description of changes:**

**Testing**


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1757,11 +1808,74 @@ export class DemoMeetingApp implements
...
+      return this.chosenDevice.transformDevice;
+    }
+
+    const proc = this.videoFilterToProcessor(this.videoFilter);
+    // different processor
+    if (this.chosenDevice.transformDevice) {
+      await this.chosenDevice.transformDevice.stop();
+    }
+
+    const transformDevice = new DefaultVideoTransformDevice(new ConsoleLogger('Demo'), intrinsicDevice, [proc]);
```
richnew10, line 1866: Reuse `this.logger`.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1757,11 +1808,74 @@ export class DemoMeetingApp implements
...
+    return this.chosenDevice.transformDevice;
+  }
+
+  private async videoInputSelectionToDevice(
+    value: string,
+  ): Promise<VideoInputDevice> {
+    const intrinsicDevice = this.videoInputSelectionToIntrinsicDevice(value);
+    if (intrinsicDevice === null) {
+      return null;
+    }
```
richnew10, line 1877: Question: is it meaningful to transform the null device?

demos/browser/app/meetingV2/videofilter/CircularCut.ts
```diff
@@ -0,0 +1,71 @@
...
+  private canvasVideoFrameBuffer = new CanvasVideoFrameBuffer(this.targetCanvas);
+  private sourceWidth: number = 0;
+  private sourceHeight: number = 0;
+
+  constructor(private radius: number = 150) {}
+
+  name(): string {
+    return 'CircularCut';
+  }
+
```
richnew10, line 28: I would remove this and see how it changes your demo app.

demos/browser/app/meetingV2/videofilter/CircularCut.ts
```diff
@@ -0,0 +1,71 @@
...
+  destroy(): Promise<void> {
+    return;
+  }
```
richnew10, line 31: Discard the canvas, context, and buffer.

demos/browser/app/meetingV2/videofilter/ResizeProcessor.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  constructor(private displayAspectRatio: number) {}
+
+  name(): string {
+    return 'Resize';
+  }
+
+  destroy(): Promise<void> {
+    return;
+  }
```
richnew10, line 33: Discard stuff.

src/videoframeprocessor/DefaultVideoFrameProcessorPipeline.ts
```diff
@@ -0,0 +1,251 @@
...
+  ): void {
+    for (const observer of this.observers) {
+      setTimeout(() => {
+        observerFunc(observer);
+      }, 0);
+    }
+  }
+
+  private destroyInputMediaStreamAndBuffers(): void {
+    if (!!this.inputVideoStream) {
```
richnew10, line 239: No need for this boolean trick.

src/videoframeprocessor/DefaultVideoFrameProcessorPipeline.ts
```diff
@@ -0,0 +1,251 @@
...
+        });
+      }
+    }
+
+    // measures time
+    const processVideoLatency = performance.now() - processVideoStart;
+    const leave = (1000 * 2) / this.framerate - processVideoLatency; // half fps
+    const nextFrameDelay = Math.max(0, 1000 / this.framerate - processVideoLatency);
+
+    if (leave < 0) {
```
richnew10, line 216: Should this be `<=`?

src/videoframeprocessor/DefaultVideoTransformDevice.ts
```diff
@@ -0,0 +1,231 @@
...
+  private async createVideoFrameProcessorPipeline(): Promise<VideoFrameProcessorPipeline> {
+    if (!this.pipe) {
+      this.pipe = new DefaultVideoFrameProcessorPipeline(this.logger);
+      this.pipe.addObserver(this);
+    }
+
+    this.pipe.processors = this.processors;
+
+    return this.pipe;
+  }
```
richnew10, line 169: Why doesn't the pipeline own the processors?

richnew10 (changes requested):

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -104,6 +109,20 @@ const VOICE_FOCUS_SPEC = {
   paths: VOICE_FOCUS_PATHS,
 };
 
+enum VideoFilterNameEnum {
+  EMOJIFY = 'Emojify',
+  CIRCULAR_CUT = 'CircularCut',
+  NO_OP = 'NoOp',
+}
```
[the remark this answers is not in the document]
XHatan: sounds good, just curious, why is "type" preferred over "enum" ?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -104,6 +109,20 @@ const VOICE_FOCUS_SPEC = {
...
+
+const VIDEO_FILTERS = [
+  'None', 
+  VideoFilterNameEnum.EMOJIFY, 
+  VideoFilterNameEnum.CIRCULAR_CUT, 
+  VideoFilterNameEnum.NO_OP
```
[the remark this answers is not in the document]
XHatan: With `None`  we don't use transform device, with `NoOp` we use transform device with a pass-through processor.

demos/browser/app/meetingV2/meetingV2.ts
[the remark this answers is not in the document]
davejlane: @XHatan here's a good article on this subject https://blog.logrocket.com/why-typescript-enums-suck/

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -541,10 +561,11 @@ export class DemoMeetingApp implements
             await this.openVideoInputFromSelection(camera, false);
             this.audioVideo.startLocalVideoTile();
           } catch (err) {
-            this.log('no video input device selected');
+            this.log(`failed to start local video tile due to ${err}`);
           }
         } else {
           this.audioVideo.stopLocalVideoTile();
+          await this.cleanUpChosenTransformDevice();
```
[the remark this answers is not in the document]
XHatan: We could only calls `stop` on the transform device and keep it ?

demos/browser/app/meetingV2/meetingV2.ts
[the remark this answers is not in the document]
richnew10: It's a long story! In short, for the kind of thing you're doing here — listing a restricted set of names — doing this explicitly by naming points in the space of strings is both clearer and simpler.

Enums with associated strings are so often a bad pattern: devs use them as `#define`s, you're tempted to use them for 'parsing', etc. etc.

Either use a pure enum to identify points in a value space:

```typescript
enum VideoFilters {
  emojify,
  circularCut,
  noOp,
}
```

or use an enum to associate points in the enum space with values:

```typescript
enum VideoFilterConstructors {
  emojify = EmojiFilter,
  circularCut = CircularCutFilter,
  noOp = NoOpFilter,
}
```

or use type unions to identify points in a label space (see above).

Don't use an enum with associated types to implement the "const string" pattern

```c
#define VIDEO_FILTER_UPPER_CASE "CapitalCamelCase"
```

in TypeScript as

```typescript
enum VideoFilters {
  UPPER_CASE = "CapitalCamelCase"
}
```

Make sense?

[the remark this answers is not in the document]
XHatan: Thanks for the explanation @richnew10  and @davejlane , I still need to take a bit of time to soak up the information

demos/browser/app/meetingV2/meetingV2.ts
XHatan: There can be a use case where a user would want to generate a scene and send the stream, in that case, I imagine the inner device is `null`.
device controller will be confused by the `null` and this use case is not supported right now.

src/videoframeprocessor/DefaultVideoTransformDevice.ts
XHatan: updated.  `pipe.destroy` will destroy the processors,  so `transformDevice.stop` will release pipe + processors, does this help ?

src/videoframeprocessor/DefaultVideoFrameProcessorPipeline.ts
XHatan: changed

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1757,11 +1808,74 @@ export class DemoMeetingApp implements
...
+    }
+
+    return null;
+  }
+
+  private chosenDevice: { transformDevice: DefaultVideoTransformDevice | undefined; processor: VideoFrameProcessor;  innerDevice: Device} = {
+    transformDevice: undefined,
+    processor: undefined,
+    innerDevice: undefined
+  };
```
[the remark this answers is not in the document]
XHatan: updated, so demo keeps `{ filterName, processor }` pair and a `chosenDevice` for transform device.

demos/browser/app/meetingV2/videofilter/CircularCut.ts
XHatan: updated

demos/browser/app/meetingV2/videofilter/ResizeProcessor.ts
XHatan: done

src/videoframeprocessor/DefaultVideoFrameProcessorPipeline.ts
XHatan: updated

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1374,9 +1395,40 @@ export class DemoMeetingApp implements
...
 
+  private videoFilter: string = 'None';
+  private async populateVideoFilterInputList(): Promise<void> {
+    const genericName = 'Filter';
+    let filters: string[] = [];
+    if (this.defaultBrowserBehaviour.supportsVideoProcessing() && this.enableUnifiedPlanForChromiumBasedBrowsers) {
+      filters = VIDEO_FILTERS;
+    } else {
+      filters.push('None');
+    }
```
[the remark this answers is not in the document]
XHatan: updated

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1757,11 +1808,74 @@ export class DemoMeetingApp implements
...
+  };
+
+  private async cleanUpChosenTransformDevice(): Promise<void> {
+    if (this.chosenDevice && this.chosenDevice.transformDevice) {
+      await this.chosenDevice.transformDevice.stop();
+    }
+    this.chosenDevice.transformDevice = null;
+    this.chosenDevice.processor = null;
+    this.chosenDevice.innerDevice = null;
+  }
```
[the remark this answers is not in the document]
XHatan: updated so demo calls  `chosenDevice.stop` when it is done with the device.

demos/browser/app/meetingV2/videofilter/CircularCut.ts
XHatan: updated and demo do not call this function.

demos/browser/app/meetingV2/meetingV2.ts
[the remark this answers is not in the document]
richnew10: I suspect something like that. See where we call

```
this.voiceFocusDevice?.stop();
```

in the existing code.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1374,9 +1389,38 @@ export class DemoMeetingApp implements
...
+    await this.populateVideoFilterInputList();
     await this.populateAudioOutputList();
   }
 
+  private videoFilter: string = 'None';
+  private async populateVideoFilterInputList(): Promise<void> {
+    const genericName = 'Filter';
+    let filters: string[] = ['None'];
+    if (this.defaultBrowserBehaviour.supportsVideoProcessing() && this.enableUnifiedPlanForChromiumBasedBrowsers) {
+      filters.concat(VIDEO_FILTERS);
```
richnew10, line 1401: `concat` returns a new array, so this won't do anything. Did you test this?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1757,11 +1800,69 @@ export class DemoMeetingApp implements
...
+
+  private videoFilterToProcessor(videoFilter: string): VideoFrameProcessor | null {
+    this.log(`Choosing video filter ${this.videoFilter}`);
+
+    if (videoFilter === 'Emojify') {
+      return new EmojifyVideoFrameProcessor('🚀');
     }
 
-    return value;
+    if (videoFilter === 'CicurlarCut') {
```
richnew10, line 1813: Typo. Did you test this?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1757,11 +1800,69 @@ export class DemoMeetingApp implements
...
+      return new NoOpVideoFrameProcessor();
+    }
+
+    return null;
+  }
+
+  private chosenProcessor: { name: string | undefined; processor: VideoFrameProcessor | undefined; } = {
+    name: undefined,
+    processor: undefined
+  };
```
richnew10, line 1827: You never use `processor` — it's adopted by the transform device. Just track the name: `selectedVideoFilter`.

demos/browser/app/meetingV2/meetingV2.ts
XHatan: thanks, updated

src/devicecontroller/VideoTransformDevice.ts
```diff
@@ -23,6 +23,10 @@ export default interface VideoTransformDevice {
    */
   applyProcessors(mediaStream?: MediaStream): Promise<MediaStream>;
 
+  /**
+   * `onOutputStreamDisconnect` is called when device controller disconnects the transformed video stream.
+   */
+  onOutputStreamDisconnect(): void;
```
richnew10, line 29: Newline after.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1757,11 +1800,64 @@ export class DemoMeetingApp implements
...
+      return new NoOpVideoFrameProcessor();
+    }
+
+    return null;
+  }
+
+  private chosenDevice: DefaultVideoTransformDevice;
+  private chosenProcessor: string;
+
+  private async cleanUpChosenTransformDevice(): Promise<void> {
```
richnew10, line 1827: `stopVideoTransformDevice`?

demos/browser/app/meetingV2/meetingV2.ts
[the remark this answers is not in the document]
richnew10: `stop` for a transform device is defined as "free any resources associated with the device" — it's not something that you can restart, so you should move this call (inlined, probably) to `leave` so we can reuse the selected device when turning video on and off.

src/videoframeprocessor/DefaultVideoTransformDevice.ts
```diff
@@ -0,0 +1,224 @@
...
+  private observers: Set<DefaultVideoTransformDeviceObserver> = new Set<
+    DefaultVideoTransformDeviceObserver
+  >();
+
+  constructor(
+    private logger: Logger,
+    private device: Device,
+    private processors: VideoFrameProcessor[],
+    private browserBehavior: DefaultBrowserBehavior = new DefaultBrowserBehavior(),
+    private pipe: DefaultVideoFrameProcessorPipeline | null = null
```
richnew10, line 30: Why do you need to take a pipeline as an argument, when you're taking a list of processors as an argument? Just create the pipeline once, and pass the processors to its constructor.

src/videoframeprocessor/DefaultVideoTransformDevice.ts
```diff
@@ -0,0 +1,224 @@
...
+   * Create {@link VideoFrameProcessorPipeline} if there is not a existing one and start video processors.
+   * Returns output `MediaStream` produced by {@link VideoFrameProcessorPipeline}.
+   */
+  async applyProcessors(mediaStream?: MediaStream): Promise<MediaStream> {
+    await this.createVideoFrameProcessorPipeline();
+    await this.pipe.setInputMediaStream(mediaStream);
+    this.inputMediaStream = mediaStream;
+    return this.pipe.outputMediaStream;
+  }
+
```
richnew10, line 107: If the pipeline is empty, will it pass through the stream unmolested (e.g., defaulting to a `NoOp` step)?

If the pipeline is absent (I suggest you prevent that case by not having it as an argument!), can you make `DVTD` just pass through the `mediaStream` here?

src/videoframeprocessor/VideoFrameProcessor.ts
```diff
@@ -0,0 +1,26 @@
...
+ * [[VideoFrameProcessor]] is defined as a unit of video processing.
+ * It transforms an array of {@link VideoFrameBuffer} and produces an array of {@link VideoFrameBuffer}.
+ * It can be chained together to be used with {@link VideoFrameProcessorPipeline}.
+ */
+export default interface VideoFrameProcessor {
+  /**
+   * Returns the name of the processor
+   */
+  name(): string;
+
```
richnew10, line 16: Is this used?

src/videoframeprocessor/DefaultVideoTransformDevice.ts
```diff
@@ -0,0 +1,224 @@
...
+  private async createVideoFrameProcessorPipeline(): Promise<VideoFrameProcessorPipeline> {
+    if (!this.pipe) {
+      this.pipe = new DefaultVideoFrameProcessorPipeline(this.logger);
+      this.pipe.addObserver(this);
+    }
+
+    // transfer processor owner to pipe
+    this.pipe.processors = this.processors;
```
richnew10, line 159: Why isn't this passed to `DVFPP`'s constructor just above?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1744,8 +1795,8 @@ export class DemoMeetingApp implements
     return this.audioInputSelectionWithOptionalVoiceFocus(inner);
   }
 
-  private videoInputSelectionToDevice(value: string): VideoInputDevice {
-    if (this.isRecorder() || this.isBroadcaster()) {
+  private videoInputSelectionToIntrinsicDevice(value: string): Device {
+    if (this.isRecorder() || this.isBroadcaster() || value === 'None') {
```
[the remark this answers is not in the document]
richnew10: ⬆️

src/videoframeprocessor/DefaultVideoTransformDevice.ts
XHatan: Yes, if the pipeline is empty, it will pass through the stream like a no op step.

I move the pipeline out of argument. And `DVTD` should always have a pipeline if it is to be used by device controller

demos/browser/app/meetingV2/meetingV2.ts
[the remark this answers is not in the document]
XHatan: Could you explain a little bit ? What's leaking ?

I think 'None' as video device is intentional to test chooseVideoInputDevice(null).

demos/browser/app/meetingV2/meetingV2.ts
XHatan: sounds good, also move this call to `leave()`

src/videoframeprocessor/VideoFrameProcessor.ts
XHatan: not in demo. I kind of like processor has identifier but I could remove it if you think it's too extra.

demos/browser/app/meetingV2/meetingV2.ts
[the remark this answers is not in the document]
richnew10: I mean: the idea of whether you have picked `'None'` in a picker is 'above' the decision about intrinsic versus transformed, right? (Perhaps the recorder/broadcaster checks are, too?)

From that perspective, you shouldn't even get to `vISTID` if you are trying to pick the null device. See what I mean?

src/videoframeprocessor/VideoFrameProcessor.ts
richnew10: I think a processor could implement `toString` if it wants to, right? If you're logging or otherwise inspecting this thing, you might want to print `<Crop 80x80>`, no?

This is the danger of defining APIs without having some concrete uses in mind — `name` might be too rigid a concept.

XHatan: I see. Make sense. Removed

demos/browser/app/meetingV2/meetingV2.ts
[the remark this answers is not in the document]
XHatan: I see.
it should be something like 
```

  private async videoInputSelectionToDevice(
    value: string,
  ): Promise<VideoInputDevice> {
    if (this.isRecorder() || this.isBroadcaster() || value === 'None') {
      return null;
    }
    const intrinsicDevice = this.videoInputSelectionToIntrinsicDevice(value);
    return await this.videoInputSelectionWithOptionalFilter(intrinsicDevice);
  }
```

?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1757,11 +1796,64 @@ export class DemoMeetingApp implements
...
+    await this.chosenDevice?.stop();
+    this.chosenDevice = null;
+  }
+
+  private async videoInputSelectionWithOptionalFilter(innerDevice: Device): Promise<VideoInputDevice> {
+    if (this.videoFilter === 'None') {
+      return innerDevice;
+    }
+    
+    if (this.chosenDevice && this.videoFilter === this.chosenProcessor) {
```
richnew10, line 1833: You have a bit of a collision here between `videoFilter` (the one selected in the dropdown) and `chosen*` (the one that's active).

I suggest renaming `videoFilter` to `selectedVideoFilterItem`, and using `chosenVideoFilter` for the latter.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1757,11 +1796,64 @@ export class DemoMeetingApp implements
...
+    }
+
+    const proc = this.videoFilterToProcessor(this.videoFilter);
+    this.chosenProcessor = this.videoFilter;
+    // different processor
+    if (this.chosenDevice) {
+      await this.chosenDevice.stop();
+    }
+
+    this.chosenDevice = new DefaultVideoTransformDevice(new ConsoleLogger('Demo'), innerDevice, [proc]);
```
richnew10, line 1845: Try not to mix solidifying the selection of the name with the update of the device. Do one first and then the other.

demos/browser/package.json
```diff
@@ -28,11 +28,18 @@
     "ts-loader": "^8.0.4",
     "typescript": "^4.0.0",
     "webpack": "^4.44.2",
-    "webpack-cli": "^3.3.12"
+    "webpack-cli": "^3.3.12",
+    "source-map-loader": "^1.0.0"
```
richnew10, line 32: Please add these in order.

demos/browser/package.json
```diff
@@ -28,11 +28,18 @@
...
   },
   "dependencies": {
+    "@tensorflow-models/body-pix": "^2.0.5",
+    "@tensorflow/tfjs": "^2.7.0",
+    "@tensorflow/tfjs-backend-webgl": "^2.7.0",
+    "@tensorflow/tfjs-converter": "^1.7.4",
+    "@tensorflow/tfjs-core": "^2.7.0",
+    "@types/webgl2": "0.0.5",
```
richnew10, line 40: Do you need all of these?

integration/js/utils/SdkBaseTest.js
```diff
@@ -28,6 +28,10 @@ class SdkBaseTest extends KiteBaseTest {
     if (this.useSimulcast) {
       this.testName += 'Simulcast';
     }
+    this.useVideoProcessor = !!this.payload.useVideoProcessor ? true : false;
```
richnew10, line 31: `!!this.payload.useVideoProcessor` is sufficient.

src/audiovideocontroller/AudioVideoController.ts
```diff
@@ -34,6 +34,15 @@ export default interface AudioVideoController extends AudioVideoControllerFacade
...
   restartLocalVideo(callback: () => void): void;
 
+  /**
+   * Replaces the local video device without a WebRTC negotiation on the sender side and
+   * avoids that all video subscribers to re-negotiate at the same time.
+   * This is an optimized way of replacing video stream and requires `WebRTC Unified Plan`.
+   * See {@link MeetingSessionConfiguration.enableUnifiedPlanForChromiumBasedBrowsers} and
+   * {@link BrowserBehavior.requiresUnifiedPlan}.
+   */
+  replaceLocalVideo?(callback: () => void): Promise<void>;
```
richnew10, line 44: It is insane to both return a `Promise` and take a callback. Do one or the other. I recommend killing the callback.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -745,22 +838,40 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+      if (fromVideoTransformDevice) {
+        // similar to `useWebaudio`, either Device or VideoTransformDevice, `this.activeDevices['video']` tracks the supplied inner Device.
+        // Upon in-meeting switching to VideoTransformDevice, device controller releases old "supplied" stream and
+        // calls replaceLocalVideo to avoid a full stop-start update.
+        await this.boundAudioVideoController.replaceLocalVideo(() => {
+          this.logger.info('successfully replaced video track');
+          if (oldStream && oldStream.active) {
+            this.logger.warn('previous media stream is not stopped during restart video');
+            this.releaseMediaStream(oldStream);
+          }
```
richnew10, line 860: Just turn the callback into code after the `await`!

richnew10 (changes requested):
Nearly done!

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -541,7 +555,7 @@ export class DemoMeetingApp implements
             await this.openVideoInputFromSelection(camera, false);
             this.audioVideo.startLocalVideoTile();
           } catch (err) {
-            this.log('no video input device selected');
+            this.log(`failed to start local video tile due to ${err}`);
```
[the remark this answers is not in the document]
XHatan: is `` more costly ?

demos/browser/package.json
XHatan: removed.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -992,6 +1006,7 @@ export class DemoMeetingApp implements
     this.statsCollector.resetStats();
     this.audioVideo.stop();
     this.voiceFocusDevice?.stop();
+    this.stopChosenTransformDevice();
     this.voiceFocusDevice = undefined;
```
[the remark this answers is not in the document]
XHatan: oops, the method is removed and cleaning up is done inline

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1757,11 +1796,64 @@ export class DemoMeetingApp implements
...
+
+    if (videoFilter === 'NoOp') {
+      return new NoOpVideoFrameProcessor();
+    }
+
+    return null;
+  }
+
+  private chosenDevice: DefaultVideoTransformDevice;
+  private chosenProcessor: string;
```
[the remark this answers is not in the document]
XHatan: sounds good, I ended up declaring the type using as suggested here https://github.com/aws/amazon-chime-sdk-js/pull/911#discussion_r533606928

demos/browser/app/meetingV2/meetingV2.ts
XHatan: fixed.

demos/browser/app/meetingV2/meetingV2.ts
XHatan: 👍

src/audiovideocontroller/AudioVideoController.ts
XHatan: fixed.

src/devicecontroller/DefaultDeviceController.ts
XHatan: fixed.

demos/browser/app/meetingV2/meetingV2.ts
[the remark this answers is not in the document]
richnew10: It turns into a string concatenation, which isn't free, but cost isn't the main motivation here.

When a browser console logs an object (especially an error!), you get much more useful output: colorized, with error stacks, with clickable and explorable objects. You can assign the error to a variable, navigate to code, etc.

Screen Shot 2020-12-02 at 9 29 24 AM


The SDK logging is pretty frustrating because it doesn't leverage this, but at least in the demo app we can do better.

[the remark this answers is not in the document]
XHatan: 👍  make sense. thanks

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -1037,25 +1035,300 @@ describe('DefaultDeviceController', () => {
...
+      });
+
+      new TimeoutScheduler(250).start(() => {
+        deviceController.chooseVideoInputDevice(null);
+      });
+
+      new TimeoutScheduler(300).start(() => {
+        device.stop();
+        done();
+      });
```
richnew10, line 1086: I don't really know what you're trying to exercise with this test. You're setting up two calls to race, but reducing the likelihood of them racing by putting a constructor call in the middle. You're then triggering two delayed operations 50ms apart, and stopping the device after, but there's no guarantee there of ordering at all — the first once might take 500ms to run, and will race with the second, and the stop might complete before the previous two calls. What are you testing here?

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -1037,25 +1035,300 @@ describe('DefaultDeviceController', () => {
...
+      const mockVideoTrack = new MediaStreamTrack('test', 'video');
+      mockVideoStream.addTrack(mockVideoTrack);
+      domMockBehavior.createElementCaptureStream = mockVideoStream;
+      deviceController.chooseVideoInputDevice(mockVideoStream).then(() => {});
+      const processor = new NoOpVideoFrameProcessor();
+      const device = new DefaultVideoTransformDevice(logger, mockVideoStream, [processor]);
+      deviceController.chooseVideoInputDevice(device).then(() => {});
+      new TimeoutScheduler(400).start(() => {
+        deviceController.chooseVideoInputDevice(null);
+      });
```
richnew10, line 1143: I don't get it. Are you trying to `await choose(); await choose(null); await device.stop()`?

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -1037,25 +1035,300 @@ describe('DefaultDeviceController', () => {
...
+      const device = new DefaultVideoTransformDevice(logger, 'test', [processor]);
+      deviceController.chooseVideoInputDevice(device).then(() => {});
+      new TimeoutScheduler(400).start(() => {
+        deviceController.chooseVideoInputDevice(null);
+      });
+
+      new TimeoutScheduler(450).start(() => {
+        device.stop();
+        expect(spy.called).to.equal(true);
+        done();
```
richnew10, line 1186: Generally speaking I want to ban the use of `TimeoutScheduler` in tests unless you are specifically assessing a time-based piece of code.

test/dommock/DOMMockBuilder.ts
```diff
@@ -1115,6 +1099,125 @@ export default class DOMMockBuilder {
...
+
+          new TimeoutScheduler(mockBehavior.videoElementSetWidthHeightAttiributeDelay).start(() => {
+            this.videoWidth = 1280;
+            this.videoHeight = 720;
+          });
+        }
+      }
+      fired = false;
+      load(): void {
+        if (!!this.refSrcObject) {
```
richnew10, line 1154: You never need `!!` inside a conditional. Remove them all.

test/videoframeprocessor/DefaultVideoFrameProcessorPipeline.test.ts
```diff
@@ -0,0 +1,432 @@
...
+      let stopped = 0;
+      class PipeObserver implements VideoFrameProcessorPipelineObserver {
+        processingDidStart(): void {
+          started = 1;
+        }
+
+        processingDidStop(): void {
+          stopped = 1;
+        }
+      }
```
richnew10, line 124: Just use sinon like you do elsewhere in the file!

test/videoframeprocessor/DefaultVideoFrameProcessorPipeline.test.ts
```diff
@@ -0,0 +1,432 @@
...
+      pipe.setInputMediaStream(mockVideoStream);
+      new TimeoutScheduler(300).start(() => {
+        expect(pipeObserver.processingDidStart.called).to.eq(true);
+        pipe.setInputMediaStream(null);
+      });
+
+      new TimeoutScheduler(350).start(() => {
+        expect(pipeObserver.processingDidStop.called).to.eq(true);
+        done();
+      });
```
richnew10, line 110: Are you doing this because you're trying to wait for the observers to have been called?

If so, scheduling like this is absolutely not the way to do it, unless you're trying to test that 100% of the time this callback will have been invoked within 350ms.

Either use promises or callbacks to have `done` called or have the test return when both observer methods are called. (See `Promise.all`.)

test/videoframeprocessor/DefaultVideoFrameProcessorPipeline.test.ts
```diff
@@ -0,0 +1,432 @@
...
+      pipe.addObserver(obs);
+      await pipe.setInputMediaStream(mockVideoStream);
+
+      await new Promise(resolve => setTimeout(resolve, 150));
+
+      expect(obs.processingDidStart.called).to.equal(true);
+      pipe.destroy();
+      await new Promise(resolve => setTimeout(resolve, 150));
+      expect(obs.processingDidStop.called).to.equal(true);
+      expect(procs[0].destroy.called).to.equal(true);
```
richnew10, line 429: Every time you do this kind of thing you're making our test suite take longer to run.

test/videoframeprocessor/DefaultVideoTransformDevice.test.ts
```diff
@@ -0,0 +1,467 @@
...
+
+      // a little bit of time to let processing run
+      await new Promise(resolve => setTimeout(resolve, 200));
+
+      expect(obs.processingDidStart.called).to.equal(true);
+      assert.exists(device.outputMediaStream);
+
+      await device.stop();
+      await new Promise(resolve => setTimeout(resolve, 50));
+      expect(obs.processingDidStop.called).to.equal(true);
```
richnew10, line 113: This really makes me raise my eyebrows.

test/videoframeprocessor/DefaultVideoTransformDevice.test.ts
```diff
@@ -0,0 +1,467 @@
...
+
+    it('can start processing on different media stream', async () => {
+      const obs = new MockObserver();
+      const processor = new NoOpVideoFrameProcessor();
+      const device = new DefaultVideoTransformDevice(logger, 'test', [processor]);
+      device.addObserver(obs);
+      domMockBehavior.createElementCaptureStream = mockVideoStream;
+      const outputStream = await device.applyProcessors(mockVideoStream);
+      // a little bit of time to let processing run
+      await new Promise(resolve => setTimeout(resolve, 200));
```
richnew10, line 125: Why?! Why 200ms?!

test/videoframeprocessor/DefaultVideoTransformDevice.test.ts
```diff
@@ -0,0 +1,467 @@
...
+
+    it('processingLatencyTooHigh', done => {
+      class Observer2 implements DefaultVideoTransformDeviceObserver {}
+      const obs = new MockObserver();
+
+      class WrongProcessor extends NoOpVideoFrameProcessor {
+        async process(buffers: VideoFrameBuffer[]): Promise<VideoFrameBuffer[]> {
+          await new Promise(resolve => setTimeout(resolve, (1000 / 15) * 3));
+          return buffers;
+        }
```
richnew10, line 451: Now *this* is a legit use of timing.

test/videoframeprocessor/DefaultVideoTransformDevice.test.ts
```diff
@@ -0,0 +1,467 @@
...
+
+      const procs = [new WrongProcessor()];
+      const device = new DefaultVideoTransformDevice(logger, 'test', procs);
+      device.addObserver(obs);
+      device.addObserver(new Observer2());
+      device.applyProcessors(mockVideoStream);
+      // simulate high latency, has to wait long enough for the callback to be fired.
+      new TimeoutScheduler(800).start(() => {
+        expect(obs.processingLatencyTooHigh.called).to.equal(true);
+        device.stop();
```
richnew10, line 462: … but note that you could just wait on the observer instead of setting a timer here. The whole test runs in a timeout.

richnew10 (changes requested):

src/browserbehavior/BrowserBehavior.ts
```diff
@@ -96,4 +96,9 @@ export default interface BrowserBehavior {
    * Returns whether browser supports setSinkId operation
    */
   supportsSetSinkId(): boolean;
+
+  /**
+   * Returns whether browser supports video processing
+   */
+  supportsVideoProcessing(): boolean;
```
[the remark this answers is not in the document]
XHatan: This is actually to disable iOS safari because https://bugs.webkit.org/show_bug.cgi?id=181663

test/videoframeprocessor/DefaultVideoTransformDevice.test.ts
XHatan: how do I await on observer ? 

something like 
```
      await new Promise<void>(async(resolve) => {
        class Observer implements VideoFrameProcessorPipelineObserver {
          processingDidStart(): void {
            resolve();
          }
        }
        const obs = new Observer();
        pipe.addObserver(obs);
        await pipe.setInputMediaStream(mockVideoStream);
      });

``` 
?

XHatan: how do I await on two callbacks ?

richnew10: For example (untested):

```typescript
class MyObserver extends Whatever {
  onFoo: sinon.stub(),
  onBar: sinon.stub(),
}

const observer = new MyObserver();

// Abstracting this into helpers is trivial.
const firstObserverCalled = new Promise((resolve, reject) => {
  observer.onFoo.callsFake(resolve);
});
const secondObserverCalled = new Promise((resolve, reject) => {
  observer.onBar.callsFake(resolve);
});

// If you want to impose a timeout, you can add Promise.race here.
const bothCalled = Promise.all([firstObserverCalled, secondObserverCalled]);

// Now either…
bothCalled.then(done);

// or, for an async test function…
await bothCalled;
```

src/browserbehavior/BrowserBehavior.ts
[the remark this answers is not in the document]
richnew10: Then consider whether you want "processing", "input transform", or something else. "Video processing" is pretty broad.

test/videoframeprocessor/DefaultVideoTransformDevice.test.ts
richnew10:
```typescript
function timeoutPromise(timeout: number): Promise<void> {
  return new Promise((_resolve, reject) => {
    setTimeout(timeout, reject);
  };
}

function called(stub: SinonStub, timeout: number = 1_000): Promise<void> {
  return Promise.race([
    new Promise((resolve, _reject) => { stub.callsFake(resolve); }),
    timeoutPromise(timeout),
  ]);
}

// Returns a promise that resolves if both observers are called within one second,
// and rejects if either takes longer than a second.
const bothCalled = Promise.all([called(observer.onFoo), called(observer.onBar)]);
```

XHatan: coolio, works like a charm, now I can remove all the timeout

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1374,9 +1396,38 @@ export class DemoMeetingApp implements
...
     await this.populateAudioInputList();
     await this.populateVideoInputList();
+    await this.populateVideoFilterInputList();
     await this.populateAudioOutputList();
   }
 
+
+  private async populateVideoFilterInputList(): Promise<void> {
+    const genericName = 'Filter';
+    let filters: string[] = ['None'];
```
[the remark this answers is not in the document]
XHatan: fixed.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -174,11 +178,85 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+    if (device === this.chosenVideoTransformDevice) {
+      this.logger.info('Reuse same VideoTransformDevice');
+      return;
+    }
+
+    const prevVideoTransformDevice = this.chosenVideoTransformDevice;
+    if (prevVideoTransformDevice) {
+      this.logger.info('Switched from previous VideoTransformDevice');
+    }
+    const wasUsingTransformDevice = !!this.chosenVideoTransformDevice;
```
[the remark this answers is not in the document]
XHatan: done.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -174,11 +178,85 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+      this.logger.info('video transform device needs new intrinsic device');
+      if (wasUsingTransformDevice) {
+        // detach input media stream - turn off the camera or leave it be if inner is media stream
+        prevVideoTransformDevice.onOutputStreamDisconnect();
+      }
+      this.chosenVideoTransformDevice = device;
+      // VideoTransformDevice owns input MediaStream
+      this.activeDevices['video'] = null;
+      await this.chooseInputIntrinsicDevice('video', inner, false, true);
+    } else {
```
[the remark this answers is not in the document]
XHatan: 👍

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -947,7 +1058,13 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     this.activeDevices[kind] = newDevice;
 
     if (kind === 'video') {
-      this.restartLocalVideoAfterSelection(oldStream, fromAcquire);
+      // attempts to mirror `this.useWebAudio`. The difference is that audio destination stream stays the same
+      // but video sending needs to switch streams.
+      if (this.chosenVideoInputIsTransformDevice()) {
+        this.logger.info('apply processors to transform');
+        await this.chosenVideoTransformDevice.applyProcessors(this.activeDevices['video'].stream);
```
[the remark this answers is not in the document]
XHatan: changed to "transformStream"

src/videoframeprocessor/NoOpVideoFrameProcessor.ts
```diff
@@ -0,0 +1,22 @@
...
+import VideoFrameProcessor from './VideoFrameProcessor';
+
+/**
+ * [[NoOpVideoFrameProcessor]] implements [[VideoFrameProcessor]].
+ * It's a no-op processor and input is passed to output directly.
+ */
+export default class NoOpVideoFrameProcessor implements VideoFrameProcessor {
+  name(): string {
+    return 'NoOp';
+  }
```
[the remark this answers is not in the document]
XHatan: done

test/devicecontroller/DefaultDeviceController.test.ts
XHatan: did a pass to remove all the TimeoutScheduler, this test is just try to choose same device twice.

test/devicecontroller/DefaultDeviceController.test.ts
XHatan: removed the timeout scheduler

test/dommock/DOMMockBuilder.ts
XHatan: fixed.

test/videoframeprocessor/DefaultVideoTransformDevice.test.ts
XHatan: removed

test/videoframeprocessor/DefaultVideoTransformDevice.test.ts
XHatan: removed eyebrow-raising timeout scheduler.

src/videoframeprocessor/DefaultVideoFrameProcessorPipeline.ts
```diff
@@ -0,0 +1,261 @@
...
+    // null input media stream stops the pipeline.
+    this.videoInput.removeEventListener('loadedmetadata', this.process);
+    this.videoInput.srcObject = null;
+    // clean input stream and buffers
+    this.destroyInputMediaStreamAndBuffers();
+    if (this.outputMediaStream) {
+      for (const track of this.outputMediaStream.getVideoTracks()) {
+        track.stop();
+      }
+    }
```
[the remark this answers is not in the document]
XHatan: em, don't think so

test/videoframeprocessor/NoOpVideoFrameProcessor.test.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  describe('process', () => {
+    it('passes through the buffers', async () => {
+      const buffers: VideoFrameBuffer[] = [];
+      let output = await proc.process(buffers);
+      expect(output).to.deep.equal(buffers);
+      buffers.push(new CanvasVideoFrameBuffer(null));
+      output = await proc.process(buffers);
+      expect(output).to.deep.equal(buffers);
```
richnew10, line 30: I *think* you should be using `to.be` or `to.eql` here — `NoOpVideoFrameProcessor` passes the exact input to the output, right?

test/videoframeprocessor/DefaultVideoTransformDevice.test.ts
```diff
@@ -0,0 +1,444 @@
...
+    //   const processor = new NoOpVideoFrameProcessor();
+    //   const device = new DefaultVideoTransformDevice(logger, 'test', [processor]);
+    //   device.transformStream(mockVideoStream);
+
+    //   new TimeoutScheduler(200).start(() => {
+    //     device.stop();
+    //   });
+    //   await device.stop();
+    // });
+
```
richnew10, line 92: ?

test/videoframeprocessor/DefaultVideoTransformDevice.test.ts
```diff
@@ -0,0 +1,444 @@
...
+      device.removeObserver(obs);
+    });
+
+    it('processingLatencyTooHigh', async () => {
+      class Observer2 implements DefaultVideoTransformDeviceObserver {}
+      const obs = new MockObserver();
+
+      const latencyCallback = new Promise(resolve => {
+        obs.processingLatencyTooHigh.callsFake(resolve);
+      });
```
richnew10, line 426: See my comment earlier for a helper function to remove a bunch of duplicated code around each of these.

test/videoframeprocessor/CanvasVideoFrameBuffer.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  let domMockBehavior: DOMMockBehavior;
+  let domMockBuilder: DOMMockBuilder;
+
+  beforeEach(() => {
+    domMockBehavior = new DOMMockBehavior();
+    domMockBuilder = new DOMMockBuilder(domMockBehavior);
+  });
+
+  afterEach(() => {
+    if (domMockBuilder) {
```
richnew10, line 22: I'm surprised you need this condition.

richnew10 (commented):
Please go through and examine and resolve all previous comments.

src/videoframeprocessor/DefaultVideoTransformDevice.ts
XHatan: now pipeline constructor takes processors as arguments.

test/videoframeprocessor/CanvasVideoFrameBuffer.test.ts
XHatan: removed

src/browserbehavior/BrowserBehavior.ts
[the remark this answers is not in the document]
XHatan: How about "supportsCanvasCapturedStreamPlayback" to reflect the webkit bug ?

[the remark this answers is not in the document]
richnew10: Sounds descriptive to me!

test/videoframeprocessor/DefaultVideoTransformDevice.test.ts
XHatan: removed

src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -62,6 +62,10 @@ export default class DefaultBrowserBehavior implements BrowserBehavior {
     return this.isFirefox();
   }
 
+  supportsVideoProcessing(): boolean {
+    return !this.isIOSSafari();
+  }
+
```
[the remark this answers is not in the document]
XHatan: changed to "supportsCanvasCapturedStreamPlayback"

test/videoframeprocessor/NoOpVideoFrameProcessor.test.ts
XHatan: yes, changed

demos/browser/app/meetingV2/meetingV2.ts
XHatan: we don't have a this.logger in DemoMeetingApp.

we have `MultiLogger`, `MeetingSessionPOSTLogger` created on demand, do you mind if I do another PR after this to clean that up a bit ?

src/browserbehavior/BrowserBehavior.ts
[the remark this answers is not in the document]
XHatan: k, changed

PR #913, opened 2020-12-01
Fix uncaught promise exception for bindAudioOutput API

anuranduttaroy:
**Issue #:** https://github.com/aws/amazon-chime-sdk-js/issues/907

**Description of changes:** Checking for `browserBehavior.supportsSetSinkId()` before calling `AsyncScheduler() -> this.bindAudioOutput();` in [`DefaultDeviceController`](https://github.com/aws/amazon-chime-sdk-js/blob/59bf60c30c8450451d9d1e36ec30a5b67e787607/src/devicecontroller/DefaultDeviceController.ts#L402)

**Testing**

1. Have you successfully run `npm run build:release` locally?
Yes.

2. How did you test these changes?
Tested by running the `serverless demo` & the `meeting readiness checker` demo in Firefox & Safari browser to check if there is any issue while
    - Starting/Joining a meeting.
    - Choosing audio input device, mute, unmute operation.
    - Enable/Disable speaker from the meeting canvas.
    - Set the `media.setsinkid.enabled` preference to `true` in Firefox. This enables the use of `SetSinkId` API as mentioned here https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/setSinkId and verified this error does not show up in the browser console.

3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it?
Yes. Serverless demo.

4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved?
NA

5. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved?
NA

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #930, opened 2020-12-05
Don't throw an error when binding audio if the builder is not trying to specify a particular output device.

richnew10:
This commit makes `DefaultAudioMixController` more nuanced: by looking
at the element and the chosen device, we can determine whether the
caller is attempting to specify an output device when the browser does
not support doing so, and throw in only that case.

This commit also amends the demo app to use `supportsSetSinkId` as
documented, setting a good example for developers, and also fixes the
docs that were not updated to match that API change in 2.0.

Additionally, this commit moves `stripInternal` into the correct
`tsconfig` file to cause it to apply to generated docs, preventing the
generation of docs for internal interfaces.

**Issue #:**

Ongoing from #913.

**Testing**

Yes.

Loading the demo app in both Safari and Chrome, and exercising device switching.

Browser.

No. This changes the circumstances in which an error is thrown, but only in the sense of fixing a bug.

No.


src/audiomixcontroller/DefaultAudioMixController.ts
```diff
@@ -86,13 +103,17 @@ export default class DefaultAudioMixController implements AudioMixController {
...
-    } catch (error) {
-      this.logger?.error(`Failed to set sinkId for audio element: ${error}`);
-      throw error;
+
+    if (shouldSetSinkId) {
+      try {
+        // @ts-ignore
+        await existingAudioElement.setSinkId(newSinkId);
+      } catch (error) {
+        this.logger?.error(`Failed to set sinkId for audio element: ${error}`);
```
anuranduttaroy, line 112: Nitpick: Do we want to change this to like we did here in [bindAudioStream](https://github.com/aws/amazon-chime-sdk-js/pull/930/files?file-filters%5B%5D=.json&file-filters%5B%5D=.ts#diff-646676a4d8e0ab238513a20bd428ed47017c8ef824e7a99f69731a5dbf315983L40)

```
/* istanbul ignore else */
if (this.logger) {
  this.logger.error(`Failed to set sinkId for audio element: ${error}`);
}
```

src/audiomixcontroller/DefaultAudioMixController.ts
```diff
@@ -47,7 +55,7 @@ export default class DefaultAudioMixController implements AudioMixController {
      * Throw error if browser doesn't even support setSinkId
      * Read more: https://caniuse.com/?search=setSinkId
      */
-    if (!this.browserBehavior.supportsSetSinkId()) {
+    if (device && !this.browserBehavior.supportsSetSinkId()) {
```
anuranduttaroy, line 58: Question: If we do not do anything if it's a `null` device, then why do we even allow passing `null` device here in this function?

src/audiomixcontroller/DefaultAudioMixController.ts
rnewman: The only reason I split it out is because the no-logger case wasn't covered by tests, and it was not worth creating another test case just to exercise the non-presence of a log item if the logger was not provided, which is kinda meaningless.

src/audiomixcontroller/DefaultAudioMixController.ts
rnewman: In this function I think you might be right! These multiple entry points and shared logic are definitely DRY with lots of shared state, so it's hard to tell which steps can be elided. For example, `bindAudioMix` bails out if the old sink ID and new sink ID are the same, and I'm not sure it should!

anuranduttaroy (approved):

anuranduttaroy:
Related: https://github.com/aws/amazon-chime-sdk-js/issues/932

PR #940, opened 2020-12-09
Add fatal error catching to the demo.

richnew10:
This change introduces the Red Screen of Death, which will be shown if
an unexpected error is thrown during execution, even if that error would
otherwise have been caught by code in the demo app.

Screen Shot 2020-12-09 at Dec 9 11 41 20

This is largely manual: each time you add a catch block that *should
never be hit*, add a call to `fatal(err)`. Catch blocks that catch
routine failures (e.g., the browser not supporting Voice Focus) should
not add a call to `fatal`.

However, this change also adds a top-level unhandled promise rejection
listener, and a top-level error listener. These will cause fatal errors
to be detected in e.g., async tasks.

You can test this with something like:

```
setTimeout(() => Promise.reject(new Error('oh no')), 2000)
```

This change should be sufficient to alert developers and integration
tests to new programming errors.

This commit adds the following URL query params to the demo:

* `testfatal=1`: if provided, this will show the fatal error page.
* `fatal=0`: if provided, fatal errors will be ignored, even for
  localhost.
* `fatal=1`: if provided, fatal errors will cause a Red Screen of Death,
  even when not testing locally.

The default is for the Red Screen of Death to be shown if loading from
`127.0.0.1:8080` only.

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.html
+++ b/demos/browser/app/meetingV2/meetingV2.html
@@ -8,8 +8,16 @@
 </head>
 <body>
 
-<!-- Initial meeting authentication screen with meeting and name inputs -->
+<div id="flow-fatal" class="flow">
+  <div style="vertical-align: middle; text-align: left; width: 40%; margin: 10em 30% 10em 30%;">
+  <h1>Fatal error</h1>
+  <p>An error was thrown that should not occur in ordinary use. This should be considered a fatal error for testing and canary purposes.</p>
+  <p>This error will ordinarily be muffled, but you should consider removing this code if you are building an app based on this demo.</p>
+  <code id="stack"></code>
+  </div>
+</div>
 
+<!-- Initial meeting authentication screen with meeting and name inputs -->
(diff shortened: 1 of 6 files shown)


demos/browser/app/meetingV2/meetingV2.html
```diff
@@ -8,8 +8,16 @@
 </head>
 <body>
 
-<!-- Initial meeting authentication screen with meeting and name inputs -->
+<div id="flow-fatal" class="flow">
+  <div style="vertical-align: middle; text-align: left; width: 40%; margin: 10em 30% 10em 30%;">
+  <h1>Fatal error</h1>
+  <p>An error was thrown that should not occur in ordinary use. This should be considered a fatal error for testing and canary purposes.</p>
```
anuranduttaroy, line 14: Do we want to mention the word `canary` here?

richnew10: I see no reason why not.

richnew10:
Closing in favor of #941 because our integration tests still don't run against forks.

PR #941, opened 2020-12-09
Add fatal error catching to the demo.

richnew10:
This change introduces the Red Screen of Death, which will be shown if
an unexpected error is thrown during execution, even if that error would
otherwise have been caught by code in the demo app.

Screen Shot 2020-12-09 at Dec 9 11 41 20

This is largely manual: each time you add a catch block that *should
never be hit*, add a call to `fatal(err)`. Catch blocks that catch
routine failures (e.g., the browser not supporting Voice Focus) should
not add a call to `fatal`.

However, this change also adds a top-level unhandled promise rejection
listener, and a top-level error listener. These will cause fatal errors
to be detected in e.g., async tasks.

You can test this with something like:

```
setTimeout(() => Promise.reject(new Error('oh no')), 2000)
```

This change should be sufficient to alert developers and integration
tests to new programming errors.

This commit adds the following URL query params to the demo:

* `testfatal=1`: if provided, this will show the fatal error page.
* `fatal=0`: if provided, fatal errors will be ignored, even for
  localhost.
* `fatal=1`: if provided, fatal errors will cause a Red Screen of Death,
  even when not testing locally.

The default is for the Red Screen of Death to be shown if loading from
`127.0.0.1:8080` only.

(#940 pushed to `origin`)

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -46,7 +46,17 @@ import {
   VoiceFocusTransformDevice,
   isAudioTransformDevice,
 } from '../../../../src/index';
+
 import WebRTCStatsCollector from './webrtcstatscollector/WebRTCStatsCollector';
+
+const SHOULD_DIE_ON_FATALS = (() => {
+  const isLocal = document.location.host === '127.0.0.1:8080';
+  const fatalYes = document.location.search.includes('fatal=1');
+  const fatalNo = document.location.search.includes('fatal=0');
+  return fatalYes || (isLocal && !fatalNo);
+})();
+
+let fatal: (e: Error) => void;
 
 class DemoTileOrganizer {
   // this is index instead of length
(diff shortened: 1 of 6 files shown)


richnew10:
N.B., canaries should be updated to specify `fatal=1`, and the same should be done when QA tests a release.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -248,6 +260,18 @@ export class DemoMeetingApp implements
...
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     (global as any).app = this;
+
+    fatal = this.fatal.bind(this);
+
+    // Listen for unhandled errors, too.
+    window.addEventListener('error', (event) => { fatal(event.error); });
+    window.onunhandledrejection = (event: PromiseRejectionEvent) => { fatal(event.reason); };
+
+    if (document.location.search.includes('testfatal=1')) {
```
ltrung, line 270: Does that mean it will always trigger if testfatal=1?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1628,18 +1680,20 @@ export class DemoMeetingApp implements
   }
 
   async openAudioOutputFromSelection(): Promise<void> {
-    if (this.defaultBrowserBehaviour.supportsSetSinkId()) {
+    if (true || this.defaultBrowserBehaviour.supportsSetSinkId()) {
```
ltrung, line 1683: Is this intentional?

richnew10: Nope! Good catch. This was me provoking the error.

demos/browser/app/meetingV2/meetingV2.ts
richnew10: Yes. If you load with `testfatal=1` in the URL, a fatal is triggered automatically. This allows us to at least test that the fatal UI loads when an unhandled !

ltrung (approved):

Issue #904, opened 2020-11-26
Need a Chime log to indicate that signalling has negotiated WebRTC over TCP


tthorpe-git:
Although most of our video meetings using Chime see video media carried over UDP, we have an issue where our network set-up is causing signalling negotiation to fallback to carry video media for some meetings over TCP, which is not supportable. It took us a while to figure out that this is happening, partly because there was no logged indication from Chime that this fallback to TCP is happening. 

I have captured the Chime logs for 2 video sessions, each included as an excel spreadsheet in the uploaded compressed tar file 'chime_logs.tar.gz'.

1. video media carried over UDP: graylog-udp-mary-faulkner.bournemouth-sso-ui-only-longer-period-2020-11-19T09_25_00.000Z-2020-11-19T09_55_00.000Z

2. video media carried over TCP: graylog-tcp-andy-grice-clay.wolverhampton-sso-ui-only-longer-period-2020-11-19T14_10_00.000Z-2020-11-19T14_36_00.000Z

In each file, there are a mixture of Chime logs & our own 'wrapper' service logs; the former are identifiable by the '[BROWSER-CONDUCT]' tag.
In each case, the user has actually started multiple video meetings - look for "Staff Entered Landing Page Accessed" log for the start of each new cycle.
In each case, 2 people attended the meeting, who we call 'conductor' (in a DWP office) & 'attendee' (UK citizen on the public internet), but I have only uploaded those logs for the conductor side of each meeting, as this is the side that is sometimes negotiating to carry the video media traffic over TCP, due to the route from that particular DWP office to the Chime UDP servers, through the DWP network, being blocked for some reason.

[chime_logs.tar.gz](https://github.com/aws/amazon-chime-sdk-js/files/5601245/chime_logs.tar.gz)

Meeting and attendee IDs: 

Video media carried over UDP:
meeting id: 2d10ea9f-3a6a-4271-a0f2-70265db8e1f0
attendee id: f9c35b5a-6f06-d70c-40f3-61aa1ebfaed1

Video media carried over TCP:
meeting id: a030f62c-4ed3-4386-94e2-9aae58461822
attendee id: 64bad67e-9dd6-50e4-d332-be4fc2e53556

**Timeline**

Video media carried over UDP:
* 2020-11-19 09:34:15 DWP staff member enters DWP's own meeting code
* 2020-11-19 09:34:21 Signalling begins
* 2020-11-19 09:34:27 Video meeting begins

Video media carried over TCP:
[...]

PR #948, opened 2020-12-14
Add logging for UDP or TCP/TLS protocol being used for audio/video

anuranduttaroy:
**Issue #:** https://github.com/aws/amazon-chime-sdk-js/issues/904

**Description of changes:** Currently in the SDK code, we do not have a way to identify whether the meeting is using `UDP` or has it fallen back to `TCP/TLS`. This PR adds a quick way to check which protocol is being used currently.

**Testing**

Yes.

Tested it by running the demo app. Also added the same logic in the `TCP` & `UDP` connectivity check in the readiness-checker and verified it works there as well.

Yes. Serverless demo app.

NA. Just demo changes.

NA

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -645,6 +645,8 @@ export class DemoMeetingApp implements
     buttonVideoStats.addEventListener('click', () => {
       if (this.isButtonOn('button-video-stats')) {
         document.querySelectorAll('.stats-info').forEach(e => e.remove());
+      } else {
+        this.getRelayProtocol();
       }
       this.toggleButton('button-video-stats');
     });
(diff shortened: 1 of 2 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -903,7 +905,23 @@ export class DemoMeetingApp implements
...
     }
+  }
 
+  async getRelayProtocol() {
+    const rawStats = await this.audioVideo.getRTCPeerConnectionStats();
+    if (rawStats) {
+      rawStats.forEach(report => {
+        if (report.type === 'local-candidate') {
+          this.log(`Local WebRTC Ice Candidate stats: ${JSON.stringify(report)}`);
+          let relayProtcol = report.relayProtocol;
```
simmkyu, line 916: - Typo: `relayProtcol` => `relayProtocol`
- Use `const`
- `relayProtocol` is optional, and calling `toUpperCase()` can throw an error. Check if the type is string.
  https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidateStats

```
if (typeof relayProtcol === 'string') {
```

anuranduttaroy: Fixed

simmkyu (dismissed):

simmkyu (approved):

PR #955, opened 2020-12-17
Add release script

ltrung:
**Description of changes:**
Add a new release script and GitHub action to publish to NPM for each release.
**Testing**

2. How did you test these changes? Manually running the script


script/release
```diff
@@ -99,6 +96,6 @@ x = STDIN.gets
 exit(1) unless x.strip == 'yes'
 puts
 
-verbose("git push origin")
-verbose("git push origin -f #{tag}")
-verbose("git push origin -f #{semvertag}")
+verbose("git push origin HEAD:release --follow-tags -f")
+
+puts "Remember to tag the release with #{semvertag}"
```
michhyun1, line 101: Should this script create the PR with the tag?

script/release
```diff
@@ -86,11 +86,8 @@ verbose("git tag -d #{tag} || true")
 verbose("git tag #{tag}")
 verbose("git tag #{semvertag} -m \"Release #{semvertag}.\"")
```
michhyun1, line 88: why is there two tags?

script/release
ltrung: I can remove it. Just a reminder to use which tag to tag the release when drafting new release.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,25 @@
...
+          node-version: 12
+          registry-url: https://registry.npmjs.org/
+      - name: NPM Install
+        run: npm install
+      - name: NPM run build
+        run: npm run build
+      - name: Publish to NPM
+        run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
```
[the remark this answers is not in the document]
ltrung: I will shadow you on Monday and will add them then :)

script/release
ltrung: This is from the old publish script. We already tag twice with the old one. Richard and Andrew said we need to make sure nobody uses them before removing one of them so i leave it there.

michhyun1 (approved):

PR #956, opened 2020-12-17
Add PublishConfig

ltrung:
**Description of changes:**
Add PublishConfig so that new NPM publish to 1.x is not tagged with latest.
**Testing**

2. How did you test these changes? Npm run build

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -19,64 +19,28 @@ jobs:
   build:
     name: Build and Run Unit Tests
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node-version: [12.x]
     steps:
       - name: Checkout Package
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Using Node.js ${{ matrix.node-version }}
+      - name: Using Node.js 12.x
         uses: actions/setup-node@v1
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version: 12.x
       - name: Clean Install
         run: npm ci
       - name: Build
         run: npm run build:release
 
-  check_integ_test_required:
-    name: Check if Integ Tests are Required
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -2,6 +2,9 @@
   "name": "amazon-chime-sdk-js",
   "version": "1.22.0",
   "description": "Amazon Chime SDK for JavaScript",
+  "publishConfig": {
+    "tag": "release-1.x"
```
devalevenkatesh, line 6: So when installing through NPM if we do not specify anything then the latest tagged version from the release-1.x branch will be installed right?

ltrung: So this tag is for amazon-chime-sdk-js npm package. If we do not specify the tag, when we release for 1.x, it will get latest tag so anyone who do npm update amazon-chime-sdk-js will pull in this version instead of the latest 2.x version.

devalevenkatesh: Yup yup 👍

.github/workflows/continuous-integration.yml
```diff
@@ -19,64 +19,28 @@ jobs:
...
-
-    strategy:
-      matrix:
-        node-version: [12.x]
-
     steps:
       - name: Create a Job ID
         id: create-job-id
-        uses: filipstefansson/uuid-action@v1
+        uses: filipstefansson/uuid-action@ce29ebbb0981ac2448c2e406e848bfaa30ddf04c
```
devalevenkatesh, line 43: Any reason for the specific commit? It would be good if we document it somehow.

ltrung: That is how it is in the master branch https://github.com/aws/amazon-chime-sdk-js/blob/master/.github/workflows/continuous-integration.yml#L48 so I just copied it over. @michhyun1 can provide more details :D

devalevenkatesh (approved):

PR #966, opened 2020-12-22
Add Video Test App to integ test suite

michhyun1:
**Issue #:**

**Description of changes:**
Adds a integ test for bind and unbind video element apis.

**Testing**
2. How did you test these changes? Ran the video test locally

### DIFF
--- a/integration/js/script/run-test
+++ b/integration/js/script/run-test
@@ -115,6 +115,7 @@ case $1 in
       r video_test_processor.config.json
       #  TODO: Re-enable simulcast test after it has been fixed
       # r video_test_simulcast.config.json
+      r test_ap_video_test.config.json
       r app_quit_video_test.config.json
       r meeting_leave_video_test.config.json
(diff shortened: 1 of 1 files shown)


integration/js/script/run-test
```diff
@@ -115,6 +115,7 @@ case $1 in
       r video_test_processor.config.json
       #  TODO: Re-enable simulcast test after it has been fixed
       # r video_test_simulcast.config.json
+      r test_ap_video_test.config.json
```
ltrung, line 118: This wont work. Remember that all these video tests are using the meeting v2 demo while this one uses your new demo. You either need to break this out or try to relaunch the demo (not recommended).
Also there is an error in the GitHub test run (we need to identify why it did not fail the check).
```
java.io.FileNotFoundException: configs/test_ap_video_test.config.json (No such file or directory)
```

ltrung (changes requested):

Issue #974, opened 2020-12-24
Make realtimeUnsubscribeFromVolumeIndicator function able to remove specific callback


asc0910:
The realtimeSubscribeToVolumeIndicator function add a callback to a attendee
And realtimeUnsubscribeFromVolumeIndicator remove all callback.
How about to change realtimeUnsubscribeFromVolumeIndicator function that can be able to remove specific callback?

This is the current realtimeUnsubscribeFromVolumeIndicator function

```js
realtimeUnsubscribeFromVolumeIndicator(attendeeId: string): void {
  try {
    delete this.state.volumeIndicatorCallbacks[attendeeId];
  } catch (e) {
    this.onError(e);
  }
}
```

We can change like this
```js
  realtimeUnsubscribeFromVolumeIndicator(attendeeId: string, callback?: (
    attendeeId: string,
    volume: number | null,
    muted: boolean | null,
    signalStrength: number | null,
    externalUserId?: string
  ) => void): void {
    if (!callback) {
      try {
        delete this.state.volumeIndicatorCallbacks[attendeeId];
      } catch (e) {
        this.onError(e);
      }
    } else {
      if(this.state.volumeIndicatorCallbacks[attendeeId]) {
        this.state.volumeIndicatorCallbacks[attendeeId] = this.state.volumeIndicatorCallbacks[attendeeId].filter(f => f !== callback);
      }
    }
  }
```

p-foucht:
Closing in favor #942

PR #975, opened 2020-12-24
Add optional callback parameter to realtimeUnsubscribeFromVolumeIndicator

asc0910:
**Issue #:**
https://github.com/aws/amazon-chime-sdk-js/issues/974
**Description of changes:**
Add callback parameter to realtimeUnsubscribeFromVolumeIndicator 
**Testing**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  `VideoFrameProcessor`, `VideoFrameProcessorPipeline` and `VideoFrameBuffer` interfaces
  are added to support `DefaultVideoTransformDevice` and allow processing steps to be applied to device.
  The method `chooseVideoInputDevice` in `DefaultDeviceController` can handle `VideoTransformDevice` now.
+- add callback parameter to `realtimeUnsubscribeFromVolumeIndicator`
 
 ### Changed
 
--- a/src/audiovideofacade/DefaultAudioVideoFacade.ts
+++ b/src/audiovideofacade/DefaultAudioVideoFacade.ts
@@ -251,8 +251,17 @@ export default class DefaultAudioVideoFacade implements AudioVideoFacade {
     this.trace('realtimeSubscribeToVolumeIndicator', attendeeId);
   }
 
-  realtimeUnsubscribeFromVolumeIndicator(attendeeId: string): void {
-    this.realtimeController.realtimeUnsubscribeFromVolumeIndicator(attendeeId);
+  realtimeUnsubscribeFromVolumeIndicator(
+    attendeeId: string,
+    callback?: (
+      attendeeId: string,
+      volume: number | null,
+      muted: boolean | null,
+      signalStrength: number | null,
+      externalUserId?: string
+    ) => void
+  ): void {
+    this.realtimeController.realtimeUnsubscribeFromVolumeIndicator(attendeeId, callback);
     this.trace('realtimeUnsubscribeFromVolumeIndicator', attendeeId);
   }
 
--- a/src/realtimecontroller/DefaultRealtimeController.ts
+++ b/src/realtimecontroller/DefaultRealtimeController.ts
@@ -267,12 +267,26 @@ export default class DefaultRealtimeController implements RealtimeController {
       this.onError(e);
     }
   }
-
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  `VideoFrameProcessor`, `VideoFrameProcessorPipeline` and `VideoFrameBuffer` interfaces
  are added to support `DefaultVideoTransformDevice` and allow processing steps to be applied to device.
  The method `chooseVideoInputDevice` in `DefaultDeviceController` can handle `VideoTransformDevice` now.
+- add callback parameter to `realtimeUnsubscribeFromVolumeIndicator`
```
richnew10, line 17: This will need to move into the 'Unreleased' section of the changelog.

src/audiovideofacade/DefaultAudioVideoFacade.ts
```diff
@@ -251,8 +251,17 @@ export default class DefaultAudioVideoFacade implements AudioVideoFacade {
...
-    this.realtimeController.realtimeUnsubscribeFromVolumeIndicator(attendeeId);
+  realtimeUnsubscribeFromVolumeIndicator(
+    attendeeId: string,
+    callback?: (
+      attendeeId: string,
+      volume: number | null,
+      muted: boolean | null,
+      signalStrength: number | null,
+      externalUserId?: string
+    ) => void
```
richnew10, line 262: It would be nice if this type signature were extracted and shared between the sub/unsub functions, here and in the realtime controller. It's now repeated in four places.

src/realtimecontroller/DefaultRealtimeController.ts
```diff
@@ -267,12 +267,26 @@ export default class DefaultRealtimeController implements RealtimeController {
...
+    ) => void
+  ): void {
+    if (!callback) {
+      try {
+        delete this.state.volumeIndicatorCallbacks[attendeeId];
+      } catch (e) {
+        this.onError(e);
+      }
+    } else {
+      if(this.state.volumeIndicatorCallbacks[attendeeId]) {
```
richnew10, line 287: Nit: space after `if`, and prefer early return rather than `else`. Did you run `build:release` or `lint`?

src/realtimecontroller/DefaultRealtimeController.ts
```diff
@@ -267,12 +267,26 @@ export default class DefaultRealtimeController implements RealtimeController {
...
+        this.state.volumeIndicatorCallbacks[attendeeId] = this.state.volumeIndicatorCallbacks[attendeeId].filter(f => f !== callback);
```
richnew10, line 288: You should use `splice` here. See how it's done on line 366.

richnew10, line 288: This branch needs `try..catch`, too. It is vanishingly unlikely that anything would go wrong in this expression, but this class is written to be as bulletproof as possible — see the block comment at the top. Best to copy the patterns in the other removal methods.

richnew10 (changes requested):
Thanks for the pull request! I left some specific comments inline.

Please note that we have a requirement for 100% test coverage on this project, and by inspection I don't think that's preserved by your change. You'll need to extend the existing tests for the method you're changing to cover removing with and without a specific callback, and make sure that your test runs end with 100% coverage.

You will need to run `npm run build:release`, and commit the changes it makes, both to your code during linting and also the generated HTML docs. When that goes green, please fold all of your commits together and give them a descriptive commit message, like:

```
Add an optional callback argument to realtimeUnsubscribeFromVolumeIndicator.

This change allows a caller to unsubscribe only one callback from the array
of subscribed callbacks. Calling `realtimeUnsubscribeFromVolumeIndicator`
with no callback argument removes all subscribers for the supplied attendee ID.
```

Thanks!

src/realtimecontroller/DefaultRealtimeController.ts
richnew10: (Yes, strictly speaking `filter` will catch the cases where you subscribe multiple times with the same callback and want it removed in one call. `splice` is the approach we've taken elsewhere for symmetry, so N subscribes need N removals.)

bradmanAWS:
This fix has been implemented in https://github.com/aws/amazon-chime-sdk-js/pull/979. Closing.

PR #978, opened 2020-12-29
Import current version of Amazon Voice Focus code: Safari Technology Preview support and multiple quanta execution

richnew10:
This adds the ability to run inline work across multiple Web Audio
quanta, extending Amazon Voice Focus to lower-end devices, and
additionally supports browsers that do not support streaming
WebAssembly compilation.

This code change also supports Safari Technology Preview. Note that
Safari clients need to use the default Amazon Voice Focus revision, or
be sure to use a new revision that supports Safari.

This commit adjusts the `VoiceFocusDeviceTransformer` test to
accommodate new User-Agent checks for Safari, which does not support
sending WebAssembly modules to worklets via `postMessage`. It also
resolves some Safari-specific issues around error handling in the demo
app.

Yes.

There are several aspects to this change.

Behavior with current CDN contents is expected to be unchanged for Firefox, Chrome, and Safari, and **to fail** (throwing a fatal in test mode, or silently failing for builders) with Safari Technology Preview.

With unreleased CDN contents, all browsers except for release Safari should work, and inline execution should occur in three chunks rather than one.

These combinations were manually tested.

Additionally, chunked execution is covered by unit tests in the Amazon Voice Focus repo, where the bulk of this change was reviewed. The unit tests for Amazon Voice Focus in the SDK were amended to match, and integration tests will run in Firefox and Chrome as part of this PR to verify that existing CDN contents continue to work with this code.

Final testing of this change in the SDK plus the updated CDN contents will have to wait until those changes hit a prerelease CDN.

meetingV2.

No.

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Add support for Amazon Voice Focus support in Safari Technology Preview for macOS.
+  Builders using an explicit revision or asset group must make sure to use a
+  revision no earlier than this; an error will be thrown in Safari if older
+  revisions are used.
 
 ### Changed
 
-- Corrected `null` type on `DefaultVideoFrameProcessorPipeline` and `DefaultVideoTransformDevice`
+- Corrected `null` type on `DefaultVideoFrameProcessorPipeline` and `DefaultVideoTransformDevice`.
+- Amazon Voice Focus now makes better use of available CPU resources,
+  extending support to lower-end devices and improving quality on higher-end
+  devices.
 
 ### Removed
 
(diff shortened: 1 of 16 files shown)


rnewman:
I will rebase this when #979 lands.

CHANGELOG.md
```diff
@@ -8,10 +8,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Add support for Amazon Voice Focus support in Safari Technology Preview for macOS.
```
XHatan, line 11: maybe add which version of the technology review

XHatan (approved):

CHANGELOG.md
richnew10: That's in the support table, at least.

PR #982, opened 2020-12-30
Remove possibility of video processor output media streams being null.

richnew10:
@XHatan, here's a sketch of the solution I propose. Tests pass, but coverage needs a little enhancement.


rnewman:
Finished in #983.

PR #983, opened 2020-12-30
Corrected `outputMediaStream` getter return type on `DefaultVideoTransformDevice`

XHatan:
**Issue #:**

**Description of changes:**

built upon #982 to correct `outputMediaStream` getter return type on `DefaultVideoTransformDevice` and fix small warts

**Testing**

2. How did you test these changes? demo and unit test
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? yes


test/videoframeprocessor/DefaultVideoTransformDevice.test.ts
```diff
@@ -73,7 +73,7 @@ describe('DefaultVideoTransformDevice', () => {
     it('returns null if device is not started', async () => {
       const processor = new NoOpVideoFrameProcessor();
       const device = new DefaultVideoTransformDevice(logger, 'test', [processor]);
-      expect(await device.outputMediaStream).to.be.null;
+      expect(await device.outputMediaStream).to.be.undefined;
     });
```
richnew10, line 77: This isn't right. The description of the test is no longer accurate, but more importantly it means `NoOpVideoFrameProcessor` is not returning a stream when the type signature demands that it should.

Further to our earlier conversation, this is exactly the kind of thing that strict typing *would* catch, because` outputMediaStream` is not allowed to be `undefined`.

richnew10 (changes requested):

richnew10 (approved):

PR #996, opened 2021-01-07
Mute video elements correctly.

richnew10:
There are two ways the SDK allows content to be shared:

* From the browser via `getDisplayMedia`, for a tab or the whole screen.
* From a `video` element (`content-share-video` in the demo), either
  from the web or from a local file.

Unfortunately, if audio is played in the content share stream, it will
be echoed back in the local conference via the main video element.

This is because the code to mute video elements on bind was setting the
`muted` _attribute_, which reflects to `defaultMuted` on the instance
itself.

This commit does the following:

* Correctly uses `element.muted = true` to set mute state.
* Sets `playinline` on the hidden video element in the demo app, for
  completeness.

Additionally, this commit refactors the demo app to correctly set the
sink ID on the `content-share-video` element when choosing an output
device. This forces non-tab content to play audio through the same
device as the rest of the conference, which will prevent echo in some
circumstances.

This change was tested by sharing the Test Video and a tab with audio in
Chrome, changing the output device, and consuming from Firefox.

Unit tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Documentation] Corrected name for `voiceFocusInsufficientResources` in documentation.
 - Allow for `realtimeUnsubscribeFromVolumeIndicator` to unsubscribe from specific callbacks.
+- Correctly mute video elements when bound, preventing local echo when sharing tabs via content
+  share.
+- [Demo] Local content share (e.g., video files) now plays audio through the selected audio
+  output device, rather than the default device, in browsers that support `setSinkId`.
 
 ## [2.3.0] - 2020-12-21
 
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -1744,13 +1744,24 @@ export class DemoMeetingApp
           return;
         }
         try {
-          await this.audioVideo.chooseAudioOutputDevice(name);
+          await this.chooseAudioOutputDevice(name);
         } catch (e) {
           fatal(e);
           this.log('Failed to chooseAudioOutputDevice', e);
         }
       }
     );
+  }
+
+  private async chooseAudioOutputDevice(device: string): Promise<void> {
+    // Set it for the content share stream if we can.
+    const videoElem = document.getElementById('content-share-video') as HTMLVideoElement;
+    if (this.defaultBrowserBehaviour.supportsSetSinkId()) {
+      // @ts-ignore
+      videoElem.setSinkId(device);
+    }
+
+    await this.audioVideo.chooseAudioOutputDevice(device);
   }
 
   private analyserNodeCallback = () => {};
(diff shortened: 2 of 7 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2205,6 +2216,16 @@ export class DemoMeetingApp
...
+    // In this case, we need to mute it, otherwise we'll hear our own shared audio.
+    // If it's not, we unmute the element — it should always be silent anyway, but
+    // this is a consistent default.
+    const us = this.meetingSession.configuration.credentials.attendeeId;
+    const isOurContent = tileState.isContent && tileState.boundAttendeeId === `${us}#content`;
+    this.log(`Is our content, and thus should be muted? ${isOurContent}`);
+    videoElement.muted = isOurContent;
+
     this.log(`binding video tile ${tileState.tileId} to ${videoElement.id}`);
     this.audioVideo.bindVideoElement(tileState.tileId, videoElement);
```
XHatan, line 2230: I would expect `bindVideoElement` to mute the video element https://github.com/aws/amazon-chime-sdk-js/blob/master/src/videotile/DefaultVideoTile.ts#L43 ?

richnew10: That's what I thought, but it's missing a bunch of the information it needs to make that determination, so I would need to alter the tile info structure to get to that, right?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1737,7 +1737,7 @@ export class DemoMeetingApp
           return;
         }
         try {
-          await this.audioVideo.chooseAudioOutputDevice(name);
+          this.chooseAudioOutputDevice(name);
```
XHatan, line 1740: does this need `await` ?

richnew10: Good catch, thanks!

demos/browser/app/meetingV2/meetingV2.ts
ltrung: I do not follow. Like Xuan said, bindVideoElement should just call connectVideoStreamToVideoElement and check whether the videoElement is muted and if it is not, append the 'muted' attribute. Is that not the case?

richnew10: Unless we want to always mute every video element (which seems to perhaps be the case!) we don't know within that code whether this is _our_ content stream, and thus we should mute it.

There is already code in `DefaultVideoTile` in `connectVideoStreamToVideoElement` to set `muted` by default, so perhaps that's not being called?

richnew10:
Ah, I found the real bug. Fix coming.

XHatan (approved):

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Documentation] Corrected name for `voiceFocusInsufficientResources` in documentation.
 - Allow for `realtimeUnsubscribeFromVolumeIndicator` to unsubscribe from specific callbacks.
+- Correctly mute video elements when bound, preventing local echo when sharing tabs via content
+  share.
+- [Demo] Local content share (e.g., video files) now plays audio through the selected audio
```
ltrung, line 31: FYI, we do not need to add CHANGELOG for demo or test related changes anymore.

richnew10: Yeah, this was by choice.

ltrung (approved):

Issue #706, opened 2020-09-23
audioVideo.stop does not stop/clean everything


rlebosse:
* For my test, I have blocked the domain `chime.aws` with the Little Snitch app (real use case : the user is on a network that is blocking the domain `chime.aws`)

* In the browser, I initialize, set some observers and start the meeting session :
```js
const logger = new ConsoleLogger('SDK', LogLevel.WARN)
const deviceController = new DefaultDeviceController(logger)
const configuration = new MeetingSessionConfiguration(
    meetingData,
    attendeeData
)
const meetingSession = new DefaultMeetingSession(
    configuration,
    logger,
    deviceController
)
const audioVideo = meetingSession.audioVideo
const audioVideoObserver = {
    audioVideoDidStartConnecting(reconnecting) {
        console.log('RLE audioVideoDidStartConnecting', {reconnecting})
    },
    audioVideoDidStart() {
        console.log('RLE audioVideoDidStart')
    },
    audioVideoDidStop(sessionStatus: MeetingSessionStatus) {
        console.log('RLE audioVideoDidStop', {
            sessionStatus,
            statusCode: sessionStatus?.statusCode(),
        })
    },
}
audioVideo.addObserver(audioVideoObserver)
audioVideo.start()
```

* I receive the `audioVideoDidStartConnecting` with `reconnecting = false`
* I can see in the Chrome DevTools that the `amazon-chime-sdk-js` is trying to connect to the Chime Meeting (in vain), multiple times
* After some time, in my code, I call `audioVideo.stop()`
* I expect from the `amazon-chime-sdk-js` to stop trying to connect to the Chime Meeting, but it keeps doing it !

Moreover, if stop blocking the domain `chime.aws` after calling `audioVideo.stop()`, the other attendees will see me joining the Chime Meeting even if I explicitly stopped it.

Would it be possible to make the call to `audioVideo.stop()` to stop trying to connect to the Chime meeting please ?
I think there should be a way to clean stop everything.

Thanks in advance !

**Platform**
 - OS: macOS 10.15.6
 - Browser: Google Chrome 85
 - Amazon Chime SDK version: 1.18.0

rlebosse:
Any news on this issue ?

bradmanAWS:
Hi @rlebosse I just looked into this issue and I cannot reproduce this behavior. Running `audioVideo.stop()` kills the meeting session, as expected. Can you confirm this is still an issue for you? Sorry for the delay.

[...]

PR #1002, opened 2021-01-12
Disable reconnecting in AudioVideoControllerFacade's stop method.

devalevenkatesh:
**Issue #:**
#706 

**Description of changes:**
- In reconnection, due to current priorities and transitions of the meeting session state and controller, some time the `AudioVideoStart` and `AudioVideoReconnect` will keep on running without any effect of `AudioVideoStop` and `AudioVideoClean` even if explicitly called. This leads to a meeting re-join even if someone leaves the meeting.
- There may be still cases when `AudioVideoStart` may start due to transitions and session's state as implemented in [`DefaultSessionStateController`](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/sessionstatecontroller/DefaultSessionStateController.ts).
- This change solves one case which should stop the meeting session reconnecting attempts once `audioVideo.stop()` is called. This will also clean the meeting session resources and avoid re-joing the same meeting once left.

**Testing**

2. How did you test these changes? Added unit test and tried locally.
Testing steps:
2.1. Run the JS SDK demo meeting app.
2.2. Join a meeting successfully.
2.3. Turn off your internet, JS SDK will start reconnecting.
2.4. Now calling `app.audioVideo.stop()` in the browser console should stop the reconnecting attempts and do not re-join the meeting.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes, the demo browser app.
4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? I have modified a public method which is AudioVideoControllerFacade's `stop` method. Will add Kyu as reviewer as well here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Amazon Voice Focus now makes better use of available CPU resources,
   extending support to lower-end devices and improving quality on higher-end
   devices.
+- Disable reconnecting in AudioVideoControllerFacade's `stop` method.
+  Add documentation for the `stop` method.
 
 ### Removed
 
--- a/docs/classes/defaultaudiovideocontroller.html
+++ b/docs/classes/defaultaudiovideocontroller.html
@@ -604,7 +604,7 @@
 						<li class="tsd-description">
 							<aside class="tsd-sources">
 								<ul>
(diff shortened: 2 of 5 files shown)


src/audiovideocontroller/AudioVideoControllerFacade.ts
```diff
@@ -9,6 +9,14 @@ export default interface AudioVideoControllerFacade {
   addObserver(observer: AudioVideoObserver): void;
   removeObserver(observer: AudioVideoObserver): void;
   start(): void;
+  /**
```
simmkyu, line 12: Nice explanation, but I think you can move this comment into the `stop()` method in the implementation.

devalevenkatesh: Cool, thanks! I have updated it now.

simmkyu (dismissed):

RidipDe (approved):

Issue #985, opened 2021-01-03
realtimeSubscribeToAttendeeIdPresence doesn't capture attendees who have left while network failure


abhijeet-toptal:
**Timeline**
Outline the main actions the user performed and when:

* User A joined
* User B joined
* User A's internet went haywire (simulated by turning wifi off)
* User B observed that user A is no longer in the meeting, which is good and work as expected.
* User B left the meeting.
* User A's internet is back on.
* realtimeSubscribeToAttendeeIdPresence didn't fire with information that User B has left the meeting. It only fired for local user (User A itself) with present: true.

**Describe the bug**
realtimeSubscribeToAttendeeIdPresence doesn't get fired for attendees who have left while network failure.
Ideally when User A comes back online, there should be events fired to keep the roster up to date.

So how can we keep the roster up to date in case of network failures? 

**Platform**
 - OS: Windows 10
 - Browser: Chrome
 - Browser Version: Latest
 - Amazon Chime SDK GitHub Commit: Latest

abhijeet-toptal:
Logs:

```
instrument.js:118 2021-01-03T07:15:15.564Z [INFO] SDK - running task AudioVideoReconnect/8eda3966-9562-4168-8491-3e72b4f545d0/9213d0d3-f07f-6c19-3910-2890af5087e9/Timeout15000ms/Media
instrument.js:118 2021-01-03T07:15:15.565Z [INFO] SDK - serial group task AudioVideoReconnect/8eda3966-9562-4168-8491-3e72b4f545d0/9213d0d3-f07f-6c19-3910-2890af5087e9/Timeout15000ms/Media running subtask AudioVideoReconnect/8eda3966-9562-4168-8491-3e72b4f545d0/9213d0d3-f07f-6c19-3910-2890af5087e9/Timeout15000ms/Media/CleanRestartedSessionTask
instrument.js:118 2021-01-03T07:15:15.565Z [INFO] SDK - running task AudioVideoReconnect/8eda3966-9562-4168-8491-3e72b4f545d0/9213d0d3-f07f-6c19-3910-2890af5087e9/Timeout15000ms/Media/CleanRestartedSessionTask
[...]

PR #1012, opened 2021-01-13
Fix dropped attendee presence during network reconnects

ajselvar:
**Issue #:**
#985 

**Description of changes:**

- The webscoket events for remote attendees who left or dropped during network reconnects are missed by the local attendee because of the broken websocket. This results in inconsistency in the local attendee's roster.
- After a successful network reconnect, the backend sends a snapshot of list of current attendees.
- This change uses the snapshot and calculate delta between local and remote snapshots to find the attendees who left or dropped during the network reconnect.

**Testing**

1. Have you successfully run `npm run build:release` locally? **Yes**
2.1. Go to the demo/browser application and run `npm run build`.
2.2. go to the demo/serverless application. Follow through README file to deploy the serverless meetingV2 application.
2.3. Using the generated endpoint, join the same meeting from 5 machines machines(A,B,C,D,E).
2.4. Disconnect network in one of the machine A.
2.5. Leave meeting from machine B and C.
2.6. Reconnect the network in machine A.
2.7. Check for the reception of the snapshot event.
2.8. Verify that the roster is updated with attendees B and C removed.
2.9. Rejoin the meeting as attendee F. Verify the roster is updated with attendee F.
2.10 Make attendee F leave. Verify attendee F is removed from roster.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? **browser, serverless**


src/volumeindicatoradapter/DefaultVolumeIndicatorAdapter.ts
```diff
@@ -85,6 +101,35 @@ export default class DefaultVolumeIndicatorAdapter implements VolumeIndicatorAda
...
+      delete this.streamIdToAttendeeId[streamId];
+      delete this.streamIdToExternalUserId[streamId];
+      delete this.warnedAboutMissingStreamIdMapping[streamId];
+      delete this.attendeeIdToStreamId[deletedAttendeeId];
+
+      this.realtimeController.realtimeSetAttendeeIdPresence(
+        deletedAttendeeId,
+        false,
+        externalUserId,
+        false,
```
aburnson, line 129: Technically, we don't know whether they dropped or not. It might be worth just adding a comment to that effect that given the information in protocol, dropped/left is unknowable.

aburnson (dismissed):

src/volumeindicatoradapter/DefaultVolumeIndicatorAdapter.ts
ajselvar: Added comments here.

aburnson (approved):

PR #1025, opened 2021-01-22
Add hotfix option to release script

ltrung:
Description of changes:
Add hotfix option for the release script

Testing
- Have you successfully run npm run build:release locally? Yes
- How did you test these changes? Run the script in a fork private repo
- Can these changes be tested using the demo application? If yes, which demo application can be used to test it? No
- Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? No
- Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved? No


script/release
```diff
@@ -62,13 +54,14 @@ puts
...
 x = STDIN.gets
 exit(1) unless x.strip == 'yes'
 puts
 
 verbose('git fetch origin')
-verbose('git reset --hard origin/master')
+verbose('git reset --hard origin/master') if release_option != "5"
 verbose('git clean -ffxd .')
 
 File.write(changelog_file, File.read(changelog_file).gsub(/\[Unreleased\]/, "[#{version_string}] - #{Date.today}"))
```
michhyun1, line 67: Since we only have one commit on top of the last publish tagged commit, we shouldnt have any changelog items here right?

script/release
```diff
@@ -89,5 +82,11 @@ puts
 
 verbose("git push origin HEAD:release -f")
 
+if release_option != "5"
+  verbose("git push origin HEAD:release -f")
+else
+  verbose("git push -f")
```
michhyun1, line 88: which branch would this push to?

ltrung: The current branch which is hotfix.

script/release
ltrung: We do. We will have the hotfix commit before this. This is the release commit on top of the hotfix.

script/release
michhyun1: Can we explicitly push to hotfix branch? Instead of assuming that the use is on the hotfix branch when they run this script?

script/release
```diff
@@ -89,5 +82,11 @@ puts
 
 verbose("git push origin HEAD:release -f")
 
+if release_option != "5"
+  verbose("git push origin HEAD:release -f")
```
anuranduttaroy, line 86: This is already being done on Line 83

script/release
```diff
@@ -62,13 +54,14 @@ puts
...
 puts
+
 puts "Type 'yes' to continue"
 x = STDIN.gets
 exit(1) unless x.strip == 'yes'
 puts
 
 verbose('git fetch origin')
-verbose('git reset --hard origin/master')
+verbose('git reset --hard origin/master') if release_option != "5"
```
michhyun1, line 64: should we do reset --hard origin/hotfix?

ltrung: we do not fix the branch name for hotfix so this would not be possible.

michhyun1 (approved):

PR #1050, opened 2021-02-09
Video test app integration test

ldai1:
**Issue #:**

**Description of changes:**
Add end-to-end test for video test app. 
Tests cover apis:
1. addVideoTile 
2. startLocalVideoTile
3. hasStartedLocalVideoTile
4. haveVideoTilesWithStreams
5. haveVideoTileForAttendeeId
6. getLocalVideoTile
7. getAllVideoTiles
8. getAllRemoteVideoTiles 
9. removeAllVideoTiles
10. stopLocalVideoTile 
11. sendTileStateUpdate

Test cover scenarios:
    * 2 attendees join from 2 different browser sessions
    * Both attendees turn on their local video
    * Verify that both attendees can see their local video on
    * Verify that both attendees can see the other attendee’s remote video
    * Both attendees turn off their video
    * Verify that both attendees’ local videos off
    * Verify that both attendees’ remote videos off

**Testing**

2. How did you test these changes? manullly and run the chromDrive
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Video_test


michhyun1:
FAIL: Requires CHANGELOG, but no CHANGELOG.md entries found since origin/master.

You need to add changelog entry

integration/js/app/video_test/video_test.html
```diff
@@ -62,6 +62,76 @@ <h2>Join a meeting</h2>
           Unbind Video Element
         </button>
       </form>
+
+       
+      <form id="add-video-tile-form">
+        <button id="add-video-tile-btn" type="submit">
+          addVideoTile
+        </button>
+        <label for="add-video-tile-label">Tile Id: </label>
```
michhyun1, line 71: Can we put "Added tile ID: " instead of just "Tile ID:"

integration/js/app/video_test/video_test.html
```diff
@@ -62,6 +62,76 @@ <h2>Join a meeting</h2>
...
+        </button>
+        <label for="has-started-local-video-tile-label">Value: </label>
+        <span id="has-started-local-video-tile-boolean"></span>
+      </form>
+
+      <form id="get-local-video-tile-form">
+        <button id="get-local-video-tile-btn" type="submit">
+          getLocalVideoTile
+        </button>
+        <label for="local-video-tile-label">Tile Id: </label>
```
michhyun1, line 87: Does this only return the Tile ID?

integration/js/app/video_test/video_test.html
```diff
@@ -62,6 +62,76 @@ <h2>Join a meeting</h2>
...
+        <span id="local-video-tile-id"></span>
+      </form>
+
+      <form id="get-all-video-tiles-form">
+        <button id="get-all-video-tiles-btn" type="submit">
+          getAllVideoTiles
+        </button>
+        <label for="total-video-tiles-label">Total Tiles Count: </label>
```
michhyun1, line 95: Maybe it would be useful to return the actual VideoTile Objects here instead of just the count.

integration/js/app/video_test/video_test.html
```diff
@@ -62,6 +62,76 @@ <h2>Join a meeting</h2>
...
+        <span id="total-video-tiles-count"></span>
+      </form>
+
+      <form id="get-all-remote-video-tiles-form">
+        <button id="get-all-remote-video-tiles-btn" type="submit">
+          getAllRemoteVideoTiles
+        </button>
+        <label for="total-remote-video-tiles-label">Total Remote Tiles Count: </label>
```
michhyun1, line 103: Same here

integration/js/app/video_test/video_test.html
```diff
@@ -62,6 +62,76 @@ <h2>Join a meeting</h2>
...
+        <span id="total-remote-video-tiles-count"></span>
+      </form>
+
+      <form id="have-video-tile-for-attendeeId-form">
+        <button id="have-video-tile-for-attendeeId-btn" type="submit">
+          haveVideoTileForAttendeeId
```
michhyun1, line 109: Is there an input where I can specify the attendee Id?

integration/js/app/video_test/video_test.html
```diff
@@ -62,6 +62,76 @@ <h2>Join a meeting</h2>
...
+        <button id="have-video-tiles-with-streams-btn" type="submit">
+          haveVideoTilesWithStreams
+        </button>
+        <label for="have-video-tiles-with-streams-label">Value: </label>
+        <span id="have-video-tiles-with-streams-boolean"></span>
+      </form>
+
+      <form id="send-tile-state-update-form">
+        <button id="send-tile-state-update-btn" type="submit">
+          sendTileStateUpdate
```
michhyun1, line 125: If we do include this API, I think we may want to be able to pass in a TileState parameter

integration/js/app/video_test/video_test.ts
```diff
@@ -113,6 +113,90 @@ export class DemoMeetingApp implements AudioVideoObserver {
...
+      const localVideoTile = this.audioVideo.getLocalVideoTile().id();
+      let videoTileIdObject = document.getElementById(`local-video-tile-id`);
+      videoTileIdObject.innerText = JSON.stringify(localVideoTile, null, '  ');;
+      getLocalVideoTileForm.reset();
+    });
+
+    const getAllVideoTilesForm = document.getElementById('get-all-video-tiles-form') as HTMLFormElement;
+    getAllVideoTilesForm.addEventListener('submit', e => {
+      e.preventDefault();
+      const totalVideoTilesCount = this.audioVideo.getAllVideoTiles().length;
```
michhyun1, line 146: Maybe we can show the video tiles, not just the length of the array?

integration/js/app/video_test/video_test.ts
```diff
@@ -113,6 +113,90 @@ export class DemoMeetingApp implements AudioVideoObserver {
...
+      let videoTilesCountObject = document.getElementById(`total-video-tiles-count`);
+      videoTilesCountObject.innerText = JSON.stringify(totalVideoTilesCount, null, '  ');;
+      getAllVideoTilesForm.reset();
+    });
+
+    const getAllRemoteVideoTilesForm = document.getElementById('get-all-remote-video-tiles-form') as HTMLFormElement;
+    getAllRemoteVideoTilesForm.addEventListener('submit', e => {
+      e.preventDefault();
+      const totalRemoteVideoTilesCount = this.audioVideo.getAllRemoteVideoTiles().length;
```
michhyun1, line 155: same here

integration/js/app/video_test/video_test.ts
```diff
@@ -113,6 +113,90 @@ export class DemoMeetingApp implements AudioVideoObserver {
...
+      let remoteVideoTilesCountObject = document.getElementById(`total-remote-video-tiles-count`);
+      remoteVideoTilesCountObject.innerText = JSON.stringify(totalRemoteVideoTilesCount, null, '  ');;
+      getAllRemoteVideoTilesForm.reset();
+    });
+
+
+    const haveVideoTileForAttendeeIdForm = document.getElementById('have-video-tile-for-attendeeId-form') as HTMLFormElement;
+    haveVideoTileForAttendeeIdForm.addEventListener('submit', e => {
+      e.preventDefault();
+      const attendeeId = this.meetingSession.configuration.credentials.attendeeId;
```
michhyun1, line 165: I think we should let the user decide what the attendee ID should be, so that we can write dynamic tests

integration/js/app/video_test/video_test.ts
```diff
@@ -113,6 +113,90 @@ export class DemoMeetingApp implements AudioVideoObserver {
...
+      let haveVideoTilesWithStreamsBooleanObject = document.getElementById(`have-video-tiles-with-streams-boolean`);
+      haveVideoTilesWithStreamsBooleanObject.innerText = JSON.stringify(haveVideoTilesWithStreamsBoolean, null, '  ');;
+      haveVideoTilesWithStreamsForm.reset();
+    });
+
+    const sendTileStateUpdateForm = document.getElementById('send-tile-state-update-form') as HTMLFormElement;
+    sendTileStateUpdateForm.addEventListener('submit', e => {
+      e.preventDefault();
+      // @ts-ignore
+      const videoTile = this.audioVideo.videoTileController.sendTileStateUpdate();
```
michhyun1, line 187: We may not need this API, this should be called when we update any video tile, so we should have coverage on this API without explicitly adding this API to the video_test integ test

integration/js/pages/TestAppPage.js
```diff
@@ -53,6 +70,12 @@ class TestAppPage {
     await attendeeNameInputBox.sendKeys(attendeeName);
   }
 
+  async enterMeetingTitle(meetingTitle) {
+    await this.enterMeetingId(meetingTitle);
+    let joinMeetingButton = await this.driver.findElement(elements.joinMeetingButton);
+    await joinMeetingButton.click();
+  }
+
```
michhyun1, line 78: This looks the same as authenticate(). We can use one and remove the other?

integration/js/pages/TestAppPage.js
```diff
@@ -110,6 +133,153 @@ class TestAppPage {
...
+    let hasStartedLocalVideoTileBooleanElement = await this.driver.findElement(elements.hasStartedLocalVideoTileElementId);
+    if (!hasStartedLocalVideoTileBooleanElement) {
+      return false;
+    }
+    let hasStartedLocalVideoTileBoolean =  await hasStartedLocalVideoTileBooleanElement.getText();
+    
+    if (expectedHasStartedLocalVideoTileBoolean === hasStartedLocalVideoTileBoolean) {
+      return true;
+    }
+    return false;
```
michhyun1, line 210: Seems like we could use a helper function to reduce repetitive code here

integration/js/steps/JoinVideoTestMeetingStep.js
```diff
@@ -0,0 +1,47 @@
+const {KiteTestError, Status} = require('kite-common');
+const AppTestStep = require('../utils/AppTestStep');
+
+class JoinVideoTestMeetingStep extends AppTestStep {
```
michhyun1, line 4: Dont we already have a Step that joins a meeting?

integration/js/app/video_test/video_test.html
ldai1: Sure, will do.

integration/js/app/video_test/video_test.html
ldai1: It returns VideoTile but I only use id to verify the response.

integration/js/app/video_test/video_test.html
ldai1: You mean compare the Two objects with the expected two objects to verify the response? Not sure if it is necessary but just compare the count could probably be enough since in the corresponding step in the videoTest, we only compare the number on roster to verify both attendees have joined the meeting.

integration/js/app/video_test/video_test.html
ldai1: Same reason above. Let me know if you have a different thought.

integration/js/app/video_test/video_test.html
ldai1: The attendeeId would be the local attendeeId, which I get from the video_test.js to fetch from meeting session.

integration/js/app/video_test/video_test.ts
ldai1: Make sense. Will remove.

integration/js/pages/TestAppPage.js
ldai1: But the authenticate() has step to create meetingId randomly. In this end_to_end test, we are relying on sever.js to create attendeeId and meetingID for us, then we type in these values into the UI text input.

integration/js/pages/TestAppPage.js
ldai1: Good point. Will change it.

integration/js/app/video_test/video_test.html
michhyun1: I think the main reason is for validation purposes. The VideoTile is an object full of metadata that would make it easier to see what is happening to those videoTile objects behind the scenes as the video tile states are changing

integration/js/VideoEnd2EndTest.js
```diff
@@ -0,0 +1,67 @@
...
+
+  async runIntegrationTest() {
+    this.numberOfParticipant = 2;
+    const attendeeId = uuidv4();
+    const meetingTitle = 'video_end2end_test_' + uuidv4() ;
+    const session = this.seleniumSessions[0];
+    const useSimulcast = this.useSimulcast;
+    await OpenAppStep.executeStep(this, session);
+    // local user join meeting step 
+    await JoinVideoTestMeetingStep.executeStep(this, session, attendeeId, meetingTitle, useSimulcast);
```
michhyun1, line 20: I think we can jsut use AuthenitcateUserStep here, you dont need to pass in a meeting Title

integration/js/app/video_test/video_test.html
michhyun1: But what if I want to check if there is a video tile for an attendee that is not the local attendee?

integration/js/pages/TestAppPage.js
michhyun1: I noticed that you stil create the meeting id randomly in integration/js/VideoEnd2EndTest.js. The code looks almost identical for AuthenticateUserStep

integration/js/VideoEnd2EndTest.js
ldai1: If I use AuthenitcateUserStep, the existing authenticate method in TestAppPage will create a new meetingTitle, which is not we want. In the existing workflow of video test, we enter the meetingTitle in the authenticate() method where we create a new meetingTitle on the fly. But what we want here is to use the meetingTitle created by the server and enter that meetingTitle in the authenticate step. In order not to break the existing test, I create this JoinVideoTestMeetingStep to pass the meetingTitle to a new method enterMeetingTitle in TestAppPage to enter the meetingTitle.

integration/js/steps/JoinVideoTestMeetingStep.js
ldai1: Yes, but I need a customized method to pass in the attendeeId and meetingTitle so that we can enter them in the UI input boxes.

integration/js/app/video_test/video_test.ts
ldai1: Yes you are right. Will make it as an input for dynamic tests

PR #1056, opened 2021-02-10
Set attendeePresenceTimeoutMs to use value passed as parameter in the URL

RidipDe:
**Issue #:**
N/A

**Description of changes:**
Set attendeePresenceTimeoutMs to use value passed as parameter in the URL 

**Testing**

Yes

Manually

Yes, meeting demo app

No

No

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -1128,7 +1128,11 @@ export class DemoMeetingApp
       enableWebAudio: this.enableWebAudio,
     });
     configuration.enableUnifiedPlanForChromiumBasedBrowsers = this.enableUnifiedPlanForChromiumBasedBrowsers;
-    configuration.attendeePresenceTimeoutMs = 5000;
+    const urlParameters = new URL(window.location.href).searchParams;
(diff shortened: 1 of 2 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1128,7 +1128,10 @@ export class DemoMeetingApp
       enableWebAudio: this.enableWebAudio,
     });
     configuration.enableUnifiedPlanForChromiumBasedBrowsers = this.enableUnifiedPlanForChromiumBasedBrowsers;
-    configuration.attendeePresenceTimeoutMs = 5000;
+    const urlParameters = new URL(window.location.href).searchParams;
+    if (urlParameters.has('attendee-presence-timeout-ms')) {
```
simmkyu, line 1132: Use `isNaN` to check if the query string can be converted to a string.

```
const timeoutMs = urlParameters.get('attendee-presence-timeout-ms');
if (!isNaN(timeoutMs)) {
  configuration.attendeePresenceTimeoutMs = Number(timeoutMs);
}
```

RidipDe: Removed the check urlParameters.has() and used isNaN instead.

simmkyu (approved):

Issue #1051, opened 2021-02-09
Speaker Test in Demo App is unreliable on iOS 14.4


chuckntaylor:
**Timeline**
Outline the main actions the user performed and when:

1. Ran `npm run start` to start up the demo meeting application.
2. Used ngrok to make port 8080 accessible on my phone
3. Opened the app with Safari on an iPhone 12 pro max running iOS14.4
4. Provided meeting information on the first screen (meeting name, username, etc).
5. On the 'select devices' screen, there is a 'Test' button for checking the audio output. The audio tone barely works. It sometimes plays, sometimes doesn't. And always, after enough use, it simply stops playing any tone.

On my laptop macOS 11.2 running Chrome version 88, there is absolutely no problem whatsoever. Plays beautifully. It is repeatable, and has no bugs.

Further Question:
Is there a reason why the audio test doesn't simply play an audio file? Is there a reason for generating a tone using the `AudioContext` API?

As far as I can gather, the issue must be in here (part of the `TestSound` class:

```
...
const audioMixController = new DefaultAudioMixController(this.logger)
    if (new DefaultBrowserBehavior().supportsSetSinkId()) {
      try {
        await audioMixController.bindAudioDevice({ deviceId: this.sinkId })
      } catch (e) {
        this.logger?.error(`Failed to bind audio device: ${e}`)
      }
    }
    try {
      await audioMixController.bindAudioElement(new Audio())
    } catch (e) {
      this.logger?.error(`Failed to bind audio element: ${e}`)
    }
    await audioMixController.bindAudioStream(destinationStream.stream)
    new TimeoutScheduler((this.rampSec * 2 + this.durationSec + 1) * 1000).start(() => {
      audioContext.close()
    })
...
```

**Platform**
 - OS: iOS 14.4
 - Browser: iOS Safari
 - Browser Version: cannot see
 - Amazon Chime SDK GitHub Commit: V2.4.1

chuckntaylor:
I would like to add that the speaker test in the `meetingReadinessChecker` app works without any issue on my iPhone. I believe in that app's situation, it works a bit differently in that we call `meetingReadinessChecker.checkAudioOutput()` to get the tone.

ltrung:
@chuckntaylor Thanks for reporting. I have been able to reproduce the issue on my iphone 12.

chuckntaylor:
@ltrung Thanks for looking into it! I just checked out the changes made in the merge and tried it out on my end. Works perfectly!

PR #1058, opened 2021-02-11
[Demo] Fix test sound in Safari

ltrung:
**Issue #1051 :**
Fix test sound button in the preview page sometimes does not play sound in Safari (macOS and iOS).

**Description of changes:**
Create an audio element for testing audio output in preview page instead of creating a new HTMLAudioElement everytime the test sound button is clicked which seems to cause issues in Safari.
 
**Testing**

2. How did you test these changes? Manually clicked on the test sound button many times in Safari MacOS and iOS and verified that there was sound.
Yes, this can be tested using the meeting demo test sound button in preview page. 

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -124,6 +124,8 @@ type VideoFilterName = 'Emojify' | 'CircularCut' | 'NoOp' | 'None';
 const VIDEO_FILTERS: VideoFilterName[] = ['Emojify', 'CircularCut', 'NoOp'];
 
 class TestSound {
+  static testAudioElement = new Audio();
+
   constructor(
     private logger: Logger,
     private sinkId: string | null,
@@ -164,7 +166,7 @@ class TestSound {
       }
     }
     try {
-      await audioMixController.bindAudioElement(new Audio());
+      await audioMixController.bindAudioElement(TestSound.testAudioElement);
     } catch (e) {
       fatal(e);
       this.logger?.error(`Failed to bind audio element: ${e}`);


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -164,13 +166,14 @@ class TestSound {
       }
     }
     try {
-      await audioMixController.bindAudioElement(new Audio());
+      await audioMixController.bindAudioElement(this.audioElem);
```
richnew10, line 169: If the fix here is to retain a reference to the audio element…

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -587,7 +590,12 @@ export class DemoMeetingApp
     document.getElementById('button-test-sound').addEventListener('click', async e => {
       e.preventDefault();
       const audioOutput = document.getElementById('audio-output') as HTMLSelectElement;
-      const testSound = new TestSound(this.meetingEventPOSTLogger, audioOutput.value);
+      const testAudioElem = document.getElementById('test-audio') as HTMLAudioElement;
```
richnew10, line 593: … do you actually need to make this change?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -164,13 +166,14 @@ class TestSound {
...
     try {
-      await audioMixController.bindAudioElement(new Audio());
+      await audioMixController.bindAudioElement(this.audioElem);
     } catch (e) {
       fatal(e);
       this.logger?.error(`Failed to bind audio element: ${e}`);
     }
     await audioMixController.bindAudioStream(destinationStream.stream);
     new TimeoutScheduler((this.rampSec * 2 + this.durationSec + 1) * 1000).start(() => {
+      console.log('closing audio context');
```
richnew10, line 176: Remove debug log.

richnew10 (changes requested):

demos/browser/app/meetingV2/meetingV2.ts
ltrung: No, it is fine with just a constant AudioElement (which I had originally). I just thought to do it this way since we also had an audio element specified for the main meeting page.

richnew10 (approved):

PR #1060, opened 2021-02-12
Add support for iOS Chrome and iOS Firefox

michhyun1:
**Issue #:**

**Description of changes:**

Add support for ios chome and ios firefox browsers.

**Testing**

2. How did you test these changes? locally and QA Sanity tests
Yes, to test you can checkout this code and deploy a serverless demo.


src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -141,7 +161,7 @@ export default class DefaultBrowserBehavior implements BrowserBehavior {
   }
 
   screenShareUnsupported(): boolean {
-    if (this.isSafari()) {
+    if (this.isSafari() || this.isIOSChrome() || this.isIOSFirefox()) {
```
ltrung, line 164: This is for the legacy screenshare so we should remove it instead.

src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -90,15 +109,16 @@ export default class DefaultBrowserBehavior implements BrowserBehavior {
...
   }
 
   requiresIceCandidateGatheringTimeoutWorkaround(): boolean {
     return this.hasChromiumWebRTC();
   }
 
   requiresUnifiedPlanMunging(): boolean {
-    let shouldRequire = this.isSafari() && this.isUnifiedPlanSupported();
+    let shouldRequire =
+      (this.isSafari() && this.isUnifiedPlanSupported()) || this.hasWebKitWebRTC();
```
ltrung, line 121: Doesn't Safari also use WebKit? Should we group this under something like WebKitBasedBrowsers (similar to ChromiumBasedBrowsers?

simmkyu: Yes, macOS and iOS Safari are WebKit-based browsers

src/browserbehavior/DefaultBrowserBehavior.ts
michhyun1: I just put a todo comment to remove it in the next major version update, in case anyone might be using it.

src/browserbehavior/DefaultBrowserBehavior.ts
michhyun1: Added Safari to the array of Webkit browsers

src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -140,6 +156,7 @@ export default class DefaultBrowserBehavior implements BrowserBehavior {
     return this.isChrome() || this.isEdge();
   }
 
+  // TODO: Deprecated, needs to be removed
```
ltrung, line 159: We probably wants to at least to do like console.warn and say this is no longer supported and just return false all the time. This API is already wrong anyway.

ltrung (approved):

Issue #1062, opened 2021-02-12
Chrome - VoiceFocus - The AudioContext was not allowed to start.


peili:
Steps to reproduce:

Connect to an AWS Chime Meeting (VoiceFocus) enabled, without any user gesture on the page.
Common use case, e.g. page refresh.

`
const voiceFocusDevice = await transformer.createTransformDevice(deviceId)
meetingSession.audioVideo.chooseAudioInputDevice(voiceFocusDevice)
`

This will show a warning, but also entire meeting audio fails.
`DefaultDeviceController.js?66c1:1156 The AudioContext was not allowed to start.`
See: https://developers.google.com/web/updates/2017/09/autoplay-policy-changes#webaudio

Selecting a deviceId without creating transform device, works without any issues. 
`meetingSession.audioVideo.chooseAudioInputDevice(mediaDevice.deviceId)`

It would be an improvement if either `createTransformDevice` or `chooseAudioInputDevice` throw an error in such a case (because entire media will fail).
Current workaround is to create your own user gesture listener to detect if `createTransformDevice` can be used.

michhyun1:
Hi @peili, could you please clarify what you mean by user gesture? Also, could you please specify the SDK version that you are using? Thanks

rnewman:
Hi @peili,

Thanks for filing an issue.

In general it's expected that audio input and output configuration occurs after a user gesture — this restriction also applies to `getUserMedia` calls and playing output audio, so I'm a little surprised that you report that `chooseAudioInputDevice` is working without a gesture.

Is there any reason you think that resuming the `AudioContext` is insufficient here?

peili:
> In general it's expected that audio input and output configuration occurs after a user gesture
Agree in general, however on a page refresh I think it's reasonable to skip the previous user gesture requirement and get right back into the meeting. That has been working so far without using VoiceFocus.  So I thought I'd report the inconsistent behaviour when enabling VoiceFocus. Feel free to close this ticket, if you think that browser policy is communicated clearly. It wasn't obvious to me first.

rnewman:
Browser policy here is not only not communicated clearly, but it also [continues to evolve over time](https://developers.google.com/web/updates/2018/11/web-audio-autoplay), so I definitely want to get to the bottom of what you're seeing and find out if there's something we can smooth over in the Chime SDK, either through documentation or code.

[...]

PR #1066, opened 2021-02-16
Resume a suspended audio context when selecting a transform device.

richnew10:
In some circumstances, most notably creating a transform device prior to the
user performing a gesture (e.g., clicking), the audio context through which
the `DefaultDeviceController` pipes audio will be suspended.

In this commit the device controller will attempt to resume the context if
this occurs.

An exception is present for the unusual circumstance of the context being an
`OfflineAudioContext`.

Tests exercise the new behavior.

This commit also fixes some CHANGELOG entries to be more descriptive.

**Issue #:** #1062

**Description of changes:**

**Testing**

Yes.

Only via unit tests.

No: our demo applications all involve user gestures.

No.

No.

### DIFF
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -165,6 +165,25 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     }
 
     const context = DefaultDeviceController.getAudioContext();
+
+    if (context instanceof OfflineAudioContext) {
+      // Nothing to do.
+    } else {
+      switch (context.state) {
+        case 'running':
+          // Nothing to do.
+          break;
+        case 'closed':
+          // A closed context cannot be used for creating nodes, so the correct
+          // thing to do is to raise a descriptive error sooner.
+          throw new Error('Cannot choose a transform device with a closed audio context.');
+        case 'suspended':
+          // A context might be suspended after page load. We try to resume it
+          // here, otherwise audio won't work.
+          await context.resume();
+      }
+    }
+
     let nodes;
     try {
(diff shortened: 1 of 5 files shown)


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1338,7 +1357,11 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
 
   static closeAudioContext(): void {
     if (DefaultDeviceController.audioContext) {
-      DefaultDeviceController.audioContext.close();
+      try {
+        DefaultDeviceController.audioContext.close();
+      } catch (e) {
+        // Nothing we can do.
+      }
```
richnew10, line 1364: This prevents failures if the context is somehow an `OfflineAudioContext`, which throws when you try to close it.

p-foucht (approved):

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -165,6 +165,25 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     }
 
     const context = DefaultDeviceController.getAudioContext();
+
+    if (context instanceof OfflineAudioContext) {
+      // Nothing to do.
```
ltrung, line 170: nit: Do we need this if statement at all?

richnew10: We need the `else` that comes after it 😂 

I find it's clearer to write `instanceof` in the positive form and explain that the block is intentionally empty than to write `if (!(context instanceof OfflineAudioContext)) {` with a comment above.

ltrung (approved):

Issue #1052, opened 2021-02-09
Cannot read property 'toLowerCasePropertyNames' of undefined


gsudhanshu:
Description: I get meeting Response from server as json object as shown below in after createMeeting. and I get attendee response in json object format as shown below. When I try to instantiate a MeetingSessionConfiguration using these responses, I get error ERROR TypeError: Cannot read property 'toLowerCasePropertyNames' of undefined
    at toLowerCasePropertyNames (MeetingSessionConfiguration.js:157)

after createMeeting :
meetings.service.ts:19 
{Meeting: {…}, @metadata: {…}}@metadata: {statusCode: 201, effectiveUri: "https://chime.us-east-1.amazonaws.com/meetings", headers: {…}, transferStats: {…}}Meeting: ExternalMeetingId: "Meeting1"MediaPlacement: {AudioHostUrl: "79f1ea3279da5e1eb710020f92578ae5.k.m3.as1.app.chime.aws:3478", AudioFallbackUrl: "wss://haxrp.m3.as1.app.chime.aws:443/calls/9f8e3030-7d32-42d9-b425-efaee9ef2766", ScreenDataUrl: "wss://bitpw.m3.as1.app.chime.aws:443/v2/screen/9f8e3030-7d32-42d9-b425-efaee9ef2766", ScreenSharingUrl: "wss://bitpw.m3.as1.app.chime.aws:443/v2/screen/9f8e3030-7d32-42d9-b425-efaee9ef2766", ScreenViewingUrl: "wss://bitpw.m3.as1.app.chime.aws:443/ws/connect?pa…tHub-Call-Id=9f8e3030-7d32-42d9-b425-efaee9ef2766", …}MediaRegion: "ap-southeast-1"MeetingId: "9f8e3030-7d32-42d9-b425-efaee9ef2766"__proto__: Object__proto__: Objectconstructor: ƒ Object()hasOwnProperty: ƒ hasOwnProperty()isPrototypeOf: ƒ isPrototypeOf()propertyIsEnumerable: ƒ propertyIsEnumerable()toLocaleString: ƒ toLocaleString()toString: ƒ ()valueOf: ƒ valueOf()__defineGetter__: ƒ __defineGetter__()__defineSetter__: ƒ __defineSetter__()__lookupGetter__: ƒ __lookupGetter__()__lookupSetter__: ƒ __lookupSetter__()get __proto__: ƒ __proto__()set __proto__: ƒ __proto__()

[...]

PR #1072, opened 2021-02-17
Fix crash issue "Cannot read property 'toLowerCasePropertyNames' of undefined"

narek-king:
Related issue link: https://github.com/aws/amazon-chime-sdk-js/issues/1052

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `attendeePresenceReceived` meeting event.
 - Add the file sharing workaround for Chrome 88 in FAQs.
 - Add support for Chrome for iOS and Firefox for iOS.
+- Fix Cannot read property 'toLowerCasePropertyNames' of undefined
 
 ### Changed
 - [Demo] Set `attendeePresenceTimeoutMs` to use value passed as parameter in the URL.
--- a/src/meetingsession/MeetingSessionConfiguration.ts
+++ b/src/meetingsession/MeetingSessionConfiguration.ts
@@ -173,7 +173,12 @@ export default class MeetingSessionConfiguration {
     } else if (typeof input !== 'object') {
       return input;
     } else if (Array.isArray(input)) {
-      return input.map(this.toLowerCasePropertyNames);
+        if (input.length) {
+            return input.map(this.toLowerCasePropertyNames);
+        } else {
+            return input;
+        }
+      
     }
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     return Object.keys(input).reduce((result: any, key: string) => {
--- a/src/meetingsession/MeetingSessionConfiguration.ts
+++ b/src/meetingsession/MeetingSessionConfiguration.ts
@@ -173,12 +173,11 @@ export default class MeetingSessionConfiguration {
     } else if (typeof input !== 'object') {
       return input;
     } else if (Array.isArray(input)) {
-        if (input.length) {
-            return input.map(this.toLowerCasePropertyNames);
-        } else {
-            return input;
-        }
-      
+      if (input.length) {
+        return input.map(this.toLowerCasePropertyNames);
+      } else {
+        return input;
+      }
     }
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     return Object.keys(input).reduce((result: any, key: string) => {
--- a/test/meetingsession/MeetingSessionConfiguration.test.ts
+++ b/test/meetingsession/MeetingSessionConfiguration.test.ts
@@ -299,5 +299,31 @@ describe('MeetingSessionConfiguration', () => {
       expect(configuration.credentials.attendeeId).to.eq('attendee-id');
       expect(configuration.credentials.joinToken).to.eq('join-token');
     });
+
+    it('an attribute contains empty array', () => {
+        const configuration = new MeetingSessionConfiguration(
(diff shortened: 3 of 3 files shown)


ltrung:
@narek-king Thank for fixing this. Do you mind updating the unit test for the case that causes the crash here https://github.com/aws/amazon-chime-sdk-js/blob/master/test/meetingsession/MeetingSessionConfiguration.test.ts ?

narek-king:
@ltrung I fixed the unit tests. The CI fails and I am not sure why. Can you please review it?

ltrung:
> @ltrung I fixed the unit tests. The CI fails and I am not sure why. Can you please review it?

@narek-king  currently forked PR would not be able to trigger our integration test as it cannot access GitHub secret. We are working on fixing it.
Regarding your PR, could you clarify what the issue is? I could run the unit test you added successfully without the fix. [].map should not cause any issue.

narek-king:
> > @ltrung I fixed the unit tests. The CI fails and I am not sure why. Can you please review it?
> 
> @narek-king currently forked PR would not be able to trigger our integration test as it cannot access GitHub secret. We are working on fixing it.
> Regarding your PR, could you clarify what the issue is? I could run the unit test you added successfully without the fix. [].map should not cause any issue.

@ltrung I agree that [].map should not cause any issue but I have a real case that it didn't work. This fixed my problem and I am currently using the SDK from my forked repo. In my opinion, the problem is not related to [].map but more of recursively calling the `toLowerCasePropertyName` method. In the scope, the reference of the method is not accessible. I was thinking to refactor to call as a static method but in this case, it doesn't make sense. 
I was thinking of another case where the input is [[]] but didn't have time to check.
image
I am not sure why this is hapenning^

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `attendeePresenceReceived` meeting event.
 - Add the file sharing workaround for Chrome 88 in FAQs.
 - Add support for Chrome for iOS and Firefox for iOS.
+- Fix Cannot read property 'toLowerCasePropertyNames' of undefined
```
richnew10, line 18: This needs to move to the 'Unreleased' section.

src/meetingsession/MeetingSessionConfiguration.ts
```diff
@@ -173,7 +173,11 @@ export default class MeetingSessionConfiguration {
     } else if (typeof input !== 'object') {
       return input;
     } else if (Array.isArray(input)) {
-      return input.map(this.toLowerCasePropertyNames);
+      if (input.length) {
+        return input.map(this.toLowerCasePropertyNames);
+      } else {
+        return input;
+      }
```
richnew10, line 180: I'm not entirely sure what you're trying to fix here, but I suspect the correct thing to do is to move `toLowerCasePropertyNames` out as a free function. Even with your change this use of `this` is incorrect.

richnew10 (changes requested):
See inline.

nainkunal933:
> > > @ltrung I fixed the unit tests. The CI fails and I am not sure why. Can you please review it?
> > 
> > 
> > @narek-king currently forked PR would not be able to trigger our integration test as it cannot access GitHub secret. We are working on fixing it.
> > Regarding your PR, could you clarify what the issue is? I could run the unit test you added successfully without the fix. [].map should not cause any issue.
> 
> @ltrung I agree that [].map should not cause any issue but I have a real case that it didn't work. This fixed my problem and I am currently using the SDK from my forked repo. In my opinion, the problem is not related to [].map but more of recursively calling the `toLowerCasePropertyName` method. In the scope, the reference of the method is not accessible. I was thinking to refactor to call as a static method but in this case, it doesn't make sense.
> I was thinking of another case where the input is [[]] but didn't have time to check.
> image
> I am not sure why this is hapenning^

Hi @narek-king, thank you for opening this PR. I have looked through your code fix. It seems like you are checking for array length before calling `return input.map(this.toLowerCasePropertyNames);`

The reason why the code is throwing undefined error here is because of `Array.map()`'s behavior. Here is quote from the methods implementation.

>“If a thisArg parameter is provided, it will be used as callback’s this value. Otherwise, the value undefined will be used as its this value. The this value ultimately observable by callbackFn is determined according to the usual rules for determining the this seen by a function.”

If I understand correctly, this is why @richnew10 is referring to the usage of `this` as incorrect in the [comment above](https://github.com/aws/amazon-chime-sdk-js/pull/1072#discussion_r579714994).

One way to fix this will be to pass `this` as a param to the map method, like this:
Screen Shot 2021-09-09 at 12 05 25 PM

Please refer to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) for more info on this optional parameter.

I will be raising a PR soon to fix this by making the method static. I have updated the issue #1052 with my comment. Please let me know if you have any questions.

nainkunal933:
> @ltrung I fixed the unit tests. The CI fails and I am not sure why. Can you please review it?

@narek-king The CI is failing because external PRs that are coming from a fork do not have access to GitHub secrets. So, you will see that the integration tests are not running. Currently, we do not have a fix for this issue and we plan to resolve this in the future.

Because the tests will not run we will not be able to merge this PR. I am sorry for this inconvenience. I want to thank you for taking the time to open this PR. Your PR was very helpful in understanding the issue. Please refer to the issue #1052 as I will keep you updated there and feel free to review my PR 😄

Issue #1029, opened 2021-01-26
Reconnect: Video Receive Fails when Mobile Peer Reconnects


tylerlot:
Wanted to report a bug for the mobile client’s video reconnection logic between the mobile and web sdk. When we force the mobile client to reconnect to a meeting (toggling mobile device WiFi/LTE off/on, no web changes), we notice that after the mobile client reconnects, the mobile-video-send is not received by the web client (web client sees audioVideoTileWasRemoved fires after reconnection). The mobile local video preview remains intact. We are able to resolve this state by manually restarting the mobile client’s video feed. After reconnection there is no problem with the audio feed, this always resolves successful. The issue does not occur if the party who is “reconnecting” to the meeting is the web client.

**Meeting and attendee IDs:**
e289d41b-f0a2-072f-832e-0e12b8256d19 ::: desktop user attendee
6dc2e4ab-2670-82aa-49e8-81d03cf773ae ::: mobile user attendee
[Serverless demo logs.txt](https://github.com/aws/amazon-chime-sdk-js/files/5875214/Serverless.demo.logs.txt)
[Mobile logs.txt](https://github.com/aws/amazon-chime-sdk-js/files/5875215/Mobile.logs.txt)

**Describe the bug**
The bug appears as the web client not receiving any video from the mobile client after the mobile client goes through reconnection. Not sure if the issue is on the web or mobile side, so cutting across platforms.

We have been able to reproduce this issue on both iOS and Android platforms, and on separate devices.
- iOS issue: aws/amazon-chime-sdk-ios#221
- Android issue: aws/amazon-chime-sdk-android#222

We noticed the mobile app SDP contains 2 video tracks (the initial track, and a new one being offer). This is followed by logs indicating the web client receives a track event, but does not add the track as this attendee already has a tileId (logs attached). When the user attempts to workaround the bug by restarting their video, we see a log in the web client again that the stream does not belong to a known attendee. We believe the duplicate tracks to be the source of the issue here.

[TwoTrackSDPFromMobile.txt](https://github.com/aws/amazon-chime-sdk-js/files/5875255/TwoTrackSDPFromMobile.txt)

**Screenshots**
If applicable, add screenshots to help explain the problem.

**Platform**
 - OS: Mac
 - Browser: Chrome 
 - Browser Version: Version 88.0.4324.96 (Official Build) (x86_64)
 - Amazon Chime SDK GitHub Commit: 1.19

simmkyu:
[...]

Issue #561, opened 2020-07-16
For remote attendees - Video/ Screen Stream frozen for disconnected attendee during call


ankit-maheshwari:
**Describe the bug**
If an attendee is disconnected from the internet during the call, while his/her video and screen share stream is enabled. It shows frozen video streams for both video & screen share on the remote attendee's screens.  The video tiles of the disconnected attendee don't update for the remote attendees and remain on the screen.

**Platform**
 - OS: Windows, MacOS
 - Browser: Chrome/ Firefox
 - Browser Version: Latest
 - Amazon Chime SDK GitHub Commit: 1.7.0

simmkyu:
Thanks for reporting.

Repro steps in the Chime SDK demo
1. UserA opens the SDK demo in the iOS Safari and joins a meeting.
2. UserB opens the SDK demo in the desktop Chrome and joins the same meeting.
3. UserA and UserB enable local videos.
4. UserB turn Wi-Fi off in the computer.
5. In the UserA's demo, UserB's video immediately freezes. But UserB still exists in the roster.
6. After several seconds, UserB disappears from the roster in the UserA's demo. the frozen UserB video still exists.

dmtux:
Hello Dear AWS Team,

Any updates regarding this issue?

Thanks,
Denis

efortney:
Hi AWS team, we are seeing this problem as well. I have been able to reproduce by using Network Link Conditioner.

tarunkolla:
Any update on this? We are having similar issues with screen share and I could reproduce this on the demo app although it wasn't consistent

edwardbc:
Still happening on +2.0.0 , mostly on two user sessions. I wasn't able to replicate it on more three users or more. 

Another problem I noticed with this bug, is that if in some cases when UserB comes back, UserA is not able to see him due the conflict created by the frozen video tile, even if it already dissapeared.

rachinkapoor:
Do we have ETA for the fix?
Is there any workaround? 
Eg. Can we listen to any events or watch matrix to detect the frozen screen and take some steps to fix it?

p-foucht:
Some users have been referencing this thread in regards to black tiles, so just to clarify -- there are two issues here

1. The frozen videos caused by network reconnection. We are in the process of investigating this and will provide updates here

[...]

PR #1073, opened 2021-02-17
Ignore old group IDs which may correspond to a disconnected client

hensmi-amazon:
**Issue #:** https://github.com/aws/amazon-chime-sdk-ios/issues/222 https://github.com/aws/amazon-chime-sdk-android/issues/222 https://github.com/aws/amazon-chime-sdk-js/issues/1029 & partially https://github.com/aws/amazon-chime-sdk-js/issues/561 (dupes in other repos), linking so they update issue.

**Description of changes:**
Copying comment here:
```
    // In the Amazon Chime SDKs, we assume a one to one mapping of group ID to profile ID when creating
    // video tiles (multiple video sources are supported through applying a `Modality` to a given profile/session token)
    //
    // We enforce this here to mitigate any possible duplicate group IDs left from a reconnection where the previous
    // signal channel hasn't been timed out yet.  To guarantee we receive the latest stream we use the highest group ID
    // since they are monotonically increasing.
```

**Testing**

Yes

I tested by using demo client to join a meeting.  I then used two modified native clients to join the meeting with the same attendee ID, both trying to send video.  I confirmed that the JS SDK client only subscribed to the newest group ID.

Yes, the browser demo.

No

No

### DIFF
--- a/src/videostreamindex/DefaultVideoStreamIndex.ts
+++ b/src/videostreamindex/DefaultVideoStreamIndex.ts
@@ -81,11 +81,14 @@ export default class DefaultVideoStreamIndex implements VideoStreamIndex {
     // signal channel hasn't been timed out yet.  To guarantee we receive the latest stream we use the highest group ID
     // since they are monotonically increasing.
     const attendeeIdToMainGroupIdMap = new Map<string, number>();
-    this.currentIndex.sources.forEach(source => {
+    // Improve performance by not filtering sources unless
+    // we know the list will actually change
+    let attendeeWithMultipleGroupIdsExists = false;
+    for (const source of indexFrame.sources) {
       if (!attendeeIdToMainGroupIdMap.has(source.attendeeId)) {
         // We haven't see this attendee ID so just keep track of it
         attendeeIdToMainGroupIdMap.set(source.attendeeId, source.groupId);
-        return;
+        continue;
       }
 
       // Otherwise see if we should use the group ID corresponding to this source (we prefer the highest for each attendee)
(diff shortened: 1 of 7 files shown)


src/videostreamindex/DefaultVideoStreamIndex.ts
```diff
@@ -73,6 +73,34 @@ export default class DefaultVideoStreamIndex implements VideoStreamIndex {
...
+      if (currentGroupId < source.groupId) {
+        this.logger.warn(
+          `Old group ID ${currentGroupId} found for attendee ID ${source.attendeeId}, replacing with newer group ID ${source.groupId}`
+        );
+        attendeeIdToMainGroupIdMap.set(source.attendeeId, source.groupId);
+      }
+    });
+    this.currentIndex.sources = this.currentIndex.sources.filter(
+      source => attendeeIdToMainGroupIdMap.get(source.attendeeId) === source.groupId
+    );
```
richnew10, line 102: So for every index frame you're:

* Allocating a new `Map` and a closure.
* Walking all of the sources, running that closure on each.
* Walking the sources again, allocating a new array.
* Discarding the map and all of that garbage.

How big is `sources`? How likely are duplicates? How frequently is this function called?

hensmi-amazon: `sources` is number of source videos * simulcast streams (e.g. 16*2, perhaps).  Duplicates are not common, only in reconnections as noted in github issues.  This function is called when streams change, i.e. someone enabling or disabling video, which should not be common.

richnew10: OK, that makes sense.

Timings for me in Firefox, using 100K iterations:

* Your code: 1151ms
* Using `for (const source of input.sources)` instead of `forEach`: 889ms
* Tracking whether any dupes were seen, avoiding the `filter` if unnecessary: 132ms

Just add

```typescript
    let anyDupes = false;
…
      if (currentGroupId < source.groupId) {
        attendeeIdToMainGroupIdMap.set(source.attendeeId, source.groupId);
      }
      anyDupes = true;
…
      if (anyDupes) {
…        
```

hensmi-amazon: Thanks for the suggestion!  I integrated those changes, though I think in your test the `return` in the closure may have been replaced by a `break` instead of a `continue`.  I still see a solid 20-30% improvement though!

XHatan (approved):

richnew10 (approved):

PR #1078, opened 2021-02-19
Add string as a parameter for checker APIs

ltrung:
**Description of changes:**
- Add string as an acceptable parameter in checker APIs to allow passing in just the deviceId.

**Testing**

2. How did you test these changes? Manually verify 
Yes, this can be verified by running the meeting readiness checker demo.
4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? Yes, we modify checker APIs to allow string to be passed in. Checked with the team.


src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
```diff
@@ -44,7 +44,9 @@ export default class DefaultMeetingReadinessChecker implements MeetingReadinessC
     private configuration: MeetingReadinessCheckerConfiguration = new MeetingReadinessCheckerConfiguration()
   ) {}
 
-  async checkAudioInput(audioInputDeviceInfo: MediaDeviceInfo): Promise<CheckAudioInputFeedback> {
+  async checkAudioInput(
+    audioInputDeviceInfo: MediaDeviceInfo | string
+  ): Promise<CheckAudioInputFeedback> {
```
richnew10, line 49: Just change it to

```
async checkAudioInput(input: Device)
```

Any caller who was providing a `MediaDeviceInfo` was actually providing a `MediaTrackConstraints` instance, so let's just be accurate about it.

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
```diff
@@ -59,15 +61,17 @@ export default class DefaultMeetingReadinessChecker implements MeetingReadinessC
   }
 
   async checkAudioOutput(
-    audioOutputDeviceInfo: MediaDeviceInfo,
+    audioOutputDeviceInfo: MediaDeviceInfo | string,
```
richnew10, line 64: Again.

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
```diff
@@ -59,15 +61,17 @@ export default class DefaultMeetingReadinessChecker implements MeetingReadinessC
...
   ): Promise<CheckAudioOutputFeedback> {
     try {
-      const audioOutputDeviceId =
-        audioOutputDeviceInfo && audioOutputDeviceInfo.deviceId
-          ? audioOutputDeviceInfo.deviceId
-          : '';
+      let audioOutputDeviceId = '';
+      if (typeof audioOutputDeviceInfo === 'string') {
+        audioOutputDeviceId = audioOutputDeviceInfo;
+      } else if (audioOutputDeviceInfo && audioOutputDeviceInfo.deviceId) {
```
richnew10, line 72: Extract `DefaultDeviceController.getIntrinsicDeviceIdStr` and use that instead.

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
```diff
@@ -162,7 +168,7 @@ export default class DefaultMeetingReadinessChecker implements MeetingReadinessC
   }
 
   async checkCameraResolution(
-    videoInputDevice: MediaDeviceInfo,
+    videoInputDevice: MediaDeviceInfo | string,
```
richnew10, line 171: Consider whether changing this to `Device` makes sense — after all, in this function you're constructing constraints and then getting a stream, and the other kinds of `Device` are exactly those two things!

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+- Add string as a parameter for checker APIs
```
richnew10, line 9: > Allow device checker APIs to take devices as input, rather than only `MediaDeviceInfo` objects.

demos/browser/app/meetingReadinessChecker/meetingReadinessChecker.ts
```diff
@@ -181,35 +181,19 @@ export class DemoMeetingApp {
     }
   }
 
-  getAudioInputDevice = async (): Promise<MediaDeviceInfo> => {
-    const audioInputDevices = await this.deviceController.listAudioInputDevices();
+  getAudioInputDevice = async (): Promise<string> => {
```
richnew10, line 184: This all gets a lot neater if you make this `Promise<Device>` — or even `Promise<AudioInputDevice>`, if you can swing that.

richnew10 (changes requested):

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
ltrung: I actually used Device first then realized that Device also contains MediaStream which doesn't work with all check here (like the Camera one) so I just do string instead.

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
ltrung: ChooseAudioOutput API actually only accepts string and not Device. We do not use it here but using string would be consistent

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
ltrung: Good call. Will do it :)

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
richnew10: Why would we not allow testing a stream?

demos/browser/app/meetingReadinessChecker/meetingReadinessChecker.ts
ltrung: Change to Device

CHANGELOG.md
ltrung: Updated

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
ltrung: I decided to stick with just MediaDeviceInfo | string since this API is to specifically check for a camera device.

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
ltrung: Update to Device

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1273,7 +1276,7 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
 
     const settings = stream.getTracks()[0].getSettings();
     // If a device does not specify deviceId, we have to assume the stream is not reusable.
-    return settings.deviceId === this.getIntrinsicDeviceIdStr(device);
+    return settings.deviceId === DefaultDeviceController.getIntrinsicDeviceIdStr(device);
```
richnew10, line 1279: There might have been a bug here!

richnew10 (changes requested):

demos/browser/app/meetingReadinessChecker/meetingReadinessChecker.ts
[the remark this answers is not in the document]
ltrung: Updated

src/devicecontroller/DefaultDeviceController.ts
ltrung: Add a check to make sure settings.deviceId is not undefined, null, or ''.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -552,6 +552,33 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+    if (typeof device === 'string') {
+      return device;
+    }
+    if ((device as MediaStream).id) {
+      return (device as MediaStream).id;
+    }
+
+    const constraints: MediaTrackConstraints = device as MediaTrackConstraints;
+    if (!constraints.deviceId) {
+      return '';
```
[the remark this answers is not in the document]
ltrung: Update to return undefined

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -552,6 +552,33 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+    if (typeof constraints.deviceId === 'string') {
+      return constraints.deviceId;
+    }
+
+    const deviceIdConstraint: ConstrainDOMStringParameters = constraints.deviceId as ConstrainDOMStringParameters;
+    if (typeof deviceIdConstraint.exact === 'string') {
+      return deviceIdConstraint.exact;
+    }
+
+    return '';
```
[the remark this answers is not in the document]
ltrung: Updated.

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
```diff
@@ -194,18 +194,19 @@ export default class DefaultMeetingReadinessChecker implements MeetingReadinessC
   }
 
   private calculateVideoConstraint(
-    videoInputDevice: MediaDeviceInfo,
+    videoInputDevice: MediaDeviceInfo | string,
```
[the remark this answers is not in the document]
ltrung: Update to only take string as input.

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
ltrung: When we design this API, I do not think we have that use cases in mind (The name camera resolution means specifically to a device). What you mentioned is a valid use case but I think we may want to create a different API for it.

demos/browser/app/meetingReadinessChecker/meetingReadinessChecker.ts
```diff
@@ -181,35 +181,19 @@ export class DemoMeetingApp {
     }
   }
 
-  getAudioInputDevice = async (): Promise<MediaDeviceInfo> => {
-    const audioInputDevices = await this.deviceController.listAudioInputDevices();
+  getAudioInputDeviceID = async (): Promise<string> => {
```
richnew10, line 184: These functions no longer needs to return `Promise`s — they're just DOM lookups.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -552,6 +552,44 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+    }
+
+    if (typeof constraints.deviceId === 'string') {
+      return constraints.deviceId;
+    }
+
+    const deviceIdConstraint: ConstrainDOMStringParameters = constraints.deviceId as ConstrainDOMStringParameters;
+    if (typeof deviceIdConstraint.exact === 'string') {
+      return deviceIdConstraint.exact;
+    }
```
richnew10, line 588: Strictly speaking, `.deviceId` could be `instanceof Array`, and with this code it will fall through to the `undefined` case. You should probably handle that explicitly.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1273,7 +1287,10 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
 
     const settings = stream.getTracks()[0].getSettings();
     // If a device does not specify deviceId, we have to assume the stream is not reusable.
-    return settings.deviceId === this.getIntrinsicDeviceIdStr(device);
+    return (
+      settings.deviceId &&
+      settings.deviceId === DefaultDeviceController.getIntrinsicDeviceId(device)
+    );
```
richnew10, line 1293: Nice. Was there a bug hiding here?

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
```diff
@@ -59,15 +60,14 @@ export default class DefaultMeetingReadinessChecker implements MeetingReadinessC
...
     audioElement: HTMLAudioElement = null
   ): Promise<CheckAudioOutputFeedback> {
     try {
-      const audioOutputDeviceId =
-        audioOutputDeviceInfo && audioOutputDeviceInfo.deviceId
-          ? audioOutputDeviceInfo.deviceId
-          : '';
+      const audioOutputDeviceId = audioOutputDeviceInfo
+        ? DefaultDeviceController.getIntrinsicDeviceId(audioOutputDeviceInfo)
+        : '';
```
richnew10, line 70: I think you want `null` here.

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
richnew10: That's fair!

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -2106,4 +2106,38 @@ describe('DefaultDeviceController', () => {
...
+
+    it('Return string if the constraint deviceId is of type string', () => {
+      const constraints: MediaTrackConstraints = { deviceId: '' };
+      expect(DefaultDeviceController.getIntrinsicDeviceId(constraints)).to.equal('');
+    });
+
+    it('Return string if the constraint deviceId is of type ConstrainDOMStringParameters', () => {
+      const constraints: MediaTrackConstraints = { deviceId: { exact: 'device1' } };
+      expect(DefaultDeviceController.getIntrinsicDeviceId(constraints)).to.equal('device1');
+    });
```
richnew10, line 2141: You'll want to add tests for the array versions here, and probably at least one non-trivial constraint — things like `{sampleRate: 44100}`.

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -2106,4 +2106,38 @@ describe('DefaultDeviceController', () => {
...
+      expect(DefaultDeviceController.getIntrinsicDeviceId(undefined)).to.be.undefined;
+    });
+
+    it('Return null if the device is null', () => {
+      expect(DefaultDeviceController.getIntrinsicDeviceId(null)).to.be.null;
+    });
+
+    it('Return string if the device is of type string', () => {
+      expect(DefaultDeviceController.getIntrinsicDeviceId('')).to.equal('');
+    });
```
richnew10, line 2121: Also test `'default'` and a real device ID value.

richnew10 (changes requested):

src/devicecontroller/DefaultDeviceController.ts
ltrung: Tbh, I am not sure how to make deviceId from mediaTrackSettings to be null, undefined, or empty string from the demo app but yes in theory it can be any of those value so it is better to guard against it.

test/devicecontroller/DefaultDeviceController.test.ts
ltrung: Updated.

test/devicecontroller/DefaultDeviceController.test.ts
ltrung: Updated

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
ltrung: The previous value is also '' and I check bindAudioMix, it also use ''.

src/devicecontroller/DefaultDeviceController.ts
ltrung: Updated.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1271,9 +1298,20 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+    const settings = this.getMediaTrackSettings(stream);
     // If a device does not specify deviceId, we have to assume the stream is not reusable.
-    return settings.deviceId === this.getIntrinsicDeviceIdStr(device);
+    if (!settings.deviceId) {
+      return false;
+    }
+    const deviceIds = DefaultDeviceController.getIntrinsicDeviceId(device);
+    if (Array.isArray(deviceIds)) {
+      return deviceIds.includes(settings.deviceId);
+    }
```
richnew10, line 1309: Probably not unless it's a single-element array — this check determines whether we can apply new constraints to the input, and that won't necessarily be true. It's safer to just return `false` here for a multi-element array.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1011,7 +1037,8 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
           }
         });
       }
-      newDevice.groupId = this.getGroupIdFromDeviceId(kind, this.getIntrinsicDeviceIdStr(device));
+      const groupId = this.getMediaTrackSettings(newDevice.stream)?.groupId;
+      newDevice.groupId = groupId ? groupId : '';
```
richnew10, line 1041: `groupId || ''` is equivalent.

src/devicecontroller/DefaultDeviceController.ts
ltrung: Update to return false for Array in both hasSameGroupId and isMediaStreamReusableByDeviceId

src/devicecontroller/DefaultDeviceController.ts
ltrung: Updated

richnew10 (approved):

PR #1079, opened 2021-02-20
Update typedoc to fix a security vulnerability

devalevenkatesh:
**Issue #:**
Fix a security vulnerability with `marked` library. We depend on `typedoc` for our doc generation which depends on `marked`. `typedoc's` latest version fixes the security vulnerability with `marked` hence we need to bump our `typedoc` from `0.19.2` to `0.20.25`.

**Description of changes:**
- Updated `typedoc` to fix a security vulnerability associated with the `marked` library.
- Followed the migration guide under `0.20.0` release from [here](https://github.com/TypeStrong/typedoc/releases?after=v0.20.1) and the typedoc [options](https://typedoc.org/guides/options/) guide.

**Testing**

- By opening up the generated docs using `open index.html` after `npm run build` and `npm run doc`.
- There are some decorative changes I found since we bumped `typedoc` from `0.19.2` to `0.20.25`. But, all the documentation content looked similar to what was generated using the `typedoc` `0.19.2` version.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? No. This is a documentation package update.


ltrung:
Do we want to update all the doc filenames?

ltrung:
since we could specify typedoc config inside tsconfig.js, can we get rid of typedoc.js (and maybe the one inside guides folder)?

devalevenkatesh:
> since we could specify typedoc config inside tsconfig.js, can we get rid of typedoc.js (and maybe the one inside guides folder)?

Yup, so I have removed `typedoc.json` and added those options under `guides/tsconfig.json` under `typedocOptions` key. Also, we need two separate `tsconfig.json`s since our `typedoc` uses different set of keys than the build process keys.

devalevenkatesh:
> Do we want to update all the doc filenames?

According to new typedoc version they removed the support for "mode" option and introduced "entryPoints" option. If we have multiple entry points (our case), we get the module name appended to the filenames. I have added both "src/index.ts" and "guides/index.ts" as entry points but with custom module names. The exports from "src/index.ts" will be shown under "amazon-chime-sdk-js" module and similarly exports from "guides/index.ts" will show under "Guides".

script/generate-docs
```diff
@@ -19,4 +19,4 @@ STDERR.puts "Generating docs against #{remote}/#{branch}."
 # Explicitly use the guides/tsconfig.json file, because it includes
 # both the source and docs. The build itself just uses the root tsconfig,
 # which just hits src.
-`node node_modules/typedoc/bin/typedoc --tsconfig guides/tsconfig.json --options typedoc.json --gitRemote #{remote} --gitRevision #{branch}`
+`node node_modules/typedoc/bin/typedoc --tsconfig guides/tsconfig.json --gitRemote #{remote} --gitRevision #{branch}`
```
ltrung, line 22: For the library src folder does it recognize it automatically since you only specific the one from the guides folder here?

devalevenkatesh: The library src folder is picked from the typedoc options in the guides/tsconfig.json, we have specified two entry points "src/index.ts" and "guides/index.ts". Thats how it recognizes the two.

script/build-guides.js
```diff
@@ -46,9 +47,17 @@ walk('guides')
...
 
-fs.writeFileSync('./guides/docs.ts', result, 'utf8');
+let namespaceExports = `\nexport {\n ${namespaces.join(',\n ')}\n}`;
+fs.writeFileSync('./guides/docs.ts', `${result+namespaceExports};`, 'utf8');
+const customModuleNameAnnotation =
+`/**
+ * @packageDocumentation
+ * @module Guides
+ */`;
+fs.writeFileSync('./guides/index.ts', `${customModuleNameAnnotation}\n${namespaceExports} from './docs';`, 'utf8');
```
ltrung, line 60: What is the difference between index.ts and doc.ts?

devalevenkatesh: To give the "guides/docs.ts" a different custom name like "Guides", I had to export the namespaces that are generated at runtime in the "docs.ts". Once exported I had to re-export them from "index.ts" where I could give a custom module name to that "index.ts". Giving a custom name at the start of the "docs.ts" did not helped as that file complains of not being a module. And I tried various variations but still the name in generated docs remained "guides/docs.ts" and not what I provided. So as Richard suggested to have index.ts and export everything from docs.ts, that helped.

ltrung (approved):

PR #1090, opened 2021-02-24
Allow for retrieving and specifying VoiceFocusDeviceTransformer configurations

richnew10:
A configuration is an opaque blob derived from a spec within a
particular runtime context. The configuration drives exactly
how noise suppression will be applied to an input stream.

Configurations are extremely specific to a point in time, machine
configuration, browser version, and SDK version. They should not be
persisted, transferred between browsers, or mutated.

This commit allows:

* A transformer's configuration to be retrieved via
  `transformer.getConfiguration`
* A transformer to be created with an existing configuration via a new
  argument to `VoiceFocusDeviceTransformer.create`
* A configuration to be created from a spec and options without
  instantiating a transformer via `VoiceFocusDeviceTransformer.configure`.

These new functions allow the work of creating a transformer to be split
or duplicated across execution contexts. For example, you might compute a
configuration from a spec each time your app launches, making the
subsequent initialization faster, or you might reuse a configuration in
a second window by sending an existing transformer's configuration via
`postMessage`.

Creating a configuration via `VoiceFocusDeviceTransformer.configure`
also supports pre-resolving model URLs.

Most customers will not need to use these capabilities, but in some
complex situations they can improve end-to-end latency.

**Testing**

Yes.

Good unit test coverage, and by manually adding the less common flows to the demo app for testing.

We do not have a suitable harness to automate testing of more complex network-facing flows like these.

No.

Yes:

1. `VoiceFocusDeviceTransformer.configure`
2. The additional argument to `.create`
3. The `preResolve` option.

These have not been separately reviewed.

No.


test/utils.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export function delay(timeoutMs: number): Promise<void> {
```
ltrung, line 4: Nice. Someone finally do it :D

ltrung (approved):

PR #1100, opened 2021-02-26
Make logger robust against undefined arguments.

richnew10:
I saw a log from a customer that looked something like:

```
TypeError: debugFunction is not a function
```

which seems like an easy thing to avoid bailing on.

This commit also fixes a few typos and JS bad habits.

**Testing**

Yes.

Unit tests only. The integration tests will run on push.

No.

This broadens the API signature of the POST logger to match Logger.

No.

### DIFF
--- a/src/logger/MeetingSessionPOSTLogger.ts
+++ b/src/logger/MeetingSessionPOSTLogger.ts
@@ -30,11 +30,18 @@ export default class MeetingSessionPOSTLogger {
       });
   }
 
-  debug(debugFunction: () => string): void {
+  debug(debugFunction: string | (() => string)): void {
     if (LogLevel.DEBUG < this.level) {
       return;
     }
-    this.log(LogLevel.DEBUG, debugFunction());
+
+    if (typeof debugFunction === 'string') {
+      this.log(LogLevel.DEBUG, debugFunction);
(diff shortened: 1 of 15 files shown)


src/logger/MeetingSessionPOSTLogger.ts
```diff
@@ -30,11 +30,18 @@ export default class MeetingSessionPOSTLogger {
...
       return;
     }
-    this.log(LogLevel.DEBUG, debugFunction());
+
+    if (typeof debugFunction === 'string') {
+      this.log(LogLevel.DEBUG, debugFunction);
+    } else if (debugFunction) {
+      this.log(LogLevel.DEBUG, debugFunction());
+    } else {
+      this.log(LogLevel.DEBUG, '' + debugFunction);
```
ltrung, line 43: So is this just a ''+null/undefined?

ltrung (approved):

src/logger/MeetingSessionPOSTLogger.ts
richnew10: Yes; if you call `.debug(undefined)`, it's equivalent to calling `.debug('undefined')`.

Issue #1103, opened 2021-02-28
ReceiveAudioInputTask error: undefined is not an object (evaluating 'track.label.indexOf')


TheAnkurPanchani:
**Important:**
INFO-level logs with timestamps.
[AmazonChimeLogs(iOS).txt](https://github.com/aws/amazon-chime-sdk-js/files/6057361/AmazonChimeLogs.iOS.txt)

**Meeting and attendee IDs:**
meeting ID:1b3c7be7-525a-4f3a-b3c6-77ffdd3e7516
attendee ID: 18db81d7-b13a-b410-d332-0a5cda75fdec

**Timeline**
* Just started and joined the meeting.

**Describe the bug**
[Info] 2021-02-28T17:19:31.067Z [INFO] SDK - serial group task AudioVideoStart/1b3c7be7-525a-4f3a-b3c6-77ffdd3e7516/18db81d7-b13a-b410-d332-0a5cda75fdec was canceled due to subtask AudioVideoStart/1b3c7be7-525a-4f3a-b3c6-77ffdd3e7516/18db81d7-b13a-b410-d332-0a5cda75fdec/ReceiveAudioInputTask error: undefined is not an object (evaluating 'track.label.indexOf')

**Screenshots**
No Screenshots.

**Platform**
 - OS: iOS 14.4
 - Browser: ios-webview
 - Browser Version: 605 (605.1.15)
 - Amazon Chime SDK GitHub Commit: [v2.5.0](https://github.com/aws/amazon-chime-sdk-js/releases/tag/v2.5.0)

michhyun1:
@TheAnkurPanchani Hi there, could you specify what device you are using? We'll try to reproduce this on our end and give you an update. Thanks

michhyun1:
@TheAnkurPanchani I noticed that you are using 'ios-webview' as your browser, which is currently not a platform that we are supporting right now, hence the issue you are seeing. I've taken note of this, and I've created a task on our end to add support for WKWebView browsers.

TheAnkurPanchani:
[...]

PR #1106, opened 2021-03-01
Remove audioDeviceInformation state bug fix

michhyun1:
**Issue #:**

**Description of changes:**
Remove audioDeviceInformation, since it is unused and fixes a bug.

**Testing**

2. How did you test these changes? Run integration tests, unit tests.


src/audiovideocontroller/AudioVideoControllerState.ts
```diff
@@ -110,8 +110,6 @@ export default class AudioVideoControllerState {
 
   videoInputAttachedTimestampMs: number = 0;
 
-  audioDeviceInformation: { [id: string]: string } = {};
```
XHatan, line 113: we do publish video device information through event controller.  I am more curious about how to trigger the track label undefined condition ? Any idea ?

michhyun1: This was due to a bug filed here: https://github.com/aws/amazon-chime-sdk-js/issues/1103. Looks like the device label is undefined

devalevenkatesh (approved):

XHatan (approved):

guides/index.ts
```diff
@@ -0,0 +1,17 @@
+/**
```
ltrung, line 1: Is this intentional? Doesn't seem related.

PR #1108, opened 2021-03-01
[Demo] add segmentation demo for video processor

XHatan:
**Issue #:**

**Description of changes:**

add segmentation demo for video processor

**Testing**

2. How did you test these changes? yes
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? browser demo
4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? NA

### DIFF
--- a/README.md
+++ b/README.md
@@ -1083,6 +1083,11 @@ The use of Amazon Voice Focus via this SDK involves the downloading and executio
 
 The use of Amazon Voice Focus runtime code is subject to additional notices. See [this NOTICES file](https://static.sdkassets.chime.aws/workers/NOTICES.txt) for details. You agree to make these additional notices available to all end users who use Amazon Voice Focus runtime code via this SDK.
 
+
+The browser demo applications in the [demos directory](https://github.com/aws/amazon-chime-sdk-js/tree/master/demos) use [TensorFlow.js](https://github.com/tensorflow/tfjs) and pre-trained [TensorFlow.js models](https://github.com/tensorflow/tfjs-models) for image segmentation. Use of these third party models involves downloading and execution of code at runtime from [jsDelivr](https://www.jsdelivr.com/) by end user browsers. For the jsDelivr Acceptable Use Policy, please visit this [link](https://www.jsdelivr.com/terms/acceptable-use-policy-jsdelivr-net).
+
+The use of TensorFlow runtime code referenced above may be subject to additional license requirements. See the licenses page for TensorFlow.js [here](https://github.com/tensorflow/tfjs/blob/master/LICENSE) and TensorFlow.js models [here](https://github.com/tensorflow/tfjs-models/blob/master/LICENSE) for details.
+
 ---
 
 Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -52,6 +52,11 @@ import {
 
 import CircularCut from './videofilter/CircularCut';
 import EmojifyVideoFrameProcessor from './videofilter/EmojifyVideoFrameProcessor';
+import SegmentationProcessor from './videofilter/SegmentationProcessor';
+import {
+  loadBodyPixDependency,
+  platformCanSupportBodyPixWithoutDegradation,
+} from './videofilter/SegmentationUtil';
 import WebRTCStatsCollector from './webrtcstatscollector/WebRTCStatsCollector';
 
 const SHOULD_DIE_ON_FATALS = (() => {
@@ -126,7 +131,7 @@ const VOICE_FOCUS_SPEC = {
   paths: VOICE_FOCUS_PATHS,
 };
 
-type VideoFilterName = 'Emojify' | 'CircularCut' | 'NoOp' | 'None';
+type VideoFilterName = 'Emojify' | 'CircularCut' | 'NoOp' | 'Segmentation' | 'None';
 
 const VIDEO_FILTERS: VideoFilterName[] = ['Emojify', 'CircularCut', 'NoOp'];
 
@@ -221,6 +226,11 @@ export class DemoMeetingApp
   static readonly DATA_MESSAGE_TOPIC: string = 'chat';
   static readonly DATA_MESSAGE_LIFETIME_MS: number = 300000;
 
+
+  // ideally we don't need to change this. Keep this configurable in case users have super slow network.
+  loadingBodyPixDependencyTimeoutMs: number = 10000;
+  loadingBodyPixDependencyPromise: undefined | Promise<void>;
+
   showActiveSpeakerScores = false;
   activeSpeakerLayout = true;
   meeting: string | null = null;
@@ -308,7 +318,6 @@ export class DemoMeetingApp
       this.fatal(new Error('Testing fatal.'));
       return;
     }
-
     (document.getElementById('sdk-version') as HTMLSpanElement).innerText =
       'amazon-chime-sdk-js@' + Versioning.sdkVersion;
     this.initEventListeners();
@@ -1605,6 +1614,23 @@ export class DemoMeetingApp
     await this.populateAudioOutputList();
   }
 
+  private async selectVideoFilterByName(name: VideoFilterName): Promise<void> {
+    this.selectedVideoFilterItem = name;
+    this.log(`clicking video filter ${this.selectedVideoFilterItem}`);
+    this.toggleButton(
+      'button-video-filter',
+      this.selectedVideoFilterItem === 'None' ? 'off' : 'on'
+    );
+    if (this.isButtonOn('button-camera')) {
+      try {
+        await this.openVideoInputFromSelection(this.selectedVideoInput, false);
+      } catch (err) {
+        fatal(err);
+        this.log('Failed to choose VideoTransformDevice', err);
+      }
+    }
+  }
+
   private async populateVideoFilterInputList(): Promise<void> {
     const genericName = 'Filter';
     let filters: VideoFilterName[] = ['None'];
@@ -1614,6 +1640,27 @@ export class DemoMeetingApp
       this.enableUnifiedPlanForChromiumBasedBrowsers
     ) {
       filters = filters.concat(VIDEO_FILTERS);
+      if (platformCanSupportBodyPixWithoutDegradation()) {
+        if (!this.loadingBodyPixDependencyPromise) {
+          this.loadingBodyPixDependencyPromise = loadBodyPixDependency(this.loadingBodyPixDependencyTimeoutMs);
+        }
+        // do not use `await` to avoid blocking page loading
+        this.loadingBodyPixDependencyPromise.then(() => {
+          filters.push('Segmentation');
+          this.populateInMeetingDeviceList(
+            'dropdown-menu-filter',
+            genericName,
+            [],
+            filters,
+            undefined,
+            async (name: VideoFilterName) => {
+              await this.selectVideoFilterByName(name);
+            }
+          );
+        }).catch(err => {
+          this.log('Could not load BodyPix dependency', err);
+        });
+      }
     }
 
     this.populateInMeetingDeviceList(
@@ -1623,20 +1670,7 @@ export class DemoMeetingApp
       filters,
       undefined,
       async (name: VideoFilterName) => {
-        this.selectedVideoFilterItem = name;
-        this.log(`clicking video filter ${this.selectedVideoFilterItem}`);
-        this.toggleButton(
-          'button-video-filter',
-          this.selectedVideoFilterItem === 'None' ? 'off' : 'on'
-        );
-        if (this.isButtonOn('button-camera')) {
-          try {
-            await this.openVideoInputFromSelection(this.selectedVideoInput, false);
-          } catch (err) {
-            fatal(err);
-            this.log('Failed to choose VideoTransformDevice', err);
-          }
-        }
+        await this.selectVideoFilterByName(name);
       }
     );
   }
@@ -2054,6 +2088,10 @@ export class DemoMeetingApp
 
     if (videoFilter === 'NoOp') {
       return new NoOpVideoFrameProcessor();
+    }
+
+    if (videoFilter === 'Segmentation') {
+      return new SegmentationProcessor();
     }
 
     return null;
--- a/demos/browser/app/meetingV2/videofilter/SegmentationProcessor.ts
+++ b/demos/browser/app/meetingV2/videofilter/SegmentationProcessor.ts
@@ -1,0 +1,78 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// eslint-disable-next-line
+declare var bodyPix: any;
+
+import {
+  CanvasVideoFrameBuffer,
+  VideoFrameBuffer,
+  VideoFrameProcessor,
+} from 'amazon-chime-sdk-js';
+
+/**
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -1076,6 +1076,10 @@ The use of Amazon Voice Focus via this SDK involves the downloading and executio
 
 The use of Amazon Voice Focus runtime code is subject to additional notices. See [this NOTICES file](https://static.sdkassets.chime.aws/workers/NOTICES.txt) for details. You agree to make these additional notices available to all end users who use Amazon Voice Focus runtime code via this SDK.
 
+The browser demo applications in the demos directory (https://github.com/aws/amazon-chime-sdk-js/tree/master/demos) use TensorFlow.js (https://github.com/tensorflow/tfjs) and pre-trained TensorFlow.js models (https://github.com/tensorflow/tfjs-models) for image segmentation. Use of these third party models involves downloading and execution of code at runtime from jsDelivr (https://www.jsdelivr.com/) by end user browsers. For the jsDelivr Acceptable Use Policy, please visit this link (https://www.jsdelivr.com/terms/acceptable-use-policy-jsdelivr-net).
+
+The use of TensorFlow runtime code referenced above may be subject to additional license requirements. See the licenses page for TensorFlow.js (https://github.com/tensorflow/tfjs/blob/master/LICENSE) and TensorFlow.js models (https://github.com/tensorflow/tfjs-models/blob/master/LICENSE)for details.
```
richnew10, line 1081: Nit: space before 'for' at the end of the line.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -52,6 +52,11 @@ import {
 
 import CircularCut from './videofilter/CircularCut';
 import EmojifyVideoFrameProcessor from './videofilter/EmojifyVideoFrameProcessor';
+import SegmentationProcessor from './videofilter/SegmentationProcessor';
+import {
+  loadBodyPixDependency,
+  platformCanSupportBodyPixWithoutDegradation
```
richnew10, line 58: Nit: comma at end of line.

demos/browser/app/meetingV2/videofilter/SegmentationProcessor.ts
```diff
@@ -0,0 +1,74 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// eslint-disable-next-line
+declare var bodyPix: any;
+
+import { CanvasVideoFrameBuffer, VideoFrameBuffer, VideoFrameProcessor } from 'amazon-chime-sdk-js';
```
richnew10, line 7: Multi-line, trailing commas.

demos/browser/app/meetingV2/videofilter/SegmentationProcessor.ts
```diff
@@ -0,0 +1,74 @@
...
+      // update target canvas size to match the frame size
+      this.targetCanvas.width = this.sourceWidth;
+      this.targetCanvas.height = this.sourceHeight;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let predictions = await this.model.segmentPerson(inputCanvas);
+    if (predictions) {
+      const foregroundColor = { r: 255, g: 255, b: 255, a: 255 };
+      const backgroundColor = { r: 0, g: 0, b: 0, a: 255 };
```
richnew10, line 56: Do these need to be new objects each time, or can you lift these up to a const?

demos/browser/app/meetingV2/videofilter/SegmentationProcessor.ts
```diff
@@ -0,0 +1,74 @@
...
+      this.mask = bodyPix.toMask(predictions, foregroundColor, backgroundColor, true);
+    }
+
+    if (this.mask) {
+      bodyPix.drawMask(this.targetCanvas, inputCanvas as HTMLCanvasElement, this.mask);
```
richnew10, line 61: Why do you need the cast? `.asCanvasElement()` returns `HTMLCanvasElement | null`, and you exclude `null` on line 33.

demos/browser/app/meetingV2/videofilter/SegmentationProcessor.ts
```diff
@@ -0,0 +1,74 @@
...
+      buffers[0] = this.canvasVideoFrameBuffer;
+    }
+    predictions = undefined;
```
richnew10, line 64: This line doesn't do anything.

demos/browser/app/meetingV2/videofilter/SegmentationUtil.ts
```diff
@@ -0,0 +1,60 @@
...
+    src: 'https://cdn.jsdelivr.net/npm/@tensorflow-models/body-pix@2.0.5/dist/body-pix.min.js',
+    integrity: 'sha384-dPJ/sICXCqdh39bsuGLVFcOiRyeL/XcFrwiFrJq9oh7k1TCtsUKhX6UV2X4UuKU4',
+    crossOrigin: 'anonymous',
+  },
+];
+
+
+export async function loadBodyPixDependency(timeoutMs: number): Promise<void> {
+  // the tf library loading order must be followed
+  for (let i = 0; i < SEGMENTATION_DEPENDENCIES.length; i++) {
```
richnew10, line 36:
```
for (const { src, integrity, crossOrigin } of SEGMENTATION_DEPENDENCIES) {
```

README.md
```diff
@@ -1076,6 +1076,10 @@ The use of Amazon Voice Focus via this SDK involves the downloading and executio
 
 The use of Amazon Voice Focus runtime code is subject to additional notices. See [this NOTICES file](https://static.sdkassets.chime.aws/workers/NOTICES.txt) for details. You agree to make these additional notices available to all end users who use Amazon Voice Focus runtime code via this SDK.
 
+The browser demo applications in the demos directory (https://github.com/aws/amazon-chime-sdk-js/tree/master/demos) use TensorFlow.js (https://github.com/tensorflow/tfjs) and pre-trained TensorFlow.js models (https://github.com/tensorflow/tfjs-models) for image segmentation. Use of these third party models involves downloading and execution of code at runtime from jsDelivr (https://www.jsdelivr.com/) by end user browsers. For the jsDelivr Acceptable Use Policy, please visit this link (https://www.jsdelivr.com/terms/acceptable-use-policy-jsdelivr-net).
```
devalevenkatesh, line 1079: If these are demo specific notifices, can we can move this under demos/browser README?

demos/browser/app/meetingV2/videofilter/SegmentationUtil.ts
```diff
@@ -0,0 +1,60 @@
...
+    const dep = SEGMENTATION_DEPENDENCIES[i];
+    await new Promise<void>((resolve, reject) => {
+      let script = document.createElement('script');
```
richnew10, line 39: `const`?

demos/browser/app/meetingV2/videofilter/SegmentationUtil.ts
```diff
@@ -0,0 +1,60 @@
...
+      const timer = setTimeout(() => {
+        reject(new Error(`Loading script ${dep.src} takes longer than ${timeoutMs}`));
+      }, timeoutMs);
+      script.onload = function(_ev) {
```
richnew10, line 43: Consider using `onerror` as well as, or instead of, the timeout.

richnew10: Have you tested this failure mode?

demos/browser/app/meetingV2/videofilter/SegmentationUtil.ts
```diff
@@ -0,0 +1,60 @@
...
+        clearTimeout(timer);
+        resolve();
+      };
+      script.src = dep.src;
+      script.integrity = dep.integrity;
+      script.crossOrigin = dep.crossOrigin;
```
richnew10, line 49: Set `src` last in case the browser aggressively prefetches before you set the integrity.

demos/browser/app/meetingV2/videofilter/SegmentationUtil.ts
```diff
@@ -0,0 +1,60 @@
...
+      document.body.appendChild(script);
+    });
+  }
+}
+
+export function platformCanSupportBodyPixWithoutDegradation(): boolean {
+  // https://blog.tensorflow.org/2019/11/updated-bodypix-2.html for more detail on performance
+  // https://github.com/tensorflow/tfjs/issues/3319 which results in firefox memory leak
+  const browser = detect();
```
richnew10, line 58: Have you tested this in a recent Firefox (e.g., Nightly 88)?

richnew10 (changes requested):

README.md
devalevenkatesh, line 1079: I think you want to link the browser demo here right?
Also for consistency we could markdown the links.
The [browser demo application](https://github.com/aws/amazon-chime-sdk-js/tree/master/demos/browser) uses [TensorFlow.js](https://github.com/tensorflow/tfjs) and pre-trained [TensorFlow.js models](https://github.com/tensorflow/tfjs-models) for image segmentation. Use of these third party models involves downloading and execution of code at runtime from [jsDelivr](https://www.jsdelivr.com/) by end user browsers. For the jsDelivr Acceptable Use Policy, please visit [this link](https://www.jsdelivr.com/terms/acceptable-use-policy-jsdelivr-net).

README.md
devalevenkatesh, line 1081: Same comment for the Markdown links.

devalevenkatesh (changes requested):

README.md
richnew10: No; this is a legal notice, so we don't want to bury it. It might be worth putting it in both places.

richnew10: By all means make sure the text is linkified, but keep the hyperlinks readable. This text might be copied and pasted in plain text.

devalevenkatesh: Got it!

demos/browser/app/meetingV2/videofilter/SegmentationUtil.ts
XHatan: I ran our demo and bodypix's demo https://storage.googleapis.com/tfjs-models/demos/body-pix/index.html  and it still has memory leak,

demos/browser/app/meetingV2/videofilter/SegmentationUtil.ts
XHatan: yeah. by throttling network or setting the loading timeout to 0, error is thrown

README.md
XHatan: thanks. also this is moved to demo folder readme since it's demo specific.

demos/browser/app/meetingV2/videofilter/SegmentationProcessor.ts
XHatan: done.

README.md
XHatan: just saw below comment, will add to both places.

README.md
XHatan: fixed.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1613,6 +1622,15 @@ export class DemoMeetingApp
       this.enableUnifiedPlanForChromiumBasedBrowsers
     ) {
       filters = filters.concat(VIDEO_FILTERS);
+      if (platformCanSupportBodyPixWithoutDegradation()) {
+        try {
+          await loadBodyPixDependency(this.loadingBodyPixDependencyTimeoutMs);
```
richnew10, line 1627: This is probably wrong for a couple of reasons.

Firstly, this whole promise chain is `await`ed — you're basically blocking page load. Try it out: put

```
setTimeout(resolve, 20000);
```

in `loadBodyPixDependency`, and see how long it takes to load a meeting. This is a bad pattern for customers to follow.

Secondly, each time the user passes through the authentication flow, you will re-load these files. Try it: hit the 'Leave' button in the top right of a meeting, and join a second meeting.

You will need to switch to using `.then` here to fix the first issue, and use a single, persistent load promise to fix the second issue.

richnew10 (changes requested):

demos/browser/app/meetingV2/meetingV2.ts
XHatan: Sorry I am not familiar with UI best practice. 
So in this case, the dependencies should be loaded before the filter item is added to the drop down menu, that's why it's await-ed. With `then`, we don't suspend but (re)populate the filter list in the body of `then`, kind of like 
```
private async populateVideoFilterInputList (): void Promise<void> {
        loadDepedencies.then(() => {
           add segmentation filter to list and populate filter list
        });
        populate other filters.
}
```
So user could see a drop down menu without a segmentation filter but click a second time it show up ?

richnew10: You would just add 'Segmentation' to the list in the `.then` block, correct. You will also need a `.catch` block to handle failure. This will allow the rest of the initialization work to occur in parallel.

XHatan: Thanks. For the second issue, https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/browser/app/meetingV2/meetingV2.ts#L2245 page is redirected here when session is stopped. 
It makes sense in general to have a global loading of dependency, but it seems like it does nothing in our demo because page is refreshed when user leaves. Do you mind if I leave persistent loading out for now ?

richnew10: It's preferable to show common best practices if you can.

This should be as simple as somewhere at the top of the file:

```
const bodyPixLoad: Promise<void>;
```

and down here:

```
if (!this.bodyPixLoad) {
  this.bodyPixLoad = loadBodyPixDependency(…);
}
this.bodyPixLoad.then(() => {
  // Add to menu.
}).catch(e => {
  this.logger.warn('Could not load bodyPix:' + e);
});
```

XHatan: When page is re-directed, new demo object is created. I tried moving this out of demo class but it seems like the variable is not retained.

richnew10: Yes, you will get a new page load for the second meeting, so anything you store won't be retained. The point of doing this work is that the demo app doesn't have to work this way, and many customer apps won't — they will join multiple calls within the same JS environment and page scope.

If you want to test that you got it right, call `populateVideoFilterInputList` twice during page load. You should only get one fetch for each BodyPix resource.

XHatan: that makes sense. Changed.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -221,6 +226,11 @@ export class DemoMeetingApp
   static readonly DATA_MESSAGE_TOPIC: string = 'chat';
   static readonly DATA_MESSAGE_LIFETIME_MS: number = 300000;
 
+
+  // ideally we don't need to change this. Keep this configurable in case users have super slow network.
+  loadingBodyPixDependencyTimeoutMs: number = 10000;
+  loadingBodyPixDependencyPromise: Promise<void>;
```
richnew10, line 232: Strictly speaking, this should be `: undefined | Promise<void>;`. At some point we will turn on strict mode and this will be an error.

richnew10 (dismissed):
Add to CHANGELOG and rebase!

demos/browser/app/meetingV2/meetingV2.ts
XHatan: thanks ! will fix

devalevenkatesh (approved):

PR #1110, opened 2021-03-02
Log TURN credentials response URIs and username

devalevenkatesh:
**Issue #:**
- We need to catch if the TURN credentials are empty / incorrect if we see failure in joining meetings.

**Description of changes:**
 - Adding TURN credentials logging for TURN URIs and username after we fetch them from TURN control URL.
 - Removed `guides/index.ts` which was committed in last commit.
 
**Testing**

2. How did you test these changes? Checked the logs by running the demo app.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? The demo/browser meeting app. It is just the addition of logging.
4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? No.


tenge-amzn (approved):

src/task/ReceiveTURNCredentialsTask.ts
```diff
@@ -104,5 +106,7 @@ export default class ReceiveTURNCredentialsTask extends BaseTask {
         return !!uri;
       });
     this.context.turnCredentials.username = responseBodyJson.username;
+    this.context.logger.info(`TURN URIs: ${this.context.turnCredentials.uris}`);
+    this.context.logger.info(`Username: ${this.context.turnCredentials.username}`);
```
XHatan, line 110: maybe add to same log line so we can easily grep them

devalevenkatesh: Yeah, good call. Updated to be in the same log line.

ltrung (approved):

PR #1112, opened 2021-03-04
Fix demo pause/unpause button event listener behavior

hensmi-amazon:
**Issue #:** None

**Description of changes:** Before this change, every pause/unpause would add another event listener so it would eventually be spamming the signal channel with pause/unpause signals.  With this change we track the event listeners so we can remove them if one already exists.

**Testing**

2. How did you test these changes? Before pause/unpause would spam depending on how many times the button had been pressed.  After change it just sends the expected signal once.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Browser demo


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2246,6 +2247,18 @@ export class DemoMeetingApp
     }
   }
 
+  createPauseResumeListener(tileState: VideoTileState): (event: Event) => Promise<void> {
+      return async (event: Event): Promise<void> => {
```
devalevenkatesh, line 2251: Since the pause behavior is synchronous, so just want to know should this function execute synchronously?

hensmi-amazon: Oh hmm, don't know why I added the `async` modifier here.  Will remove.

XHatan (approved):

Issue #1115, opened 2021-03-04
Reduce leakage of objects due to global observers


richnew10:
In a few places it's hard to cause all SDK objects to be garbage collected:

* `MeetingSessionPOSTLogger` sets up a timer that isn't cleaned up on stop.
* `DefaultDeviceController` registers a global `devicechange` observer that retains the DDC itself.
* `DefaultActiveSpeakerDetector` gets retained somehow, though I haven't pinned this down.

There are probably more. This ticket tracks cleaning these up.

richnew10:
I have a branch that fixes all of the leaks I've found in the demo. It unfortunately breaks 53 tests (possibly because the tests are broken!), so it will take me some time to get it into a landable state.

richnew10:
Work is in `richnew/leaks`, in case I get hit by a bus.

PR #1126, opened 2021-03-10
Fix memory leaks

richnew10:
**Issue #:** #1115

**Description of changes:**

This PR fixes several classes to be more careful with their lifetimes and event listeners in order to minimize leaks.

It is split into multiple commits for ease of review. It will be flattened shortly before final review.

Because the audio-video controller can be stopped and restarted, as can other classes, this PR introduces a `Disposable` interface. When you are completely finished with meeting objects, call `dispose()` and discard those objects. Applications that expect a meeting to last for the life of the page don't have to bother with this, and you can achieve the same effect by carefully removing observers and deselecting devices — the demo app shows this.

Additionally, this commit causes the `ended` event to be dispatched on `MediaStreamTrack`s that are closed by the SDK itself.

As an additional bonus, this introduces `AsyncScheduler.nextTick(callback)`, which is equivalent to `new AsyncScheduler(callback).start()` but without the object creation/GC/method call overhead.

Much work was required in mocks and tests to get the behavior of the test suite to be closer to the behavior of browsers; apologies for the churn.

**Testing**

Yes.

Unit tests and manually.

Leak testing requires making a change to the meetingV2 demo and using some tooling. This was shared with the team.

Yes. This change introduces the `Disposable` interface on `DefaultActiveSpeakerDetector` and `DefaultDeviceController`, and the `nextTick` method on `AsyncScheduler`. 

Additionally, it will cause the `ended` event to be dispatched when it otherwise would not.

No.

---


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -489,10 +562,17 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+    // it manually. Do so in order to clean up listeners, if for no other reason.
+    AsyncScheduler.nextTick(() => {
+      for (const track of tracksToStop) {
+        track.dispatchEvent(new Event('ended'));
+      }
+    });
+
     for (const kind in this.activeDevices) {
       if (this.activeDevices[kind] && this.activeDevices[kind].stream === mediaStreamToRelease) {
         this.activeDevices[kind] = null;
```
michhyun1, line 578: nit: can we delete this entry instead of setting to null

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -221,16 +221,18 @@ export class DemoMeetingApp
   static testVideo: string =
     'https://upload.wikimedia.org/wikipedia/commons/transcoded/c/c0/Big_Buck_Bunny_4K.webm/Big_Buck_Bunny_4K.webm.360p.vp9.webm';
   static readonly LOGGER_BATCH_SIZE: number = 85;
-  static readonly LOGGER_INTERVAL_MS: number = 2000;
+  static readonly LOGGER_INTERVAL_MS: number = 2_000;
```
ltrung, line 224: Is this a new JS feature? Can we assume it is well-supported?

richnew10: It's a TypeScript feature, introduced in TS 2.7 in late 2017 or early 2018.

src/devicecontroller/DefaultDeviceController.ts
richnew10: Sure!

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2235,15 +2247,64 @@ export class DemoMeetingApp
     this.log(`session stopped from ${JSON.stringify(sessionStatus)}`);
     this.log(`resetting stats in WebRTCStatsCollector`);
     this.statsCollector.resetStats();
```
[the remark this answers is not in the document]
richnew10: I don't think that would be correct. This is called any time AV stops, which can happen for 23 different reasons. `onLeftMeeting` is called in exactly two of those.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2273,15 +2284,64 @@ export class DemoMeetingApp
...
+
+    /**
+     * This is approximately the inverse of the initialization method above.
+     * This work only needs to be done if you want to continue using the page; if
+     * your app navigates away or closes the tab when done, you can let the browser
+     * clean up.
+     */
+    const onLeftMeeting = async () => {
+      // Clean up the timers for this.
+      this.audioVideo.unsubscribeFromActiveSpeakerDetector(this.activeSpeakerHandler);
```
ltrung, line 2307: Is there anyway we could automatically unsubscribe once the meeting session stop? I feel It is a bit overwhelming for builders to remember to cleanup everything themselves.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2273,15 +2284,64 @@ export class DemoMeetingApp
...
+
+      // Stop listening to attendee presence.
+      this.audioVideo.realtimeUnsubscribeToAttendeeIdPresence(this.attendeeIdPresenceHandler);
+
+      // Stop watching device changes in the UI.
+      this.audioVideo.removeDeviceChangeObserver(this);
+
+      // Drop the audio input and output.
+      await this.audioVideo.chooseAudioInputDevice(null);
```
ltrung, line 2316: Don't we already do it here https://github.com/aws/amazon-chime-sdk-js/blob/master/src/task/CleanStoppedSessionTask.ts#L59 ?

demos/browser/app/meetingV2/meetingV2.ts
richnew10: Remember that you can start, stop, start an AV session.

The active speaker detector doesn't know whether the session is started or stopped, and the detector is created and passed as an argument to the AV facade.

At one point I had `DefaultAudioVideoController.stop` disposing of the active speaker detector, but that would require that the caller re-subscribe before the next `start`, and that's simply not our API contract.

Further, unsubscribing here helps to avoid leaking `this.activeSpeakerHandler`, which is retained in the detector.

Happy to have someone who knows more about the guts here suggest a way to get this to avoid leaking — maybe there's a realtime observer notification that would allow the detector to discard the timers and recreate them?

demos/browser/app/meetingV2/meetingV2.ts
richnew10: Possibly. I think I didn't have this line to start with, but it seemed odd to have the video one but not this…

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2273,15 +2284,64 @@ export class DemoMeetingApp
...
+      await this.audioVideo.chooseAudioOutputDevice(null);
+      this.audioVideo.unbindAudioElement();
+
+      // Clean up the loggers so they don't keep their `onload` listeners around.
+      setTimeout(async () => {
+        await this.meetingEventPOSTLogger?.dispose();
+        await this.meetingSessionPOSTLogger?.dispose();
```
ltrung, line 2323: Doesn't this also call device controller dispose method which also dispose the audio io and video input? So do we need the code above?

demos/browser/app/meetingV2/meetingV2.ts
ltrung: Do we need to release video input device here too?

demos/browser/app/meetingV2/meetingV2.ts
ltrung: For simplicity, can we just have a dispose/clean up function for AV to unsubscribe everything ?

src/scheduler/AsyncScheduler.ts
```diff
@@ -11,4 +11,20 @@ export default class AsyncScheduler extends TimeoutScheduler {
   constructor() {
     super(0);
   }
+
+  /**
+   * Execute the provided callback on the next tick of the event loop.
+   * This is semantically equivalent to
+   *
+   * ```typescript
+   * new AsyncScheduler(callback).start();
```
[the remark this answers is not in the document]
richnew10: You can't cancel it, correct.

These two lines are equivalent:

```
AsyncScheduler.nextTick(callback);
```

and

```
new AsyncScheduler(callback).start();
```

Note that I did *not* draw an equivalence to

```
const scheduler = new AsyncScheduler(callback);
scheduler.start();
// Do other things, potentially including calling `stop` in the same event loop tick
```

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -212,6 +215,7 @@ export default class DefaultAudioVideoController
   }
 
   start(): void {
+    this.activeSpeakerDetector;
```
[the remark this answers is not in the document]
richnew10: Yes. Lazy getter properties are the devil, but that's the API we're working with.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -436,6 +440,11 @@ export default class DefaultAudioVideoController
       this._reconnectController.disableReconnect();
       this.logger.info('attendee left meeting, session will not be reconnected');
       this.actionDisconnect(new MeetingSessionStatus(MeetingSessionStatusCode.Left), false, null);
+
+      /* istanbul ignore if */
+      if (!this._activeSpeakerDetector || !isDisposable(this._activeSpeakerDetector)) {
```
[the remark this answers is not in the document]
richnew10: Dead code; it's removed in my next push.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -106,6 +108,73 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+    this.onDeviceChangeCallback = undefined;
+  }
+
+  private shouldObserveDeviceChanges(): boolean {
+    if (this.deviceChangeObservers.size) {
+      return true;
+    }
+
+    const activeDevices =
+      (this.activeDevices['audio'] && this.activeDevices['audio'].constraints !== null) ||
```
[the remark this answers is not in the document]
richnew10: You might be right that we need to check for null constraints for video devices too.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -489,22 +559,29 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     }
 
     for (const track of tracksToStop) {
-      this.logger.info(`stopping ${track.kind} track`);
       track.stop();
     }
 
+    // Calling `stop` does not actually trigger `onended` callbacks unless we do
+    // it manually. Do so in order to clean up listeners, if for no other reason.
+    AsyncScheduler.nextTick(() => {
```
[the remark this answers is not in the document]
richnew10: Correct. This is indeed a change in behavior, which is why I called it out in the PR.

We have three choices:

* Don't call `ended`, and leak.
* Do call `ended` when _the SDK_ ends a track (e.g., when a device is discarded), which makes the SDK roughly equivalent to the browser.
* Figure out some other way to throw away callbacks that are attached to the stream's `ended` event.

demos/browser/app/meetingV2/meetingV2.ts
richnew10: Why would you think it does that?

ltrung: oops. I misread this to meetingSession.dispose. Should we call meetingSession.dispose then? https://github.com/aws/amazon-chime-sdk-js/pull/1126/commits/8ab1cce7ffe35d8de673463a17715db4ee234577?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.ts#diff-bade3a8271704aa5eaa7df2a595d9aa8b18f8ebb3e1aa1fdf65f443c3abbdd37R114
You introduced several dispose but i did not see we use it here.

src/devicecontroller/DefaultDeviceController.ts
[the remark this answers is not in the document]
ltrung: Can we store the callback reference in activeDevices[kind] and call removeEventListener here (given that we have activeDevices[kind].stream reference already)?

demos/browser/app/meetingV2/meetingV2.ts
richnew10: My aim was to make it so existing apps that unregister in all the right places wouldn't also need to call dispose — the first draft of this papered over the problem by allowing a force cleanup, but avoiding leaks without that is better.

src/devicecontroller/DefaultDeviceController.ts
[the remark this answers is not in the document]
richnew10: That's a good idea! I'll give it a shot.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -106,6 +108,73 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+    }
+  }
+
+  async dispose(): Promise<void> {
+    // Remove device change callbacks.
+    this.stopWatchingDeviceChanges();
+
+    // Deselect any audio input devices and throw away the streams.
+    // Discard the current video device, if there is one.
+    await this.chooseAudioInputDevice(null);
```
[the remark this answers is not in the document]
richnew10: The reason I did this is:

* There is a bunch of cleanup that's done when you select the null device — e.g., disposing of transforms.
* You could conceivably call `dispose` with an audio device still selected — e.g., create a device controller, choose an input, and then throw the controller away without even starting a meeting, let alone ending one.

The audio output is taken care of by unbinding just below.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -153,6 +148,14 @@ export default class DefaultAudioVideoController
   }
 
   get activeSpeakerDetector(): ActiveSpeakerDetector {
+    // Lazy init.
+    if (!this._activeSpeakerDetector) {
```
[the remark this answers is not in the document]
richnew10: It means we only actually create the detector when we try to use it, which is in `start`, but with none of the safety concerns that come with tying creation and destruction to particular lifecycle events.

demos/browser/app/meetingV2/meetingV2.ts
richnew10: Oh, I just realized — we cannot rely on `CleanStoppedSessionTask` having run, in part because they might be disposing before the task gets to run, and in part because they might never have started the meeting in the first place!

demos/browser/app/meetingV2/meetingV2.ts
richnew10: I don't think that's as easy as we might want.

We couldn't just unsubscribe, because we need to know which handlers to unsubscribe, and this is what the code looks like when it does!

We could add an `unsubscribeAll` method to a bunch of places, but that makes our interface bigger and isn't feasible for classes that don't keep a reference to their callbacks — `activeSpeakerHandler` does, but not all do.

Implementing `destroy` on the facade is trickier than you'd think. Because of the facade pattern, we would need `isDestroyable` checks and `destroy` calls on every component member of `DefaultAudioVideoFacade`. Each of the default classes would need to implement `Destroyable` and recurse down.

There are also confusing object references in the demo — e.g., `this.meetingSession` holds a reference to the device controller, which cleaned up by a destroyer — and so we would save ~5 lines of code here, but still need to do some cleanup as you can see below.

I am not opposed to doing this work, but I think it's a good intermediate step to make manual cleanup like this function correctly, and then consider whether making `DefaultAudioVideoFacade implements Destroyable` is a worthwhile step.

CHANGELOG.md
```diff
@@ -24,7 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [2.6.1] - 2021-03-17
 
 ### Fixed
-- Fix infinite loop when calling `chooseAudioInputDevice` with a `MediaDeviceInfo` instance.
+- Fix infinite loop when calling `chooseAudioInputDevice` with a
+  `MediaDeviceInfo` instance.
+- Fewer observers are now retained after meetings end. This should reduce
```
ltrung, line 34: This should be moved to the Unreleased section

src/contentsharecontroller/ContentShareMediaStreamBroker.ts
```diff
@@ -109,7 +109,8 @@ export default class ContentShareMediaStreamBroker implements MediaStreamBroker
   cleanup(): void {
     if (this.mediaStream) {
       for (let i = 0; i < this.mediaStream.getTracks().length; i++) {
-        this.mediaStream.getTracks()[i].stop();
+        const track = this.mediaStream.getTracks()[i];
```
ltrung, line 112: is this just coding style change?

src/task/CleanStoppedSessionTask.ts
```diff
@@ -56,6 +56,9 @@ export default class CleanStoppedSessionTask extends BaseTask {
       this.context.turnCredentials = null;
       this.context.videoSubscriptions = null;
       this.context.transceiverController.reset();
+
+      // This should really be a _device selection_ operation that chooses no device,
+      // allowing the device controller to clean up any selected transform device.
```
ltrung, line 61: Should we add a TODO here if we want to defer this?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -965,9 +1178,11 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
 
     if (device === null && kind === 'video') {
       this.lastNoVideoInputDeviceCount = this.inputDeviceCount;
-      if (this.activeDevices[kind]) {
-        this.releaseMediaStream(this.activeDevices[kind].stream);
+      const active = this.activeDevices[kind];
```
ltrung, line 1181: nit: Could we rename this to activeDevice? active sounds like a boolean variable :)

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -954,6 +1150,23 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     }
   }
 
+  private releaseActiveDevice(device: DeviceSelection): void {
```
ltrung, line 1153: Nice

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -724,7 +912,7 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
   }
 
   private async listDevicesOfKind(deviceKind: string): Promise<MediaDeviceInfo[]> {
-    if (this.deviceInfoCache === null) {
+    if (this.deviceInfoCache === null || !this.isWatchingForDeviceChanges()) {
```
ltrung, line 915: Is this really necessary? In what cases can this happen?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -27,7 +28,90 @@ import VideoInputDevice from './VideoInputDevice';
...
+  scheduler.start(onTick);
+
+  const listener = (): void => {
+    scheduler.stop();
+  };
+
+  // This event listener will leak unless you remove it.
+  stream.getVideoTracks()[0].addEventListener('ended', listener);
+
+  return { listener, scheduler, stream };
```
ltrung, line 110: Do we need to return listener and scheduler? We only get stream in the caller function.

src/devicecontroller/DefaultDeviceController.ts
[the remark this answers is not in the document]
ltrung: Sorry maybe I miss it but what line number where we do unbind audiooutput?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -106,6 +190,73 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+  }
+
+  async destroy(): Promise<void> {
+    // Remove device change callbacks.
+    this.stopWatchingDeviceChanges();
+
+    // Deselect any audio input devices and throw away the streams.
+    // Discard the current video device, if there is one.
+    await this.chooseAudioInputDevice(null);
+    if (this.activeDevices['video']) {
```
ltrung, line 244: Can we call chooseVideoInputDevice(null) here? It should execute a similar code:
```
    if (device === null && kind === 'video') {
      this.lastNoVideoInputDeviceCount = this.inputDeviceCount;
      if (this.activeDevices[kind]) {
        this.releaseMediaStream(this.activeDevices[kind].stream);
        delete this.activeDevices[kind];
      }
      return;
    }
```

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -462,48 +621,127 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+    /* istanbul ignore next */
+    const audioEnded = this.activeDevices['audio']?.endedCallback;
+    for (const track of this.audioInputSourceNode.mediaStream.getTracks()) {
+      track.stop();
+
+      /* istanbul ignore else */
+      if (audioEnded) {
+        track.removeEventListener('ended', audioEnded);
+        delete this.activeDevices['audio'].endedCallback;
+        delete this.activeDevices['audio'];
```
ltrung, line 652: Just curious. What happen if we just delete `this.activeDevices['audio']` and not the `endedCallback`?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -298,26 +301,46 @@ export class DemoMeetingApp
 
   meetingLogger: Logger | undefined = undefined;
 
-  constructor() {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    (global as any).app = this;
+  // If you want to make this a repeatable SPA, change this to 'spa'
```
ltrung, line 304: Could we add more documentation here? Such as if you do Shift + Leave Meeting it will trigger halt to clean up the demo app?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2273,15 +2310,89 @@ export class DemoMeetingApp
...
+
+      // Stop watching device changes in the UI.
+      this.audioVideo.removeDeviceChangeObserver(this);
+
+      // Stop content share and local video.
+      await this.audioVideo.stopLocalVideoTile();
+      await this.audioVideo.stopContentShare();
+
+      // Drop the audio output.
+      await this.audioVideo.chooseAudioOutputDevice(null);
```
ltrung, line 2362: Don't we want to also chooseAudioInput and VideoInput to null? Or can we call audioVideo.deviceController.destroy here?

devalevenkatesh (changes requested):

CHANGELOG.md
richnew10: We really need to fix this whole changelog thing at some point.

demos/browser/app/meetingV2/meetingV2.ts
richnew10: We discussed this in earlier versions.

See three lines below your comment: leaving the meeting will do this for inputs.

src/devicecontroller/DefaultDeviceController.ts
richnew10: If the previous value of `this.activeDevices['audio']` were still present anywhere — including in console logs, debugger-stored variables, etc. — then the callback would leak. Deleting the callback from the device itself helps to aggressively prune the object graph.

This is also a piece of code that is directly related to cleaning up references to this callback, so it's nice to be explicit about where the leaks might lie.

src/devicecontroller/DefaultDeviceController.ts
richnew10: The point of this is to make it super explicit to anyone reading this code that it's going to leak, and all of these things are relevant.

src/devicecontroller/DefaultDeviceController.ts
ltrung: Got it

src/devicecontroller/DefaultDeviceController.ts
richnew10: `listDevicesOfKind` is called by `listAudioDevices`, which is a public API that can be called at any time — including when there are no active devices and no device change observers. For example: immediately after creating the controller!

```typescript
const d = new DefaultDeviceController(…);

// `d` has no global `devicechange` observer so that it doesn't leak.

let audio = await d.listAudioDevices();
…
// We still have no observer, so this needs to update the device info cache again.
audio = await d.listAudioDevices();
```

If we have no device change observers and no active devices, this thing is 'dormant', and has no observers and thus won't leak if discarded… but in this case we need to update our device cache every time, just in case something changed since the last call. If we have observers our device info cache is kept up to date, and we can skip the work.

src/devicecontroller/DefaultDeviceController.ts
richnew10: Does it matter to you as a reader that you can see that the variable only exists for the next four lines?

My general rules of thumb:

* If you can see the start and end of the variable scope without moving your eye, use `a`, `d`, `c`, etc. This is like math: `F = ma` is a lot more readable in context than `acceleratingForceVector = massOfObject * accelerationVector`.
* If you can see the start and end of scope without scrolling, use just enough word to disambiguate from others: `active`, `device`, `controller`.
* If your scope is longer than that, be very explicit in one or both of name and type declaration: `activeDevice`, `CONTROLLER_TIMEOUT_MILLIS: number`, etc.

src/task/CleanStoppedSessionTask.ts
richnew10: It's not a TODO, it's an observation. We can't really fix it, but a reader will learn context and perspective by reading this comment.

src/contentsharecontroller/ContentShareMediaStreamBroker.ts
richnew10: It's not a semantic change, correct.

I made the change in order to add logging (later removed), but I do find it slightly more readable — it splits the `()[].` sequence — and the extra line adds an easy breakpoint for those who use debuggers.

src/task/CleanStoppedSessionTask.ts
ltrung: What do you mean we can't fix it? When you say "This should be" I thought it would be something we want to do (maybe in v3)

src/devicecontroller/DefaultDeviceController.ts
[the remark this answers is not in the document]
richnew10: You didn't miss it — it's no longer there!

src/task/CleanStoppedSessionTask.ts
richnew10: We can't fix it within the current API because `CSST` only knows about `MediaStreamBroker`, not about `DeviceController` — it only knows how to release media streams that are tracked in the `AudioVideoControllerState`, not how to unselect a device.

The issue here is that _we now work with much more than streams_, and this API hasn't kept pace with the complexity of the SDK (something we missed in v2.0).

It's currently up to the developer's application to manage which device is currently selected (peep the complexity in the demos) and `DDC` has to figure out from the stream passed here which device to clean up.

Fixing either of those things can't be done until v3.0 at least, and it's a much larger scope than I would normally put in a TODO.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1130,16 +1157,17 @@ export class DemoMeetingApp
...
+      this.meetingSessionPOSTLogger = new MeetingSessionPOSTLogger(
+        'SDK',
+        configuration,
+        DemoMeetingApp.LOGGER_BATCH_SIZE,
+        DemoMeetingApp.LOGGER_INTERVAL_MS,
+        `${DemoMeetingApp.BASE_URL}logs`,
+        logLevel
+      );
       this.meetingLogger = new MultiLogger(
+        this.meetingEventPOSTLogger,
```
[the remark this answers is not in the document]
richnew10: Yes! Thanks.

richnew10:
@devalevenkatesh @ltrung back to you…

src/devicecontroller/DefaultDeviceController.ts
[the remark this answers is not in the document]
ltrung: So do we no longer try to clean up the audio output binding anymore?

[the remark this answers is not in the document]
richnew10: Correct; the output binding is actually part of the AudioVideoController, so if you want to clean that up you should call `audioVideo.unbindAudioElement()`.

src/meetingsession/DefaultMeetingSession.ts
```diff
@@ -108,6 +109,15 @@ export default class DefaultMeetingSession implements MeetingSession {
     return this._deviceController;
   }
 
+  async destroy(): Promise<void> {
+    if (isDestroyable(this._deviceController)) {
+      await this._deviceController.destroy();
+    }
+    this._deviceController = undefined;
+    this.audioVideoController = undefined;
```
devalevenkatesh, line 117: Should we call `this.audioVideoController.stop()` here and then set to undefined?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -786,7 +812,10 @@ export class DemoMeetingApp
     });
 
     const buttonMeetingLeave = document.getElementById('button-meeting-leave');
-    buttonMeetingLeave.addEventListener('click', _e => {
+    buttonMeetingLeave.addEventListener('click', e => {
+      if (e.shiftKey) {
+        this.behaviorAfterLeave = 'halt';
+      };
       new AsyncScheduler().start(async () => {
```
devalevenkatesh, line 819: `AsyncScheduler.nextTick` here as well?

src/meetingsession/DefaultMeetingSession.ts
richnew10: I actually feel like it should do `audioVideoController` (because the session created it) and _not_ the device controller, which was passed in as an argument.

What do you think?

devalevenkatesh: If you mean destroy the `audioVideoController` then yes, I think so it would be right to destroy what meeting session creates avoiding `deviceController` cleanup which is demo created. Further, the `cleanUpResources` function in the demo has some function calls which could be moved here like the `this.audioVideo.stopLocalVideoTile()` and so on which are not dependent on the objects created in the demo.

But I think, this cleanup at two places would be confusing. Also,  we have to explicitly document to call this at last as builders may call `meetingSession.destroy()` first and then this would set the `audioVideo` object to `undefined` and the demo cleanup calls will internally fail. So instead of `meetingSession.destroy()` should we just call `this.deviceController.destroy()` in the `cleanUpResources` itself in the demo?

richnew10: In writing this code I wanted it to be possible to use the SDK without explicitly destroying most of these core classes: it should be enough to remove observers and stop them as appropriate, and the demo serves to model that behavior. 

These `destroy` methods are a hammer to use when it is hard or impossible to untangle callbacks. Again the meeting demo showcases this when it explicitly destroys the POST loggers, which cannot otherwise be correctly disposed of.

Because the SDK classes typically promise to be reusable — that is, you can create a meeting, stop it, and then start it again — it's important that we clean up as we go, and do so as incrementally as possible. It should not be necessary for the demo to destroy its device controller in order to stop other classes from leaking.

demos/browser/app/meetingV2/meetingV2.ts
richnew10: Done, throughout.

src/meetingsession/DefaultMeetingSession.ts
devalevenkatesh: Okok got it! So, in-order to keep the re-usability here, will you be removing the `deviceController.destory()` call here or we need it to fix the leak?

richnew10: I've removed it and added destroys for the two owned controllers. This code is not called from the demo, so it's not necessary to fix leaks, because the demo cleans up on its own.

demos/browser/package.json
```diff
@@ -26,15 +26,15 @@
     "sass": "^1.26.11",
     "sass-loader": "^10.0.2",
     "style-loader": "^1.2.1",
-    "ts-loader": "^8.0.4",
-    "typescript": "^4.0.0",
+    "ts-loader": "^8.1.0",
+    "typescript": "^4.2.3",
     "webpack": "^4.44.2",
-    "webpack-cli": "^3.3.12",
+    "webpack-cli": "^4.6.0",
```
devalevenkatesh, line 32: Does anything change here due to the major version upgrade? I checked [here](https://webpack.js.org/migrate/4/#cli) but looks like we should be good.

demos/serverless/deploy.js
```diff
@@ -147,7 +147,10 @@ function ensureApp(appName) {
 function ensureTools() {
   spawnOrFail('aws', ['--version']);
   spawnOrFail('sam', ['--version']);
-  spawnOrFail('npm', ['install']);
+
+  // Work around https://github.com/awslabs/aws-embedded-metrics-node/issues/84.
+  // When that issue is fixed, you can remove --legacy-peer-deps.
+  spawnOrFail('npm', ['install', '--legacy-peer-deps']);
```
devalevenkatesh, line 153: Should we use `--no-save` here and down below as well?

From NPM blog on NPM 7 update:
"Running npm install with npm 7 in a project with a v1 lockfile will replace that lockfile with the new v2 format. To avoid this, you can run npm install --no-save."

So for people running on npm 7 using this dont generate the v2 lock file.

richnew10: Possibly. I'd rather we just… fix it. This change was just necessary for me to test a deployed version, because the machine I do my deploys on is running npm 7.

demos/browser/package.json
ltrung: I assume we have verified that npm run start still works after this :D

richnew10: Correct :D

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2273,15 +2312,95 @@ export class DemoMeetingApp
...
+      await this.audioVideo.chooseAudioOutputDevice(null);
+      this.audioVideo.unbindAudioElement();
+
+      // Stop any video processor.
+      await this.chosenVideoTransformDevice?.stop();
+
+      // Stop Voice Focus.
+      await this.voiceFocusDevice?.stop();
+
+      // If you joined and left the meeting, `CleanStoppedSessionTask` will have deselected
```
ltrung, line 2373: Should this comment need to be moved up to line 2363?

richnew10: It's valid anywhere in this block, and we did just stop the two transform devices, so I think here is fine.

demos/serverless/deploy.js
devalevenkatesh: Fixing as in upgrading to NPM 7 😅  ?

richnew10: Not necessarily — just upgrading to lockfile version 2. npm 6 and npm 7 both support that version.

devalevenkatesh: Cool, sounds good as it is back-compatible as well.

devalevenkatesh (approved):

richnew10:
Spurious integration test failure due to Chrome bug; those integration tests pass locally if hardware accel is disabled in Chrome.

PR #1138, opened 2021-03-16
Fix infinite loop when passing in MediaDeviceInfo when choosing Audio Input

ltrung:
**Issue**
If you pass in MediaDeviceInfo object into `chooseAudioInputDevice`, it will cause infinite loop. 
After requesting a device from getUserMedia, `chooseAudioInputDevice` calls `restartLocalAudio` which in turns call `chooseInputIntrinsicDevice` again and it relies on the [existing device check](https://github.com/aws/amazon-chime-sdk-js/blob/v2.6.0/src/devicecontroller/DefaultDeviceController.ts#L987) to quit early. If this check returns false, we will try to select the device again.

The root cause issue here is that the logic for [calculating media track constraints](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/devicecontroller/DefaultDeviceController.ts#L1221) will return the constraints without any device info if the input is MediaDeviceInfo. Without the device info (deviceId and groupId), getUserMedia might not return the device requested.

In 2.5.0 and earlier, there is actually a bug when selecting input device. The code flow would be in Chrome assume there are 2 devices (device 1 - default, device 2)
```
chooseAudioInputDevice(deviceInfo for device 2)
->  chooseInputIntrinsicDevice(deviceInfo for device 2)
->    getUserMedia({ constraints with no deviceInfo}) = stream from default device 1
->    set activeDevices groupId from input which is device2's groupId <- incorrect
->    restartLocalAudio
->      chooseInputIntrinsicDevice(existing constraint with no deviceInfo)
->      hasSameGroupId -> returns false since the groupId recorded does not match the constraint which is ''
->      getUserMedia({ constraints with no deviceInfo}) = stream from default device 1
->      set activeDevices groupId from input which is now ''
->      restartLocalAudio
->      chooseInputIntrinsicDevice(existing constraint with no deviceInfo)
->      hasSameGroupId -> returns true since the groupId recorded is now also ''
```
End result: device1 is selected instead of device 2.

In 2.6.0, we updated the logic to actually get the groupId from the acquired stream (in PR #1078) which together with the existing constraint bug causes the infinite loop.
```
chooseAudioInputDevice(deviceInfo for device 2)
->  chooseInputIntrinsicDevice(deviceInfo for device 2)
->    getUserMedia({ constraints with no deviceInfo}) = stream from default device 1
->    set activeDevices groupId from acquired stream which is default device 1's groupId
->    restartLocalAudio
->      chooseInputIntrinsicDevice(existing constraint with no deviceInfo)
->      hasSameGroupId -> returns false since the groupId recorded does not match the constraint which is ''
->      getUserMedia({ constraints with no deviceInfo}) = stream from default device 1
->      set activeDevices groupId from acquired stream which is default device 1's groupId
->      restartLocalAudio
->      chooseInputIntrinsicDevice(existing constraint with no deviceInfo)
->      hasSameGroupId -> returns false since the groupId recorded does not match the constraint which is ''
->      getUserMedia({ constraints with no deviceInfo}) = stream from default device 1
->      <the loop continue>
```
**Description of changes:**
Fixing the constraint bug if MediaDeviceInfo is passed to chooseAudioInputDevice.
Although MediaDeviceInfo is not a valid input to chooseAudioInputDevice, it works in previous version so we try to keep the same behavior. 
**Testing**

2. How did you test these changes? Manually run demo app and call chooseAudioInput with mediaDeviceInfo
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes
1. Open console and call `devices = await app.audioVideo.listAudioInputDevices();`
2. Call await app.audioVideo.listAudioInputDevices(devices[2]);
3. Make sure that the call returns.


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Improve some unit tests.
+- Fix infinite loop when passing in MediaDeviceInfo when choosing Audio Input
```
richnew10, line 23: Be specific:

```
- Fix infinite loop when calling `chooseAudioInputDevice` with a `MediaDeviceInfo` instance.
```

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1170,6 +1170,10 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     } else if (stream) {
       // @ts-ignore - create a fake track constraint using the stream id
       trackConstraints.streamId = stream.id;
+    } else if (DefaultDeviceController.isMediaDeviceInfo(device)) {
+      // trackConstraints.deviceId = device.deviceId;
+      // trackConstraints.groupId = device.groupId;
```
richnew10, line 1175: ?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1424,4 +1428,15 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+  private static isMediaDeviceInfo(device: unknown): device is MediaDeviceInfo {
+    return (
+      !!device &&
+      typeof device === 'object' &&
+      'deviceId' in device &&
+      'groupId' in device &&
+      'kind' in device &&
+      'label' in device
+    );
+  }
```
richnew10, line 1441: Make this a free function rather than a static method.

You also don't need `!!` here.

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -760,6 +760,18 @@ describe('DefaultDeviceController', () => {
...
+    it('Can chooseAudioInputDevice if pass in MediaDeviceInfo', async () => {
+      const spy = sinon.spy(navigator.mediaDevices, 'getUserMedia');
+      try {
+        await deviceController.chooseAudioInputDevice(
+          getMediaDeviceInfo('device-Id', 'audioinput', 'label', 'group-Id')
+        );
+      } catch (e) {
+        throw new Error('This line should not be reached.');
+      }
+      expect(spy.calledOnce).to.be.true;
```
richnew10, line 772: Nice test.

Could you add a spy on one of the functions that was involved in the recursion, too, and verify that it's only called once?

richnew10 (changes requested):

src/devicecontroller/DefaultDeviceController.ts
ltrung: My bad. I commented it out to test unit test and forgot to revert it back

CHANGELOG.md
ltrung: Updated

test/devicecontroller/DefaultDeviceController.test.ts
ltrung: getUserMedia is the call that happens in the recursion :)

src/devicecontroller/DefaultDeviceController.ts
ltrung: Updated. But I followed the example here https://github.com/aws/amazon-chime-sdk-js/blob/master/src/devicecontroller/VideoTransformDevice.ts#L44. Do we need !! there?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1425,3 +1428,13 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
     this.logger.info(s);
   }
 }
+
+export function isMediaDeviceInfo(device: unknown): device is MediaDeviceInfo {
```
richnew10, line 1432: You don't need to export this.

test/devicecontroller/DefaultDeviceController.test.ts
richnew10: I meant one of our functions, like `chooseInputIntrinsicDevice`, which also appears in the recursion.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1425,3 +1428,13 @@ export default class DefaultDeviceController implements DeviceControllerBasedMed
...
+  return (
+    typeof device === 'object' &&
+    'deviceId' in device &&
+    'groupId' in device &&
+    'kind' in device &&
+    'label' in device
```
richnew10, line 1438: Are these four fields always present in all `MediaDeviceInfo` instances?

test/devicecontroller/DefaultDeviceController.test.ts
ltrung: Added `chooseInputIntrinsicDevice`

src/devicecontroller/DefaultDeviceController.ts
ltrung: Verified in Chrome, FF, Safari, Safari iOS. All four fields are present (although groupId = "" in Safari).

richnew10 (approved):

PR #1149, opened 2021-03-25
Adding call to realtimeSetLocalAudioInput after restartLocalAudio

anuranduttaroy:
**Issue #:** NA

**Description of changes:** 
  
*Issue:* There is no logic to update the attachment of the realtime controller to the newly selected audio device. This caused the local mute/unmute to break if a new device was selected after joining the meeting. 

*Solution:* Calling `realtimeSetLocalAudioInput` as part of `AudioVideoController.restartLocalAudio()` to 
  fix local mute/unmute issue while switching audio devices.

**Testing**

Yes.

Used the serverless demo to test `realtimeSetLocalAudioInput` is being called on every device switch.

Yes. Meeting serverless demo.

NA.

NA

### DIFF
--- a/src/audiovideocontroller/DefaultAudioVideoController.ts
+++ b/src/audiovideocontroller/DefaultAudioVideoController.ts
@@ -564,7 +564,6 @@ export default class DefaultAudioVideoController
     if (!audioStream || audioStream.getAudioTracks().length < 1) {
       throw new Error('could not acquire audio track');
     }
-
     this.connectionHealthData.reset();
     this.connectionHealthData.setConnectionStartTime();
 
@@ -584,6 +583,7 @@ export default class DefaultAudioVideoController
         audioTrack
       );
     }
+    this._realtimeController.realtimeSetLocalAudioInput(audioStream);
     this.meetingSessionContext.activeAudioInput = audioStream;
     callback();
     if (replaceTrackSuccess) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Improve some unit tests.
+- Calling `realtimeSetLocalAudioInput` as part of `AudioVideoController.restartLocalAudio()` to 
+  fix local mute/unmute issue while switching audio devices.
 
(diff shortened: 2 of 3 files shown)


src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -564,7 +564,7 @@ export default class DefaultAudioVideoController
     if (!audioStream || audioStream.getAudioTracks().length < 1) {
       throw new Error('could not acquire audio track');
     }
-
```
aburnson, line 567: Suggest moving this to after the following `awaits` to `replaceAudioTrack` to ensure that the realtime controller is only updated if the track is actually replaced:

```
this._realtimeController.realtimeSetLocalAudioInput(audioStream);
this.meetingSessionContext.activeAudioInput = audioStream;
```

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -564,7 +564,7 @@ export default class DefaultAudioVideoController
...
+    this._realtimeController.realtimeSetLocalAudioInput(audioStream);
```
aburnson, line 567: Suggest moving this to after the following `awaits` to `replaceAudioTrack` to ensure that the realtime controller is only updated if the track is actually replaced:

```
this._realtimeController.realtimeSetLocalAudioInput(audioStream);
this.meetingSessionContext.activeAudioInput = audioStream;
```

ltrung: When we reach this point in the code, the local audio input stream (or active audio input in deviceController) already switches to the new stream right. So I thought it would make sense to do it here

aburnson: What matters however is the track that is attached to the peer connection.

ltrung: @aburnson  Correct me if I am wrong but let's say the replaceTrack failed for some reason and we did not update realtime local audio input. At this point the local audio input device already switches to a new device (Let's say to device B from device A). Since the realtime local audio input is still A, realtime Mute will try to mute that instead of B.
Isn't that the bug we are fixing here?

aburnson (approved):

Issue #1153, opened 2021-03-25
Crash when connecting to a meeting with an attendee already in meeting with content share


rlebosse:
In 2 different windows of Google Chrome : WINDOW_1 and WINDOW_2 :
* WINDOW_1 : connects to a Chime Meeting and **starts a content share** (`startContentShareFromScreenCapture`)
* WINDOW_2 : connects to the same Chime Meeting with the same attendee credentials

Results :
* ✅ WINDOW_1 receives `audioVideoDidStop` event with status code `MeetingSessionStatusCode.AudioJoinedFromAnotherDevice`
* ❌ WINDOW_2 : there is the following error in the DevTools console :
![Capture d’écran 2021-03-25 à 15 02 12](https://user-images.githubusercontent.com/2794137/112486598-50715d80-8d7c-11eb-8704-5cb144930651.png)
* ❌ WINDOW_2 receives `audioVideoDidStop` event with status code `MeetingSessionStatusCode.RealtimeApiFailed`
* ✅ WINDOW_2 : if I try to reconnect to the Chime Meeting : it works !

Expected :
* The first time WINDOW_2 tries to connect to the Chime Meeting with the same attendee credentials as WINDOW_1, he should be able to connect to the Chime Meeting

**Platform**
 - OS: macOS 11.2.1
 - Browser: Google Chrome 89
 - Amazon Chime version: 2.6.0

salihsagdilek:
After upgrading to 2.6.2, we started getting this error very often.

![image](https://user-images.githubusercontent.com/9050352/113498180-433e3680-9513-11eb-8e39-c81b63c4467c.png)

ltrung:
@salihsagdilek Do you have the same workflow as @rlebosse which is re-joining with the same attendeeId?

salihsagdilek:
@ltrung  no we're getting this error sometimes trying open camera.
this error not predictable, getting sometimes 
example flows ->
user joined meeting -> open camera -> close camera -> open camera -> getting error sometimes

simmkyu:
Your stack trace indicates that the `getVideoTracks of null` error occurred in the callback passed to the following real-time API:
https://github.com/aws/amazon-chime-sdk-js/blob/8d6e38c23a2e959953d127628b412e86d9522704/src/realtimecontroller/RealtimeController.ts#L59-L71

In the Chime SDK for JavaScript, I could not find any possible path to this error except the following content share API.
[...]

PR #1154, opened 2021-03-25
Fix crash when connecting to a meeting with an attendee already in meeting with content share

rlebosse:
**Issue #:** 1153

**Description of changes:** Check that the content share stream if defined before getting its media tracks


src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -147,7 +147,7 @@ export default class DefaultContentShareController
           return;
         }
         const stream = this.mediaStreamBroker.mediaStream;
-        if (stream.getVideoTracks().length > 0) {
+        if (stream && stream.getVideoTracks().length > 0) {
```
richnew10, line 150: This can be more simply worded as

```
if (stream?.getVideoTracks().length) {
```

richnew10 (commented):
From reading #1153 it looks like the confusion here is that the SDK is conflating "I have started content share and thus have a media stream" with "there is an attendee in the meeting with my attendee ID that's sharing content", and the latter is only an approximation of the former.

This fix might avoid one particular bug, but it's probably worth looking more deeply to consider whether there is a more precise phrasing (e.g., adding `|| !this.mediaStreamBroker.mediaStream` to the early return above) or a deeper check that needs to occur elsewhere.

src/contentsharecontroller/DefaultContentShareController.ts
rlebosse: Totally, I have made the modification. Thanks for your feedback @richnew10 !

ltrung:
@rlebosse Thanks for fixing this :). I have created a new PR #1265 with your changes and rebased it and merged it in.

PR #1157, opened 2021-03-25
Don't select unavailable device - show notification when trying to st…

p-foucht:
- Don't select "unavailable device" when no devices are present
- Don't toggle camera UI until after our AV calls
- Show notification if user tries to start video with "unavailable device"
- remove js-sdk version tag from bottom right
- Fix grey "active" state buttons that should be white

**Issue #:**

**Description of changes:**

**Testing**

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -36,7 +36,6 @@ import {
   RemovableAnalyserNode,
   SimulcastLayers,
   TimeoutScheduler,
-  Versioning,
   VideoFrameProcessor,
   VideoInputDevice,
   VideoSource,
@@ -129,6 +128,8 @@ const VOICE_FOCUS_SPEC = {
   revisionID: VOICE_FOCUS_REVISION_ID,
   paths: VOICE_FOCUS_PATHS,
 };
+
+const DEVICE_UNAVAILABLE = 'Device selection unavailable';
 
 type VideoFilterName = 'CircularCut' | 'Segmentation' | 'None';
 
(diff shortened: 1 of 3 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -574,22 +573,41 @@ export class DemoMeetingApp
     });
 
     const buttonVideo = document.getElementById('button-camera');
+    const videoNotif = document.getElementById('no-video-notif');
+    let notifTimeout: number | null = null;
```
richnew10, line 577: This should be `undefined | ReturnType<typeof setTimeout>`.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1544,16 +1562,18 @@ export class DemoMeetingApp
...
-      await this.audioVideo.listVideoInputDevices(),
+      videoDevices,
       []
     );
     this.populateInMeetingDeviceList(
       'dropdown-menu-camera',
       genericName,
-      await this.audioVideo.listVideoInputDevices(),
+      videoDevices,
       [],
```
richnew10, line 1577: Some of these fixes we really need to backport…

richnew10 (approved):

demos/browser/app/meetingV2/meetingV2.ts
p-foucht: Hmm, Typescript tells me `setTimeout` returns a `number`. At least when calling `window.setTimeout` compared to just `setTimeout`, which I think the compiler thinks it might be running in a node environment that has a different return type or something?

richnew10: In a browser `setTimeout` returns a `number`. In Node it returns a structure. `ReturnType` means you don't have to worry about it (you cannot strictly type your TypeScript code for both environments without it!) and it's a nice expressive shorthand to say exactly what you mean, which is you'll use this slot to hold whatever it gives you.

p-foucht: Sweet, haven't ran into `ReturnType` before but will definitely start using it

PR #1162, opened 2021-03-30
Possible workaround for Chromium AudioWorklet issue

richnew10:
Note that this PR is built on top of #1126.

This commit factors out `ExtendedBrowserBehavior` and uses it to decide
whether to recreate a Web Audio context when selecting a transform
device that might use an Audio Worklet. This avoids a Chromium bug:

https://bugs.chromium.org/p/chromium/issues/detail?id=1173656

This commit also reworks the browser detection tests, and adds optional
logging for packets-per-second sent over the WebRTC peer connection.

**Testing**

In pieces.

By hand in the demo app, plus unit tests. More testing needed to find out if the workaround helps.

Yes; see Chromium bug report.

This change factors out `ExtendedBrowserBehavior` and adds `browserBehavior` as an optional third constructor argument to `DefaultDeviceController`. This fixes that important omitted extension point.

It adds a new method to `ExtendedBrowserBehavior` to control this workaround.

No.

---


test/browserbehavior/DefaultBrowserBehavior.test.ts
```diff
@@ -48,103 +71,103 @@ describe('DefaultBrowserBehavior', () => {
...
-      expect(new DefaultBrowserBehavior().requiresNoExactMediaStreamConstraints()).to.eq(false);
+      expect(new DefaultBrowserBehavior().requiresNoExactMediaStreamConstraints()).to.be.false;
       expect(new DefaultBrowserBehavior().supportsSenderSideBandwidthEstimation()).to.be.true;
       const enableUnifiedPlan = true;
       expect(
         new DefaultBrowserBehavior({
           enableUnifiedPlanForChromiumBasedBrowsers: enableUnifiedPlan,
         }).requiresUnifiedPlan()
-      ).to.eq(true);
+      ).to.be.true;
```
devalevenkatesh, line 110: Is `be` a best practice over `eq`?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -138,6 +138,50 @@ export default class DefaultAudioVideoController
...
+    // To log packets per second numbers, set
+    // ```
+    // window.CHIME_LOG_PPS = true;
+    // ```
+    // before creating the AV controller.
+    //
+    // This is not an ongoing supported interface, but it is useful for testing.
+
+    /* istanbul ignore next */
+    if (((global as unknown) as { CHIME_LOG_PPS: boolean }).CHIME_LOG_PPS) {
```
devalevenkatesh, line 151: Could we avoid this in `DefaultAudioVideoController` and add in the demo itself? We have done earlier at few places like: [1](https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/browser/app/meetingV2/meetingV2.ts#L1083), [2](https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/browser/app/meetingV2/meetingV2.ts#L1053).

src/browserbehavior/ExtendedBrowserBehavior.ts
```diff
@@ -0,0 +1,19 @@
...
+ * This interface contains methods on {@link DefaultBrowserBehavior} that were
+ * incorrectly omitted from {@link BrowserBehavior}, plus new methods that
+ * were added since the last major version.
+ *
+ * Members of this interface can change without a major version bump to accommodate new browser
+ * bugs and capabilities. If you extend this type, you might need to rework your code for new minor
+ * versions of this library.
+ */
+export default interface ExtendedBrowserBehavior extends BrowserBehavior {
+  requiresResolutionAlignment(width: number, height: number): [number, number];
```
devalevenkatesh, line 16: Can these methods added as optional methods in the `BrowserBehavior` itself? I think these methods are now in both the interfaces right?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -152,9 +153,11 @@ export default class DefaultDeviceController
   private inputDeviceCount: number = 0;
   private lastNoVideoInputDeviceCount: number;
 
-  private browserBehavior: DefaultBrowserBehavior = new DefaultBrowserBehavior();
-
-  constructor(private logger: Logger, options?: { enableWebAudio?: boolean }) {
+  constructor(
+    private logger: Logger,
+    options?: { enableWebAudio?: boolean },
+    private browserBehavior: ExtendedBrowserBehavior = new DefaultBrowserBehavior()
```
devalevenkatesh, line 159: This can be optional as well or since we are providing `DefaultBrowserBehavior` object it is not a breaking change, correct?

Could you please explain in what cases we want to override and have this option in constructor? In the description you have said "This fixes that important omitted extension point.", but could not understand it.

devalevenkatesh: I see you have a unit test with this case, but trying to understand from builders perspective is this required?

ltrung (commented):
What more testing we need to do to verify that this workaround works? If it needs QA testing, can we deploy a test link to QA to verify before merging to main branch?

test/browserbehavior/DefaultBrowserBehavior.test.ts
richnew10: For booleans, yes.

src/audiovideocontroller/DefaultAudioVideoController.ts
richnew10: I'm not sure if I was able to get it to work without poking directly at the sender. I want customers to have an easy way to determine whether they are affected by this issue. If we can make it a handful of lines of code to add, then sure, but otherwise I think it makes sense to be in core.

src/browserbehavior/ExtendedBrowserBehavior.ts
richnew10: `DDC` is written to only work with a `DefaultBrowserBehavior`. It requires these methods, and generalizing the `DDC` constructor to allow the behavior object to be specified requires either making the argument a `DBB` or introducing an interface. 

Making them optional in the existing public interface doesn't actually help much, because not implementing the methods means that DDC doesn't know what to do.

We _could_ make an additional interface called `MoreBrowserBehavior`, and make the argument to `DDC` be of type `BrowserBehavior & MoreBrowserBehavior`, but I chose to just extend the type instead.

You could view `ExtendedBrowserBehavior` as a shorter name for `TheKindOfBrowserBehaviorThatOurDeviceControllerActuallyRequiresToFunction`.

src/devicecontroller/DefaultDeviceController.ts
richnew10: This is an optional argument — that's implied by the `=`. And yes, it's not a breaking change because we are defaulting to `DBB`.

I have _almost_ made this change a bunch of times, because I kept needing to supply different behavior — e.g., this would eliminate the need for much of our DOM mocking in tests!

A builder might provide the argument here if they want a different implementation of any of the methods in `BB`. For example, they might know that their transform devices don't use Audio Worklet, and want to return `false` from the new method I added because they don't need this workaround.

Or they might want to work around a bug in the SDK's feature detection (e.g., to insist that screen share is supported) or force a specific video codec.

The alternative to this is that they have to implement their own `DeviceController`, or break encapsulation and poke at the internals of `DDC`.

richnew10:
> What more testing we need to do to verify that this workaround works? If it needs QA testing, can we deploy a test link to QA to verify before merging to main branch?

I have manually tested and the workaround seems to work. Mostly I want to make sure that it doesn't regress anything else.

src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -153,6 +153,36 @@ export default class DefaultBrowserBehavior implements BrowserBehavior {
     return this.isSamsungInternet();
   }
 
+  requiresContextRecreationForAudioWorklet(): boolean {
+    // Definitely not Chrome; no worries.
+    if (!('chrome' in global)) {
+      return false;
+    }
+
+    const mac = navigator.userAgent.match(/Macintosh/) !== null;
```
ltrung, line 162: Isn't it simpler to use the detect-browser dependency this.browser.os?

src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -153,6 +153,36 @@ export default class DefaultBrowserBehavior implements BrowserBehavior {
...
+
+    // Everything seems to work fine on platforms other than macOS.
+    if (!mac) {
+      return false;
+    }
+
+    // Electron counts… but Electron apps often have messy UAs, so detect it here first.
+    const electron = navigator.userAgent.match(/Electron\/([0-9]+)/) !== null;
```
ltrung, line 170: Isn't Electron identified as "Chrome"?

src/browserbehavior/ExtendedBrowserBehavior.ts
```diff
@@ -0,0 +1,19 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { BrowserBehavior } from '..';
+
+/**
+ * This interface contains methods on {@link DefaultBrowserBehavior} that were
+ * incorrectly omitted from {@link BrowserBehavior}, plus new methods that
+ * were added since the last major version.
+ *
+ * Members of this interface can change without a major version bump to accommodate new browser
```
ltrung, line 11: Could you elaborate on what this means? Aren't we allowed to modify any interface before a major version bump?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -288,11 +291,67 @@ export default class DefaultDeviceController
...
+       * This is not safe in the general case: an application that already
+       * retrieved the audio context in order to build an audio graph for some other purpose
+       * will fail at this point as we pull the context out from under it.
+       *
+       * An application that always uses the supplied context in an
+       * `AudioTransformDevice.createAudioNode` call should work correctly.
+       *
+       * If you are confident that your application does not use AudioWorklet, does not run in
+       * an un-fixed Chromium version, or will never be used with sample-rate-switching Bluetooth
+       * devices, you can leave this hack disabled.
```
ltrung, line 321: How do one disable this? Do you mean they create their own device controller implementation without this?

src/browserbehavior/DefaultBrowserBehavior.ts
richnew10: `detect-browser` seems to be wrong more often than it's right, so I chose to be very specific here instead.

src/browserbehavior/DefaultBrowserBehavior.ts
richnew10: Lemme see if I can cut this down.

src/devicecontroller/DefaultDeviceController.ts
richnew10: See conversation above about passing in a new `ExtendedBrowserBehavior` instance! I'll leave a note.

src/browserbehavior/ExtendedBrowserBehavior.ts
richnew10: This is a note to customers warning them that this interface can change without a major version bump.

src/browserbehavior/ExtendedBrowserBehavior.ts
devalevenkatesh: Ok got it! Thanks for the explanation.

src/devicecontroller/DefaultDeviceController.ts
devalevenkatesh: Yup, I agree with the flexibility this introduces. But, if builders provide their `DBB` implementation to the `DDC` then the same browser behavior should be passed to other classes like the `DAV` cause, now if any method implementation contradicts then the behavior would be different and a flow may run into error, right? We have used different instances of `DBB` for e.g. [1](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/audiovideocontroller/DefaultAudioVideoController.ts#L122), [2](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/audiovideocontroller/DefaultAudioVideoController.ts#L235) which is why I am little cautious here.

richnew10:
This isn't quite ready.

richnew10:
Two new fixes here:

* I unbusted `npm run start:hot` in some cases.
* This now unhooks the old transform device from the context. Tested switching VF on and off, switching devices, etc.

src/devicecontroller/DefaultDeviceController.ts
richnew10: If we are currently relying on asking the same question in two or three places and getting the same answer, we should probably not do that.

Yes, we should allow passing the customer's provided behavior implementation around. No, I don't intend to do that in this PR because it'll keep getting bigger :)

devalevenkatesh: So would it be better to do the constructor change in `DDC` when we actually update the rest of the classes which need `DBB`? I agree that it would be a bigger PR and can be taken up later as well as we really need to use `DBB` as a singleton to avoid this issue.

richnew10: No, because a customer who wants to opt out of this workaround needs a way to do so.

richnew10: (Also: no singletons!)

devalevenkatesh: Yeah it should be same object no singleton. Thanks for clarifying the use here.

demos/browser/package.json
```diff
@@ -7,7 +7,7 @@
     "build:fast": "tsc && webpack --config ./webpack.config.js",
     "build": "npm run deps && npm install && npm run build:fast",
     "start:fast": "npm run build:fast && node serve.js",
-    "start:hot": "tsc && webpack-dev-server --config ./webpack.config.js",
```
devalevenkatesh, line 10: Should we also update the `package.json` to remove the `webpack-dev-server` dependency if we are removing it from here?

richnew10: I didn't remove it because I didn't have the time to verify whether it was still needed. Happy to tackle that in a follow-up.

devalevenkatesh: Sounds good!

ltrung (approved):

PR #1171, opened 2021-04-02
Documentation fixes and exclude private members from public API documentation

richnew10:
I've been applying some tooling, and found a bunch of broken links, malformed syntax, etc. in our documentation comments.

This PR fixes essentially every one of them.

Additionally, this PR adds the `--excludePrivate` flag to our TypeDoc configuration, which prevents private members and methods from being included in generated documentation. This is a significant improvement in clarity.

---


richnew10:
One problem here is that TypeDoc doesn't actually support the TSDoc standard. For example TSDoc says that params should have a hyphen:

```typescript
/**
 * Frob the bar.
 * @param foo - the foo to use.
 */
function frob(foo: number): void {
}
```

while TypeDoc expects there to be no hyphen, and actually renders it into the doc output.

I'm going to split up this PR into fixes that will work with TypeDoc so we can land them now, and keep the rest for later.

richnew10:
Content Share test passed (as you would expect — there are no code changes in this PR) but ended with a cancel.

src/signalingprotocol/SignalingProtocol.d.ts
```diff
@@ -66,7 +66,7 @@ export class SdkSignalFrame implements ISdkSignalFrame {
 
     /**
      * Constructs a new SdkSignalFrame.
-     * @param [properties] Properties to set
+     * @param properties Properties to set
```
XHatan, line 69: is this change manual ? SignalingProtocol.d.ts is auto-generated through script/generate-signaling-protocol, maybe we should update the script ?

richnew10: It is a manual change.

The protobufjs version used in that script is outdated, as is its dependency on jsdoc — and neither of them produce the correct output.

generate-signaling-protocol isn't run automatically.

I will make a note to write a postprocessing script as a follow-up to make it generate correct output.

src/backoff/BackoffFactory.ts
```diff
@@ -6,13 +6,11 @@ import Backoff from './Backoff';
 export default interface BackoffFactory {
   /**
    * Backoff factory method
-   * @returns {Backoff}
```
devalevenkatesh, line 9: This is returning `Backoff` but, we are removing this?

src/datamessage/DataMessage.ts
```diff
@@ -31,7 +31,7 @@ export default class DataMessage {
   /**
    * true if server throttled or rejected message
    * false if server has posted the message to its recipients
-   * Throttling soft limit: Rate:100, Burst:200
+   * Throttling soft limit: Rate: 100, Burst:200
```
devalevenkatesh, line 34: Need space in front of 200 as well here, right?

Burest: 200.

src/signalingclient/SignalingClient.ts
```diff
@@ -38,29 +38,29 @@ export default interface SignalingClient {
...
-   * @param {SignalingClientJoin} settings How to configure the Join frame.
+   * @param settings How to configure the Join frame.
    */
   join(settings: SignalingClientJoin): void;
 
   /**
    * Sends a subscribe frame with the given settings.
    *
-   * @param {SignalingClientSubscribe} settings How to configure the Subscribe frame.
+   * @param settings How to configure the Subscribe frame.
```
devalevenkatesh, line 63: So, does it still display the parameter type name or its fine as we already show in the method signature. Does it still highlight or linkify if users want to click and go to it?

We are removing in other classes as well so just trying to understand the end output here.

typedoc.json
```diff
@@ -1,6 +1,7 @@
 {
   "out": "./docs",
   "readme": "README.md",
+  "excludePrivate": true,
   "exclude": [
     "src/signalingprotocol/SignalingProtocol.js",
     "src/signalingprotocol/SignalingProtocol.d.ts",
```
devalevenkatesh, line 7: Since these two files are excluded should we make changes to them?

richnew10: They are still processed by analysis tools like api-extractor, so I fixed them.

src/signalingclient/SignalingClient.ts
richnew10: The syntax I'm fixing here is JSDoc. JavaScript doesn't have a way to type-annotate parameters, so it uses this `{syntax}` to indicate the type.

TypeScript has type annotations, so TypeDoc and other tools _already know_ that `settings` is a `SignalingClientSubscribe` instance, and can link directly to the correct type.

The output is exactly the same; TypeDoc is ignoring the `{SignalingClientSubscribe}` here.

richnew10: Here's what it looks like.

Screen Shot 2021-04-08 at 2 45 46 PM

devalevenkatesh: Cool, thanks for the explanation!

typedoc.json
devalevenkatesh: Got it, so no configuration on the api-extractor level in that case?

richnew10: We'll get there when we get there :joy:

devalevenkatesh (approved):

PR #1184, opened 2021-04-07
Populate video filters after checking for Amazon Voice Focus compatibility.

richnew10:
Previously we would update the device lists multiple times, and update the list of video filters — including loading TensorFlow.js from the CDN — at the same time as performing estimation for Voice Focus.

On the SauceLabs resource-constrained Windows machines, this contention for resources caused the canary to fail about 45% of the time, because the VF estimator correctly reported that the browser was too slow. The canary test expects that the browser is fast enough, and thus that VF not being supported means that there's a bug. That's not the case: we're just accidentally doing too much at one time.

Eliminating the parallel loading of video filters, and reducing the redundant device refreshes, should make the canary less likely to fail for spurious performance reasons.

---

### Have you successfully run npm run build:release locally?

Yes.

### How did you test these changes?

This is a demo-only change. Tested by hand. Menus are populated correctly, including filters in Chrome.

### Can these changes be tested using the demo application? If yes, which demo application can be used to test it?

By definition.

### Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved?

No.

### Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved?

No.

---

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -417,7 +417,6 @@ export class DemoMeetingApp
         if (this.supportsVoiceFocus) {
           logger.info('[DEMO] Amazon Voice Focus is supported.');
           document.getElementById('voice-focus-setting').classList.remove('hidden');
-          await this.populateAllDeviceLists();
           return;
         }
       }
@@ -428,7 +427,6 @@ export class DemoMeetingApp
     logger.warn('[DEMO] Does not support Amazon Voice Focus.');
     this.supportsVoiceFocus = false;
     document.getElementById('voice-focus-setting').classList.toggle('hidden', true);
-    await this.populateAllDeviceLists();
   }
 
   private async onVoiceFocusSettingChanged(): Promise<void> {
@@ -497,6 +495,8 @@ export class DemoMeetingApp
           (document.getElementById('info-name') as HTMLSpanElement).innerText = this.name;
 
           await this.initVoiceFocus();
+          await this.populateAllDeviceLists();
+          await this.populateVideoFilterInputList();
 
           this.switchToFlow('flow-devices');
           await this.openAudioInputFromSelectionAndPreview();
@@ -1257,7 +1257,6 @@ export class DemoMeetingApp
     }
     this.audioVideo.addDeviceChangeObserver(this);
     this.setupDeviceLabelTrigger();
-    await this.populateAllDeviceLists();
     this.setupMuteHandler();
     this.setupCanUnmuteHandler();
     this.setupSubscribeToAttendeeIdPresenceHandler();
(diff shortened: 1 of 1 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1694,7 +1694,6 @@ export class DemoMeetingApp
   async populateAllDeviceLists(): Promise<void> {
     await this.populateAudioInputList();
     await this.populateVideoInputList();
-    await this.populateVideoFilterInputList();
```
devalevenkatesh, line 1697: The `populateAllDeviceLists()`  is also called in the `initializeMeetingSession` method [here[(https://github.com/aws/amazon-chime-sdk-js/blob/c13b6cce5e2d35841b7cb4fac2df2f184a4d45b2/demos/browser/app/meetingV2/meetingV2.ts#L1260) in the demo. So, should we keep it and remove `await this.populateVideoFilterInputList();` from the `onVoiceFocusSettingChanged()`?

richnew10: The change isn't in `onVoiceFocusSettingChanged`, it's in the `authenticate` handler that's configured in `initEventListeners`; GitHub is showing context sub-optimally.

With this change we init VF, the device lists, and the video filter after you authenticate. When you authenticate we _also_ init the device lists. If anything, it's that second `populateAllDeviceLists` call that can go.

richnew10: I removed the additional call in `initializeMeetingSession` and tested.

devalevenkatesh: Got it, Github did not show the change under `initEventListeners()`. Thanks!

ltrung (commented):
Could you fill in the template for the PR and give more details about the changes?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -497,6 +495,8 @@ export class DemoMeetingApp
           (document.getElementById('info-name') as HTMLSpanElement).innerText = this.name;
 
           await this.initVoiceFocus();
+          await this.populateAllDeviceLists();
```
ltrung, line 498: So it seems we only call `populateAllDeviceLists` once together with `populateVideoFilterInputList`. So do we need to separate `populateVideoFilterInputList` from `populateAllDeviceLists`?

richnew10: The same line of reasoning would suggest combining `initVoiceFocus` with `populateAllDeviceLists`, no? Those are also both called only once and called right here.

I think the names of these functions now accurately describe what they do, each is <50 lines long and is individually understandable, and the point at which they are called accurately represents the order of execution. This is the ideal structure for code.

ltrung (approved):

PR #1188, opened 2021-04-09
[Integration test] Update README with KITE setup build error instruct…

devalevenkatesh:
…ions

**Issue #:**
- KITE installation runs into an maven build error while building the `KITEAppRTC_test` as per current steps in the `integration/js` README. Thus, we cannot run any integration test locally using local grid or saucelabs.

**Description of changes:**
- A README update to add instructions to solve the above issue.

**Testing**

2. How did you test these changes? By triggering few integration test runs on SauceLabs from local machine.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? NA.


integration/js/README.md
```diff
@@ -66,6 +66,30 @@ export PATH="$PATH:$KITE_HOME/scripts/mac/path/"
...
+cd ~/.m2
+touch settings.xml
+```
+2. Put below content in the created `settings.xml` file.
+```xml
+<settings>
+  <mirrors>
+    <mirror>
+      <id>internal-repository</id>
+      <name>Maven Repository Manager running on repo.mycompany.com</name>
```
richnew10, line 84: This doesn't seem right.

integration/js/README.md
```diff
@@ -66,6 +66,30 @@ export PATH="$PATH:$KITE_HOME/scripts/mac/path/"
 c
 ```
 
+If you run into a build error here, do the following:
+
+Credits: https://github.com/webrtc/KITE/issues/181
+1. Create the `settings.xml` file in the `.m2` directory.
```
richnew10, line 72: The only thing in my `settings.xml` is my GitHub username, and I last successfully ran KITE tests a couple of weeks ago. Are you sure you need to put springsource in here?

richnew10: Oh, I see the ticket. Why is the springsource mirror blocked?

devalevenkatesh: Not sure why it is blocked. I checked into the error but did not found much reported around KITE and then found that github issue. Also, I did not have any `settings.xml` file.

This is using the internal repository in `.m2` as a mirror rather than external, so is this not an appropriate fix here? I could not found any other solution to get KITE running locally following the setup instruction steps mentioned in the `integration/js` README.

Issue #1172, opened 2021-04-02
Eliminate Node.js 10.x support from documentation (and code)


richnew10:
Node.js 10.x entered LTS in October 2018, entered maintenance in April 2020, and hit EOL in April 2021.

We do not test against Node.js 10, and we should not recommend that customers use versions past their end of life.

We should recommend that customers use Node.js 14.x, which is in active support until October 2021, or Node.js 12.x, which will remain in maintenance mode until April 2022.

We should also consider using the `engines` field in package.json to specify our supported Node.js versions.

PR #1191, opened 2021-04-09
Move config files into /config; use v2 lockfile format

richnew10:
This PR consists of two commits.

The first moves most of our configuration files — `tsconfig.json`, `eslint` configuration, etc. — into a subdirectory. This is modern style, because as you can see these files tend to proliferate.

I also fixed some malformed JSON.

The second does several related things:

1. It shifts to using the current npm lockfile format, which is v2. This is supported by npm v6 (Node 14) and npm v7 (Node 15). Using v7 will automatically update the files, so this has to be done at some point.
2. It shifts CI to use Node 14 rather than Node 12, because it's not 2019.
3. It shifts the `build` step to use `npm ci` rather than `npm install`. This has several benefits:
   * It prevents colliding `npm install` calls on v7 and v6 from flip-flopping the lockfile back to v1. Use npm v7 to install packages and perform periodic updates. Node 14 is in "Active LTS" and enters maintenance mode in October 2021. Node 15 is the current version as of October 2020.
   * It makes builds reproducible: it will no longer be the case that just building the software can accidentally change which dependencies are in use. You should _deliberately_ upgrade dependencies by running `npm install` and committing the result. CI and developers will now use the lockfile as intended.
4. It switches our `npm audit` behavior to _fail the build_ rather than _modify `package-lock.json`_ if there are vulnerabilities. In both cases the build will fail in CI (the latter after five minutes because there are uncommitted changes!), but this is quicker and avoids `npm audit` flip-flopping the lockfile to v1. This forces us to be aware of and more careful about making audit changes without introducing any risk. Dependabot will continue to send PRs.

**Testing**

Yes.

Build-only. Tests and docs pass, demos build.

N/A.

N/A.

N/A.

---


.eslintignore
```diff
@@ -13,7 +14,6 @@ guides
 index.ts
 integration
 modules
-node_modules
```
richnew10, line 16: `node_modules` is implicitly ignored by eslint.

package.json
```diff
@@ -26,7 +26,7 @@
     "doc": "script/generate-docs",
     "postdoc": "node script/update-typedoc-link.js",
     "test": "npm run test:fast",
-    "test:glob": "cross-env TS_NODE_CACHE=false nyc mocha -r esm -rv ts-node/register -- ",
+    "test:glob": "cross-env TS_NODE_PROJECT=test/tsconfig.json TS_NODE_CACHE=false nyc mocha -v -r esm -rv ts-node/register -- ",
```
richnew10, line 29: Turns out this has been using the default config 🤷

ltrung (commented):
Will it be slower since we basically do a clean install of all the dependencies every time we build? From what i understand, npm install only install dependencies that are in package.json but not in node_modules folder.

.github/workflows/build.yml
```diff
@@ -19,10 +19,10 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Using Node.js 12.x
+      - name: Using Node.js 14.x
         uses: actions/setup-node@v1
         with:
-          node-version: 12.x
+          node-version: 14.x
```
ltrung, line 25: Do we want to update this in other GitHub workflows too?

devalevenkatesh (commented):
We are changing the Node version to 14.X while building so can this be extended to introduce the `engines` field in the `package.json`?

richnew10:
Last I checked, `engines` was kinda useless — has that changed?

.github/workflows/build.yml
richnew10: We should!

richnew10:
> Will it be slower since we basically do a clean install of all the dependencies every time we build? From what i understand, npm install only install dependencies that are in package.json but not in node_modules folder.

@ltrung yes, if the lockfile is up to date, `npm ci` is slower because it recreates `node_modules` every time. On a fresh checkout `npm ci` is much faster.

I think the right thing to do here is to _not run either of these steps_ as part of the build.

Local developers should run either `npm install` or `npm ci` whenever `package-lock.json` changes or they begin development. 

CI should run `npm ci` prior to starting a build.

`npm run build` should just perform build steps using what's in `node_modules`.

The issue then is how to make sure that developers run `npm ci` again when `package-lock.json` changes. To do that we need to store a hash of the lockfile; the presence and correctness of that hash can drive whether `npm ci` is called. I'll write that today.

script/conditional-install
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env ruby
+require 'digest'
+
+def verbose command
+  puts("--> #{command}") || system(command) || fail("Failed: #{command}")
+end
+
+# Why not read node_modules/.package-lock.json? Because it's not the same.
```
ltrung, line 8: I confuse here. What is .package-lock.json?

.github/workflows/build.yml
ltrung: Do we want to update them in this PR :)

ltrung:
> Last I checked, `engines` was kinda useless — has that changed?

I think Venky referred to issue #1172  that you created.

>We should also consider using the engines field in package.json to specify our supported Node.js versions.

script/conditional-install
richnew10: It's a file that `npm` itself creates, which _you would think_ corresponds to the `package-lock.json` that was used to populate `node_modules`… but it's not. I left that comment for the next person who stumbles on this script and thinks "why not just use that?".

richnew10:
> > Last I checked, `engines` was kinda useless — has that changed?
> 
> I think Venky referred to issue #1172 that you created.
> 
> > We should also consider using the engines field in package.json to specify our supported Node.js versions.

Gotcha; I thought you were suggesting it for some functional reason.

script/conditional-install
ltrung: I checked my local node_modules folder but did not see any .package-lock.json file. Is it something in NPM 7 (I am using 6).

package.json
```diff
@@ -8,25 +8,30 @@
     "libs",
     "build"
   ],
+  "engines": {
+    "node": "^12 || ^14 || ^15",
```
ltrung, line 12: The README says Node 10 or later. Should we update it to 12?

script/conditional-install
richnew10: I don't think so; I'm using 6.14 on this machine and it's there.

ltrung (approved):

PR #1196, opened 2021-04-13
[Integration test] Update integration test server and add attendee presence timeout to audio, video test only

devalevenkatesh:
**Issue #:**
- Currently, some canary runs are failing with the error:
```
Unable to create meeting, server error: Error: Request failed with status code 403
```

- This is due to the incorrect handling of baseURL in the integration server.

The current demo URL is (incorrect URL hence error from server):
```
<server>/?attendee-presence-timeout-ms=5000join?title=<title>&name=<name>&region=<region>
```
After this fix it will be:
```
<server>/join?attendee-presence-timeout-ms=5000&title=<title>&name=<name>&region=<region>
```

**Description of changes:**
- This fix handles the query params if any and accordingly generates the create meeting URL.
- Adds attendee presence timeout to the demo url when running Audio and Video test only. The canary demo link is now changed back to just the URL with no query params.
- The `integration/js/server.js` change can be reverted because we reverted the query parameter on the gamma demo URL, but it will catch such cases where we add any other query parameter to demo URLs and run test against it.

**Testing**
- By running the integration tests on the local selenium grid.

3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? NA. This is integration test change only.


integration/js/server.js
```diff
@@ -33,13 +33,19 @@ createMeeting = async (baseUrl) => {
   }
   meetingCreationStatus = "creating";
   meetingTitle = uuidv4();
-  const createMeetingUrl = `${baseUrl}join?title=${meetingTitle}&name=MeetingOwner&region=us-east-1`;
+  let url = `${baseUrl}`;
+  let index = url.indexOf('?');
```
michhyun1, line 37: url.includes('?')

michhyun1: Can we use this instead?

devalevenkatesh: I needed the index for the `substring` so used `indexOf`.

integration/js/utils/SdkBaseTest.js
```diff
@@ -20,6 +20,10 @@ class SdkBaseTest extends KiteBaseTest {
       this.userArn = this.payload.userArn;
     }
 
+    if (['Video', 'Audio'].includes(testName)) {
```
michhyun1, line 23: Just double checking - this is not backwards? Will 'testname' be Video or Audio?

michhyun1: Just checked - our Audio and Video E2E test use the name "Audio" and "Video". Thanks

michhyun1 (approved):

PR #1202, opened 2021-04-16
Parallelize tasks; enable early signaling connection establishment.

richnew10:
Opening the signaling websocket is one of the biggest sources of delay
when joining a meeting: even a fast connection to a nearby region can
take over 300ms.

This commit introduces an argument to `start` to allow a 'pre-start'
behavior to occur: call `start({ signalingOnly: true })` to begin signaling, and then call
`start()` to complete peer connection setup and media negotiation.

This change also slightly parallelizes some start tasks.

**Testing**

Yes.

With the meeting demo.

See above.

Yes: `AudioVideoFacade.start` now takes an options map as an argument, allowing a signaling-only start to occur.

The alternative is to add a new method to the facade and controller interfaces.

No, though this does cause the meeting app to establish a signaling connection earlier — in the lobby. Customer applications doing so will cause a small increase in load on the service.

---


src/audiovideofacade/DefaultAudioVideoFacade.ts
```diff
@@ -48,8 +48,8 @@ export default class DefaultAudioVideoFacade implements AudioVideoFacade {
     this.audioVideoController.setAudioProfile(audioProfile);
   }
 
-  start(): void {
-    this.audioVideoController.start();
+  start(options?: { signalingOnly?: boolean }): void {
```
aburnson, line 51: Benefits of this approach over adding an explicit top-level `prestart()` method?

richnew10: Firstly, it avoids creating an optional method on a core interface -- a customer's `AudioVideoController` subclass doesn't need to change even if the facade does, and that's true even without having to check `if (this.audioVideoController.prestart) { ... }`.

Secondly, this is a one-to-two transition, from "just start" to "maybe there are several kinds or phases of starting", and so I'm looking ahead to the next: check-in mode, full connection but don't start sending media yet, whatever else pops up.

I don't feel strongly about it, and you have both deeper knowledge of the signaling flow and longer experience with the design intent of this API, so happy to receive guidance.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -359,47 +390,53 @@ export default class DefaultAudioVideoController
...
+                new ListenForVolumeIndicatorsTask(this.meetingSessionContext),
+                new SendAndReceiveDataMessagesTask(this.meetingSessionContext),
+                new JoinAndReceiveIndexTask(this.meetingSessionContext),
+                new ReceiveTURNCredentialsTask(this.meetingSessionContext),
+                // TODO: ensure index handler does not race with incoming index update
+                new ReceiveVideoStreamIndexTask(this.meetingSessionContext),
+              ]),
+              new SerialGroupTask(this.logger, 'Peer', [
+                new CreatePeerConnectionTask(this.meetingSessionContext),
+                new AttachMediaInputTask(this.meetingSessionContext),
```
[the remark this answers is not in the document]
richnew10: In an ambitious alternative future we rework all of these tasks and all of the fields they use (e.g., the signaling client itself) to be a tree of promises, allowing the dependency tree and timeouts to control order of execution and parallelism. Maybe not today!

The approach I took first was to define `PromiseTask`, which just waits on an already-established promise. As you saw, I reworked that to `OnceTask`, allowing the same `Task` to be present multiple times in the task tree. We can use that same trick here and elsewhere by representing a resolves-before relationship by putting the dependent task earlier in the same serial group:

```typescript
  new SerialGroupTask(this.logger, 'Peer', [
    new CreatePeerConnectionTask(this.meetingSessionContext),
    previousReceiveAudioInputTask,
    new AttachMediaInputTask(this.meetingSessionContext),
```

This gives us a synchronization barrier without much work. Thoughts?

richnew10:
Evaluating performance of this:

* By default, the only change will be to parallelize meeting join. The signaling change is behind a checkbox that is off by default. Specify `earlyConnect=1` in the URL to turn it on, or check it by hand.
* The entire connection flow is in a serial task labeled `AudioVideoStart`.
* If you check the checkbox, `OpenSignalingConnectionTask` will not contribute to `AudioVideoStart` in most cases.

So: compare the `AudioVideoStart took` values between this and a previous release, and then compare with `?earlyConnect=1` in the URL to see the preconnect improvement.

richnew10:
Three failures to work on.

richnew10:
Testing in Sydney from the west coast of the US:

```
Early join, promises:
AudioVideoStart took 976 ms
AudioVideoStart took 952 ms

No early join, promises:
AudioVideoStart took 1712 ms
AudioVideoStart took 1757 ms
AudioVideoStart took 1959 ms

No early join, task:
AudioVideoStart took 1745 ms
AudioVideoStart took 1912 ms
AudioVideoStart took 2011 ms
```

It looks like the promise-based approach in this Chrome fast path (permissions already granted etc.) shaves off 30–50msec, and the early signaling connection establishment saves that entire time, which is about a second to Sydney.

The long and the short of it: this change *halves* the time to enter the meeting for people in the most distant regions, and saves about 300ms for those close to their meeting region.

richnew10:
N.B., I will write some additional guidance docs for this as a follow-up, but I didn't want to block review (and potential landing) on that.

I suggest flipping on the checkbox in the meeting demo for the majority of integration tests: it will speed up the tests in general, but leave a few to take the slow path.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -104,8 +105,12 @@ export default class DefaultAudioVideoController
   private enableSimulcast: boolean = false;
   private totalRetryCount = 0;
   private startAudioVideoTimestamp: number = 0;
+  private signalingTask: Task;
   destroyed = false;
 
+  /** @internal */
+  usePromises: boolean = true;
+
```
richnew10, line 113: Leaving this in for a little while to have an easier way to revert this change if needed.

src/task/NoOpTask.ts
```diff
@@ -11,7 +11,7 @@ export default class NoOpTask implements Task {
   }
 
   run(): Promise<void> {
-    return;
+    return Promise.resolve();
   }
```
richnew10, line 15: This fix is the result of a disappointingly large amount of wasted time wondering why my tests were broken!

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -242,6 +244,166 @@ export default class DefaultAudioVideoController
...
+                ]),
+                this.meetingSessionContext.meetingSessionConfiguration.attendeePresenceTimeoutMs
+              )
+            : /* istanbul ignore next */ new SetRemoteDescriptionTask(context),
+        ])
+      ),
+    ]);
+  }
+
+  private connectWithTasks(needsToWaitForAttendeePresence: boolean): Task {
```
richnew10, line 360: This is the old implementation. Will remove when we're confident.

src/task/OnceTask.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (this.promise) {
+      return this.promise;
+    }
+
+    const dependencies = this.dependencies
+      ? Promise.all(this.dependencies.map(d => d?.run()))
+      : Promise.resolve();
+
+    this.ongoing = dependencies.then(() => {
+      this.logDependencies();
```
devalevenkatesh, line 97: Understanding: The `logDependencies` function will be executed once all dependency promises are resolved right ? as it is getting called in `then`?

richnew10: Correct.

CHANGELOG.md
```diff
@@ -9,12 +9,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `DefaultAudioVideoFacade.start` now takes an options argument. You can use
+  this to trigger a signaling socket connection prior to device selection: call
+  `audioVideo.start({ signalingOnly: true })`, and then later call
+  `audioVideo.start()` as usual.
+
 ### Changed
 - `startVideoPreviewForVideoInput` uses the active video input stream instead of calling `getUserMedia` again
```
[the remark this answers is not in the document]
richnew10: It got rebased into the wrong place: should be one line higher.

src/task/OnceTask.ts
devalevenkatesh: So then the `logDependencies` function logs 'waiting for dependencies:..', should it be logging 'completed'?

richnew10: Switched this to log after triggering the deps.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -242,6 +244,166 @@ export default class DefaultAudioVideoController
...
+      // Just-in-case error handling.
+      .catch(
+        /* istanbul ignore next */
+        e => {
+          this.logger.error(`start failed: ${e}`);
+        }
+      );
+  }
+
+  // For tests only.
```
ltrung, line 275: nit: Is this accurate since we called it under start()?

src/task/OnceTask.ts
```diff
@@ -0,0 +1,109 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import Logger from '../logger/Logger';
+import LogLevel from '../logger/LogLevel';
+import Task from './Task';
+
+function nextTick(f: () => void): void {
```
ltrung, line 8: Should we use AsyncScheduler here since you introduced nextTick there already?

integration/configs/app_quit_video_test.config.json
```diff
@@ -10,7 +10,7 @@
       "noOfThreads": 1,
       "testImpl": "AppQuitVideoTest.js",
       "payload": {
-        "url": "http://localhost:8080/",
+        "url": "http://localhost:8080/?earlyConnect=1",
```
ltrung, line 13: Is there any specific reason we enable this for these particular tests but not the others?

test/audiovideocontroller/DefaultAudioVideoController.test.ts
```diff
@@ -333,6 +336,123 @@ describe('DefaultAudioVideoController', () => {
...
+      webSocketAdapter.send(makeAudioStreamIdInfoFrame());
+      webSocketAdapter.send(makeAudioMetadataFrame());
+
+      expect(sessionStarted).to.be.true;
+      expect(sessionConnecting).to.be.true;
+
+      await stop();
+    });
+
+    it('can be started with a pre-start', async () => {
```
ltrung, line 391: Do we need a test case for prestart and then stop and make sure we clean up everything correctly?

integration/configs/app_quit_video_test.config.json
richnew10: See earlier commentary — it makes sure we get some coverage on both paths without having to do something redundant and costly like running the entire suite twice.

ltrung (commented):
Could we update the PR Description with how testing should be done?

src/task/OnceTask.ts
richnew10: Sure.

guides/03_API_Overview.md
```diff
@@ -142,14 +142,14 @@ You may optionally listen to the following callbacks to monitor aspects of conne
...
 
-To start the meeting session, call meetingSession.audioVideo.[start()](https://aws.github.io/amazon-chime-sdk-js/interfaces/audiovideofacade.html#start). This method will initialize all underlying components, set up connections, and immediately start sending and receiving audio.
-
-To stop the meeting session, call meetingSession.audioVideo.[stop()](https://aws.github.io/amazon-chime-sdk-js/interfaces/audiovideofacade.html#stop).
+To stop the meeting session, call `meetingSession.audioVideo.[stop()](https://aws.github.io/amazon-chime-sdk-js/interfaces/audiovideofacade.html#stop)`.
 
 The `stop()` method does not clean up observers. You can start and stop a session multiple times using the same observers. In other words observers are not tied to the lifecycle of the session.
 
+You can specify a `signalingOnly` option when calling `start` to cause only the initial signaling connection to be established, then complete meeting join with a second call to `start` after choosing audio and video sources. This can improve meeting join latency.
```
ltrung, line 151: When do builders want to do this? Could we give a more specific use case?

ltrung, line 151: When do builders want to do this? Could we give a more specific use case?

integration/configs/app_quit_video_test.config.json
ltrung: Got it. I understand the reasoning just wonder if there is any specific reason to choose these specific tests :)

guides/03_API_Overview.md
richnew10: It's discussed a little in the FAQ.

src/audiovideocontroller/DefaultAudioVideoController.ts
richnew10: The comment implies "this is public for tests only". I'll clarify 'cos evidently the implication isn't clear enough!

integration/configs/app_quit_video_test.config.json
richnew10: Nope, no reason. I just picked a few at random.

src/audiovideofacade/DefaultAudioVideoFacade.ts
ltrung: Do we see that any app would want to do this as best practice to reduce join time? If so, it doesn't feel like it should be an option.
Is there any plan to change this in future (maybe v3)?

test/audiovideocontroller/DefaultAudioVideoController.test.ts
richnew10: I'll add one.

src/audiovideofacade/DefaultAudioVideoFacade.ts
richnew10: Most apps will want to, but knowing _when_ is up to the application.

guides/03_API_Overview.md
richnew10: Added an example.

aburnson (approved):

devalevenkatesh (approved):

PR #1208, opened 2021-04-19
Add optional header parameter to the MeetingPOSTLogger

RidipDe:
**Description of changes:**
Add optional header parameter to the MeetingPOSTLogger

**Testing**
Unit tests

Yes
Unit tests
Yes, meeting
Yes - added an optional parameter
Yes.
No

### DIFF
--- a/src/logger/MeetingSessionPOSTLogger.ts
+++ b/src/logger/MeetingSessionPOSTLogger.ts
@@ -27,7 +27,8 @@ export default class MeetingSessionPOSTLogger implements Destroyable {
     private batchSize: number,
     private intervalMs: number,
     private url: string,
-    private level = LogLevel.WARN
+    private level = LogLevel.WARN,
+    private headers?: Record<string, string>
   ) {
     this.startLogPublishScheduler(this.batchSize);
 
@@ -105,6 +106,11 @@ export default class MeetingSessionPOSTLogger implements Destroyable {
         const response = await fetch(this.url, {
           method: 'POST',
           body,
+          ...(this.headers
+            ? {
+                headers: this.headers,
+              }
+            : {}),
         });
         if (response.status === 200) {
           this.logCapture = this.logCapture.slice(batch.length);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Add optional header parameter to the MeetingPOSTLogger
 
 ### Changed
 
--- a/docs/classes/meetingsessionpostlogger.html
+++ b/docs/classes/meetingsessionpostlogger.html
@@ -126,7 +126,7 @@
 					<a name="constructor" class="tsd-anchor"></a>
 					<h3>constructor</h3>
 					<ul class="tsd-signatures tsd-kind-constructor tsd-parent-kind-class">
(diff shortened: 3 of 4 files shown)


src/logger/MeetingSessionPOSTLogger.ts
```diff
@@ -27,7 +27,8 @@ export default class MeetingSessionPOSTLogger implements Destroyable {
     private batchSize: number,
     private intervalMs: number,
     private url: string,
-    private level = LogLevel.WARN
+    private level = LogLevel.WARN,
+    private headers?: Record<string, string>
```
simmkyu, line 31: In TypeScript DOM types, `headers` is of the `HeadersInit` type.
https://github.com/microsoft/TypeScript/blob/c671fe1bc2878e1ba60f4199856d1a57c3cc99ff/lib/lib.dom.d.ts#L1618

```
type HeadersInit = Headers | string[][] | Record<string, string>;
```

src/logger/MeetingSessionPOSTLogger.ts
```diff
@@ -102,10 +103,16 @@ export default class MeetingSessionPOSTLogger implements Destroyable {
       const batch = this.logCapture.slice(0, batchSize);
       const body = this.makeRequestBody(batch);
       try {
-        const response = await fetch(this.url, {
-          method: 'POST',
-          body,
-        });
+        const response = this.headers
```
simmkyu, line 106: tip: You can write the code using one `fetch`.

```
const response = await fetch(this.url, {
  method: 'POST',
  body,
  ...(this.headers ? {
    headers: this.headers
  } : {})
});
```

simmkyu (commented):
You need additional tests for `Headers` and `string[][]`. See my comment in the post logger.

src/logger/MeetingSessionPOSTLogger.ts
RidipDe: Discussed with Kyu offline, based on the existing pattern in the JS SDK https://github.com/aws/amazon-chime-sdk-js/blob/6596622238e69c5202c3e697aeeedd54f2a0d35b/libs/voicefocus/types.d.ts#L125 we will follow only the Record<string, string> type as of now.

src/logger/MeetingSessionPOSTLogger.ts
RidipDe: Thanks, great to know!
Updated to this format.

package.json
```diff
@@ -79,6 +79,7 @@
   "dependencies": {
     "@types/dom-mediacapture-record": "^1.0.7",
     "@types/ua-parser-js": "^0.7.35",
+    "aws-embedded-metrics": "^2.0.4",
```
simmkyu, line 82: This dependency doesn't seem to be related to this PR?

src/logger/MeetingSessionPOSTLogger.ts
```diff
@@ -105,6 +106,11 @@ export default class MeetingSessionPOSTLogger implements Destroyable {
         const response = await fetch(this.url, {
           method: 'POST',
           body,
+          ...(this.headers
```
simmkyu, line 109: nit: for readability.. Ignore this commit if the Prettier has automatically fixed your code.

```
...(this.headers ? {
  headers: this.headers
} : {}),
```

package.json
RidipDe: Hmm, thats weird. Have removed it.

src/logger/MeetingSessionPOSTLogger.ts
RidipDe: npm run build:release is fixing that

simmkyu (approved):

PR #1211, opened 2021-04-21
Add extra logging for Content Share

RidipDe:
**Description of changes:**
Adding extra logs

**Testing**

Yes
Manual
Yes, meeting
No
No

### DIFF
--- a/src/contentsharecontroller/ContentShareMediaStreamBroker.ts
+++ b/src/contentsharecontroller/ContentShareMediaStreamBroker.ts
@@ -23,6 +23,7 @@ export default class ContentShareMediaStreamBroker implements MediaStreamBroker 
 
   async acquireAudioInputStream(): Promise<MediaStream> {
     if (this._mediaStream.getAudioTracks().length === 0) {
+      this.logger.info('No audio stream available. Synthesizing an audio stream.');
       return DefaultDeviceController.synthesizeAudioDevice(0) as MediaStream;
     }
     return this._mediaStream;
--- a/src/contentsharecontroller/DefaultContentShareController.ts
+++ b/src/contentsharecontroller/DefaultContentShareController.ts
@@ -164,6 +164,9 @@ export default class DefaultContentShareController
           return;
         }
         const stream = this.mediaStreamBroker.mediaStream;
+        this.contentAudioVideo.logger.info(
+          `The content share stream has ${stream.getVideoTracks().length} video track(s)`
+        );
         if (stream.getVideoTracks().length > 0) {
           this.contentShareTile = this.attendeeAudioVideo.videoTileController.addVideoTile();
           const track = stream.getVideoTracks()[0];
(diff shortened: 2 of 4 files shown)


src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -187,6 +188,7 @@ export default class DefaultContentShareController
             this.contentAudioVideo.configuration.credentials.externalUserId
           );
         }
+        this.contentAudioVideo.logger.info('Setting up the ContentShareDidStart observer');
```
ltrung, line 191: I don't think this add any value.

src/contentsharecontroller/ContentShareMediaStreamBroker.ts
```diff
@@ -23,6 +23,7 @@ export default class ContentShareMediaStreamBroker implements MediaStreamBroker
 
   async acquireAudioInputStream(): Promise<MediaStream> {
     if (this._mediaStream.getAudioTracks().length === 0) {
+      this.logger.info('Synthesizing fake audio stream');
```
ltrung, line 26: Should we not use the word fake? Maybe something like 'No audio stream available. Synthesize an audio stream'

simmkyu, line 26: We use the term "empty audio."
https://github.com/aws/amazon-chime-sdk-js/blob/7d7d0cbebdfc1a11113001ca76f2b7eddf379d7a/src/devicecontroller/DefaultDeviceController.ts#L943-L945

How about the following message?
> Create an empty audio device for the content share

src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -165,6 +165,7 @@ export default class DefaultContentShareController
         }
         const stream = this.mediaStreamBroker.mediaStream;
         if (stream.getVideoTracks().length > 0) {
+          this.contentAudioVideo.logger.info('Stream has video tracks');
```
simmkyu, line 168: To distinguish from the regular audio-video logs, include the term "content share." e.g., `The content share stream has ${stream.getVideoTracks().length} video track(s)`

src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -164,6 +164,9 @@ export default class DefaultContentShareController
           return;
         }
         const stream = this.mediaStreamBroker.mediaStream;
+        this.contentAudioVideo.logger.info(
```
ltrung, line 167: This should be just this.logger.info

src/contentsharecontroller/ContentShareMediaStreamBroker.ts
RidipDe: Changed!!

src/contentsharecontroller/DefaultContentShareController.ts
RidipDe: Added this log line.

src/contentsharecontroller/DefaultContentShareController.ts
RidipDe: DefaultContentShareController.ts does not have direct reference to the logger.

simmkyu (approved):

PR #1216, opened 2021-04-22
Adding priority based downlink policy

lzongkun:
**Issue #:**
Adding priority based downlink policy
**Description of changes:**
Adding priority based downlink policy
Related PR: https://github.com/XHatan/mirror-amazon-chime-js-sdk/pull/21
**Testing**

2. How did you test these changes? Tested with demo app
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes, meetingV2
4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? No public API/interface changes


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -211,6 +214,8 @@ interface Toggle {
   action: () => void;
 }
 
+var demoApp:DemoMeetingApp;
```
RidipDe, line 217: nit: var -> let

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2461,6 +2481,55 @@ export class DemoMeetingApp
         }
   }
 
+  videoTilePin(e: any) {
```
RidipDe, line 2484: nit: can we define the variable 'e'

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2461,6 +2481,55 @@ export class DemoMeetingApp
...
+    const pinButtonElement = e.target; //document.getElementById(`video-pin-${tileIndex}`) as HTMLButtonElement;
```
RidipDe, line 2485: Pls remove the comment

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2461,6 +2481,55 @@ export class DemoMeetingApp
...
+      } else {
+        pinButtonElement.innerText = 'Unpin';
+        this.roster[attendeeId].pinned = true;
+        this.updateRoster();
+      }
+      this.updateDownlinkPreference();
+  }
+
+  updateDownlinkPreference(): void {
+    let videoPreferences: VideoPreference[] = [];
```
RidipDe, line 2505: can be const

integration/js/VideoPinTest.js
```diff
@@ -0,0 +1,32 @@
+const {TestUtils} = require('kite-common');
+const SdkBaseTest = require('./utils/SdkBaseTest');
+const {OpenAppStep, AuthenticateUserStep, ClickStartLocalVideoButton, ClickStopLocalVideoButton, ClickUnbindVideoElementButton, ClickBindVideoElementButton} = require('./steps');
```
RidipDe, line 3: `
const { OpenAppStep, AuthenticateUserStep, ClickStartLocalVideoButton, ClickStopLocalVideoButton, ClickUnbindVideoElementButton, ClickBindVideoElementButton } = require('./steps');
`

Need spaces { abc }
npm run build:release will prettify the code.

integration/js/VideoPinTest.js
```diff
@@ -0,0 +1,32 @@
...
+    await VideoPreferenceCheck.executeStep(this, session, 'attendeeId', 'priority', 'targetSize');
+    await this.waitAllSteps();
+  }
+}
+
+module.exports = VideoPinTest;
+
+(async () => {
+  const kiteConfig = await TestUtils.getKiteConfig(__dirname);
+  let test = new VideoPinTest('Test App Video test', kiteConfig);
```
RidipDe, line 30: nit: let -> const

RidipDe: npm run build:release should fix these lint errors

integration/js/app/video_test/video_test.ts
```diff
@@ -113,6 +118,36 @@ export class DemoMeetingApp implements AudioVideoObserver {
...
+        videoPreferenceObject.innerText += JSON.stringify(preference, null, '  ');
+      }
+      pinVideoTileForm.reset();
+    });
+
+    const unpinVideoTileForm = document.getElementById('unpin-video-tile-form') as HTMLFormElement;
+    const unpinAttendeeId = document.getElementById('unpin-attendee-id') as HTMLInputElement;
+    unpinVideoTileForm.addEventListener('submit', e => {
+      e.preventDefault();
+      let videoPreferences: VideoPreference[] = [];
```
RidipDe, line 140: let -> const

integration/js/app/video_test/video_test.ts
```diff
@@ -113,6 +118,36 @@ export class DemoMeetingApp implements AudioVideoObserver {
...
+      videoPreferences.push(new VideoPreference(unpinAttendeeId.value, 2, TargetDisplaySize.High));
+      this.downlinkPolicy.chooseRemoteVideoSources(videoPreferences);
+      let videoPreferenceObject = document.getElementById(`video-preference`);
```
RidipDe, line 143: let -> const

integration/js/pages/TestAppPage.js
```diff
@@ -135,6 +154,23 @@ class TestAppPage {
     return false;
   }
 
+  async videoPreferenceCheck(tileStateElementId, tileStateAttribute, tileStateValue) {
+    let localVideoTileState = await this.driver.findElement(By.id(tileStateElementId));
```
RidipDe, line 158: let -> const,
Applicable to all the variables below

RidipDe (commented):
Few suggestions:
1. Run 'npm run build:release' before the push to run the unit tests locally. This will also run the prettifier and eslint.
2.  You might have to add the following lines at the top of every newly created file or else the build will fail.
```
```

richnew10 (changes requested):
* Please run `npm run build:release`.
* Please squash all of your commits.
* Please write a thorough PR description that explains what you're landing. [Here is an example](https://github.com/aws/amazon-chime-sdk-js/pull/1090).
* Please review all of my comments on the staging PR. I didn't write them for my own amusement. Some of the basics, like the existence of `demoApp`, really should be addressed.

integration/js/VideoPinTest.js
lzongkun: Sorry I might miss something here. 
I did run npm run build:release and it did not seem to correct the format or prettify the code under integration?

src/videodownlinkbandwidthpolicy/VideoPreferences.ts
```diff
@@ -0,0 +1,234 @@
...
+      return;
+    }
+    this.cow();
+    this.items.add(item);
+  }
+
+  replaceFirst(newItem: T, f: (item: T) => boolean): void {
+    // Don't actually need to COW unless the item is there to add
+    // or the previous is there to remove
+    if (this.items.has(newItem) && !this.items.some(f)) {
```
richnew10, line 126: Careful! 

What's the outcome of

```
[1, 2, 3, 4].replaceFirst(1, isEven);
```

If it's 

```
[1, 3, 4]
```

then you can't do this check as written.

richnew10 (changes requested):

src/videodownlinkbandwidthpolicy/VideoPreferences.ts
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  add(item: T): void {
+    this.items.push(item);
+  }
+
+  replaceFirst(newItem: T, f: (item: T) => boolean): void {
+    this.items = this.items.filter(a => !f(a));
+    this.add(newItem);
+  }
```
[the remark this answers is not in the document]
hensmi-amazon: Ended up making small modification to handle duplicates:
```
    const pos = this.items.findIndex(f);
    if (pos === -1) {
      // If this is used elsewhere, there needs to be a duplicate check here
      this.items.push(newItem);
    } else if (!this.has(newItem)) {
      this.items[pos] = newItem;
    } else {
      this.items.splice(pos, 1);
    }
```

src/videodownlinkbandwidthpolicy/VideoPreferences.ts
hensmi-amazon: Hmm, i think this is fine?  I did rewrite the comment since before it describes `!(!this.items.has(newItem) || this.items.some(f))` which is a bit more verbose then was used.

In the above case `!this.items.some(f) === false` so it won't enter the branch.  I added a handful more unit tests, including the one you mention.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2675,6 +2758,39 @@ export class DemoMeetingApp
...
+    const attendeeId = this.audioVideo.getVideoTile(tileId)?.state().boundAttendeeId;
+    this.roster[attendeeId].bandwidthConstrained = true;
+    this.updateRoster();
+  }
+
+  tileWillBeUnpausedByDownlinkPolicy(tileId: number): void {
+    this.log(`Tile ${tileId} will be resumed due to sufficient bandwidth`);
+    const attendeeId = this.audioVideo.getVideoTile(tileId)?.state().boundAttendeeId;
+    this.roster[attendeeId].bandwidthConstrained = false;
+    this.updateRoster();
```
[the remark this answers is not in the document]
hensmi-amazon: Worth noting that as is, `chooseRemoteVideoSources` won't end up immediately pausing or unpausing videos, since that is still driven by the `wantsSubscribe` check from the facade every 2 seconds (@lzongkun can confirm?  I feel like we may want to see if it its possible to trigger a (throttled) immediate resubscribe for applications like pagination.  

Onto your main point this demo, the roster is updated:
* When video sources change -> we don't have any video send indicator so this is unnecessary.  Will remove.
* Network pause/unpause so we set status message to `Paused` -> I don't think this is necessary, will remove
* Pinned/unpinned so we set status message to `Pinned` -> Even more unnecessary, will remove

`videoTileDidUpdate` doesn't actually call `updateRoster` because as mentioned, we don't have any video send indication on current demo.

src/videodownlinkbandwidthpolicy/VideoPreferences.ts
richnew10: I think as written you'll get

```
[1, 2, 3, 4]
```

so the intent to replace the first even item is not fulfilled because `1` is already present in the array.

hensmi-amazon: Since its `&&` though (again I think maybe comment caused confusion because i used `or`) it won't hit the branch.  I confirmed with the following test I added (shouldn't have posted comments before pushing, `build:release` always takes longer then i expect :P) :
```
  it('replaces only first occurance', () => {
    const b1: MutableVideoPreferences = VideoPreferences.prepare();
    b1.add(new VideoPreference('foo', 1));
    b1.add(new VideoPreference('foo', 2));
    b1.add(new VideoPreference('foo', 3));
    b1.add(new VideoPreference('foo', 4));
    b1.replaceFirst(new VideoPreference('foo', 1), (item: VideoPreference) => {
      return item.priority % 2 === 0;
    });
    const p1 = b1.build();

    const b2 = VideoPreferences.prepare();
    b2.add(new VideoPreference('foo', 1));
    b2.add(new VideoPreference('foo', 3));
    b2.add(new VideoPreference('foo', 4));
    const p2 = b2.build();

    expect(p1.equals(p2)).to.be.true;
  });
 ```

richnew10 (dismissed):
I think this is close enough.

I would encourage you to amend the CHANGELOG entry, the guide, and the doc comment for the policy class itself to indicate that the exact behavior (and even the external API)) of the policy can change in future releases.

richnew10 (approved):

richnew10 (approved):

PR #1233, opened 2021-04-29
Add a build step to verify the TypeScript correctness of built artifacts.

richnew10:
**Description of changes:**

This change adds a `postbuild` step that ensures that the stuff we just put in `build/` can actually be imported by a TypeScript program.

**Testing**

Yes.

By building with and without 8d6e38c23a2e959953d127628b412e86d9522704, verifying that this build script fails.

No.

No.

No.

---

### DIFF
--- a/config/imports.test.ts
+++ b/config/imports.test.ts
@@ -1,0 +1,4 @@
+import * as Foo from 'amazon-chime-sdk-js';
+
+console.log('Imported:', Foo);
+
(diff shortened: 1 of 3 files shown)


ltrung:
Unrelated but should we remove Eq and PartialOrd from here https://github.com/aws/amazon-chime-sdk-js/blob/master/script/barrelize.js#L39

ltrung (approved):

PR #1236, opened 2021-04-30
startVideoPreviewForVideoInput uses the active video input stream instead of calling getUserMedia

ltrung:
**Description of changes:**
Currently `startVideoPreviewForVideoInput` will call `getUserMedia` to re-select the active chosen video device input.
This isn't necessary as we already call `getUserMedia` in `chooseVideoInputDevice` which is required before calling `startVideoPreviewForVideoInput`.
This also causes synthesized video input stream (such as SMTP bar) to not working for video preview because `getUserMedia` will just replace it with another video stream.

**Testing**

2. How did you test these changes? In Video Preview screen in the meeting demo, switch to `Blue` or `SMTP` options and verify that the video preview shows up correctly.
Yes, meeting demo.


richnew10:
@ltrung could you rebase this? It looks like it has some churn.

ltrung:
> @ltrung could you rebase this? It looks like it has some churn.

Could you clarify? I double check the committed files and they look correct to me.

richnew10:
> > @ltrung could you rebase this? It looks like it has some churn.
> 
> Could you clarify? I double check the committed files and they look correct to me.

nm — Git was showing a stale diff for some reason.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- `startVideoPreviewForVideoInput` uses the active video input stream instead of calling `GetUserMedia` again
```
richnew10, line 13: Nit: `getUserMedia`.

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -2257,21 +2257,10 @@ describe('DefaultDeviceController', () => {
...
-      spy.restore();
-    });
-
-    it('does not disconnect or connect the video stream for preview if no active video exsits', async () => {
+    it('does not disconnect or connect the video stream for preview if no active video exists', () => {
       const spy1 = sinon.spy(DefaultVideoTile, 'disconnectVideoStreamFromVideoElement');
       const spy2 = sinon.spy(DefaultVideoTile, 'connectVideoStreamToVideoElement');
       deviceController.startVideoPreviewForVideoInput(element);
-      await delay(100);
       expect(spy1.called).to.be.false;
```
richnew10, line 2264: Can we get a test to prove that it only calls `gUM` once?

richnew10 (changes requested):

CHANGELOG.md
ltrung: Fixed.

test/devicecontroller/DefaultDeviceController.test.ts
ltrung: We remove the getUserMedia call so there would be no call right?

richnew10: There should be a single call when they choose the video input, and no additional call when they start the preview.

This is easy to do with sinon.spy; see elsewhere in the codebase.

simmkyu: Not in this `it` block, you can validate that in the previous `it`. In this `it`, you can ensure that `gUM` is not called.

ltrung: Thanks. Updated in both test case to check for getUserMedia.

simmkyu (approved):

simmkyu (approved):

richnew10 (approved):

PR #1247, opened 2021-05-03
Fix npm scripts in the browser demo.

richnew10:
`npm run start` and its variants now correctly load the selected app.

Additionally, this commit has the build only run `tsc` in `npm run
build`, which makes everyday development faster, but still catches
issues that the Webpack loader will not.

**Testing**

N/A.

Running each variant script.

N/A.

No.

No.

---

### DIFF
--- a/demos/browser/script/webpack-wrapper.js
+++ b/demos/browser/script/webpack-wrapper.js
@@ -18,11 +18,17 @@ const options = {
   },
 };
 
+const serve = argv.indexOf('--serve') !== -1;
+
+console.info('Should start local server:', serve);
+
 // This won't be present with npm 6.
 // Exclude file paths.
-const lastIndex = argv.length - 1;
+const remainingArgs = argv.filter(v => v !== '--serve');
+const lastIndex = remainingArgs.length - 1;
+let possibleApp;
 if (lastIndex >= 2) {
-  let possibleApp = argv[lastIndex];
+  possibleApp = remainingArgs[lastIndex];
   if (!possibleApp.startsWith('/')) {
     if (possibleApp.startsWith('app=')) {
       possibleApp = possibleApp.substring(4);
@@ -31,11 +37,22 @@ if (lastIndex >= 2) {
   }
 }
 
+console.info('Possible app:', possibleApp);
+
 const webpack = spawn(command, args, options);
 
 webpack.stdout.on('data', (data) => console.log(data.toString()));
 webpack.stderr.on('data', (data) => console.error(data.toString()));
 webpack.on('close', (status) => {
   console.info('Webpack existed with status', status);
-  process.exit(status);
+
+  if (serve) {
+    console.info('Starting server.');
+    if (possibleApp) {
+      process.env.npm_config_app = possibleApp;
+    }
+    require('../serve.js');
+  } else {
+    process.exit(status);
+  }
 });
--- a/demos/browser/package.json
+++ b/demos/browser/package.json
@@ -5,9 +5,9 @@
   "scripts": {
     "deps": "cd ../.. && npm run build",
     "build:fast": "node script/webpack-wrapper.js",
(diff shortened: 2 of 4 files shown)


ltrung:
In NPM 7, if I run `npm run build:fast -- --env app=messagingSession`, the output would be which packed meetingV2 instead of messagingSession
```

ARGV: [
  '/usr/local/Cellar/node/16.0.0_1/bin/node',
  '/Volumes/Unix/workplace/chimeSDK/amazon-chime-sdk-js/demos/browser/script/webpack-wrapper.js',
  '--env',
  'app=messagingSession'
]
Should start local server: false
Env: {
  "WEBPACK_BUNDLE": true,
  "WEBPACK_BUILD": true
}

App: undefined
Using app meetingV2
```

demos/browser/script/webpack-wrapper.js
```diff
@@ -18,11 +18,17 @@ const options = {
...
-const lastIndex = argv.length - 1;
+const remainingArgs = argv.filter(v => v !== '--serve');
+const lastIndex = args.length - 1;
+let possibleApp;
 if (lastIndex >= 2) {
-  let possibleApp = argv[lastIndex];
+  possibleApp = remainingArgs[lastIndex];
   if (!possibleApp.startsWith('/')) {
     if (possibleApp.startsWith('app=')) {
       possibleApp = possibleApp.substring(4);
```
ltrung, line 34: Should we only set `options.env.npm_config_app = possibleApp;` inside the if?

ltrung (approved):

demos/browser/script/webpack-wrapper.js
richnew10: I chose to set it at point of use when we are about to reuse this execution for the server.

ltrung: So what is the difference between `options.env.npm_config_app` and `process.env.npm_config_app`? You set `process.env.npm_config_app` down below but inside this if you set `options.env.npm_config_app`.

richnew10: `options` is the dictionary we use when spawning the Webpack process. `options.env` sets an environment variable for the spawned process.

`process.env` is the env for the _current_ process, so we need to change that before we `require` the server process.

ltrung: Sorry, i meant to ask whether this line https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/browser/script/webpack-wrapper.js#L30 should be set only if `possibleApp.startsWith('app=')`?

ltrung (commented):
Should we update this with env.app also https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/browser/webpack.config.js#L12?

demos/browser/script/webpack-wrapper.js
richnew10:
```js
if (lastIndex >= 2) {
  possibleApp = remainingArgs[lastIndex];
  if (!possibleApp.startsWith('/')) {
    if (possibleApp.startsWith('app=')) {
      possibleApp = possibleApp.substring(4);
    }
    options.env.npm_config_app = possibleApp;
  }
}
```

The first and second `if`s won't be hit unless there's a possible app — the arg count check is for that purpose.

The third `if` strips any leading `key=`.

Webpack is always invoked, so as soon as we know what the possible app is, we set it in the options field.

Does that answer your question?

richnew10: I don't set the `process` value until we know whether we're going to start the server, after `webpack` is done. Setting it to `undefined` is harmless.

richnew10:
> Should we update this with env.app also https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/browser/webpack.config.js#L12?

The end result is logged two lines below; it's written this way to give unambiguous information about where the value came from.

ltrung (approved):

PR #1291, opened 2021-05-13
[Integ] Add a GitHub workflow to check previous version

ltrung:
**Description of changes:**
Add a scheduled GitHub action to verify that previous version is working by running integration tests.
Temporarily set it to run every 2 hour  at 30 minute mark to easily verify the action works.
Will do a subsequent PR to change it to run once a day.

**Testing**

### DIFF
--- a/.github/script/get-prev-version
+++ b/.github/script/get-prev-version
@@ -1,0 +1,11 @@
+#!/usr/bin/env ruby
+
+versions = `npm view amazon-chime-sdk-js versions --json`.strip.split("\n")
+# The output would be something like
+#[
+#  "1.0.0",
+#  ...
+#  "2.8.0", <- previous version
+#  "2.9.0"  <- latest version
+#]
+puts versions[versions.size-3][3..-3]
--- a/integration/js/script/run-test
+++ b/integration/js/script/run-test
@@ -126,4 +126,7 @@ case $1 in
       ;;
 esac
 
+#Kill the current local demo process
+kill -9 $(lsof -i:8080 | grep LISTEN | awk '{print $2}')
+
 ./js/script/failure-check
--- a/.github/workflows/prev-version-integration.yaml
+++ b/.github/workflows/prev-version-integration.yaml
@@ -1,0 +1,80 @@
+name: Previous Version Integration Tests Workflow
+
+on:
+  schedule:
+    - cron: '30 */2 * * *'
+
+env:
+ SELENIUM_GRID_PROVIDER: saucelabs
+ CLOUD_WATCH_METRIC: false
+ TEST_TYPE: Github-Action
+ SAUCE_USERNAME: ${{secrets.SAUCE_USERNAME}}
+ SAUCE_ACCESS_KEY: ${{secrets.SAUCE_ACCESS_KEY}}
+
+jobs:
+  prev-version-integ:
+    name: Previous Version Integration Tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Setup Node.js - 14.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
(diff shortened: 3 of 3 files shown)


.github/workflows/release-1x-integration.yaml
```diff
@@ -0,0 +1,61 @@
...
+          path: current
+      - name: Checkout Package
+        uses: actions/checkout@v2
+        with:
+          ref: release-1.x
+          path: prev
+      - name: Setup Node.js - 14.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
```
devalevenkatesh, line 52: Should this be 15.x?

.github/workflows/release-1x-integration.yaml
```diff
@@ -0,0 +1,61 @@
...
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+      - name: Setup Sauce Connect
+        uses: saucelabs/sauce-connect-action@v1
+        with:
+          username: ${{ secrets.SAUCE_USERNAME }}
+          accessKey: ${{ secrets.SAUCE_ACCESS_KEY }}
+          noSSLBumpDomains: all
+          tunnelIdentifier: ${{ steps.create-job-id.outputs.uuid }}
+      - name: Checkout Package
```
devalevenkatesh, line 40: Could not understand these two checkouts `current` and `prev`? Could you please explain a little?

.github/workflows/release-1x-integration.yaml
vidya-mohan: Did i understand that this one is testing 1.x ? I was looking for current version (which is 2.x), and last 2 versions, say 2.x-1 and 2.x-2

.github/workflows/release-1x-integration.yaml
ltrung: No longer using this but the idea is to checkout previous repo in prev folder and current repo in current folder. The issue is i made some changes to run-test script but its not available in 2.8.0 so need to copy the file over.

.github/workflows/release-1x-integration.yaml
ltrung: Revise it to previous current version

ltrung: We should not use 15 here as it is already deprecated in favor of 16 but node 16 has issue with sass compiler

.github/script/get-prev-version
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env ruby
```
devalevenkatesh, line 1: Can we use JS instead of ruby 😅 ? Will help us remove ruby step if we can.

integration/js/script/run-test
```diff
@@ -126,4 +126,7 @@ case $1 in
       ;;
 esac
 
-./js/script/failure-check
+#Kill the current local demo process
```
devalevenkatesh, line 129: Nit: space after `#`.

.github/workflows/release-1x-integration.yaml
devalevenkatesh: So this will currently test the last version of latest major version only and we will extend it other versions later right as required?

ltrung: Yep

.github/script/get-prev-version
ltrung: Well, we try to not use any ruby script in our build process. This is just for GitHub action so I don't see the need :D

integration/js/script/run-test
ltrung: Will fix this if I end up doing another commit :)

devalevenkatesh (approved):

PR #1319, opened 2021-05-26
Add additional hardening to main video integration test, abort on reconnect on specific tests

hensmi-amazon:
Disconnections have significant impact on end-user experience as not only do we have to reset the state, but in most cases the reconnect may take 15 seconds (though we intend to reduce this delay where possible).   However our current video integration test can easily survive a reconnect, regardless of whether it is due to client or backend issues, making it difficult to spot problems pre-emptively.

Additionally toggling camera feed covers a significant amount of client and backend paths (e.g. reusing previously inactive transceivers and media sections, restarting SRTP streams, etc) not hit by simply toggling on, and we would like coverage of those paths as well, again, to better be able to spot problems pre-emptively.

**Issue #:** None

**Description of changes:** Adds the following changes:
* Added a 'abort-on-reconnect' query parameter to demo URL to trigger fatal.  Use this param in `Audio` and `Video` test.
* Add additional toggle of video on video test.  This will lengthen the tests a touch (1 second locally) but I think it should be fine, lmk if we want to be conservative with this.
* Add additional quick roster check at end of `Video` test to check for anything unusual.

**Testing**

2. How did you test these changes? Confirmed query parameter works as expected when used in demo.  Confirmed I could get integration tests to fail on new paths.  Confirmed they didn't fail on normal operation.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes, meetingv2.


integration/js/VideoTest.js
```diff
@@ -26,15 +29,27 @@ class VideoTest extends SdkBaseTest {
...
-    await LocalVideoCheck.executeStep(this, session, 'VIDEO_ON');
-    await WaitForRemoteParticipantsToTurnVideoOn.executeStep(this, session);
-    await RemoteVideoCheck.executeStep(this, session, 'VIDEO_ON');
-    await WaitForRemoteVideoCheckToComplete.executeStep(this, session);
-    await ClickVideoButton.executeStep(this, session);
-    await LocalVideoCheck.executeStep(this, session, 'VIDEO_OFF');
-    await WaitForRemoteParticipantsToTurnVideoOff.executeStep(this, session);
-    await RemoteVideoCheck.executeStep(this, session, 'VIDEO_OFF');
+    for (let i = 0; i < videoToggleCount; i++) {
+        // Note since these wait functions will be called twice, we reset the respective variable after each one
```
hensmi-amazon, line 33: Not super pretty imo but is simply a requirement of the way this state is held, couldn't find a cleaner way (reset in step seems like a weird side effect, creating a reset function for every variable seemed like overkill) that didn't require more significant refactoring of these tests.

ltrung (commented):
Could you give a bit of context why do we want to modify the test ?

hensmi-amazon:
> Could you give a bit of context why do we want to modify the test ?

Added context to test

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2501,6 +2505,9 @@ export class DemoMeetingApp
 
   audioVideoDidStartConnecting(reconnecting: boolean): void {
     this.log(`session connecting. reconnecting: ${reconnecting}`);
+    if (reconnecting && this.isAbortingOnReconnect()) {
+        fatal(Error("reconnect occured with abort-on-reconnect set to true"));
```
simmkyu, line 2509: nit: Use single quotes `'` for consistency with other strings

integration/js/VideoTest.js
```diff
@@ -1,10 +1,13 @@
 const {OpenAppStep, JoinMeetingStep, AuthenticateUserStep, ClickVideoButton, ClickVideoFilterButton, WaitForRemoteVideoCheckToComplete, WaitForRemoteParticipantsToTurnVideoOff, WaitForRemoteParticipantsToTurnVideoOn, WaitForRemoteParticipantsToJoinMeeting, WaitForMeetingToBeCreated} = require('./steps');
-const {UserJoinedMeetingCheck, LocalVideoCheck, RemoteVideoCheck, UserAuthenticationCheck, RosterCheck} = require('./checks');
+const {UserJoinedMeetingCheck, LocalVideoCheck, RemoteVideoCheck, UserAuthenticationCheck, RosterCheck, RosterCheckConfig} = require('./checks');
 const {AppPage} = require('./pages/AppPage');
 const {TestUtils} = require('./node_modules/kite-common');
 const SdkBaseTest = require('./utils/SdkBaseTest');
 const { v4: uuidv4 } = require('uuid');
 
+// We will toggle the video on and off multiple times
+const videoToggleCount = 2
```
simmkyu, line 9: nit: missing semi-colon

integration/js/VideoTest.js
```diff
@@ -26,15 +29,27 @@ class VideoTest extends SdkBaseTest {
...
-    await ClickVideoButton.executeStep(this, session);
-    await LocalVideoCheck.executeStep(this, session, 'VIDEO_OFF');
-    await WaitForRemoteParticipantsToTurnVideoOff.executeStep(this, session);
-    await RemoteVideoCheck.executeStep(this, session, 'VIDEO_OFF');
+    for (let i = 0; i < videoToggleCount; i++) {
+        // Note since these wait functions will be called twice, we reset the respective variable after each one
+        await ClickVideoButton.executeStep(this, session);
+        await LocalVideoCheck.executeStep(this, session, 'VIDEO_ON');
+        await WaitForRemoteParticipantsToTurnVideoOn.executeStep(this, session);
+        this.numVideoRemoteOn = 1;
```
ltrung, line 37: Do we want to manually set these values? They are set via io socket server in SdkBaseTest. Maybe what you want is to reset these variable after each loop instead?

hensmi-amazon: This is in a way 'resetting' as the default values here are 1 (it 'includes' self as 1, as these steps compare to `numParticipants`).  I would still need to reset `numOfParticipantsCompletedVideoCheck` mid loops before the second call to `WaitForRemoteParticipantsToTurnVideoOff` so i just followed the same pattern of immediately resetting the values after each Step which checks them.

hensmi-amazon: As mentioned above, I'm not a huge fan of this but refactoring this doesn't seem trivial and would indirectly touch a lot of tests.

ltrung (approved):

Issue #1283, opened 2021-05-12
Callback to listen for SignallingChannel ready state


hesawant:
### To summarize

We rely on `AudioVideoFacade.realtimeSendDataMessage(...)` function to send messages to other attendees in the call. But some times sending data, results in `RealtimeApiFailed and error: Signaling client is not ready` which terminates the session for that attendee.

### What are you trying to do?

**Looking for a callback that would notify when the `SignalingChannel` changes state from connected to disconnected and vice versa.** Also receiving the reason for the change in state would be very helpful.

Have checked all the callbacks exposed by Chime SDK, did not find any promising way to determine if `SignalingClient` is ready or not. This new call back will help to determine when not to send message and prevent abrupt session termination.

### Some other observations/conditions around when we see this failure

1. We get warning logs that indicate that client will be reconnecting.
    a. `will retry due to status code ConnectionHealthReconnect` → triggered at [DefaultAudioVideoController.ts](https://github.com/aws/amazon-chime-sdk-js/blob/7077353c0a5e036a2afffe0a1639cf80abdfb13a/src/audiovideocontroller/DefaultAudioVideoController.ts)
    b. Followed by `AudioVideoObserver.audioVideoDidStartConnecting(true /* reconnecting */ )` getting called.
2. The warning log → `missed pong 1 time(s)` is fired.
3. AudioVideoObserver.connectionDidSuggestStopVideo()` is called.
4. Few instances of failure without any of the above.

We can handle (1) by not invoking `realtimeSendDataMessage` because we know for sure that the connection is being re-established. But the other cases we don't know for sure. Also it is too many places to keep track of.

ltrung:
@hesawant Thanks for reporting this. I have reproduced the issue. I don't think we should throw an error there and should log it as an error instead. However, the problem remains is how we alert consumers that the message is not sent.
I will bring this back and discuss with the team.

ltrung:
No a single throttled message does not disconnect the signaling client. It just means you reached the soft limit on our backend as detailed above (100 messages per second with the maximum burst size of 200 for a meeting).
The client disconnected only once you reached the hard limit.
I would think if you could control your rate to stay below that soft limit which should prevent this issue in the first place. If you think the soft limit rate is too low please cut an AWS Support ticket to us :)

xuesichao:
[...]

PR #1327, opened 2021-05-28
Log error if signaling client is not ready when sending data message

ltrung:
**Issue #1283:**

**Description of changes:**
Previously we threw an error if the signaling client is not ready when sending a data message. The data message is sent using a realtime API so an error thrown will be fatal and cause the meeting to disconnect completely.
This PR changes to log the error instead of throwing it.

**Testing**

- Join a meeting
- Using Network Link Conditioner to turn off the network
- Try to send a data message
- Turn back the network
- Verify that you can reconnect
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes, using the meeting demo.

### DIFF
--- a/src/task/SendAndReceiveDataMessagesTask.ts
+++ b/src/task/SendAndReceiveDataMessagesTask.ts
@@ -81,7 +81,7 @@ export default class SendAndReceiveDataMessagesTask
       messageFrame.messages = [message];
       this.context.signalingClient.sendDataMessage(messageFrame);
     } else {
-      throw new Error('Signaling client is not ready');
+      this.context.logger.error('Signaling client is not ready');
     }
   };
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added 
 
 ### Changed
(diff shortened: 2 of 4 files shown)


simmkyu (dismissed):

src/task/SendAndReceiveDataMessagesTask.ts
```diff
@@ -81,7 +81,7 @@ export default class SendAndReceiveDataMessagesTask
       messageFrame.messages = [message];
       this.context.signalingClient.sendDataMessage(messageFrame);
     } else {
-      throw new Error('Signaling client is not ready');
+      console.error('Signaling client is not ready');
```
devalevenkatesh, line 84: Should we use `this.context.logger` instead of console here ?

devalevenkatesh (approved):

richnew10:
Isn't the behavior we want here that (a) the consumer gets an error, but (b) the meeting does not end?

That is, when a customer tries to send a message, it's not fatal to the meeting, but it is an error…

Logging an error and doing nothing makes it really hard to build systems that work predictably.

ltrung:
@richnew10 As I mentioned in this PR, this is just a first part to unblock customers (thus, i did not close the issue yet). We would need to have a buffer to resend the failed message or an observer to notify consumers (or a combination).

PR #1329, opened 2021-05-28
Add more debug logging for choose input device

ltrung:
**Description of changes:**
Add more debug logging when choosing input devices

**Testing**

Change the Logging mode to DEBUG and verify that the debugging logs showed up.

### DIFF
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -1155,6 +1155,9 @@ export default class DefaultDeviceController
       return true;
     }
     const deviceIds = DefaultDeviceController.getIntrinsicDeviceId(device);
+    this.logger.debug(
+      `Checking deviceIds ${deviceIds} of type ${typeof deviceIds} with groupId ${groupId}`
+    );
     if (typeof deviceIds === 'string' && groupId === this.getGroupIdFromDeviceId(kind, deviceIds)) {
       return true;
     }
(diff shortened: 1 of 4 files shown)


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1155,21 +1155,33 @@ export default class DefaultDeviceController
...
+      `Checking deviceIds ${deviceIds} of type ${typeof deviceIds} with groupId ${groupId}`
+    );
     if (typeof deviceIds === 'string' && groupId === this.getGroupIdFromDeviceId(kind, deviceIds)) {
       return true;
     }
     return false;
   }
 
   private getGroupIdFromDeviceId(kind: string, deviceId: string): string {
+    /* istanbul ignore else */
```
xuesichao, line 1168: Can this one cover the missing else in if condition at line 1173?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1155,21 +1155,33 @@ export default class DefaultDeviceController
       return true;
     }
     const deviceIds = DefaultDeviceController.getIntrinsicDeviceId(device);
+    this.logger.debug(
+      `Checking deviceIds ${deviceIds} of type ${typeof deviceIds} with groupId ${groupId}`
```
simmkyu, line 1159: Could you ensure that you are not seeing the [Object object] for logging an array?

simmkyu (approved):

src/devicecontroller/DefaultDeviceController.ts
ltrung: Yes, i verified it. It printed out as devicesIds <list of deviceId delimited by comma>

src/devicecontroller/DefaultDeviceController.ts
ltrung: We could not write unit test for this. The only time this triggers is when the infinite looping happens for audio input.

tylerlot: Have we not seen this issue occur for the video devices as well, or is it just customers typically set audio first and thus encounter it first

ltrung: So the issue right now only happens to audio input because after select audio input, we call `restartLocalAudio` https://github.com/aws/amazon-chime-sdk-js/blob/master/src/devicecontroller/DefaultDeviceController.ts#L1513 which will try to call choose audio input device again, expecting if we can reuse the exsting audio device. However, since that matching fails, it will try to choose the audio input device again -> Causing the infinite loop.

PR #1347, opened 2021-06-04
Remove videohelpdesk demo

ltrung:
**Description of changes:**
Remove the current video help desk demo

**Testing**


xuesichao (approved):

.integration-watchlist
```diff
@@ -4,12 +4,12 @@ src/
 integration/
 
 # exceptions
+!.eslintignore
```
xuesichao, line 7: Just curious why do we add it this time?

ltrung: This is just an exception list of files we do not need to trigger integration tests when there are changes. eslintignore is not really needed to run integ tests so i added it.

PR #1348, opened 2021-06-04
Remove the minTargetBitrate logic in VideoPriorityBasedPolicy

lzongkun:
**Description of changes:**
Remove the minTargetBitrate logic in VideoPriorityBasedPolicy.

For more context, the logic we added to keep the minTargetBitrate always equal to the lowest stream's bitrate will ensure the policy to always at least subscribe to one stream. Under network constraints, if the estimated bitrate goes even below the bitrate of the lowest stream, we will still subscribe to it, which may cause packet loss and could also affect audio as well, eventually could result in reconnection. This is not the ideal behavior.

By removing this logic here, the policy will be able to pause all the streams if the estimated bitrate goes too low.
**Testing**

2. How did you test these changes? Via demo app.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes, by using meetingDemoV2.


hensmi-amazon:
What testing was done?  What does recovery look like before and after change?

lzongkun:
> What testing was done? What does recovery look like before and after change?

I have tested with two video streams enabled and downlink bandwidth limited to 300kbps.

Before, only one stream got paused and we were still trying to receive data from the other one which ended up in drop (and reconnect after constraint lifted).

After, both streams got paused and audio still worked (some kind of degraded but not enough to trigger reconnect/drop).

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -627,6 +618,7 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
     remoteInfos: VideoStreamDescription[]
   ): void {
     if (!this.tileController) {
+      this.logger.info('tileController not found!');
```
hensmi-amazon, line 621: Nit:  Feel like this should be error level

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 - Remove deprecated unwired webrtc constraints from device controller and peer connection construction.
+- Remove the minTargetBitrate logic in VideoPriorityBasedPolicy.
```
hensmi-amazon, line 29: Nit: Plain english? e.g. 'Removed unnecessary restriction on `VideoPriorityBasedPolicy` to always subscribe to at least one stream'

lzongkun:
> > What testing was done? What does recovery look like before and after change?
> 
> I have tested with two video streams enabled and downlink bandwidth limited to 300kbps.
> 
> Before, only one stream got paused and we were still trying to receive data from the other one which ended up in drop (and reconnect after constraint lifted).
> 
> After, both streams got paused and audio still worked (some kind of degraded but not enough to trigger reconnect/drop).



> > What testing was done? What does recovery look like before and after change?
> 
> I have tested with two video streams enabled and downlink bandwidth limited to 300kbps.
> 
> Before, only one stream got paused and we were still trying to receive data from the other one which ended up in drop (and reconnect after constraint lifted).
> 
> After, both streams got paused and audio still worked (some kind of degraded but not enough to trigger reconnect/drop).

Also wondering how to get to the stuck state where if no stream is subscribed the BWE never gets back to normal?
During my local testing, it was able to recover.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
XHatan: em...we tend to only use error level for uncontrollable environment-type event, maybe keep it as warn ?

hensmi-amazon (approved):
🚢

PR #1363, opened 2021-06-11
Add XCUITest test

michhyun1:
**Issue #:**

**Description of changes:**
Checks out the WKWebview demo, uses the newly deployed meeting URL and joins a meeting from a WKWebview. This needs to run after the deploy step, since we need to utilize the newly deployed meeting url.

**Testing**

1. Have you successfully run `npm run build:release` locally? This is a github action change
2. How did you test these changes? github actions were tested with the pull_request trigger first, then removed once verified that it worked.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? WKWebview demo is being used.

### DIFF
--- a/.github/workflows/xcuitest.yml
+++ b/.github/workflows/xcuitest.yml
@@ -1,0 +1,26 @@
+name: WkWebView Sample Demo E2E Test
+
+on:
+  workflow_run:
+    workflows: ["Deploy Demo App Workflow"]
+    branches: [master]
+    types: 
+      - completed
+
+jobs:
+  xcuitest:
+    runs-on: macos-11
+    env:
+      DEPLOYED_MEETING_DEMO: ${{secrets.DEPLOYED_MEETING_DEMO}}
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - name: Checkout amazon-chime-sdk repo
+        uses: actions/checkout@v2
+        with:
+          repository: 'aws-samples/amazon-chime-sdk'
+      - name: Insert meeting demo URL
+        run: |
+          sed -i '' -e 's@YOUR_SERVER_URL@${{secrets.DEPLOYED_MEETING_DEMO}}@' apps/iOS-WKWebView-sample/WkWebView\ Demo/AppConfiguration.swift
+      - name: Run XCUITest Locally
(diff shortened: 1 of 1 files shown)


hokyungh:
How did you test this?

.github/workflows/deploy.yml
```diff
@@ -38,3 +38,18 @@ jobs:
         run: script/github-action-awscli-installation
       - name: Deploy the Serverless Demo App
         run: script/deploy-canary-demo
+  xcuitest:
```
devalevenkatesh, line 41: Understand from description we need to have deployed demo. Can we [`workflow_run`](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run) Github action concept here?

ltrung: Can you clarify more @devalevenkatesh ?

.github/workflows/deploy.yml
```diff
@@ -38,3 +38,18 @@ jobs:
...
+    runs-on: macos-11
```
ltrung, line 42: Do we need this? What is the default value?

.github/workflows/deploy.yml
devalevenkatesh: Yeah sure, instead triggering the new workflow here after the deploy step, wanted to check if we can separate this newly added workflow into its own file and use workflow_run to trigger the new workflow based on previous completion state. Read here: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run.

michhyun1: Sure, I think that makes more sense.

.github/workflows/deploy.yml
michhyun1: Yes, testing on ios 14 requires big sur

michhyun1:
> How did you test this?

I created a Github Action with the trigger type: 'pull_request' which executed all of these steps. When I submitted this PR, it ran the github action and I saw the result had passed, meaning the test was successful.

.github/workflows/deploy.yml
michhyun1: done

ltrung (approved):

Issue #1357, opened 2021-06-09
Invalid camera gets selected when using MediaStream.clone() with chooseVideoInputDevice()


vsimeikoKG:
**Meeting and attendee IDs:**
Meeting ID - f6464f45-7a03-4531-9632-4c2cc0912074
Attendee ID - abf5d8cb-a3cb-dc39-642d-5f60526b5afd

**Timeline**
- Open browser with 2 video inputs available: USB and FaceTime cameras.
- Join the meeting with camera A selected.
- I can see my local video preview from camera A, but all other attendees can see output from camera B.

**Describe the bug**
Chime connects to the wrong camera when I use `chooseVideoInputDevice(existingMediaStream.clone())`. However, I can see correct camera local video preview (non-Chime video).

```js
  useEffect(() => {
    if (!meetingSession.current) {
      return;
    }

    const {audioVideo} = meetingSession.current;

    const handler = (stream) => {
      // Stop video if stream is null.
      if (!stream && audioVideo.hasStartedLocalVideoTile()) {
        audioVideo.stopLocalVideoTile();
      }

      const videoSettings = getVideoQualitySettings(/* ... */);

      if (
        stream &&
        (!audioVideo.hasStartedLocalVideoTile() ||
          !deepEqual(lastVideoQualitySettings.current, videoSettings))
      ) {
        const clone = stream.clone();

        window.clone = clone; // To debug. See console logs.

        console.log('CLONE DEVICE LABEL', clone.getTracks()[0].label);

        audioVideo.chooseVideoInputQuality(...videoSettings);
        audioVideo.listVideoInputDevices().then(() =>
          audioVideo.chooseVideoInputDevice(clone).then(() => {
            // Extra check here to avoid having "lost" MediaStream somewhere in Chime state.
            if (!audioVideo.hasStartedLocalVideoTile()) {
              audioVideo.startLocalVideoTile();
              console.log('VIDEO START');
            }
          })
        );

        lastVideoQualitySettings.current = videoSettings;
      }
    };

    MediaStreamSingleton.subscribeToVideoStream(handler);

    return () => MediaStreamSingleton.unsubscribeFromVideoStream(handler);
  }, [meetingSession.current, /* ... */]);
```

**Platform**
 - OS: macOS Catalina 10.15.7 (19H114)
 - Browser: Chrome
 - Browser Version: 91.0.4472.101 (Official Build) (x86_64)
[...]

PR #1366, opened 2021-06-16
Fix choosing input device API when passing in a media stream

ltrung:
**Issue #1357:**

**Description of changes:**
Previously, passing in a media stream to `choose<Audio/Video>InputDevice` would not work:
- for audio input, it can result in infinite loop.
- for video input, once you start local video, it can result in a different media stream.

The issue is with the current logic in `matchesDeviceSelection` only consider deviceId and not media stream. Thus, it always fails for media stream.
Update the logic to compare the streamId.
**Testing**

- Add a unit test
- In meeting demo, join in a meeting and open developer console.
- Using the commands to verify that you get the correct camera:
```
// Assume you have 2 cameras
devices = await app.audioVideo.listVideoInputDevices();
const stream = await navigator.mediaDevices.getUserMedia({ video: { deviceId: { exact: devices[1].deviceId }}});
await app.audioVideo.chooseVideoInputDevice(stream);
app.audioVideo.startLocalVideoTile();
// Verify that you get the correct camera
```
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? No, you would need to use developer console as the current meeting demo does not allow to select media stream.

### DIFF
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -1163,6 +1163,28 @@ export default class DefaultDeviceController
     return device && device.id ? device : null;
   }
 
+  private hasSameMediaStreamId(
+    kind: string,
+    selection: DeviceSelection,
+    proposedConstraints: MediaStreamConstraints
+  ): boolean {
+    // Checking for stream using the fake constraint created in calculateMediaStreamConstraints
+    let streamId;
+    if (kind === 'audio') {
+      // @ts-ignore
+      streamId = proposedConstraints?.audio.streamId;
+      /* istanbul ignore next */
+      // @ts-ignore
+      return !!streamId && streamId === selection.constraints?.audio?.streamId;
+    }
+    /* istanbul ignore next */
+    // @ts-ignore
+    streamId = proposedConstraints?.video.streamId;
+    /* istanbul ignore next */
+    // @ts-ignore
+    return !!streamId && streamId === selection.constraints?.video?.streamId;
+  }
+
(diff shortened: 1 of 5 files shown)


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1163,6 +1163,28 @@ export default class DefaultDeviceController
...
+    proposedConstraints: MediaStreamConstraints
+  ): boolean {
+    // Checking for stream using the fake constraint created in calculateMediaStreamConstraints
+    let streamId;
+    if (kind === 'audio') {
+      // @ts-ignore
+      streamId = proposedConstraints?.audio.streamId;
+      /* istanbul ignore next */
+      // @ts-ignore
+      return streamId && streamId === selection.constraints?.audio?.streamId;
```
simmkyu, line 1178: The return type of this function is `boolean` but it may return `undefined`. You can change the first check to:
```
!!streamId &&
```

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1163,6 +1163,28 @@ export default class DefaultDeviceController
...
+    }
+    /* istanbul ignore next */
```
simmkyu, line 1180: Can we add tests to cover all scenarios?

ltrung: As discussed, it is just to be defensive and guard in all case. In reality, for video constraints for example, it is only not set if video is previously null which quit right away and if constraints is set, video property will be set.

simmkyu (approved):

PR #1370, opened 2021-06-17
Add meeting events ingestion

devalevenkatesh:
**Issue #:**
NA

**Description of changes:**
- Add event ingestion components, documentation, unit tests to ingest events to Amazon chime backend.
- Add integration test for now so that it can report events to ingestion URL and then later we will remove the check for this in demo so that all canaries can report it.

**Testing**

- Tested locally and by deploying the browser demo app.
- Run the browser demo app and join a meeting.
- Select various audio inputs, video inputs, trigger re-connection for 5s and connect back and then leave meeting.
- All meeting events will be sent to the ingestion backend and you can monitor the request in browser devtools.

The browser demo app.
I have added public API which have been reviewed on the private repo.
No.


guides/12_Client_Event_Ingestion.md
```diff
@@ -0,0 +1,45 @@
+# Event Ingestion
+
+We send the [Amazon Chime SDK meeting events](https://aws.github.io/amazon-chime-sdk-js/modules/meetingevents.html#meeting-events-and-attributes) to the Amazon Chime backend to analyze meeting health trends or identify common failures. This helps to improve your meeting experience.
```
ltrung, line 3: Do we need a guide for this? Seem like just a use case in the README would suffice.

guides/12_Client_Event_Ingestion.md
```diff
@@ -0,0 +1,45 @@
...
+
+The Amazon Chime SDK for JavaScript will not send below sensitive attributes to the Amazon Chime backend.
+|Attribute|Description
+|--|--
+|`externalMeetingId`|The Amazon Chime SDK external meeting ID.
+|`externalUserId`|The Amazon Chime SDK external user ID that can indicate an identity managed by your application.
+   
+## Opt out of Event Ingestion
+   
+To opt out of event ingestion, provide `NoOpEventReporter` to `DefaultMeetingSession` while creating the meeting session.
```
ltrung, line 15: Is it simpler to just have a flag to disable event reporter?

devalevenkatesh: I had it first but then we thought it would be confusing to say disable it because then we have to clarify if builders can enable it or not and based on what factors and so on handle the logic around it. I synced with mobile and planned to go ahead with this one approach to keep it consistent within both platforms.

guides/12_Client_Event_Ingestion.md
devalevenkatesh: Since, we currently only internally send the events governed by MeetingHistoryState and do not support sending custom events hence did not add a use case for this.

src/useragentparser/DefaultUserAgentParser.ts
```diff
@@ -0,0 +1,55 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { UAParser } from 'ua-parser-js';
```
ltrung, line 4: What is the difference of using this vs BrowserBehavior?

devalevenkatesh: ua-parser-js is better at getting more specific information that detect-browser used in BrowserBehavior. We started using ua-parser-js when we released meeting events. I separated it so that it can be more generally used within SDK.

src/meetingsession/DefaultMeetingSession.ts
```diff
@@ -116,13 +129,47 @@ export default class DefaultMeetingSession implements MeetingSession, Destroyabl
...
+    eventReporter?: EventReporter
+  ): void {
+    if (eventReporter) {
+      this._eventReporter = eventReporter;
+    } else {
+      const {
+        urls: { eventsIngestionURL },
+      } = configuration;
+      if (eventsIngestionURL) {
+        const {
```
ltrung, line 157: What is the advantage of using this syntax instead of just assign it as const eventsIngestionURL = configuration.urls.eventsIngestionURL? Tbh, i found this very hard to read :)

src/eventbuffer/filterAttributes.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ *
+ * @param attributes received with the reported event.
+ * @param attributeToFilter set of attributes like 'externalMeetingId', 'externalUserId' to filter.
+ * @returns filtered attributes.
+ */
+const filterAttributes = (
```
ltrung, line 10: Does this need its own routine like this? If so, this seems like a helper function so should we rename it to eventBufferUtil.ts or something?

src/eventreporter/EventData.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import MeetingHistoryState from '../eventcontroller/MeetingHistoryState';
+
+/**
+ * [[EventData]] describes event information types.
+ */
+export default interface EventData {
+  name: MeetingHistoryState;
```
ltrung, line 10: Can we document what these properties mean? And maybe give an example of how it is supposed to look?

src/useragentparser/DefaultUserAgentParser.ts
ltrung: Didn't know that. Can you document it here then? Maybe we should add an action item to migrate browser behavior to ua-parser-js in v3?

src/meetingsession/DefaultMeetingSession.ts
devalevenkatesh: 😅 I had it as you say, but got comments we can do this to avoid I think more lines so changed it to follow this.

src/eventbuffer/filterAttributes.ts
devalevenkatesh: Yes, it was separated so that we can test the function separately; add unit tests just for it. It is under `eventbuffer` so I think the name should be fine and the parameters make is easy to use more generally hence the name is general. For now, it has restricted types which can be expanded to support other more complex types in future. So should I rename it then?

src/meetingsession/DefaultMeetingSession.ts
devalevenkatesh: Sorry my bad, I can change it to one line :)

```
const eventsIngestionURL = configuration.urls.eventsIngestionURL?
```

src/eventbuffer/filterAttributes.ts
ltrung: It seems like it is used in only one place right now so its strange to separate it out just for testing. It is not useful to use in other places anyway since it is nested in event buffer.  It also increases the public API unnecessarily.

src/eventreporter/EventData.ts
devalevenkatesh: Yes, updated with few examples and added doc for each property in this interface. Thanks for the catch!

src/eventbuffer/filterAttributes.ts
devalevenkatesh: Ok separated out as a private function in the buffer itself.

src/useragentparser/DefaultUserAgentParser.ts
devalevenkatesh: Yeah sounds good. Added a line over this import and will track the v3 one. Thanks!

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2504,7 +2521,21 @@ export class DemoMeetingApp
   async authenticate(): Promise<string> {
     const joinInfo = (await this.joinMeeting()).JoinInfo;
     const configuration = new MeetingSessionConfiguration(joinInfo.Meeting, joinInfo.Attendee);
-    await this.initializeMeetingSession(configuration);
+    const { Meeting: { MeetingId, MediaPlacement: { EventIngestionUrl } }} = joinInfo.Meeting;
```
ltrung, line 2524: nit: Can this be const { MeetingId, MediaPlacement: { EventIngestionUrl } } = jobInfo.Meeting.Meeting like below?

src/meetingsession/DefaultMeetingSession.ts
```diff
@@ -116,13 +129,45 @@ export default class DefaultMeetingSession implements MeetingSession, Destroyabl
...
+      this._eventReporter = eventReporter;
+    } else {
+      const eventsIngestionURL = configuration.urls.eventsIngestionURL;
+      if (eventsIngestionURL) {
+        const {
+          meetingId,
+          credentials: { attendeeId, joinToken },
+        } = configuration;
+        const meetingEventsClientConfiguration: EventsClientConfiguration = new MeetingEventsClientConfiguration(
+          meetingId,
```
ltrung, line 160: nit: Can we just use configuration.meetingId, configuration.credentials.atendeeId, etc. here? Don't think we need the above line.

src/eventreporter/EventData.ts
```diff
@@ -0,0 +1,23 @@
...
+export default interface EventData {
+  /**
+   * Name of an event with the [[MeetingHistoryState]] type.
+   * For example, "signalingDropped", "meetingReconnected", "meetingStartRequested".
+   */
+  name: MeetingHistoryState;
+  /**
+   * Event generated timestamp in milliseconds.
+   */
+  ts: number;
```
ltrung, line 18: nit: Is there a reason why we do not just call it timestamp?

devalevenkatesh: Yeah kept it in backend accepted format. They have `ts` as the timestamp attribute.

src/meetingsession/DefaultMeetingSession.ts
devalevenkatesh: Here it avoids actually three lines and repetitive `configuration.credentials` for both the `joinToken` and `attendeeId`. Also avoids `configuration.` few times once for `meetingId` and twice for `credentials`.

devalevenkatesh: Also, I had it earlier in normal where I got comment that I can use de-structring here.

demos/browser/app/meetingV2/meetingV2.ts
devalevenkatesh: Since the URL is now from meeting session configuration, this has been removed.

demos/browser/server.js
```diff
@@ -30,6 +30,8 @@ console.info('Using endpoint', endpoint);
 
 chime.endpoint = new AWS.Endpoint(endpoint);
 
+const INGESTION_URL = process.env.INGESTION_URL;
```
ltrung, line 33: Do we need this anymore?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1358,8 +1365,18 @@ export class DemoMeetingApp
       await Promise.all([
         this.createLogStream(configuration, 'create_log_stream'),
         this.createLogStream(configuration, 'create_browser_event_log_stream'),
+        this.createLogStream(configuration, 'create_browser_event_ingestion_log_stream'),
       ]);
 
+      const eventReportingPOSTLogger = new MeetingSessionPOSTLogger(
```
ltrung, line 1371: nit: Can we move it down below the other 2 so the order matches with the above init?

demos/browser/server.js
devalevenkatesh: Thanks for the catch. Have removed it. It is not needed now.

demos/browser/app/meetingV2/meetingV2.ts
devalevenkatesh: Update, moved below the existing `meetingSessionPOSTLogger`. Thanks!

.github/workflows/deploy.yml
```diff
@@ -23,6 +23,8 @@ jobs:
         DEVO_AWS_SECRET_ACCESS_KEY: ${{secrets.DEVO_AWS_SECRET_ACCESS_KEY}}
         AWS_ACCESS_KEY_ID: ${{secrets.AWS_ACCESS_KEY}}
         AWS_SECRET_ACCESS_KEY: ${{secrets.AWS_SECRET_ACCESS_KEY}}
+        BETA_EVENT_INGESTION_URL: ${{secrets.BETA_EVENT_INGESTION_URL}}
```
ltrung, line 26: init: This needs to remove?

devalevenkatesh: Yup removed from here and the canary deploy script as well. Cherry picked hence dint see this change to remove later. Good catch!

ltrung (approved):

PR #1396, opened 2021-06-28
Instantiate event reporting in demo to pass POST logger in DEBUG mode

devalevenkatesh:
**Issue #:**

**Description of changes:**
Currently, the meeting session is created a POST logger in DEBUG mode,
hence all debug logs will make a POST request with the POST logger.
The filtering whether the debug log is related to event reporting was
handled in serverless handler. Instead, setup the event reporter in demo
to pass the POST logger just for it and not the entire meeting session.
This avoids POST request throttling with so many debug log requests.

**Testing**

2. How did you test these changes? Deployed the demo app and tested if only event reporting logs are getting posted to the log group and not all. 
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? The browser meeting demo.

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -56,7 +56,10 @@ import {
   isAudioTransformDevice,
   NoOpEventReporter,
   EventReporter,
-  isDestroyable
+  isDestroyable,
+  MeetingEventsClientConfiguration,
+  EventIngestionConfiguration,
+  DefaultMeetingEventReporter
 } from 'amazon-chime-sdk-js';
 
 import CircularCut from './videofilter/CircularCut';
@@ -1365,7 +1368,6 @@ export class DemoMeetingApp
       await Promise.all([
         this.createLogStream(configuration, 'create_log_stream'),
         this.createLogStream(configuration, 'create_browser_event_log_stream'),
-        this.createLogStream(configuration, 'create_browser_event_ingestion_log_stream'),
       ]);
       this.meetingSessionPOSTLogger = new MeetingSessionPOSTLogger(
         'SDK',
@@ -1375,18 +1377,9 @@ export class DemoMeetingApp
         `${DemoMeetingApp.BASE_URL}logs`,
         logLevel
       );
-      const eventReportingPOSTLogger = new MeetingSessionPOSTLogger(
-        'SDKEventIngestion',
-        configuration,
-        DemoMeetingApp.LOGGER_BATCH_SIZE,
-        DemoMeetingApp.LOGGER_INTERVAL_MS,
-        `${DemoMeetingApp.BASE_URL}log_event_ingestion`,
-        LogLevel.DEBUG
-      );
       this.meetingLogger = new MultiLogger(
         consoleLogger,
         this.meetingSessionPOSTLogger,
-        eventReportingPOSTLogger
       );
       this.meetingEventPOSTLogger = new MeetingSessionPOSTLogger(
         'SDKEvent',
@@ -1397,6 +1390,7 @@ export class DemoMeetingApp
         logLevel
       );
     }
+    this.eventReporter = await this.setAndGetEventReporter(configuration);
     const deviceController = new DefaultDeviceController(this.meetingLogger, {
       enableWebAudio: this.enableWebAudio,
     });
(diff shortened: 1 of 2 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1413,20 +1407,12 @@ export class DemoMeetingApp
...
-        deviceController
-      );  
-    }
+    this.meetingSession = new DefaultMeetingSession(
+      configuration,
+      this.meetingLogger,
+      deviceController,
+      this.eventReporter
+    );
     this.eventReporter = this.meetingSession.eventReporter;
```
ltrung, line 1416: Do we need this line anymore?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1453,6 +1439,46 @@ export class DemoMeetingApp
     this.initContentShareDropDownItems();
   }
 
+  async setAndGetEventReporter(configuration: MeetingSessionConfiguration): Promise<EventReporter> {
```
ltrung, line 1442: nit: Maybe just call it setupEventReporter?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1453,6 +1439,46 @@ export class DemoMeetingApp
...
+    let eventReporter: EventReporter;
+    const ingestionURL = configuration.urls.eventIngestionURL;
+    if (ingestionURL) {
```
ltrung, line 1445: we can just do a if (!ingestionURL || !this.enableEventReporting) return to reduce the nested if below.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1453,6 +1439,46 @@ export class DemoMeetingApp
...
+        const meetingEventClientConfig = new MeetingEventsClientConfiguration(
+          configuration.meetingId,
+          configuration.credentials.attendeeId,
+          configuration.credentials.joinToken
+        );
+        const eventIngestionConfiguration = new EventIngestionConfiguration(
+          meetingEventClientConfig,
+          ingestionURL
+        );
+        if (!['localhost', '127.0.0.1'].includes(location.hostname)) {
```
ltrung, line 1457: We have multiple place where we do this check? Can you move it to its own routine so we could reuse it?

demos/browser/app/meetingV2/meetingV2.ts
devalevenkatesh: Cool, makes sense, I still need to make decision for both if else for enableEventReporting but returning when ingestionURL is null avoids the nested if. Thanks! Have updated the code.

demos/browser/app/meetingV2/meetingV2.ts
devalevenkatesh: Added an `isLocalHost` function and replaces two places where I got this while initializing the meeting session and setting up post logger for event reporter.

devalevenkatesh: Good callout!

demos/browser/app/meetingV2/meetingV2.ts
devalevenkatesh: Sure, changed the name. I was getting the event reporter hence had that name.

demos/browser/app/meetingV2/meetingV2.ts
devalevenkatesh: Thanks for the catch, no this is not needed anymore. Removed it.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1453,6 +1438,51 @@ export class DemoMeetingApp
...
+          `${DemoMeetingApp.BASE_URL}log_event_ingestion`,
+          LogLevel.DEBUG
+        );
+        const multiEventReportingLogger = new MultiLogger(
+          eventReportingLogger,
+          eventReportingPOSTLogger,
+        );
+        eventReporter = new DefaultMeetingEventReporter(eventIngestionConfiguration, multiEventReportingLogger);
+      }
+    } else {
```
ltrung, line 1476: this could move up also right? If event reporting is not enabled then return noOp reporter and quit right away?

devalevenkatesh: Yes, can avoid the else block here :) Looks more cleaner :) Thanks, have updated it.

ltrung (approved):

PR #1410, opened 2021-07-02
Lookup groupId from device cache

ltrung:
**Description of changes:**
- Change the logic to retrieve groupId from directly in the media stream we get from getUserMedia to retrieve from device cache.
**Testing**

2. How did you test these changes? Regression test.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Meeting demo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - [Documentation] Update priority based downlink policy guide.
+- Lookup `groupId` from device cache instead of directly from media stream when selecting input device.
 
 ### Removed
 
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -202,7 +202,10 @@ export default class DefaultDeviceController
       return;
     }
     this.logger.info('Starting devicechange listener.');
-    this.onDeviceChangeCallback = () => this.handleDeviceChange();
+    this.onDeviceChangeCallback = () => {
+      this.logger.info('Device change event callback is triggered');
+      this.handleDeviceChange();
+    };
     this.mediaDeviceWrapper.addEventListener('devicechange', this.onDeviceChangeCallback);
   }
 
@@ -1053,9 +1056,7 @@ export default class DefaultDeviceController
       this.deviceInfoCache = [];
       return;
     }
-
     let devices = await navigator.mediaDevices.enumerateDevices();
-
     let hasDeviceLabels = true;
     for (const device of devices) {
       if (!device.label) {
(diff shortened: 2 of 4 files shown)


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1053,7 +1056,7 @@ export default class DefaultDeviceController
       this.deviceInfoCache = [];
       return;
     }
-
+    this.logger.debug('Update device info cache');
```
devalevenkatesh, line 1059: Should we move this below after getting the devices and also print the `devices`, because we will not know updated devices until builders implement `audioInputsChanged` observer then and hence might miss in the logs.

ltrung: Updated

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1442,7 +1444,10 @@ export default class DefaultDeviceController
         };
         track.addEventListener('ended', newDevice.endedCallback, { once: true });
       }
-      newDevice.groupId = this.getMediaTrackSettings(newDevice.stream)?.groupId || '';
+      newDevice.groupId = this.getGroupIdFromDeviceId(
+        kind,
+        this.getMediaTrackSettings(newDevice.stream)?.deviceId || ''
+      );
```
michhyun1, line 1450: So we are choosing the groupId from the device cache? Not sure if this would help though because the infinite loop is caused because the groupId that getUserMedia gives us is not within our device cache anyways. And should the source of truth for the groupId be the device cache or the stream we get from getUserMedia?

ltrung: Why does it not help? It should match the one in the cache in the next check and terminate the loop.

michhyun1: @ltrung yeah, its hard to say because we're not sure why gUM returns this groupId that isn't listed by listAudioInputsDevices (in the case of the loop). If the groupId provided by gUM is not found in the device cache, then I suspect that the deviceId we get from gUM might not have a groupId within the device cache anyways. In that case, getGroupIdFromDeviceId might return ''. Correct me if I'm wrong.

ltrung: We only see this behavior with default so the deviceId should be there. If it is not then groupId = '' and the match check will terminate so no infinite loop.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Lookup groupId from device cache instead of directly from media stream when selecting input device.
```
devalevenkatesh, line 30: Can we backtick groupdId? Also, we need to update this to [Unreleased] Changed, currently, its under [2.13.0].

devalevenkatesh (approved):

PR #1426, opened 2021-07-09
Adding 25 tile support for demo app

lzongkun:
**Description of changes:**
Adding 25 tile support for demo app and update doc for priority based downlink policy
**Testing**
Ran npm run build:release locally and will deploy demo app for QA testing.
1. Have you successfully run `npm run build:release` locally? yes.
2. How did you test these changes? Ran some basic video testing and will ask help from QA team to test 25 tiles.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes, meetingV2.
4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `VideoPriorityBasedPolicyConfig` to control video downlink policy with network event response and recovery delays. Check [User Guide for Priority-based Downlink Policy](https://aws.github.io/amazon-chime-sdk-js/modules/prioritybased_downlink_policy.html#user-guide-for-priority-based-downlink-policy) for more information.
 - Amazon Chime SDK Project Board Overview and Guide.
 
-- Added 25 video tile support for demo app
+- Added 25 video tile support for demo app.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- [Documentation] Update priority based downlink policy guide.
+
 ### Removed
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `VideoPriorityBasedPolicyConfig` to control video downlink policy with network event response and recovery delays. Check [User Guide for Priority-based Downlink Policy](https://aws.github.io/amazon-chime-sdk-js/modules/prioritybased_downlink_policy.html#user-guide-for-priority-based-downlink-policy) for more information.
 - Amazon Chime SDK Project Board Overview and Guide.
 
+- Added 25 video tile support for demo app
+
 ### Changed
 
 ### Removed
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -95,7 +95,7 @@ declare global {
 
 class DemoTileOrganizer {
   // this is index instead of length
-  static MAX_TILES = 27;
+  static MAX_TILES = 26;
   tiles: { [id: number]: number } = {};
   tileStates: { [id: number]: boolean } = {};
   remoteTileCount = 0;
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -95,7 +95,7 @@ declare global {
 
 class DemoTileOrganizer {
   // this is index instead of length
-  static MAX_TILES = 17;
+  static MAX_TILES = 27;
   tiles: { [id: number]: number } = {};
   tileStates: { [id: number]: boolean } = {};
   remoteTileCount = 0;
@@ -2802,7 +2802,7 @@ export class DemoMeetingApp
       return;
     }
     const tileIndex = tileState.localTile
-      ? 16
+      ? 27
       : this.tileOrganizer.acquireTileIndex(tileState.tileId);
     const tileElement = document.getElementById(`tile-${tileIndex}`) as HTMLDivElement;
     const videoElement = document.getElementById(`video-${tileIndex}`) as HTMLVideoElement;
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -95,7 +95,7 @@ declare global {
 
 class DemoTileOrganizer {
   // this is index instead of length
-  static MAX_TILES = 26;
+  static MAX_TILES = 27;
   tiles: { [id: number]: number } = {};
   tileStates: { [id: number]: boolean } = {};
   remoteTileCount = 0;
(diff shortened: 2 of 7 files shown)


ltrung:
Need to update the limit here too https://github.com/aws/amazon-chime-sdk-js/blob/master/integration/js/checks/LocalVideoCheck.js#L25

ltrung:
Why looking at why the video test failed, i found this message `video filter button is [object object]`(
https://github.com/aws/amazon-chime-sdk-js/blob/master/integration/js/pages/AppPage.js#L184). Can you update and print out the values instead?

CHANGELOG.md
```diff
@@ -11,8 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `VideoPriorityBasedPolicyConfig` to control video downlink policy with network event response and recovery delays. Check [User Guide for Priority-based Downlink Policy](https://aws.github.io/amazon-chime-sdk-js/modules/prioritybased_downlink_policy.html#user-guide-for-priority-based-downlink-policy) for more information.
```
devalevenkatesh, line 13: nit: You can remove this extra newline before this log.

CHANGELOG.md
```diff
@@ -11,8 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+- Added 25 video tile support for demo app
```
devalevenkatesh, line 14: Should we end this log with a "dot", to keep consistency with other logs here?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -95,7 +95,7 @@ declare global {
 
 class DemoTileOrganizer {
   // this is index instead of length
-  static MAX_TILES = 17;
+  static MAX_TILES = 27;
```
devalevenkatesh, line 98: Can you please help me understand this bump?
Old: 17 = 16 local video tiles + 1 content share tile.
New: 27 = 25 video tiles + 2 content share tiles? 
I think, this demo does not showcase simultaneous 2 content share tiles.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2769,7 +2769,7 @@ export class DemoMeetingApp
       return;
     }
     const tileIndex = tileState.localTile
-      ? 16
+      ? 27
```
devalevenkatesh, line 2772: Should this be 25 instead of 27? Checking since, the earlier number was just the total number of local video tiles.

demos/browser/app/meetingV2/meetingV2.ts
devalevenkatesh: Yes, that is correct.

demos/browser/app/meetingV2/meetingV2.ts
lzongkun: Another question:
In the [video integration test](https://github.com/aws/amazon-chime-sdk-js/blob/master/integration/js/checks/LocalVideoCheck.js#L25), when we are checking the local tile, we actually check the tile with index 17. How does it correlate with tileIndex 16 here?

demos/browser/app/meetingV2/meetingV2.ts
ltrung: No our demo supports 2 content share if you enable a flag in the URL https://github.com/aws/amazon-chime-sdk-js/blob/0d538ef5b9a05442867557d6c646a4cc5ccae599/demos/browser/app/meetingV2/meetingV2.ts#L2945

demos/browser/app/meetingV2/meetingV2.ts
ltrung: I don't think the previous number is correct. Content share also uses video tile and also got assigned one in this function so 27 sounds more correct to me.

demos/browser/app/meetingV2/meetingV2.ts
devalevenkatesh: Ohh, dint knew we had that I think discussed this in standup sometime. Anyways, thanks for correcting. So, then we need also check that support as well as MAX_TILES actually limit. I am not sure how the current `videoTileDidUpdate` binds the second video tile then.

ltrung: There would be no different in term of binding. When we allow 2 content share, the third person would not be able to start another content share.

demos/browser/app/meetingV2/meetingV2.ts
ltrung: iirc, it is due to the difference in logic between the demo and integration test. For integration test, we do not have tileOrganizer and it basically just counts the number of video elements in the page (https://github.com/aws/amazon-chime-sdk-js/blob/master/integration/js/pages/AppPage.js#L389). 
We have a [`content-share-video` element ](https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/browser/app/meetingV2/meetingV2.html#L527) which seems to be counted first, thus the index is shifted by one (so 16 to 17).

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -95,7 +95,7 @@ declare global {
 
 class DemoTileOrganizer {
   // this is index instead of length
-  static MAX_TILES = 17;
+  static MAX_TILES = 26;
```
ltrung, line 98: 27 is correct, why changes to 26?

integration/js/checks/LocalVideoCheck.js
```diff
@@ -22,7 +22,7 @@ class LocalVideoCheck extends AppTestStep {
 
   async run() {
     try {
-      let result = await this.page.videoCheck(this, 17, this.expectedState);
+      let result = await this.page.videoCheck(this, 27, this.expectedState);
```
ltrung, line 25: I assume this should be 28 and not 27 :)

demos/browser/app/meetingV2/styleV2.scss
```diff
@@ -664,6 +676,24 @@ a.markdown:active {
...
 
+.v-grid.featured.size-18,
+.v-grid.featured.size-19,
+.v-grid.featured.size-20,
+.v-grid.featured.size-21,
+.v-grid.featured.size-22,
+.v-grid.featured.size-23,
+.v-grid.featured.size-24,
+.v-grid.featured.size-25,
+.v-grid.featured.size-26 {
```
ltrung, line 687: Should this be 27?

demos/browser/app/meetingV2/styleV2.scss
```diff
@@ -698,6 +728,21 @@ a.markdown:active {
...
+  .v-grid.size-19,
+  .v-grid.size-20,
+  .v-grid.size-21,
+  .v-grid.size-22,
+  .v-grid.size-23,
+  .v-grid.size-24,
+  .v-grid.size-25,
+  .v-grid.size-26,
+  .v-grid.size-27,
+  .v-grid.size-28 {
```
ltrung, line 742: Should this be 27?

demos/browser/app/meetingV2/styleV2.scss
```diff
@@ -749,7 +794,9 @@ a.markdown:active {
   .v-grid.featured.size-14,
   .v-grid.featured.size-15,
   .v-grid.featured.size-16,
-  .v-grid.featured.size-17 {
+  .v-grid.featured.size-17,
+  .v-grid.featured.size-18,
+  .v-grid.featured.size-19 {
```
ltrung, line 799: Why is this only up to 19?

ltrung (approved):

PR #1428, opened 2021-07-13
Add media capture to serverless and browser demo

achris:
**Issue #:** N/A

**Description of changes:**

This change adds support for media capture pipelines to the demo meeting applications.

**Testing**

Yes

I ran both the browser demo and serverless demos and joined a meeting in multiple regions, starting and stopping a media capture session.

Yes, the meeting demo app

No

No


demos/serverless/template.yaml
```diff
@@ -11,6 +11,13 @@ Parameters:
     Description: The AWS SDK Chime endpoint
     Default: "https://service.chime.aws.amazon.com"
     Type: String
+  ChimeServicePrincipal:
+    Description: "The principal that will capture to the S3 bucket"
+    Default: "chime.amazonaws.com"
+    Type: String
+  ChimeMediaCaptureS3BucketPrefix:
+    Description: Prefix of S3 bucket two write capture artifacts.  The bucket name will be suffixed with the region.
```
ltrung, line 19: typo `two`?

demos/serverless/deploy.js
```diff
@@ -20,8 +41,11 @@ function usage() {
   console.log(`  -a, --application                    Browser application to deploy, default '${app}'`);
   console.log(`  -e, --event-bridge                   Enable EventBridge integration, default is no integration`);
   console.log(`  -c, --chime-endpoint                 AWS SDK Chime endpoint, default is '${chimeEndpoint}'`);
+  console.log(`  -p, --service-principal              Service principal for meeting related resources, default is '${chimeServicePrincipal}'`)
   console.log(`  -t, --enable-termination-protection  Enable termination protection for the Cloudformation stack, default is false`);
   console.log(`  -l, --disable-printing-logs          Disable printing logs`);
+  console.log(`  -o, --capture-output-prefix          Prefix for S3 bucket name`);
+  console.log(`  -i, --opt-in-regions                 Comma separated list of additional opt-in regions to enable for media capture`);
```
ltrung, line 48: Not familiar with capturing so could you explain more about this option?

demos/serverless/deploy.js
```diff
@@ -20,8 +41,11 @@ function usage() {
   console.log(`  -a, --application                    Browser application to deploy, default '${app}'`);
   console.log(`  -e, --event-bridge                   Enable EventBridge integration, default is no integration`);
   console.log(`  -c, --chime-endpoint                 AWS SDK Chime endpoint, default is '${chimeEndpoint}'`);
+  console.log(`  -p, --service-principal              Service principal for meeting related resources, default is '${chimeServicePrincipal}'`)
```
ltrung, line 44: This is just for capturing right? If so, should we say so? Meeting related resources seem to include other things too.

demos/serverless/deploy.js
```diff
@@ -144,6 +178,53 @@ function ensureTools() {
   spawnOrFail('npm', ['install']);
 }
 
+function createCaptureS3Buckets(bucketPrefix, regions) {
+  console.log(`Creating S3 buckets for media capture pipelines artifacts.  Bucket prefix: ${bucketPrefix} Regions:[${regions}]`);
```
ltrung, line 182: Can this not be created from the SAM template?

ltrung:
Do we want to update the package README to point to other resources such as developer guide for media capturing?

demos/serverless/deploy.js
achris: This would be used anywhere in the SAM template that currently points to the `chime.amazonaws.com` service principal.  It is used for testing.

demos/serverless/deploy.js
achris: The `-o` option is used for specifying the name of the S3 buckets.  It's a prefix because we append the region since we need a bucket in each region.
The `-i` option allows to specify additional regions beyond the initial set.   This is because we support meetings in `af-south-1` and `eu-south-1` which are opt-in and require customers to enable those regions in their AWS accounts before they can be used.  If customers want media capture in those regions they need to specify them here to create the S3 buckets.

demos/serverless/template.yaml
achris: Good catch, thanks.

demos/serverless/deploy.js
achris: Unfortunately not because the SAM template can not create resources in more than one region.

demos/serverless/deploy.js
ltrung: Could we add notes about this in the README?

ltrung (approved):

PR #1432, opened 2021-07-15
Add metrics for event ingestion

devalevenkatesh:
**Issue #:**

**Description of changes:**
- This adds metrics based on logs posted by POSTLogger for the event ingestion.
- We add a success, a retry and a error metric using `aws-embedded-metrics` `putMetric` function.
- This is just the serverless demo change.

**Testing**

1. Have you successfully run `npm run build:release` locally? Yes. But, this is just a demo change.
2. How did you test these changes? I deployed the serverless demo to my AWS account and checked if the metrics appear in the AWS CW metrics for my account when the deployed demo is ran a couple of times.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Browser meeting demo.

### DIFF
--- a/demos/serverless/src/handlers.js
+++ b/demos/serverless/src/handlers.js
@@ -157,6 +157,7 @@ exports.log_event_ingestion = async (event, context) => {
         message,
         timestamp: log.timestampMs
       });
+      addEventIngestionMetricsToCloudWatch(log, meetingId, attendeeId);
     }
     return logEvents;
   });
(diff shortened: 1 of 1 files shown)


demos/serverless/src/handlers.js
```diff
@@ -312,4 +313,27 @@ function addSignalMetricsToCloudWatch(logMsg, meetingId, attendeeId) {
...
+  if (logLevel === 'ERROR') {
+    errorMetricValue = 1;
+  } else if (logLevel === 'WARN' && message.match(/Retry (count)? limit reached/g)) {
+    retryMetricValue = 1;
+  } else if (message.includes('send successful')) {
+    ingestionTriggerSuccessMetricValue = 1;
+  }
+  putMetric('EventIngestionTriggerSuccess', ingestionTriggerSuccessMetricValue, meetingId, attendeeId);
+  putMetric('EventIngestionError', errorMetricValue, meetingId, attendeeId);
+  putMetric('EventIngestionRetryCountLimitReached', retryMetricValue, meetingId, attendeeId);
```
michhyun1, line 338: So we want to report 3 metrics for every log that we get from events ingestion?

devalevenkatesh: So for event reporting canary run, it will POST event ingestion logs and hence I have to check the log message to know which is the type and accordingly metric that. As next steps, I would add alarm on the sum of these metrics based on a certain period say 15mins.

michhyun1: Dont we only need to alarm if we see `EventIngestionError` or `EventIngestionRetryCountLimitReached`? Why do we need to sum them up?

demos/serverless/src/handlers.js
```diff
@@ -312,4 +313,27 @@ function addSignalMetricsToCloudWatch(logMsg, meetingId, attendeeId) {
...
+      metrics.setProperty('MeetingId', meetingId);
+      metrics.setProperty('AttendeeId', attendeeId);
+      metrics.putMetric(metricName, metricValue);
+    });
+  const { logLevel, message } = log;
+  let errorMetricValue = 0;
+  let retryMetricValue = 0;
+  let ingestionTriggerSuccessMetricValue = 0;
+  if (logLevel === 'ERROR') {
+    errorMetricValue = 1;
```
ltrung, line 330: Can we just call putMetric right here? Why do we need these variable?

devalevenkatesh: I ran into errors when no metric was posted so tried to this way like 1 when observed 0 when not.

demos/serverless/src/handlers.js
devalevenkatesh: Retry count limit can be reached for 1 single fetch call and each fetch call can have multiple events and each demo run can trigger multiple fetch calls. Hence, it can be sum or just the count but that will be setup later when we start posting metrics. I need success to check that we see success fetch calls when the canaries run.

michhyun1: yup, sounds good to me

michhyun1 (approved):

ltrung (approved):

Issue #1384, opened 2021-06-23
"TypeError: Chime is not a constructor" via aws-sdk with MessagingSessionConfiguration


Berkyle:
### What are you trying to do?

I'm trying to initialize a messaging session on the client side using the AWS SDK and the AWS SDK's Chime client.

### How can the documentation be improved to help your use case?

Document how to utilize the AWS-SDK in the browser in order to subscribe to message events.

### What documentation have you looked at so far?

I've followed documentation within this repo as well as a few other AWS examples as to how to subscribe to a messaging session, most notably:
- https://github.com/aws/amazon-chime-sdk-js#messaging-session
- https://github.com/aws/amazon-chime-sdk-js/blob/master/demos/browser/app/messagingSession/messagingSession.ts#L21-L22

These examples work fine in a development bundle. But once I've compiled my application into a production bundle and serve the minified file from my application server, I'm seeing: `TypeError: h is not a constructor`. Here, `h` is `Chime`, imported via:
```javascript
import * as Chime from 'aws-sdk/clients/chime';
```

I've found a few other instances around the web getting this exact error ([here](https://stackoverflow.com/questions/61300302/typeerror-aws-chime-is-not-a-constructor) and [here](https://github.com/aws/amazon-chime-sdk-js/issues/303), for example), but in all of those cases the answer seemed to be "Don't import from the aws-sdk on the client side. Browser applications don't need the aws-sdk." But unfortunately the chime messaging session very much needs to use the main sdk. Is this possibly an issue with my webpack configuration, or have I wildly misunderstood how a messaging session is meant to be configured on the browser?

![image](https://user-images.githubusercontent.com/34729254/123160782-30614100-d434-11eb-84a3-c4959ea81601.png)
![image](https://user-images.githubusercontent.com/34729254/123160731-22abbb80-d434-11eb-8449-b7c4c9c01163.png)

Berkyle:
Turns out the fix is just to not use a wildcard import. It would probably be best to update this in the documentation since I imagine a lot of use cases will involve developing a chat application for the browser.

```javascript
// ❌  This breaks when compiled to vanilla js
import * as Chime from 'aws-sdk/clients/chime';

// ✅  The default export is the Chime client class anyway, no need for wildcard import
import Chime from 'aws-sdk/clients/chime';
```

devalevenkatesh:
Thanks for pointing this out and helping us correct it. We will be updating the README for this soon.

[...]

PR #1436, opened 2021-07-15
update readme

jing-chen1:
**Issue #:** #1384

**Description of changes:**
Update the readme since the wildcard imports breaks when compiled to vanilla JS. So change the 
`import * as Chime from 'aws-sdk/clients/chime';` to `import Chime from 'aws-sdk/clients/chime';`

**Testing**

2. How did you test these changes? It's just a readme change
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Nope

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ Create a messaging session in your client application to receive messages from A
 
 ```js
 import * as AWS from 'aws-sdk/global';
-import * as Chime from 'aws-sdk/clients/chime';
+import Chime from 'aws-sdk/clients/chime';
 
 import {
   ConsoleLogger,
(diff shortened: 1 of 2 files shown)


devalevenkatesh (changes requested):
The bug-report-form is merged, can you rebase and check once? These changes wont come in your PR now.

README.md
```diff
@@ -172,7 +172,7 @@ Create a messaging session in your client application to receive messages from A
 
 ```js
 import * as AWS from 'aws-sdk/global';
-import * as Chime from 'aws-sdk/clients/chime';
+import Chime from 'aws-sdk/clients/chime';
```
devalevenkatesh, line 175: I think we should test before making this change, I had check with @kmcqueen-amzn who ran into issue when running the demo chat app after using the import without wildcard import.

jing-chen1: Oh, I thought we just need to update it as you commented in that issue. 
So @kmcqueen-amzn ran into issue without wildcard import while the builder has issue with the wildcard import?

devalevenkatesh: Yup. Sorry my bad I commented that it was a documentation update as I saw a commit for the same from @kmcqueen-amzn in the issue. Can you please sync with him once and check if you run into the same issue?

jing-chen1: I saw he made a commit in his branch, but never create a PR. 
Sure, will reach out to him.

jing-chen1: Thanks for the reminder Venky! just confirmed with @kmcqueen-amz, wildcard import works with the demo as expected, but removing the wildcard *does not* work.

Will close this PR.

PR #1449, opened 2021-07-20
Remove setup test to copy run-test script

devalevenkatesh:
This is not required now as we have the run-test script since 2.10.0

**Issue #:**

**Description of changes:**
- Remove setup test to copy run-test script as it is added since 2.10.0.

**Testing**

1. Have you successfully run `npm run build:release` locally? NA.
2. How did you test these changes? NA.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? No, this is for previous version check which will be checked once the run the triggered based on the cron scheduler in the github workflow..

### DIFF
--- a/.github/workflows/prev-version-integration.yaml
+++ b/.github/workflows/prev-version-integration.yaml
@@ -56,12 +56,7 @@ jobs:
       - name: Install Kite
         run: integration/js/script/install-kite
       - name: Setup
-        # Temporarily we need to copy the run-test script over from 2.9.0. Can be removed after 2.10.0 is released
-        run: |
-          cp integration/js/script/run-test integration/js/script/run-test.bku
-          git checkout tags/v${{ steps.prev.outputs.prev_version }}
-          cp integration/js/script/run-test.bku integration/js/script/run-test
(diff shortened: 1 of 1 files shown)


.github/workflows/prev-version-integration.yaml
```diff
@@ -55,13 +55,6 @@ jobs:
           tunnelIdentifier: ${{ steps.create-job-id.outputs.uuid }}
       - name: Install Kite
         run: integration/js/script/install-kite
-      - name: Setup
-        # Temporarily we need to copy the run-test script over from 2.9.0. Can be removed after 2.10.0 is released
-        run: |
-          cp integration/js/script/run-test integration/js/script/run-test.bku
-          git checkout tags/v${{ steps.prev.outputs.prev_version }}
```
ltrung, line 62: We probably need this line to check out the previous version :D

devalevenkatesh: Yes, got it, thanks! I have kept the setup job with the checkout step now.

ltrung (approved):

Issue #1316, opened 2021-05-25
Switching video inputs causes instability on certain mobile devices


jheising:
# Problem

Calling `meetingSession.audioVideo.chooseVideoInputDevice` using a `facingMode` or `deviceID` causes the following error to occur:

`
2021-05-18T15:59:25.589Z [ERROR] SDK - failed to get video device for constraints {"video":{"deviceId":{"exact":"defd88d947ae14dba27177afb67ef6c240451813ea401fcc4a31d62093e6b2de"},"width":{"ideal":960},"height":{"ideal":540},"frameRate":{"ideal":15},"googCpuOveruseDetection":true,"googCpuOveruseEncodeUsage":true,"googCpuOveruseThreshold":85,"googCpuUnderuseThreshold":55}}: NotReadableError: Could not start video source
meetingV2.ts:2007 [DEMO] Received an event: {"name":"videoInputFailed","attributes":{"attendeeId":"4af6414a-e88a-e3c7-3920-7b107d4da593","browserMajorVersion":"89","browserName":"Chrome","browserVersion":"89.0.4389.105","deviceName":"Samsung SM-G973F","externalMeetingId":"jheising","externalUserId":"1bafe552#Browser","meetingHistory":[{"name":"audioInputSelected","timestampMs":1621353500077},{"name":"videoInputSelected","timestampMs":1621353501001},{"name":"meetingStartRequested","timestampMs":1621353504847},{"name":"videoInputUnselected","timestampMs":1621353504868},{"name":"meetingStartSucceeded","timestampMs":1621353506033},{"name":"attendeePresenceReceived","timestampMs":1621353506216},{"name":"videoInputSelected","timestampMs":1621353519076},{"name":"videoInputFailed","timestampMs":1621353531627},{"name":"videoInputFailed","timestampMs":1621353539818},{"name":"videoInputSelected","timestampMs":1621353544169},{"name":"videoInputFailed","timestampMs":1621353548259},{"name":"videoInputFailed","timestampMs":1621353561652},{"name":"videoInputFailed","timestampMs":1621353565588}],"meetingId":"56fd5d69-def0-493a-9198-0dec276c3c82","osName":"Android","osVersion":"9","sdkVersion":"2.9.0","sdkName":"amazon-chime-sdk-js","timestampMs":1621353565588,"videoInputErrorMessage":"NotReadableError: Could not start video source"}}`

# Details

[...]

PR #1452, opened 2021-07-20
Stop video tracks before switching video device input

michhyun1:
**Issue #:**
#1316 

**Description of changes:**

Previously, we releaseActiveDevice after we call getUserMedia. With our audio device switch, we call releaseActiveDevice before getUserMedia. This change will stop the video track before calling getUserMedia for the next stream so that we don't run into the NotReadableError.

Later, we may want to look into replaceTrack() api in order to allow us to switch video tracks without having to renegotiate every time we switch video streams.

**Testing**

2. How did you test these changes? Locally - QA tested. 
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes, can be tested on the meeting demo using multiple devices. Just switch between video input devices.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,23 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+[Unreleased]
+
+### Added
+
+### Changed
+
+### Removed
+
+### Fixed
+- Stop `activeDevice` video track before selecting a new device to prevent `NotReadableError` when calling `getUserMedia` for a new video input device.
+
 ## [2.14.0] - 2021-07-23
 
 ### Added
 
 - Added `VideoPriorityBasedPolicyConfig` to control video downlink policy with network event response and recovery delays. Check [User Guide for Priority-based Downlink Policy](https://aws.github.io/amazon-chime-sdk-js/modules/prioritybased_downlink_policy.html#user-guide-for-priority-based-downlink-policy) for more information.
 - Amazon Chime SDK Project Board Overview and Guide.
-
 - Added 25 video tile support for demo app.
 
 ### Changed
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -507,7 +507,7 @@ export default class DefaultDeviceController
     if (this.boundAudioVideoController?.videoTileController.hasStartedLocalVideoTile()) {
       // optimized method exists, a negotiation can be avoided
       if (this.boundAudioVideoController.replaceLocalVideo) {
-        this.restartLocalVideoAfterSelection(null, false, true);
+        this.restartLocalVideoAfterSelection(false, true);
       } else {
         // non-optimized path, a negotiation is coming
         await this.boundAudioVideoController.update();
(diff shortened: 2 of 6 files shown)


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1398,10 +1386,15 @@ export default class DefaultDeviceController
...
     }
-
-    if (kind === 'audio' && this.activeDevices[kind] && this.activeDevices[kind].stream) {
-      this.releaseActiveDevice(this.activeDevices[kind]);
+    if (this.activeDevices[kind] && this.activeDevices[kind].stream) {
+      /* istanbul ignore else */
+      if (kind === 'audio') {
+        this.releaseActiveDevice(this.activeDevices[kind]);
+      } else if (kind === 'video') {
+        this.stopTracksAndRemoveCallback('video');
```
ltrung, line 1394: So we stop tracks and release it later after getUserMedia? is it just to be safe or do you know the reason why can't we just call `releaseActiveDevice` here?

michhyun1: The reason we cant call releaseActiveDevice is because it calls stopLocalVideo tile. The problem with that is that in `restartLocalVideoAfterSelection` we only `restartLocalVideo` if `hasStartedLocalVideoTile` is true. 

Meaning we wont actually update the peerConnection if stop the localVideoTile.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1244,7 +1244,6 @@ export default class DefaultDeviceController
   }
 
   private async restartLocalVideoAfterSelection(
-    oldDevice: DeviceSelection,
```
devalevenkatesh, line 1247: Just verifying if this is updated at all places where called.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [Documentation] Update priority based downlink policy guide.
 - Lookup `groupId` from device cache instead of directly from media stream when selecting input device.
+- Stop activeDevice video track before selecting a new device to prevent NotReadableError when calling getUserMedia for a a new video input device
```
devalevenkatesh, line 21: Can we backtick the activeDevice, NotReadableError and getUserMedia here? Also, could you please end this with a dot?

src/devicecontroller/DefaultDeviceController.ts
michhyun1: yup - global search returned nothing.

CHANGELOG.md
michhyun1: will do

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Documentation] Update priority based downlink policy guide.
 - Lookup `groupId` from device cache instead of directly from media stream when selecting input device.
 - [Documentation] Update documentation for 25 video tiles.
+- Stop `activeDevice` video track before selecting a new device to prevent `NotReadableError` when calling `getUserMedia` for a a new video input device.
```
devalevenkatesh, line 22: nit: Remove an extra "a" here.

ltrung (dismissed):

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Documentation] Update priority based downlink policy guide.
 - Lookup `groupId` from device cache instead of directly from media stream when selecting input device.
 - [Documentation] Update documentation for 25 video tiles.
+- Stop `activeDevice` video track before selecting a new device to prevent `NotReadableError` when calling `getUserMedia` for a new video input device.
```
ltrung, line 22: Need to move this to `Unrelease section`

ltrung (approved):

PR #1453, opened 2021-07-20
Add debug level logs as an option in the meeting demo

michhyun1:
**Issue #:**

**Description of changes:**

Adds a checkbox to the meeting demo to enable debug level logs.
image

**Testing**

Run the demo and enable debug logs.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Use the checkbox added on the meeting join page.

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.html
+++ b/demos/browser/app/meetingV2/meetingV2.html
@@ -58,6 +58,16 @@
         <div class="col-12">
           <fieldset>
             <legend>Choose your optional features</legend>
+            <div class="custom-control" style="text-align: left;">
+              <label for="logLevel">Choose a Log Level:</label>
+              <select name="logLevel" id="logLevelSelect">
+                <option value="INFO">INFO</option>
+                <option value="DEBUG">DEBUG</option>
+                <option value="WARN">WARN</option>
+                <option value="ERROR">ERROR</option>
+                <option value="OFF">OFF</option>
+              </select>
+            </div>
             <div class="custom-control custom-checkbox" style="text-align: left;">
               <input type="checkbox" id="webaudio" class="custom-control-input">
               <label for="webaudio" class="custom-control-label">Use WebAudio</label>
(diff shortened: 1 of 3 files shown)


demos/browser/app/meetingV2/meetingV2.html
```diff
@@ -62,6 +62,10 @@ <h1 class="h3 mb-3 font-weight-normal">Join a meeting</h1>
               <input type="checkbox" id="webaudio" class="custom-control-input">
               <label for="webaudio" class="custom-control-label">Use WebAudio</label>
             </div>
+            <div class="custom-control custom-checkbox" style="text-align: left;">
```
ltrung, line 65: nit: Can we make this a category with ERROR, WARN, INFO, and DEBUG? :D

michhyun1: Yeah makes sense

devalevenkatesh (commented):
Would it be good to use a query param instead? We have checkbox options which enable-disable the features so for logging can we just do it using a `logLevel` param like we do in the react library? Just a thought good with any approach.

ltrung:
> Would it be good to use a query param instead? We have checkbox options which enable-disable the features so for logging can we just do it using a `logLevel` param like we do in the react library? Just a thought good with any approach.

Either way would be fine but putting it as an option in the demo is probably easier for QA to use.

michhyun1:
> > Would it be good to use a query param instead? We have checkbox options which enable-disable the features so for logging can we just do it using a `logLevel` param like we do in the react library? Just a thought good with any approach.
> 
> Either way would be fine but putting it as an option in the demo is probably easier for QA to use.

Yeah I was going to say the same - checkbox is just easier. For customers also

demos/browser/app/meetingV2/meetingV2.html
ltrung: Nice screenshot 👍

ltrung (approved):

PR #1465, opened 2021-07-27
Update event buffer to retry with backoff

devalevenkatesh:
**Issue #:**
- Current, retrying for failed HTTP status codes does not consider backing off.

**Description of changes:**
- Adds backoff with full jitter based on existing [`FullJitterBackoff`](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/backoff/FullJitterBackoff.ts#L38-L39) class.
- Adds a separate metric to get the network errors.
- Updates logging to log errors only when there is unexpected HTTP status codes for which wont retry and mark them immediately as failures, rest all converted to be either `DEBUG` or `WARN` logs.
- Added `meetingStartSucceeded` to important events as requested by ingestion backend team.
- Removed `JSONIngestionBufferItem` as it is not needed since the retrying is handled in the `send` function which makes the fetch call.
- Add `wait` promise based function to a new file within a new `utils` folder under `src` named `web-utils.ts` but did not export it as well as it is ignored from matching interface rule from the `check-code-style` script as this rule does not apply to utils.
- This same function from src is now used in tests. The existing tests `delay` function is removed now.

**Testing**

- Tested event ingestion with the demo browser meeting app.
- Tested retry by adding Network error as well for retry and then re-connected the network, checked if multiple retries take place.
- Also, tested the time interval which seems to be random as expected between multiple retries.

3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes, browser demo.


simmkyu (dismissed):

src/eventbuffer/InMemoryJSONEventBuffer.ts
```diff
@@ -519,15 +482,21 @@ export default class InMemoryJSONEventBuffer implements EventBuffer<EventData>,
         },
         body: data,
       });
-      return response;
+      if (response.ok || !InMemoryJSONEventBuffer.SENDING_FAILURE_CODES.has(response.status)) {
+        return response;
+      } else {
+        const newRetryCount = retryCount + 1;
+        await this.wait(this.getBackoffWaitTime(newRetryCount));
+        return this.send(data, newRetryCount);
```
asnyatkov, line 490: I suggest to avoid recursion and implement retry logic as a loop.

src/eventreporter/DefaultMeetingEventReporter.ts
```diff
@@ -21,6 +21,7 @@ export default class DefaultMeetingEventReporter implements EventReporter, Destr
     'meetingStartFailed',
     'audioInputFailed',
     'videoInputFailed',
+    'meetingStartSucceeded',
```
ltrung, line 24: This does not seem related to retry with backoff. Should we document it?

src/eventbuffer/JSONIngestionBufferItem.ts
```diff
@@ -1,9 +0,0 @@
-// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
-// SPDX-License-Identifier: Apache-2.0
-
-import JSONIngestionEvent from './JSONIngestionEvent';
-
-export default interface JSONIngestionBufferItem {
```
ltrung, line 6: Can we delete a public interface?

src/eventbuffer/InMemoryJSONEventBuffer.ts
ltrung: Agree :)

src/eventreporter/DefaultMeetingEventReporter.ts
devalevenkatesh: Yes, I have mentioned this in the description bullet point 5 :)

src/eventbuffer/JSONIngestionBufferItem.ts
ltrung: Discussed and this should be fine since we did not export it.

src/eventbuffer/InMemoryJSONEventBuffer.ts
devalevenkatesh: Cool, thanks for the explanation offline. I have updated retrying logic to use loop over recursion.

src/eventbuffer/InMemoryJSONEventBuffer.ts
```diff
@@ -488,46 +435,72 @@ export default class InMemoryJSONEventBuffer implements EventBuffer<EventData>,
...
+          headers: {
+            Authorization: `Bearer ${this.authenticationToken}`,
+          },
+          body: data,
+        });
+        if (response.ok || !InMemoryJSONEventBuffer.SENDING_FAILURE_CODES.has(response.status)) {
+          return response;
+        } else {
+          retryCount++;
+          const backoffTime = this.getBackoffWaitTime(retryCount);
```
asnyatkov, line 488: You can skip waiting step if `retryCount == this.retryCountLimit`

src/eventbuffer/InMemoryJSONEventBuffer.ts
```diff
@@ -425,53 +385,40 @@ export default class InMemoryJSONEventBuffer implements EventBuffer<EventData>,
...
-              `Event Reporting - InMemoryJSONEventBuffer - sendEventImmediately - Error reading OK response ${err}`
-            );
-          }
-        }
-      } while (
-        response &&
-        InMemoryJSONEventBuffer.SENDING_FAILURE_CODES.has(response.status) &&
-        ++retryCount < this.retryCountLimit
+      this.logger.debug(
+        `Event Reporting - InMemoryJSONEventBuffer - sendEventImmediately - body ${body}`
```
asnyatkov, line 393: Just want to confirm: do you really want to dump the entire event body in all log messages?

devalevenkatesh: Actually, it helps in debugging any failures for the rest of logs and does not need any mapping to do between the logs.

I can remove this one though, as I already print the received item at the top in this method.

src/eventbuffer/InMemoryJSONEventBuffer.ts
devalevenkatesh: Makes sense, adding a check here. Nice call!

src/eventbuffer/InMemoryJSONEventBuffer.ts
asnyatkov: No, it's ok, I just wanted to check the intent. We can make the adjustments later.

asnyatkov (dismissed):

src/eventbuffer/InMemoryJSONEventBuffer.ts
```diff
@@ -488,46 +432,75 @@ export default class InMemoryJSONEventBuffer implements EventBuffer<EventData>,
...
+    return (
+      Math.random() *
+      Math.min(
+        InMemoryJSONEventBuffer.RETRY_LONG_BACKOFF_MS,
+        InMemoryJSONEventBuffer.RETRY_SHORT_BACKOFF_MS * Math.pow(2.0, retryCount)
+      )
+    );
+  }
+
+  private wait(delayInMs: number): Promise<void> {
```
ltrung, line 466: nit: Should we move this to an util folder?

src/eventbuffer/InMemoryJSONEventBuffer.ts
```diff
@@ -488,46 +432,75 @@ export default class InMemoryJSONEventBuffer implements EventBuffer<EventData>,
...
-      const response = await fetch(this.ingestionURL, {
-        method: 'POST',
-        headers: {
-          Authorization: `Bearer ${this.authenticationToken}`,
-        },
-        body: data,
-      });
-      return response;
+      let retryCount = 0;
+      while (retryCount < this.retryCountLimit) {
```
ltrung, line 473: Could we extend the current FullJitterBackoff to handle custom retry count? Seem like useful in other place in the future too.

src/eventbuffer/InMemoryJSONEventBuffer.ts
devalevenkatesh: Actually I felt that the timeoutScheduler implementation here is a little complex than needed. So, based on the timeout I have pass a callback function when I do `timeoutScheduler.start(() => { call send again here })` hence, used this. I preferred not to separate this function since we do not have any other util folder. Could you please let me know what should I do here then?

src/eventbuffer/InMemoryJSONEventBuffer.ts
devalevenkatesh: Actually that has to be a static method and as explained in the description I want to use it at two places (sending important events and while sending normal events) hence cannot really create an object to get the different backoff time to use at two places when needed. So I implemented it in this class. Checking if I can extend existing implementation.

devalevenkatesh: I checked the implementaion further, and see there are two ways to achieve this. 
1. Add a separate class. This class wont implement the [`Backoff`](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/backoff/Backoff.ts) interface nor extend the existing [`FullJitterBackoff`](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/backoff/FullJitterBackoff.ts) as it actually wont use it. But this would be a re-usable class now.
```ts
export default class FullJitterBackoffWithRetryCount {
  private static FIXED_WAIT_MS = 0;
  private static SHORT_BACKOFF_MS = 0;
  private static LONG_BACKOFF_MS = 0;

  constructor(
    fixedWaitMs: number,
    shortBackoffMs: number,
    longBackoffMs: number
  ) {
    FullJitterBackoffWithRetryCount.FIXED_WAIT_MS = fixedWaitMs;
    FullJitterBackoffWithRetryCount.SHORT_BACKOFF_MS = shortBackoffMs;
    FullJitterBackoffWithRetryCount.LONG_BACKOFF_MS = longBackoffMs;
  }

  static nextBackoffAmountMs(retryCount: number): number {
    const fullJitterMs =
      Math.random() *
        Math.min(this.LONG_BACKOFF_MS, this.SHORT_BACKOFF_MS * Math.pow(2.0, retryCount)) +
      this.FIXED_WAIT_MS;
    return fullJitterMs;
  }
}
```

2. Add an optional `nextRetryIndex` to the Backoff interface and then change the `nextBackoffAmountMs` method implementation to handle this so for any number of objects we would get the needed correct backoff time for a particular retry count/index.
```ts
nextBackoffAmountMs(nextRetryIndex?: number): number;
```

```ts
 nextBackoffAmountMs(nextRetryIndex?: number): number {
    if (nextRetryIndex) {
      this.currentRetry = nextRetryIndex;
    }
    const fullJitterMs =
      Math.random() *
        Math.min(this.longBackoffMs, this.shortBackoffMs * Math.pow(2.0, this.currentRetry)) +
      this.fixedWaitMs;
    this.currentRetry += 1;
    return fullJitterMs;
  }
```

I prefer the first option over existing in-class implementation and the second since we should avoid an optional parameter. Let me know your thoughts on this.

src/eventbuffer/InMemoryJSONEventBuffer.ts
devalevenkatesh: As discussed have added a utils folder with this function. But we do not export it publicly.

src/eventbuffer/InMemoryJSONEventBuffer.ts
devalevenkatesh: Have updated the code to re-use the existing `FullJitterBackoff` class, thanks.

src/utils/wait.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export default function wait(waitTimeMs: number): Promise<void> {
```
ltrung, line 4: 1) Does this need to be in a folder or should it be under src as utils.ts? 
2) Can we remove this same code here https://github.com/aws/amazon-chime-sdk-js/blob/master/test/utils.ts and use this instead :)

devalevenkatesh: For the first, the SDK structure is like level 1 are all folders and then default exports, that's why I added it as a folder. So should we keep it the same? Also, I had some discussion with Kyu in initial reviews where it seems the JS SDK currently exports types as no default exports and we do not follow that convention for functions. So, what do you suggest here?

devalevenkatesh: For the second point, its a good call, but can I do it in a separate PR as there may be lot of places to update it 😅 .

devalevenkatesh: I have added a new commit addressing these two points request you to check.

src/utils/web-utils.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export function wait(waitTimeMs: number): Promise<void> {
```
ltrung, line 4: nit: We usually capitalize the file name so should be Web and does not use dash. So should be WebUtils.ts
Also should we just call it Utils.ts

devalevenkatesh: As discussed, I have updated the file name and corresponding changes. Resolving this now, thanks!

ltrung (approved):

PR #1469, opened 2021-07-28
Add NPM Release Integration Test and Backwards Compatibility Test on release branch only

michhyun1:
**Issue #:**

**Description of changes:**
1. Change the integration tests to actually generate a tarball and then use the tarball in the meeting demo before running integration tests.
2. Add Backwards compatibility workflow ONLY for the release branch. It will check out the previous meeting demo and install the new SDK tarball into the meeting demo and then run the full integ test suite

**Testing**

Locally, via Github Actions
n/a
n/a
n/a

### DIFF
--- a/.github/workflows/content-share-integration.yml
+++ b/.github/workflows/content-share-integration.yml
@@ -23,6 +23,15 @@
           uses: actions/checkout@v2
           with:
             fetch-depth: 0
+        - name: Pack the Chime SDK and install the tarball into the Demo
+          run: |
+            current_version=$(.github/script/get-current-version)
+            echo "Packing current version:" $current_version
+            npm run build
+            npm pack
+            cd demos/browser
+            npm uninstall amazon-chime-sdk-js
+            npm install ../../amazon-chime-sdk-js-$current_version.tgz
         - name: Check if needed to run
           id: test_needed
           run: |
--- a/.github/workflows/release-backwards-compatiblity.yml
+++ b/.github/workflows/release-backwards-compatiblity.yml
@@ -1,9 +1,15 @@
-name: Previous Version Integration Tests Workflow
+# For the backwards compatibility test, this should be run on the release branch only as part. 
+# The release branch should not be tagged with the new version yet (that happens when we create a Github Release)
+# so we can checkout the latest tag and install the new version tarball into the demo. 
+# For example, if we're about to release v2.14.0, first create a tarball (the relase commit should bump the pacakge.json version to v2.14.0) 
+# then check out the previous v2.13.0 demo using the v2.13.0 tag, then navigate to the v2.13.0 demo and install the v2.14.0 amazon-chime-sdk-js tarball.
+# Then run the entire integration test suite on the demo.
+name: Backwards Compatibility Test
 
 on:
-  schedule:
-    # More information on cron https://crontab.guru/
-    - cron: '0 1 * * *'
+  pull_request:
+    branches:
+      - master
 
 env:
   SELENIUM_GRID_PROVIDER: saucelabs
@@ -14,9 +20,10 @@ env:
   MESSAGING_USER_ARN: ${{secrets.MESSAGING_USER_ARN}}
 
 jobs:
-  prev-version-integ:
-    name: Previous Version Integration Tests
+  release-integ-test:
+    name: Run Backwards Compatible Integration Tests with Release Tarball
     runs-on: ubuntu-latest
+    if: ${{ github.head_ref == 'release' }}
 
     steps:
       - name: Setup Node.js - 14.x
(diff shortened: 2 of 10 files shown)


.github/workflows/release-backwards-compatiblity.yml
```diff
@@ -0,0 +1,83 @@
...
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
+      - name: Checkout Package
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Get Current Package version
+        id: package-version
+        uses: martinbeentjes/npm-get-version-action@69d4697a4cbb77459281496ef908153917e63cb2
```
devalevenkatesh, line 33: Checking if this should be `martinbeentjes/npm-get-version-action@master` based on the action [here](https://github.com/marketplace/actions/get-current-package-version).

.github/workflows/release-integration.yml
```diff
@@ -0,0 +1,79 @@
+name: Release Integration Test Workflow
```
ltrung, line 1: Is this needed? Can we just modify existing test to run on release branch PR?

.github/workflows/release-backwards-compatiblity.yml
michhyun1: We want to use the commit hash - this is for security reasons and is best practice. https://dev.to/mheap/improve-your-github-actions-security-1im7

.github/workflows/release-integration.yml
michhyun1: The only difference here is that we actually pack the tarball and use the tarball instead of `file: ../../` (the regular CI Workflow doesnt pack the tarball)

We could potentially replace our existing CI workflow to use a generated tarball by default.

.github/workflows/release-backwards-compatiblity.yml
devalevenkatesh: Any advantages using this action over adding a script ourselves to get the latest published npm version like we have for getting the previous version [here](https://github.com/aws/amazon-chime-sdk-js/blob/master/.github/script/get-prev-version)?

.github/workflows/release-backwards-compatiblity.yml
```diff
@@ -0,0 +1,83 @@
+name: Backwards Compatibility Test
```
devalevenkatesh, line 1: Would it be helpful for the team if we add some explanation at the top of these actions to help this understand more easily 😅?

.github/workflows/release-backwards-compatiblity.yml
```diff
@@ -0,0 +1,83 @@
...
+  TEST_TYPE: Github-Action
+  SAUCE_USERNAME: ${{secrets.SAUCE_USERNAME}}
+  SAUCE_ACCESS_KEY: ${{secrets.SAUCE_ACCESS_KEY}}
+  MESSAGING_USER_ARN: ${{secrets.MESSAGING_USER_ARN}}
+
+jobs:
+  release-integ-test:
+    name: Run Backwards Compatible Integration Tests with Release Tarball
+    runs-on: ubuntu-latest
+    if: ${{ github.head_ref == 'release' }}
```
devalevenkatesh, line 20: For learning: this `github.head_ref` is basically the current branch right? We do not check this out here at this line or earlier. So, do we manually run these workflows on release branches? As far as I understand, we are generating the current versions tarball from master branch itself and checking out the previous version and then trying to run the integration tests on that checked out versions demo so could you please help me understand how the release branch comes into play here?

.github/workflows/release-backwards-compatiblity.yml
michhyun1: Main benefit is that we don't need to create the script ourselves.

.github/workflows/release-backwards-compatiblity.yml
michhyun1: Yup - will add some comments on this one

.github/workflows/release-backwards-compatiblity.yml
michhyun1: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

The head_ref points to the name of the branch which is submitting the PR. The base_ref is the target branch (where it will be merged into after the PR is merged.) In this case, this workflow will only run if the PR is coming from the release branch.

This workflow should run the integ tests for the release branch - the main difference between this workflow and our regular integration workflows is that this workflow uses a `generated tarball` in the meeting demo, whereas the regular integration tests just have "file: ../.." as the dependency on the chime sdk.

Either we can switch all of our integration tests to use tarball like this workflow, or we should have a separate workflow (regular integ tests use "file: ../.." and another workflow uses tarball).

.github/workflows/content-share-integration.yml
```diff
@@ -17,12 +17,23 @@
...
       name: Content Share Integration Test
       runs-on: ubuntu-latest
+      if: ${{ github.head_ref != 'release' }}
 
       steps:
         - name: Checkout Package
           uses: actions/checkout@v2
           with:
             fetch-depth: 0
+        - name: Get Current Package version
```
devalevenkatesh, line 27: Since, this is repeating now for all steps can we add a github action or a script to generate this tarball prior and re-use everywhere?

.github/workflows/release-backwards-compatiblity.yml
devalevenkatesh: Cool, thanks for the explanation, I had searched on github head refs and found that it represent the current branch so got confused here. Thanks for clarifying.

.github/workflows/content-share-integration.yml
michhyun1: It might be easier to understand what is happening if we leave it the way it is - if we put it in a script and then run the script from the github action, it might not be as easy to see the order of steps from the github action - you'd have to go to the script and then see what the script is doing.

I'm going to replace the Get Current Package Version steps with our own script - I think its best practice to not use third party github actions if we don't need to. I'll put this one in a script since its straight forward what it does.

ltrung (approved):

devalevenkatesh (approved):

PR #1470, opened 2021-07-29
Fix priority-based downlink policy default behavior

ltrung:
**Issue:**
Currently, if priority-based downlink policy is enabled, the default behavior (i.e., builders would not manually choose any video using `chooseRemoteVideoSources`) does not work and would cause none of the remote videos to be rendered.

**Description of changes:**
Update the logic for default priority downlink policy:
- Initialize the `videoPreferences` variable to undefined instead of empty set to differentiate between initial value and when builders can set to not subscribe any remote video.
- Assign default video preference after populate it if the videoPreferences is undefined.
- Change the `VideoAdaptiveProbePolicy` to not set videoPreferences to undefined.

**Testing**

 - Unit test
 - Test in React meeting demo using the example for video priority downlink policy setup given in this doc https://github.com/aws/amazon-chime-sdk-component-library-react/blob/master/src/providers/MeetingProvider/docs/MeetingProvider.stories.mdx which uses the default behavior and verify that users can see each other remote video.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes but need to modify the meeting demo since it manually set the video priority, 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-[Unreleased]
-
-### Added
+## [Unreleased]
+### Added 
 
 ### Changed
 
@@ -15,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Stop `activeDevice` video track before selecting a new device to prevent `NotReadableError` when calling `getUserMedia` for a new video input device.
+- Fix priority-based downlink policy default behavior.
 
 ## [2.14.0] - 2021-07-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
 ### Added 
 
 ### Changed
--- a/docs/classes/videoadaptiveprobepolicy.html
+++ b/docs/classes/videoadaptiveprobepolicy.html
@@ -261,7 +261,7 @@
 								<p>Implementation of <a href="../interfaces/videodownlinkbandwidthpolicy.html">VideoDownlinkBandwidthPolicy</a>.<a href="../interfaces/videodownlinkbandwidthpolicy.html#addobserver">addObserver</a></p>
 								<p>Inherited from <a href="videoprioritybasedpolicy.html">VideoPriorityBasedPolicy</a>.<a href="videoprioritybasedpolicy.html#addobserver">addObserver</a></p>
 								<ul>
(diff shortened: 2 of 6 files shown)


lzongkun (approved):
🚀

lzongkun (approved):
🚀

franko-franicevich:
Weirdly, since we upgraded to 2.15 we've been having exactly the behaviour described in this patch; which this is supposed to fix... Instead, for us, it *causes* the issue.

We're using Simulcast, and relying on the default behaviour to prioritise all streams equally. This was working fine before. It broke now.

Reverting this patch *fixes* the behaviour for us.

Any insight? For the original dev, were you using simulcast?

After debugging this, the root of the issue seems to be the logic where "videoPreferences" is set to "undefined."

Then, the first time `VideoPriorityBasedPolicy:updateIndex` is run, it sets video preferences to  'defaultVideoPreferences' - After which, videoPreferences is *fixed* to the initial set of streams in the first index. It will never again re-calculate a default set as new participants join, as it will always check for the existence of `videoPreferences`, and since it exists, will not update it.
This is different behaviour to what was happening prior to this patch; where the default preference set was recalculated with each index update.

ltrung:
> Weirdly, since we upgraded to 2.15 we've been having exactly the behaviour described in this patch; which this is supposed to fix... Instead, for us, it _causes_ the issue.
> 
> We're using Simulcast, and relying on the default behaviour to prioritise all streams equally. This was working fine before. It broke now.
> 
> Reverting this patch _fixes_ the behaviour for us.
> 
> Any insight? For the original dev, were you using simulcast?
> 
> After debugging this, the root of the issue seems to be the logic where "videoPreferences" is set to "undefined."
> 
> Then, the first time `VideoPriorityBasedPolicy:updateIndex` is run, it sets video preferences to 'defaultVideoPreferences' - After which, videoPreferences is _fixed_ to the initial set of streams in the first index. It will never again re-calculate a default set as new participants join, as it will always check for the existence of `videoPreferences`, and since it exists, will not update it.
> This is different behaviour to what was happening prior to this patch; where the default preference set was recalculated with each index update.

It seems that was my oversight. I will take a look and fix this. However, just curious, how did the default work for you before? When I found this issue, the preference was never set to the default behavior.

franko-franicevich:
I'll try summarise as clearly as I can, please ask if I've missed something:

1. We're using Simulcast (so using `VideoAdaptiveProbePolicy` which extends from `VideoPriorityBasedPolicy`) (eg, `videoDownlinkBandwidthPolicy = new VideoAdaptiveProbePolicy(this.logger);`)
2. We never set any video preferences, which means that the old 'default' behaviour is that it will always just create a default set that includes all the current streams/participants.

This meant we didn't need to worry about the observers to be alerted to new streams, etc, and manually manage the videoPreferences. It just 'worked'.

Please don't hesitate to ask for more clarification where you need more detail.

ltrung:
@franko-franicevich Thanks. I have been able to reproduce the issue and raised a PR to fix this https://github.com/aws/amazon-chime-sdk-js/pull/1531. Can you pull the code and give it a try to see if it fixes the issue on your end?

franko-franicevich:
Thanks! Looking at the commit, it matches my understanding of the issue, I'll test this later today and get back to you.
Really appreciate the quick response on this.

ltrung:
We just release a patch 2.16.1 to fix this issue. Feel free to check it out.

franko-franicevich:
I've reverted to 2.15 and replicated the original issue I reported. Then I upgraded to 2.16.1, and confirmed that the issue is resolved. All looks good, thanks!

PR #1478, opened 2021-07-30
Fix Maven installation script

ltrung:
**Description of changes:**
The maven installation script is broken as the previous URL to download maven no longer works. Update to use the new URL and remove the usage from KITE repo.
**Testing**


ltrung:
Close as dup of #1477 1477

PR #1479, opened 2021-08-01
Replaced any type for real-time data messages with generic types.

terrasoff:
**Description of changes:**
- Replaced `any` type of data message with strict types.
- Got rid of `any` type to have strict types.
- Removed type definition duplicates and control message type in one place.

**Testing**

Yes.

I extended tests with new types.

It can be used in any demo application, which uses test messages with JSON types.

No.

No.

### DIFF
--- a/src/realtimecontroller/RealtimeDataMessage.ts
+++ b/src/realtimecontroller/RealtimeDataMessage.ts
@@ -1,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+type RealtimeDataMessage = Uint8Array | string;
+
+export default RealtimeDataMessage;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [Unreleased]
 
 ### Added
+- Added strict types for the real-time messaging. 
 
 ### Changed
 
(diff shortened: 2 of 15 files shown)


src/realtimecontroller/RealtimeDataMessage.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+type RealtimeDataMessage = Uint8Array | string;
```
ltrung, line 4: The original was `Unit8Array | string | any` right?

terrasoff: Correct. I just left types types that are compatible with `WebSocket.send()` [signature](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/send).

ltrung: We support `any` for JSON (https://github.com/aws/amazon-chime-sdk-js/blob/master/src/task/SendAndReceiveDataMessagesTask.ts#L68) so I don't think we can change that.
It is also a breaking change for the library.

ltrung:
Could you add more details in the descriptions of the PR on what the issues and what the benefits of fixing it?

src/realtimecontroller/RealtimeDataMessage.ts
terrasoff: Generally, I planned to get rid of `any` type to have strict types.
And set up messages type in one place.
Now I see it can be enhanced with a generic type of message like so:
```ts
type MyMessageType = { subject: string, message: string };
const data = { subject: 'test', message: 'test message' };
context = new AudioVideoControllerState<MyMessageType>(); // can accept any as well
context.realtimeController.realtimeSendDataMessage(topic, data, 100);
```
Would you like me to do that or just bring `any` back?

ltrung:
Closing for now due to inactivity.

PR #1491, opened 2021-08-04
Update NScale uplink policy to scale down resolution based on number of remote videos

ltrung:
**Description of changes:**
Previously, NScale video uplink policy only update the bitrate when the number of remote videos increase.  
Now, we also try to scale down the video resolution based on the number of remote videos as well as the current video capture quality, except for the active speaker who would always have no resolution scale down.

The default logic would be to scale down the capture resolution to 540p, 480p, 360p, 270, and 180p when the number of videos increases to more than 2, 4, 8, 12, 16. For less than or equal to 2 videos, the resolution will not be scale down. 

Below is the table of what the bitrate and resolution would be with the default 1400 max bitrate, 540p video resolution and 15 fps:
| Number of videos  | Max bitrates (kbps) | Resolution (p) | Total max bitrates (kbps) | Total pixels |
| ------------- | ------------- | ------------- | ------------- | ------------- | 
| 1 | 1400 | 540  |  1400 | 518,400
| 2 | 1400 | 540  | 2800 | 1,036,800
| 3 | 933 | 540  | 2799 | 1,555,200
| 4 | 933 | 540  | 3732 | 2,073,600
| 5 | 746 | 480  | 3730 | 2,048,000
| 6 | 641 | 480  | 3846 | 2,457,600
| 7 | 566 | 480  | 3962 | 2,867,200
| 8 | 509 | 480  | 4072 | 3,276,800
| 9 | 466 | 360  | 4194 | 2,073,600
| 10 | 431 | 360  | 4310 | 2,304,000
| 11| 402 | 360  |  4422 | 2,534,400
| 12 | 378 | 360  | 4536 | 2,764,800
| 13 | 358 | 270  | 4654 | 1,684,800
| 14 | 340 | 270  | 4760 | 1,814,400
| 15 | 325 | 270  | 4875 | 1,944,000
| 16 | 312 | 270  | 4992 | 2,073,600
| 17 | 301 | 180  | 5117 |  979,200
| 18 | 290 | 180 | 5220 | 1,036,800
| 19 | 281 | 180  | 5339 | 1,094,400
| 20 | 273 | 180  | 5460 | 1,152,000
| 21 | 265 | 180  | 5565 | 1,209,600
| 22 | 258 | 180  | 5676 | 1,267,200
| 23 | 252 | 180  | 5796 | 1,324,800
| 24 | 246 | 180  | 5904 | 1,382,400
| 25 | 241 | 180  | 6025 | 1,440,000

**Testing**

    - Have 25 attendees join a meeting with everyone muted.
    - Turn on 2 videos and make sure they are the same original capture resolution (e.g., 720p).
    - Turn on 2 more videos (4 total) and verify all 4 have 540p resolution.
    - Turn on 4 more videos and verify all 8 have 480p resolution.
    - Turn on 4 more videos and verify all 12 have 360p resolution
    - Turn on 4 more videos and verify all 16 have 270p resolution
    - Turn on 7 more videos and verify  all 25 have 180p resolution
    - Now repeat the test but with active speaker and verify that the resolution of the active speaker is not scaled down.
    Note that for Firefox the resolution may be different (for example, 540p will be 480p). 
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes using the meeting demo.
4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? Yes
5. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved? Yes.


src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
```diff
@@ -92,6 +93,23 @@ export default class NScaleVideoUplinkBandwidthPolicy implements VideoUplinkBand
...
+      scale = 1;
+    } else if (this.numParticipants <= 8) {
+      scale = 1.5;
+    } else if (this.numParticipants <= 16) {
+      scale = 2;
+    } else {
+      scale = 4;
+    }
+    return scale;
+  }
```
richnew10, line 111: Does it make sense to make this an abstract policy, so you could write

```typescript
class DefaultNScaleVideoUplinkBandwidthPolicy extends AbstractNScaleVideoUplinkBandwidthPolicy {
  scaleResolutionDownBy(): number {
    return …
  }
}
```

or to have the policy take a pure function of `numParticipants` as a constructor argument?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -1148,4 +1156,10 @@ export default class DefaultAudioVideoController
       );
     });
   }
+
+  private getScaleDownFactorFromUplinkPolicy(): number {
+    return this.meetingSessionContext.videoUplinkBandwidthPolicy.scaleResolutionDownBy
+      ? this.meetingSessionContext.videoUplinkBandwidthPolicy.scaleResolutionDownBy()
+      : 1;
+  }
```
richnew10, line 1164: This seems like we're going in the wrong direction for designing the interface for policies, no? Should policies be in charge of producing the `encodeParam` value or something more central, rather than returning a bunch of these individual fields that then get threaded through to the transceiver controller?

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
```diff
@@ -33,6 +33,11 @@ export default interface VideoUplinkBandwidthPolicy {
    */
   maxBandwidthKbps(): number;
 
+  /**
+   * Gets the factor by which to scale down the video during encoding.
+   */
+  scaleResolutionDownBy?(): number;
+
```
richnew10, line 40: Why isn't this baked into `chooseCaptureAndEncodeParameters` or `updateIndex` or something else?

richnew10 (commented):
Three overall comments:

1. It would be nice if the 'table' in the code matched the table in the PR description.
2. 135p is really low res. Does it actually need to go that low? How bad does the video look on a high-res display?
3. What is the total number of pixels and the total bitrate at each number of tiles? There is an ideal shape to this graph, and I want to know how close we get…

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
ltrung: So I followed the same pattern as `maxBandwidthKbps` and create a seperate function for `scaleResolutionDownBy` which is then called by `chooseCaptureAndEncodeParameters` and `updateIndex`. I think the reason these are included in the interface originally is because we do call it in `DefaultAudioVideoController.handleHasBandwidthPriority` for active speaker case.

src/audiovideocontroller/DefaultAudioVideoController.ts
ltrung: I agrees that it is confusing as defaultAudioVideoController actually tries to set these values and pass them through the transceiver, it is our existing code to handle the active speaker case so I would like to keep it that way to reduce the scope of this PR.

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
ltrung: If I understand correctly, you want to expose the scale down and bitrate calculation logic so builders can pass their own version? If so, our current recommendation is to copy the downlink policy to use their own logic.

rnewman:
What about 720p for 2 attendees?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -1026,11 +1026,18 @@ export default class DefaultAudioVideoController
...
-  private async enforceBandwidthLimitationForSender(maxBitrateKbps: number): Promise<void> {
+  private async enforceBandwidthLimitationForSender(
+    maxBitrateKbps: number
+  ): Promise<void> {
     if (this.meetingSessionContext.browserBehavior.requiresUnifiedPlan()) {
-      await this.meetingSessionContext.transceiverController.setVideoSendingBitrateKbps(
-        maxBitrateKbps
-      );
+      const encodingParams = this.meetingSessionContext.videoUplinkBandwidthPolicy.chooseEncodingParameters();
+      if (encodingParams.size <= 0) {
```
richnew10, line 1034: Can't be less than 0.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -1026,11 +1026,18 @@ export default class DefaultAudioVideoController
...
 
-  private async enforceBandwidthLimitationForSender(maxBitrateKbps: number): Promise<void> {
+  private async enforceBandwidthLimitationForSender(
+    maxBitrateKbps: number
+  ): Promise<void> {
     if (this.meetingSessionContext.browserBehavior.requiresUnifiedPlan()) {
-      await this.meetingSessionContext.transceiverController.setVideoSendingBitrateKbps(
-        maxBitrateKbps
-      );
+      const encodingParams = this.meetingSessionContext.videoUplinkBandwidthPolicy.chooseEncodingParameters();
```
richnew10, line 1033: If an empty map means "just use the max bitrate", how does the policy say "don't change anything"? Should we pass the bitrate (and the `Sender`?) as optional arguments to `chooseEncodingParameters`, and make the default policies set the `maxBitrate` encodign parameter, so only old policies take the no-entries route?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -1103,18 +1110,25 @@ export default class DefaultAudioVideoController
...
-
-      this.meetingSessionContext.videoUplinkBandwidthPolicy.setHasBandwidthPriority(
-        hasBandwidthPriority
-      );
-      const newMaxBandwidth = this.meetingSessionContext.videoUplinkBandwidthPolicy.maxBandwidthKbps();
+      if (this.meetingSessionContext.videoUplinkBandwidthPolicy.chooseEncodingParameters().size > 0) {
+        this.meetingSessionContext.videoUplinkBandwidthPolicy.setHasBandwidthPriority(
+          hasBandwidthPriority
+        );
+        await this.enforceBandwidthLimitationForSender(this.meetingSessionContext.videoUplinkBandwidthPolicy.maxBandwidthKbps());
```
richnew10, line 1117: You're computing the encoding parameters twice: once here and once in `enforceBandwidth`.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -1103,18 +1110,25 @@ export default class DefaultAudioVideoController
...
+        const oldMaxBandwidth = this.meetingSessionContext.videoUplinkBandwidthPolicy.maxBandwidthKbps();
 
-      if (oldMaxBandwidth !== newMaxBandwidth) {
-        this.logger.info(
-          `video send bandwidth priority ${hasBandwidthPriority} max has changed from ${oldMaxBandwidth} kbps to ${newMaxBandwidth} kbps`
+        this.meetingSessionContext.videoUplinkBandwidthPolicy.setHasBandwidthPriority(
+          hasBandwidthPriority
         );
-        await this.enforceBandwidthLimitationForSender(newMaxBandwidth);
+        const newMaxBandwidth = this.meetingSessionContext.videoUplinkBandwidthPolicy.maxBandwidthKbps();
```
richnew10, line 1124: This is a bit tricky. I'd _like_ to say that if the policy is returning encoding parameters, we don't also need to special-case max bandwidth, and avoid having this code.

src/transceivercontroller/DefaultTransceiverController.ts
```diff
@@ -16,8 +16,42 @@ export default class DefaultTransceiverController implements TransceiverControll
...
-  setEncodingParameters(_params: Map<string, RTCRtpEncodingParameters>): void {
-    return;
+  async setEncodingParameters(encodingParamMap: Map<string, RTCRtpEncodingParameters>): Promise<void> {
+    if (!this._localCameraTransceiver || this._localCameraTransceiver.direction !== 'sendrecv') {
+      return;
+    }
+
+    const sender = this._localCameraTransceiver.sender;
+    const newEncodingParams = Array.from(encodingParamMap.values());
+    if (newEncodingParams.length <= 0) {
```
richnew10, line 26: Never less than zero, and you can check the size of `encodingParamMap` instead of turning it into an array first.

src/transceivercontroller/DefaultTransceiverController.ts
```diff
@@ -16,8 +16,42 @@ export default class DefaultTransceiverController implements TransceiverControll
...
+              hasChanged = true;
+            }
+          }
+        }
+      }
+    }
+
+    if (hasChanged) {
+      await sender.setParameters(oldParam);
+    }
```
richnew10, line 54: Is there any reason not to just always call `setParameters`?

Can we just expect the policy to return an empty map if it gets to see `Sender` and its own state, and doesn't think anything needs to change?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -1148,4 +1172,11 @@ export default class DefaultAudioVideoController
       );
     });
   }
+
+  private useUpdateTransceiverController(): boolean {
+    return (
+      this.meetingSessionContext.browserBehavior.requiresUnifiedPlan() &&
+      !!this.meetingSessionContext.videoUplinkBandwidthPolicy.updateTransceiverController
+    );
+  }
```
richnew10, line 1181: If the policy never changes during a call, and neither does the browser behavior object, can we just set this as a private field during start rather than having it as a dynamic method?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -1103,18 +1118,27 @@ export default class DefaultAudioVideoController
...
-      );
-      const newMaxBandwidth = this.meetingSessionContext.videoUplinkBandwidthPolicy.maxBandwidthKbps();
+      if (this.useUpdateTransceiverController()) {
+        this.meetingSessionContext.videoUplinkBandwidthPolicy.setHasBandwidthPriority(
+          hasBandwidthPriority
+        );
+        await this.meetingSessionContext.videoUplinkBandwidthPolicy.updateTransceiverController(
+          this.meetingSessionContext.transceiverController
+        );
+      } else {
```
richnew10, line 1128: Maybe do this as a visually separate block with early return. That has the useful property of not touching the blame for the existing lines.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -920,6 +930,11 @@ export default class DefaultAudioVideoController
   private actionFinishUpdating(): void {
     // we do not update parameter for simulcast since they are updated in AttachMediaInputTask
     if (!this.meetingSessionContext.enableSimulcast) {
+      if (this.useUpdateTransceiverController()) {
+        this.meetingSessionContext.videoUplinkBandwidthPolicy.updateTransceiverController(
+          this.meetingSessionContext.transceiverController
+        );
+      }
       const maxBitrateKbps = this.meetingSessionContext.videoCaptureAndEncodeParameter.encodeBitrates()[0];
       this.enforceBandwidthLimitationForSender(maxBitrateKbps);
```
richnew10, line 939: Shouldn't we early return before going on to call this?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -630,6 +635,11 @@ export default class DefaultAudioVideoController
     this.signalingTask = undefined;
     this.meetingSessionContext.videoDuplexMode = SdkStreamServiceType.RX;
     if (!this.meetingSessionContext.enableSimulcast) {
+      if (this.useUpdateTransceiverController()) {
+        this.meetingSessionContext.videoUplinkBandwidthPolicy.updateTransceiverController(
+          this.meetingSessionContext.transceiverController
+        );
+      }
       this.enforceBandwidthLimitationForSender(
         this.meetingSessionContext.videoCaptureAndEncodeParameter.encodeBitrates()[0]
```
richnew10, line 644: Early return before this?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -495,6 +495,11 @@ export default class DefaultAudioVideoController
           this.configuration.credentials.attendeeId
         );
       }
+      if (this.meetingSessionContext.videoUplinkBandwidthPolicy.setTransceiverController) {
+        this.meetingSessionContext.videoUplinkBandwidthPolicy.setTransceiverController(
+          this.meetingSessionContext.transceiverController
+        );
+      }
       this.meetingSessionContext.audioProfile = this._audioProfile;
     }
```
richnew10, line 504: Just set a boolean field here.

Do you need to also check unified plan here?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -1148,4 +1172,11 @@ export default class DefaultAudioVideoController
       );
     });
   }
+
+  private useUpdateTransceiverController(): boolean {
+    return (
+      this.meetingSessionContext.browserBehavior.requiresUnifiedPlan() &&
```
richnew10, line 1178: That you can't use this with Plan B is useful information to expose in the docs for the policy! It means that N-Scale doesn't scale resolution in some Safari versions, right?

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
```diff
@@ -35,16 +44,21 @@ export default class NScaleVideoUplinkBandwidthPolicy implements VideoUplinkBand
   }
 
   updateIndex(videoIndex: VideoStreamIndex): void {
+    const hasLocalVideo = this.transceiverController
+      ? this.transceiverController.hasVideoInput()
+      : false;
```
richnew10, line 49: Nice!

richnew10 (changes requested):
Partial review before I have to hop on a customer call!

src/audiovideocontroller/DefaultAudioVideoController.ts
ltrung: I believe unified plan is for all browsers now. Plan-B is originally for Chrome but I think we plan to deprecate it soon. @hensmi-amazon Is that correct?

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
```diff
@@ -57,4 +58,16 @@ export default interface VideoUplinkBandwidthPolicy {
...
+   * Set a reference to the current transceiver controller
+   * @param {TransceiverController} transceiverController - The transceiver controller
+   */
+  setTransceiverController?(transceiverController: TransceiverController): void;
+
+  /**
+   * Update the transceiver controller
+   * @param {TransceiverController} transceiverController - The transceiver controller
+   */
+  updateTransceiverController?(transceiverController: TransceiverController): void;
```
richnew10, line 72: These need to be a lot more thorough.

* When will these methods be called?
* What will be called when the call ends, so the policy can tear down its state and unsubscribe from things? 
* What is the policy's responsibility if it implements these, and which other methods does it obsolete? (We will no longer set bandwidth for you, right?)
* What happens if Unified Plan is not enabled?

richnew10: Are these methods allowed to throw?

richnew10: Why do both methods take a transceiver controller? Can calls to `updateTC` be made with a different TC than `setTC`?

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
```diff
@@ -99,4 +113,54 @@ export default class NScaleVideoUplinkBandwidthPolicy implements VideoUplinkBand
...
+    };
+    if (
+      !this.isEncodingParameterEqual(
+        encodingParams,
+        this.encodingParamMap.get(NScaleVideoUplinkBandwidthPolicy.rid)
+      )
+    ) {
+      this.encodingParamMap.set(NScaleVideoUplinkBandwidthPolicy.rid, encodingParams);
+      this.transceiverController.setEncodingParameters(this.encodingParamMap);
+    }
```
richnew10, line 137: You're checking here for whether you need to do work at all.

Can you do so by checking your inputs (that is, same number of videos as last time and same max bitrate = no action needed)?

If we assume/require that `updateTransceiverController` implementations don't do work if they don't need it — that is, we guarantee that we will keep them up to date with whatever you picked, even if we reconnect etc — do we need additional checks prior to updating the sender?

src/transceivercontroller/DefaultTransceiverController.ts
```diff
@@ -16,8 +16,35 @@ export default class DefaultTransceiverController implements TransceiverControll
 
   constructor(protected logger: Logger, protected browserBehavior: BrowserBehavior) {}
 
-  setEncodingParameters(_params: Map<string, RTCRtpEncodingParameters>): void {
-    return;
+  async setEncodingParameters(
+    encodingParamMap: Map<string, RTCRtpEncodingParameters>
+  ): Promise<void> {
```
richnew10, line 21: You need to document the heck out of this.

* When should this method be called? What happens if you call it?
* What happens if the map is empty?
* What happens if it's the same (`===`) map as the last one you passed?
* What happens if its keys and values are equal?

src/transceivercontroller/DefaultTransceiverController.ts
```diff
@@ -16,8 +16,35 @@ export default class DefaultTransceiverController implements TransceiverControll
...
+    } else {
+      for (let i = 0; i < oldParam.encodings.length; i++) {
+        for (let j = 0; j < newEncodingParams.length; j++) {
+          if (oldParam.encodings[i].rid === newEncodingParams[j].rid) {
+            oldParam.encodings[i].maxBitrate = newEncodingParams[j].maxBitrate;
+            oldParam.encodings[i].scaleResolutionDownBy =
+              newEncodingParams[j].scaleResolutionDownBy;
+          }
+        }
+      }
```
richnew10, line 44: I'm not sure I understand this.

You take whatever was already there, and update it with precisely those two parameters.

It looks like this code (a) doesn't allow you to add new encoding parameters after the sender has been created, and (b) doesn't allow you to set new encoding parameters.

Is that deliberate? If so, why?

richnew10 (changes requested):

src/audiovideocontroller/DefaultAudioVideoController.ts
ltrung: Good callout. Updated.

src/audiovideocontroller/DefaultAudioVideoController.ts
ltrung: Updated.

src/audiovideocontroller/DefaultAudioVideoController.ts
ltrung: Oops. Do a else instead since we have more statements to execute after this.

src/audiovideocontroller/DefaultAudioVideoController.ts
ltrung: Updated.

src/transceivercontroller/DefaultTransceiverController.ts
ltrung: I mainly followed how we did it in Simulcast Transceiver Controller https://github.com/aws/amazon-chime-sdk-js/blob/master/src/transceivercontroller/SimulcastTransceiverController.ts#L47.
I am not exactly sure whether there is a reason we did that but my guess is maybe params can be changed so we cannot just do a direct assignment. Thus, just to be safe, I do it that way.

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
ltrung: There is no inputs for this method right or are you referring that we should keep check of previous bit rate and scale down? Not sure how is that different that we just do a diff of the encoding parameter?
> If we assume/require that updateTransceiverController implementations don't do work if they don't need it — that is, we guarantee that we will keep them up to date with whatever you picked, even if we reconnect etc — do we need additional checks prior to updating the sender?

Could you elaborate? This is mainly to guard against constantly need from setting encoding parameters every time active speaker changes (right now we do this check in audio video controller).

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
ltrung: Got issue with git branch this morning so I think i did not get my changes to remove the transceiver param in `updateTC` in.

src/transceivercontroller/DefaultTransceiverController.ts
richnew10: I know you need to mutate the param object given by the sender, and the `rid` is immutable, but that's not really the question — why is this code not something like this:

```typescript
for (const existing of oldParam.encodings) {
  for (const changed of newEncodingParams) {
    if (existing.rid !== changed.rid) {
      continue;
    }
    for (const key in changed) {
      // These can't be changed.
      if (key === 'rid' || key === 'codecPayloadType') {
        continue;
      }
      if (changed.hasOwnProperty(key)) {
        existing[key] = changed[key];
      }
    }
  }
}
```

which would allow the caller to change the `maxFramerate` or other properties?

src/audiovideocontroller/DefaultAudioVideoController.ts
hensmi-amazon: Moreso transition builders off/remove option to turn on before Chrome removes it completely :P, but yes we have an internal task.

src/transceivercontroller/DefaultTransceiverController.ts
ltrung: Updated the interface documentation. 
> What happens if the map is empty?
What happens if it's the same (===) map as the last one you passed?
What happens if its keys and values are equal?

Isn't this just implementation details? Do you want me to update it in `DefaultTransceiverController`?

richnew10: It's not implementation detail — it's the API contract that the builder of the policy (or even the transceiver controller!) is working with. "What happens when I do X" is a key question, and it drives policy implementation (e.g., reuse of objects, omitting unchanged keys, etc.).

richnew10: That applies to both the interface/protocol ("_all_ transceiver controllers must do X and Y") and the implementation ("this transceiver controller does X and Y").

src/audiovideocontroller/DefaultAudioVideoController.ts
ltrung: Updated.

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
ltrung: Updated.

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
ltrung: Updated the interface documentation.

src/transceivercontroller/DefaultTransceiverController.ts
ltrung: Nice.  Updated the code.

src/transceivercontroller/TransceiverController.ts
```diff
@@ -78,7 +78,13 @@ export default interface TransceiverController {
   localVideoTransceiver(): RTCRtpTransceiver;
 
   /**
-   * Set [[RTCRtpEncodingParameters]] on the sender of transceiver
+   * Set [[RTCRtpEncodingParameters]] on the sender of transceiver.
+   * This method is called when a video is enabled/disabled or when there is the active speaker changes.
+   * The input parameters should not be empty.
```
richnew10, line 83: "when the active speaker changes".

But here you're describing _what the default policy does_, right?

src/transceivercontroller/TransceiverController.ts
```diff
@@ -78,7 +78,13 @@ export default interface TransceiverController {
...
+   * The encoding parameters for sender should be retrieved using sender.getParameters and updated using
+   * sender.setParameters method.
```
richnew10, line 85: It would be worth explicitly splitting these docs into:

* When this should be called by policies or controllers.
* How this is used by the default policy.
* What is the API contract (e.g., non-empty map).
* What implementations must do.

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
```diff
@@ -57,4 +58,21 @@ export default interface VideoUplinkBandwidthPolicy {
...
+   * This method should not throw.
+   * @param {TransceiverController} transceiverController - The transceiver controller
+   */
+  setTransceiverController?(transceiverController: TransceiverController): void;
+
+  /**
+   * Update the transceiver controller that is set from setTransceiverController such as setEncodingParameters.
+   * Only used when unified plan is enabled but not available for simulcast for now.
+   * The method will be called when a video is enabled or disabled or
+   * when a new bandwidth priority event such as active speaker happens.
```
richnew10, line 74: How do I know which? Will it ever matter?

Is it OK for me, as a builder of a policy, to subscribe to `AudioVideoObserver` and `ActiveSpeakerWhatever` and just ignore this method altogether?

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
```diff
@@ -57,4 +58,21 @@ export default interface VideoUplinkBandwidthPolicy {
...
    */
   chooseMediaTrackConstraints(): MediaTrackConstraints;
+
+  /**
+   * Set a reference to the current transceiver controller.
+   * This will be called during connection and if unified plan is enabled.
+   * This method should not throw.
+   * @param {TransceiverController} transceiverController - The transceiver controller
+   */
+  setTransceiverController?(transceiverController: TransceiverController): void;
```
richnew10, line 68: How do I know when to end my subscriptions? Does my policy need to subscribe to `AudioVideoObserver`?

src/transceivercontroller/DefaultTransceiverController.ts
```diff
@@ -16,8 +16,43 @@ export default class DefaultTransceiverController implements TransceiverControll
...
+    if (!oldParam.encodings) {
+      oldParam.encodings = newEncodingParams;
+    } else {
+      for (const existing of oldParam.encodings) {
+        for (const changed of newEncodingParams) {
+          if (existing.rid !== changed.rid) {
+            continue;
+          }
+          let key: keyof RTCRtpEncodingParameters;
+          for (key in changed) {
```
richnew10, line 42: TIL!

src/transceivercontroller/DefaultTransceiverController.ts
```diff
@@ -16,8 +16,43 @@ export default class DefaultTransceiverController implements TransceiverControll
...
+            // These properties can't be changed.
+            if (key === 'rid' || key === 'codecPayloadType') {
+              continue;
+            }
+            if (changed.hasOwnProperty(key)) {
+              existing[key] = changed[key] as never;
+            }
```
richnew10, line 49: Is there a better way to model the value space of a keyof type? I haven't looked into this.

test/audiovideocontroller/DefaultAudioVideoController.test.ts
```diff
@@ -4,6 +4,7 @@
 import * as chai from 'chai';
 import * as sinon from 'sinon';
 
+import { NoVideoUplinkBandwidthPolicy } from '../../build';
```
richnew10, line 7: Fix these imports; they shouldn't import from `build`.

richnew10 (changes requested):

test/audiovideocontroller/DefaultAudioVideoController.test.ts
ltrung: Updated.

src/transceivercontroller/DefaultTransceiverController.ts
ltrung: So this is a workaround for the below TS error. Since there can be multiple types for existing[key], TS uses the intersection of all those types which is never to infer the type.
```
Type 'number | boolean | RTCDtxStatus' is not assignable to type 'never'.  
Type 'number' is not assignable to type 'never'
```

Alternatively, we can do:
- `(existing[key] as (number | boolean | RTCDtxStatus)) = changed[key];`
- or use ts-ignore.

Let me know if you can think of other ways :)

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
ltrung: Sorry what do you mean by subscription? This method is just to be set by audio video controller after connection right?

src/transceivercontroller/DefaultTransceiverController.ts
richnew10:
```typescript
type paramValue = RTCRtpEncodingParameters[keyof RTCRtpEncodingParameters];
existing[key] as paramValue = changed[key];
```

should work.

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
```diff
@@ -9,17 +10,25 @@ import VideoUplinkBandwidthPolicy from './VideoUplinkBandwidthPolicy';
 /** NScaleVideoUplinkBandwidthPolicy implements capture and encode
  *  parameters that are nearly equivalent to those chosen by the
  *  traditional native clients, except for a modification to
- *  maxBandwidthKbps described below. */
+ *  maxBandwidthKbps and scaleResolutionDownBy described below. */
 export default class NScaleVideoUplinkBandwidthPolicy implements VideoUplinkBandwidthPolicy {
+  static readonly rid = 'video';
```
richnew10, line 15: This needs to be in a docstring somewhere, right? If I write my own policy, I need this to be in the encoding parameters…

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
```diff
@@ -9,17 +10,25 @@ import VideoUplinkBandwidthPolicy from './VideoUplinkBandwidthPolicy';
...
+
   private numParticipants: number = 0;
   private optimalParameters: DefaultVideoAndEncodeParameter;
   private parametersInEffect: DefaultVideoAndEncodeParameter;
   private idealMaxBandwidthKbps = 1400;
   private hasBandwidthPriority: boolean = false;
+  private encodingParamMap = new Map<string, RTCRtpEncodingParameters>();
+  private transceiverController: TransceiverController;
 
   constructor(private selfAttendeeId: string) {
```
richnew10, line 25: You probably need a constructor argument here so that customers can turn the resolution adjustment off or on.

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
richnew10: I mean: say my policy subscribes to the active speaker detector, or some other observer interface.

It needs to unsubscribe from that interface when the call ends, and additionally drop its reference to the transceiver controller, in order to prevent leaks of call objects.

How does it know?

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+### Changed
+- Update the default behavior of NScale video uplink bandwidth policy to scale down resolution based on the number 
+  of videos.
```
richnew10, line 14: You might want more discussion about whether this is considered a breaking change, and if we don't consider it one, how to go back to the old behavior.

ltrung:
> Three overall comments:
> 
>     1. It would be nice if the 'table' in the code matched the table in the PR description.
> 
>     2. 135p is really low res. Does it actually need to go that low? How bad does the video look on a high-res display?
> 
>     3. What is the total number of pixels and the total bitrate at each number of tiles? There is an ideal shape to this graph, and I want to know how close we get…

Updated the table with total bitrates and pixels.

ltrung:
> What about 720p for 2 attendees?

For 2 videos, we do not change the resolution so they should still be 720p.

src/transceivercontroller/DefaultTransceiverController.ts
ltrung: Nice. I will update the code.

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
ltrung: So this is not relevant unless it is simulcast, the default rid is undefined or "" (in Safari). Maybe `rid` is a bad name here but this is just a dummy parameter since setEncodingParameters requires a string key.

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
ltrung: The policy should not need to subscribe to active speaker detector audio video observer. The audio video controller decides to set whenever the local video has bandwidth priority and call `setHasBandwidthPriority`.

CHANGELOG.md
ltrung: Updated

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
richnew10: I mean: I want to implement a downlink policy that makes sure that the active talker has the highest priority. How do I know who's the active talker? I make my policy subscribe to that.

But regardless, the point stands for other kinds of linkages and the leakage of the transceiver controller itself.

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
richnew10: That sounds like a great topic to cover in the developer doc for this public interface!

src/transceivercontroller/TransceiverController.ts
```diff
@@ -78,7 +78,17 @@ export default interface TransceiverController {
...
+   * This method should be called whenever the sender's encoding parameters of the local video transceiver need to
+   * be updated.
+   * For example, the default NScaleVideoUplinkBandwidthPolicy calls this method whenever a video is on/off or the
+   * active speaker changes.
+   * This method assumes that the sender of the local video transceiver is available and the input parameters should
+   * not be empty.
+   * The encoding parameters for sender should be retrieved using sender.getParameters and updated using
+   * sender.setParameters method.
+   * @param {Map<string, RTCRtpEncodingParameters>} params - The encoding parameters. If you have multiple encoding
+   * parameters for different video layers, the key should be the rid corresponding to the RTCRtpEncodingParameters.
```
richnew10, line 91: Nice, that helps.

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
```diff
@@ -9,17 +10,54 @@ import VideoUplinkBandwidthPolicy from './VideoUplinkBandwidthPolicy';
 /** NScaleVideoUplinkBandwidthPolicy implements capture and encode
  *  parameters that are nearly equivalent to those chosen by the
  *  traditional native clients, except for a modification to
- *  maxBandwidthKbps described below. */
+ *  maxBandwidthKbps and scaleResolutionDownBy described below. */
 export default class NScaleVideoUplinkBandwidthPolicy implements VideoUplinkBandwidthPolicy {
+  static readonly encondingMapKey = 'video';
```
richnew10, line 15: `encoding`

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
```diff
@@ -99,4 +153,54 @@ export default class NScaleVideoUplinkBandwidthPolicy implements VideoUplinkBand
...
+        encodingParams,
+        this.encodingParamMap.get(NScaleVideoUplinkBandwidthPolicy.encondingMapKey)
+      )
+    ) {
+      this.encodingParamMap.set(NScaleVideoUplinkBandwidthPolicy.encondingMapKey, encodingParams);
+      this.transceiverController.setEncodingParameters(this.encodingParamMap);
+    }
+  }
+
+  private isEncodingParameterEqual(
```
richnew10, line 179: `areEncodingParametersEqual`

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
```diff
@@ -99,4 +153,54 @@ export default class NScaleVideoUplinkBandwidthPolicy implements VideoUplinkBand
...
+    encoding1: RTCRtpEncodingParameters,
+    encoding2: RTCRtpEncodingParameters
+  ): boolean {
+    return (
+      encoding1.maxBitrate === encoding2.maxBitrate &&
+      encoding1.scaleResolutionDownBy === encoding2.scaleResolutionDownBy
```
richnew10, line 185: What about all of the other fields? If you're not going to compare them, probably don't use `equal` as the name — perhaps `encodingParametersHaveSameBitrateAndScaling`.

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
```diff
@@ -99,4 +153,54 @@ export default class NScaleVideoUplinkBandwidthPolicy implements VideoUplinkBand
...
+    ) {
+      this.encodingParamMap.set(NScaleVideoUplinkBandwidthPolicy.encondingMapKey, encodingParams);
+      this.transceiverController.setEncodingParameters(this.encodingParamMap);
+    }
+  }
+
+  private isEncodingParameterEqual(
+    encoding1: RTCRtpEncodingParameters,
+    encoding2: RTCRtpEncodingParameters
+  ): boolean {
```
richnew10, line 182:
```
if (encoding1 === encoding2) {
  return true;
}
```

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
```diff
@@ -99,4 +153,54 @@ export default class NScaleVideoUplinkBandwidthPolicy implements VideoUplinkBand
...
+    return (
+      encoding1.maxBitrate === encoding2.maxBitrate &&
+      encoding1.scaleResolutionDownBy === encoding2.scaleResolutionDownBy
+    );
+  }
+
+  private calculateEncodingParameters(setting: MediaTrackSettings): RTCRtpEncodingParameters {
+    const maxBitRates = this.maxBandwidthKbps() * 1000;
```
richnew10, line 190: `maxBitrate`, then you can just…

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
```diff
@@ -99,4 +153,54 @@ export default class NScaleVideoUplinkBandwidthPolicy implements VideoUplinkBand
...
+    let scale = 1;
+    if (setting && this.scaleResolution && !this.hasBandwidthPriority && this.numParticipants > 2) {
+      const targetHeight = NScaleVideoUplinkBandwidthPolicy.targetHeightArray[this.numParticipants];
+
+      scale = Math.max(setting.height / targetHeight, 1);
+    }
+    return {
+      scaleResolutionDownBy: scale,
+      maxBitrate: maxBitRates,
```
richnew10, line 199: … use `maxBitrate,` here.

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
```diff
@@ -99,4 +153,54 @@ export default class NScaleVideoUplinkBandwidthPolicy implements VideoUplinkBand
...
+    };
+  }
+
+  private getStreamCaptureSetting(): MediaTrackSettings {
```
richnew10, line 203: `| undefined`.

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
```diff
@@ -57,4 +58,23 @@ export default interface VideoUplinkBandwidthPolicy {
...
   chooseMediaTrackConstraints(): MediaTrackConstraints;
+
+  /**
+   * Set a reference to the current transceiver controller.
+   * This should be called whenever a transceiver controller is available when the call starts.
+   * The default audio video controllers calls this at the end of connection.
+   * This method should not throw.
+   * @param {TransceiverController} transceiverController - The transceiver controller
+   */
+  setTransceiverController?(transceiverController: TransceiverController): void;
```
richnew10, line 69: Again, how do I know that I'm done with this controller (that is, when `updateTransceiverController` won't ever be called again)?

Why don't you need a `stop` method on this interface, or for this to be called with `undefined` when we're done?

test/transceivercontroller/DefaultTransceiverController.test.ts
```diff
@@ -590,4 +588,189 @@ describe('DefaultTransceiverController', () => {
       expect(sender.track).to.equal(audioTrack);
     });
   });
+
+  describe('setEncodingParameters', () => {
+    it('is no-op if local transceivers are not set up', async () => {
+      tc.setEncodingParameters(
+        new Map<string, RTCRtpEncodingParameters>().set('video', {
+          scaleResolutionDownBy: 1,
+          maxBitrate: 1400000,
```
richnew10, line 597: Use underscores in numbers: `1_400_000`.

test/transceivercontroller/DefaultTransceiverController.test.ts
```diff
@@ -590,4 +588,189 @@ describe('DefaultTransceiverController', () => {
...
+        })
+      );
+      expect(setParamSpy.calledTwice).to.be.true;
+      params = localSender.getParameters();
+      expect(params.encodings.length).to.be.equal(1);
+      expect(params.encodings[0].maxBitrate).to.be.equal(600000);
+      expect(params.encodings[0].scaleResolutionDownBy).to.be.equal(2);
+      expect(params.encodings[0].codecPayloadType).to.be.equal(1);
+
+      setParamSpy.restore();
```
richnew10, line 773: FYI, there's `afterEach` which might help you kill some cleanup boilerplate.

richnew10 (changes requested):

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
ltrung: Updated to use JSON.stringify to compare the 2 params.

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
ltrung: Is it to be explicit? TS does not complain so I did not do it :p

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
ltrung: Maybe I am wrong but both the policy and transceiver should be cleaned up once the call is stopped because it is only referenced in our code in meetingContext object. So they only have to worry about it if they keep a reference to the policy on their side and if that is the case then they just need to remove that reference to the policy at the end right?

test/transceivercontroller/DefaultTransceiverController.test.ts
ltrung: Move it out to `afterEach`

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
richnew10: Yikes, no, don't do that. Not only is it extremely expensive, but the string representations of two objects aren't guaranteed to be the same.

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
richnew10: It'll complain if we ever manage to turn on strict mode. Be explicit.

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
richnew10: It's not just about our code!

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
ltrung: Got it. Updated.

src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
ltrung: This is how it is done in the current default simulcast policy. I also thought it was expensive so only compare the 2 properties that we use here.

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
richnew10: The trick is: how does this object know "the end"?

ltrung: Updated to clean up the reference to transceiver controller at `CleanStoppedSessionTask` task

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
```diff
@@ -62,11 +62,11 @@ export default interface VideoUplinkBandwidthPolicy {
   /**
    * Set a reference to the current transceiver controller.
    * This should be called whenever a transceiver controller is available when the call starts.
-   * The default audio video controllers calls this at the end of connection.
+   * The default audio video controllers calls this when the connection starts and clear it when the connection ends.
```
richnew10, line 65: "… calls this with a `TransceiverController` when the meeting starts, and calls it with `undefined` when the meeting ends. If a meeting is stopped and started repeatedly, this pair of calls will be repeated to match. All calls to `updateTransceiverController` will occur between this pair of calls."

test/audiovideocontroller/DefaultAudioVideoController.test.ts
```diff
@@ -756,6 +757,52 @@ describe('DefaultAudioVideoController', () => {
...
+        webSocketAdapter,
+        new NoOpMediaStreamBroker(),
+        reconnectController
+      );
+      await start();
+      await delay(defaultDelay);
+      audioVideoController.handleHasBandwidthPriority(true);
+    });
+
+    it('can be started and take a bandwidth update with update transceiver controller method', async () => {
```
richnew10, line 789: Where is your test that if you create a meeting and call start, then stop, then start, then stop, `setTransceiverController` is called four times with appropriate parameters?

test/audiovideocontroller/DefaultAudioVideoController.test.ts
```diff
@@ -756,6 +757,101 @@ describe('DefaultAudioVideoController', () => {
...
+      expect(spy.callCount).to.be.equal(1);
+      expect(spy.getCall(0).args[0]).not.be.undefined;
+      await stop();
+      await delay(defaultDelay);
+      expect(spy.callCount).to.be.equal(2);
+      expect(spy.getCall(1).args[0]).to.be.undefined;
+      await start();
+      await delay(defaultDelay);
+      expect(spy.callCount).to.be.equal(3);
+      expect(spy.getCall(2).args[0]).not.be.undefined;
```
richnew10, line 847: If it's the same transceiver controller, can we make sure these two are called with the `===` value?

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
```diff
@@ -57,4 +58,25 @@ export default interface VideoUplinkBandwidthPolicy {
    * Returns the selected [[MediaTrackConstraints]] to update
    */
   chooseMediaTrackConstraints(): MediaTrackConstraints;
+
+  /**
+   * Set a reference to the current transceiver controller.
+   * This should be called whenever a transceiver controller is available when the call starts.
+   * Calls this with a TransceiverController when the meeting starts, and calls it with undefined when the meeting ends.
```
richnew10, line 65: Calls => call, in both places. But note that you're writing for two audiences: people implementing a controller that will call this ("… call this when the meeting starts") and people implementing a policy ("this will be called by the audio video controller when the meeting starts"). Right now your wording falls between those two — you say "should be called" as if you're writing for the former, and then "will be repeated" as if you're writing for the latter.

richnew10 (commented):
Almost done!

test/audiovideocontroller/DefaultAudioVideoController.test.ts
ltrung: We actually initialize a new transceiver controller every time in `actionConnect` so they are not the same transceiver controller.

src/videouplinkbandwidthpolicy/VideoUplinkBandwidthPolicy.ts
ltrung: Updated

richnew10 (approved):

PR #1506, opened 2021-08-12
Expose the mute state of underlying audio input streams via DeviceChangeObserver

richnew10:
Windows allows the user to hard-mute a device.

This is exposed as `MediaStreamTrack.muted`. There are other theoretical situations in which this might be true, and they all mean "the browser thinks this device won't send you audio".

This commit ensures that an observer notification is sent whenever the
current device's hardware/OS mute state is changed, or when the device
is initially selected.

See the detailed interface documentation for more information.

Additionally, this commit fixes a bunch of test stuff to correctly
handle streams — streams must have an ID else `DefaultDeviceController`
skips them.

You can test this commit in the demo app: you'll see log lines like 

```
[DEMO] Mute state: device +B1YXuZYTA5SWHUNSD0fryKk0av3HJCWrserEYfntpI= is not muted
```

**Testing**

Yes.

Only on macOS, so only the negative case. There are unit tests.

Logging is added to the meeting demo.

Yes. A new observer method is added to `DeviceChangeObserver`. This is the review.

No.

---


ltrung:
Build failed. Seem like we have console.log in source code.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1267,6 +1267,10 @@ export class DemoMeetingApp
     }
   }
 
+  audioInputMuteStateChanged(device: string | MediaStream, muted: boolean): void {
+    this.log('Mute state: device', device, muted ? 'is muted' : 'is not muted');
```
ltrung, line 1271: Should we also update the mic button to show muted?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1440,6 +1490,23 @@ export default class DefaultDeviceController
...
+        if (track) {
+          const id = track.getSettings().deviceId || newDevice.stream;
+
+          newDevice.trackMuteCallback = (): void => {
+            this.mediaStreamMuteObserver(id, true);
+          };
+          newDevice.trackUnmuteCallback = (): void => {
+            this.mediaStreamMuteObserver(id, false);
+          };
+          track.addEventListener('mute', newDevice.trackMuteCallback, { once: false });
```
ltrung, line 1506: Isn't the default is false for `once`? Do you just want to be explicit here?

README.md
```diff
@@ -301,12 +304,18 @@ const observer = {
...
     console.log('Audio outputs updated: ', freshAudioOutputDeviceList);
   },
+
   videoInputsChanged: freshVideoInputDeviceList => {
     console.log('Video inputs updated: ', freshVideoInputDeviceList);
-  }
+  },
+
+  audioInputMuteStateChanged: (device, muted) => {
+    console.log('Device', device, muted ? 'is muted in hardware' : 'is not muted);
```
ltrung, line 317: Just a question: in a realistic application, do builders want to compare this with their app mute state and show a notification if they don't match?

src/devicecontroller/DefaultDeviceController.ts
richnew10: Yes, just being explicit.

demos/browser/app/meetingV2/meetingV2.ts
richnew10: A good follow-up is to make the mic button show muted _and be un-mutable_, and show an explanatory tooltip. I didn't have time to address those: I did the bare minimum demo work to allow us to tell whether the feature works.

README.md
richnew10: If it's a realistic concern for their application, they will want to track this state and typically enter some error state — e.g., showing the mic selector greyed out with a "!" and some warning or similar. It's not a mute that we can un-mute, so it's closer to an error state that gets fixed outside of the app.

richnew10: Note particularly that `muted` is _not_ set by the app — that's `enabled`. `muted` means "this stream is broken and we think it's because the OS muted it".

richnew10:
I have to figure out why I can't get that release call to be hit.

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -2158,6 +2158,88 @@ describe('DefaultDeviceController', () => {
...
+      const track = new MediaStreamTrack() as StoppableMediaStreamTrack;
+      track.setStreamDeviceID('foobar');
+      track.externalMute();
+      fakeStream.addTrack(track);
+      expect(!!fakeStream.id).to.be.true;
+
+      const output = new Promise((resolve, _reject) => {
+        const observer: DeviceChangeObserver = {
+          audioInputMuteStateChanged: (device, muted): void => {
+            console.info('xxx mute state changed', device, muted);
```
devalevenkatesh, line 2202: Should we remove this console log?

test/utils/Utils.test.ts
```diff
@@ -13,6 +13,8 @@ describe('Utils', () => {
     const delay = 500;
     await wait(delay);
     const endTime = new Date().getTime();
-    expect(endTime - startTime).to.gte(delay);
+
+    // Timing is imprecise. Allow 2 milliseconds of fudge. I have seen this run in 499.
+    expect(endTime - startTime).to.gte(delay - 2);
```
devalevenkatesh, line 18: Thanks for fixing this 🙏 .

test/dommock/DOMMockBuilder.ts
```diff
@@ -233,6 +247,21 @@ export default class DOMMockBuilder {
...
+      }
+
+      externalMute(): void {
+        this.muted = true;
+        this.dispatchEvent({ ...Substitute.for(), type: 'mute' });
+      }
+
+      setStreamDeviceID(id: string | undefined): void {
+        this.streamDeviceID = id;
+        console.info('Setting stream ID to', id, this.streamDeviceID);
```
devalevenkatesh, line 262: Should we remove this console log as well?

richnew10:
Cleaned up and ready for re-review.

richnew10:
If all the tests pass and coverage is good, please go ahead and merge.

ltrung (approved):

devalevenkatesh:
Build script has failed , the coverage for branches is failing.
```
ERROR: Coverage for branches (99.97%) does not meet global threshold (100%)

Statements   : 100% ( 8807/8807 )
Branches     : 99.97% ( 3889/3890 )
Functions    : 100% ( 1691/1691 )
Lines        : 100% ( 8693/8693 )
```

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -811,24 +837,34 @@ export default class DefaultDeviceController
...
 
-      if (
-        kind === 'video' &&
-        this.boundAudioVideoController?.videoTileController.hasStartedLocalVideoTile()
-      ) {
-        this.boundAudioVideoController.videoTileController.stopLocalVideoTile();
+      switch (kind) {
+        case 'audio': {
+          for (const track of mediaStreamToRelease.getAudioTracks()) {
+            track.removeEventListener('mute', activeDevice.trackMuteCallback);
```
devalevenkatesh, line 851: I see the `ended` event callback `removeEventListener` removed from above so checking if it should added here?
This shows removed.
```
tracksToStop[0].removeEventListener('ended', activeDevice.endedCallback);
```

richnew10: I think I found this to be unnecessary — this function is part of a stack that also does some cleanup work — but enough hours have passed that I'm going to have to re-check.

devalevenkatesh (approved):

PR #1507, opened 2021-08-13
Fix race condition in Safari when disconnect and connect stream from video element

ltrung:
**Issue:**
In Safari, calling `disconnectVideoStreamFromVideoElement` then call `connectVideoStreamToVideoElement` can cause the new video stream to not show up. This is because of a workaround we put in for Safari to wait for the next tick before resetting the `srcObject` property of  the videoElement to null to avoid crashing for Safari 12 and earlier. However, if another video stream is set to `srcObject` before, then this will clear out the new value.
 
**Description of changes:**
- Update the code to only do the workaround for Safari 12 as the crashing can no longer be reproduced in Safari 13 or later.
- Update the workaround to only set the `srcObject` to null if it is the same stream we want to remove.

**Testing**

    - Joined a meeting in Safari
    - In the preview window, switched the video quality and verified that the video preview showed up.

Yes, the meeting demo.

### DIFF
--- a/src/browserbehavior/DefaultBrowserBehavior.ts
+++ b/src/browserbehavior/DefaultBrowserBehavior.ts
@@ -150,7 +150,7 @@ export default class DefaultBrowserBehavior implements BrowserBehavior, Extended
   }
 
   requiresVideoElementWorkaround(): boolean {
-    return this.isSafari();
+    return this.isSafari() && this.majorVersion() === 12;
   }
 
   requiresNoExactMediaStreamConstraints(): boolean {
--- a/src/videotile/DefaultVideoTile.ts
+++ b/src/videotile/DefaultVideoTile.ts
@@ -98,8 +98,11 @@ export default class DefaultVideoTile implements DevicePixelRatioObserver, Video
       // Need to yield the message loop before clearing `srcObject` to
       // prevent Safari from crashing.
       if (new DefaultBrowserBehavior().requiresVideoElementWorkaround()) {
+        const prevSrcObject = videoElement.srcObject;
         AsyncScheduler.nextTick(() => {
-          videoElement.srcObject = null;
+          if (videoElement.srcObject === prevSrcObject) {
+            videoElement.srcObject = null;
+          }
         });
       } else {
         videoElement.srcObject = null;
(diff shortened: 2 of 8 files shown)


src/videotile/DefaultVideoTile.ts
```diff
@@ -97,9 +97,12 @@ export default class DefaultVideoTile implements DevicePixelRatioObserver, Video
 
       // Need to yield the message loop before clearing `srcObject` to
       // prevent Safari from crashing.
+      const prevSrcObject = videoElement.srcObject;
       if (new DefaultBrowserBehavior().requiresVideoElementWorkaround()) {
         AsyncScheduler.nextTick(() => {
-          videoElement.srcObject = null;
+          if (videoElement.srcObject === prevSrcObject) {
```
michhyun1, line 103: When would this not be true? On like 100, arent you setting perSrcObject to videoElement.srcObject?

ltrung: So the issue is when this code is executed, the `srcObject` in the video element may already change. So I store a reference to the previous `srcObject` to make sure we do not clear out a new one.

src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -150,7 +150,7 @@ export default class DefaultBrowserBehavior implements BrowserBehavior, Extended
   }
 
   requiresVideoElementWorkaround(): boolean {
-    return this.isSafari();
+    return this.isSafari() && this.majorVersion() <= 12;
```
devalevenkatesh, line 153: Checking since we do not officially support Safari version less than 12, should this be `== 12`?

src/videotile/DefaultVideoTile.ts
devalevenkatesh: Should we move that declaration of `prevSrcObject` inside the workaround condition since this applies only to the workaround or is that used somewhere else as well?

src/browserbehavior/DefaultBrowserBehavior.ts
ltrung: Originally i have it === 12 but change it to this to be more general.

devalevenkatesh (approved):

PR #1508, opened 2021-08-13
Add reconnect logic documentation in FAQs

ldai1:
**Issue #:**

**Description of changes:**
Add reconnect logic documentation in FAQs

**Testing**

2. How did you test these changes? Check generated HTML


guides/07_FAQs.md
```diff
@@ -145,6 +145,17 @@ The possible reasons are as follows:
...
 
+
+### What is the client behavior when WebSocket connection is terminated during connection?
+
+If the WebSocket connection is terminated when clients are in the [JoinAndReceiveIndexTask](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/task/JoinAndReceiveIndexTask.ts) or [SubscribeAndReceiveSubscribeAckTask](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/task/SubscribeAndReceiveSubscribeAckTask.ts), they will get `TaskFailed`, `InternalServerError` or `SignalingBadRequest` meeting session status code depending on failure reasons. If the status code is `TaskFailed` or `InternalServerError`, clients will reconnect immediately without waiting for connection timeout. If the status code is `SignalingBadRequest`, clients will not reconnect. There is a customizable [reconnectTimeout](https://github.com/aws/amazon-chime-sdk-js/blob/0147ed9fb76429a70fb0161405687af63b592fe0/src/meetingsession/MeetingSessionConfiguration.ts#L67) for builder to decide how long they want to timeout the reconnection. The default value is 2 minutes.
```
devalevenkatesh, line 151: Builders may not know these tasks and they wont know when they run so would it be better to explain when re-connection triggers and based on that we run these tasks which handle the WebSocket connection events. Something like:
The Amazon Chime SDK for JavaScript uses WebSocket to establish and maintain connection to Amazon Chime backend. In this process we run few tasks, some of them handle WebSocket connection failures and are explained below:
- Explain `JoinAndReceiveIndexTask`
- Explain `SubscribeAndReceiveSubscribeAckTask`

These tasks listen to WebSocket connection failures. We map these failures to meeting session status codes (You can link the status code file here). The SDK reconnects immediately if we receive `TaskFailed` or `SignalingInternalServerError` failures. If we receive `SignalingBadRequest`, the SDK will not reconnect. You can customize the reconnect behavior by changing [parameter names] in the `MeetingSessionConfiguration` required while creating the meeting session. (Since these parameters have documentation I think it is fine to drop calling defaults here since if they change in API documentation we would have to change here as well).

guides/07_FAQs.md
```diff
@@ -145,6 +145,17 @@ The possible reasons are as follows:
...
+
+### What happens if the client is in mid call and their WebSocket connection is terminated?
```
devalevenkatesh, line 153: Request you to check above comment for "client". Should we club the explanation of both of these questions together as they have related foundation? The MonitorTask can be explained below other tasks and then explain as you do in this answer.

guides/07_FAQs.md
devalevenkatesh: Should we change `InternalServerError` -> `SignalingInternalServerError`?

simmkyu: - I agree with Venky's comment that these *task*s are implementation details. We do not mention a task in other FAQs and other documentation.
- You mention *WebSocket* in the question, but *WebSocket* does not trigger these two events: `noPacketsReceivedRecently` and `hasBadAudioDelay`. These are WebRTC-based metrics. Also, I think *WebSocket* is an implementation detail, e.g., public APIs in our README do not include the term *WebSocket*. Recommend rephrasing Q/A without using WebSocket.

How about the following Q/A?
- No implementation detail
- An example to override the health configuration
- Three explicit cases + one more case for other retryable errors, (AWS document uses the term "retryable error.")

---

### When does the Amazon Chime SDK retry the connection? Can I customize this retry behavior?

The Amazon Chime SDK for JavaScript retries the connection in the following situations.
- The SDK misses consecutive pong frames from the Chime server.
- The SDK detects a high rate of audio packet loss.
- The SDK experiences a significant audio delay.
- The SDK encounters a retryable error during the session.

The SDK uses [ConnectionHealthPolicyConfiguration](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/connectionhealthpolicy/ConnectionHealthPolicyConfiguration.ts) to trigger a reconnection. We recommend using the default configuration, but you can also provide the custom [ConnectionHealthPolicyConfiguration](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/connectionhealthpolicy/ConnectionHealthPolicyConfiguration.ts) object to change this behavior.

```js
import {
  ConnectionHealthPolicyConfiguration,
  ConsoleLogger,
  DefaultDeviceController,
  DefaultMeetingSession,
  LogLevel,
  MeetingSessionConfiguration
} from 'amazon-chime-sdk-js';

const logger = new ConsoleLogger('MyLogger', LogLevel.INFO);
const deviceController = new DefaultDeviceController(logger);

const healthPolicyConfiguration = new ConnectionHealthPolicyConfiguration();
healthPolicyConfiguration.connectionUnhealthyThreshold = 50; // Default: 25 consecutive WebRTC stats with no packet
healthPolicyConfiguration.connectionWaitTimeMs = 20000; // Default: 10000 ms
healthPolicyConfiguration.maximumAudioDelayMs = 30000; // Default: 60000 ms delay
healthPolicyConfiguration.maximumAudioDelayDataPoints = 5; // Default: 10 data points
healthPolicyConfiguration.missedPongsUpperThreshold = 8; // Default: 4 missed pongs

const configuration = new MeetingSessionConfiguration(meetingResponse, attendeeResponse);

// Override the default health policy configuration.
configuration.connectionHealthPolicyConfiguration = connectionHealthPolicyConfiguration

const meetingSession = new DefaultMeetingSession(
  configuration,
  logger,
  deviceController
);
```

guides/07_FAQs.md
```diff
@@ -145,6 +145,17 @@ The possible reasons are as follows:
 2. The close code `4410` from the Chime backend indicates that an attendee has attempted to join an already-ended meeting. The Chime SDK for JavaScript throws an error with the status code `MeetingEnded`.
 3. The close code between `4500` and `4599` (inclusive) indicates an internal server error in the Amazon Chime backend. In this case, the Chime SDK for JavaScript throws an error with the status code `SignalingInternalServerError`. Please create a GitHub issue including the Chime SDK browser logs.
 
+
+### What is the client behavior when WebSocket connection is terminated during connection?
```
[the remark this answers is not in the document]
simmkyu: I recommend removing this Q/A for the same reasons in my comment below.

guides/07_FAQs.md
ldai1: Thanks Kyu and Venky for providing the great feedback with detailed examples. I have updated the doc based on your comments.

guides/07_FAQs.md
```diff
@@ -145,6 +145,24 @@ The possible reasons are as follows:
...
+The Amazon Chime SDK for JavaScript retries the connection in the following situations.
+
+* The SDK misses consecutive pong messages from the Chime server.
+* The SDK detects a high rate of audio packet loss.
+* The SDK experiences a significant audio delay.
+* The SDK encounters a retryable error during the session.
+
+The SDK uses [ConnectionHealthPolicyConfiguration](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/connectionhealthpolicy/ConnectionHealthPolicyConfiguration.ts) to trigger a reconnection. We recommend using the default configuration, but you can also provide the custom ConnectionHealthPolicyConfiguration object to change this behavior. 
+
+* Configure [missed pong threshold](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/connectionhealthpolicy/ConnectionHealthPolicyConfiguration.ts#L24): The current missed pong threshold is set to 4 in JS SDK. For example, you can set the missed pongs upper threshold to zero to force restart the session.
```
simmkyu, line 158: Instead of these points, just recommend adding the code example from the previous comment. Customers will have challenges updating these values because we don't have any example of overriding the health policy configuration.

guides/07_FAQs.md
```diff
@@ -145,6 +145,24 @@ The possible reasons are as follows:
...
+* Retryable errors are errors with [retryable response status codes](https://aws.github.io/amazon-chime-sdk-js/classes/meetingsessionstatus.html#isfailure) received regarding the session. i.e. status code with `TaskFailed` or `SignalingInternalServerError` will handled for retries.
+
+The maximum amount of time to allow for connecting is 15 seconds, which can be configurable in [MeetingSessionConfiguration](https://github.com/aws/amazon-chime-sdk-js/blob/0147ed9fb76429a70fb0161405687af63b592fe0/src/meetingsession/MeetingSessionConfiguration.ts#L40). The [reconnectTimeout](https://github.com/aws/amazon-chime-sdk-js/blob/0147ed9fb76429a70fb0161405687af63b592fe0/src/meetingsession/MeetingSessionConfiguration.ts#L67) is configurable for how long you want to timeout the reconnection. The default value is 2 minutes.
```
simmkyu, line 163: Remove this sentence? For this question, I think we can just describe the health policy configuration. (15 seconds and 2 minutes are a bit confusing.)

ldai1: yes, that make sense. It will be probably confusing for customers to understand the relation between this timeout configuration and health policy configuration. I will remove.

guides/07_FAQs.md
ldai1: Yeah, make sense. I will add the code example. Can I also keep these description text along with the code examples?

guides/07_FAQs.md
```diff
@@ -145,6 +145,64 @@ The possible reasons are as follows:
...
+The Amazon Chime SDK for JavaScript retries the connection in the following situations.
+
+* The SDK misses consecutive pong messages from the Chime server.
+* The SDK detects a high rate of audio packet loss.
+* The SDK experiences a significant audio delay.
+* The SDK encounters a retryable error during the session. Retryable errors are errors with [retryable response status codes](https://aws.github.io/amazon-chime-sdk-js/classes/meetingsessionstatus.html#isfailure) received regarding the session. i.e. status code with `TaskFailed` or `SignalingInternalServerError` will handled for retries.
+
+The SDK uses [ConnectionHealthPolicyConfiguration](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/connectionhealthpolicy/ConnectionHealthPolicyConfiguration.ts) to trigger a reconnection. We recommend using the default configuration, but you can also provide the custom ConnectionHealthPolicyConfiguration object to change this behavior. 
+
+```
```
simmkyu, line 159: Append js

```js
```js
```

simmkyu (approved):

simmkyu (approved):

PR #1516, opened 2021-08-17
Add a resource monitor to output a log message when setTimeout drifts

simmkyu:
**Issue #:**
N/A

**Description of changes:**
- Add an internal resource monitor that approximates a CPU overload.

**Testing**

2. How did you test these changes? Ran the demo app in CPU overload and ensured that the demo output a message.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes, the meeting demo.
4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? No, only an internal class and interface.

### DIFF
--- a/src/resourcemonitor/DefaultResourceMonitor.ts
+++ b/src/resourcemonitor/DefaultResourceMonitor.ts
@@ -1,0 +1,59 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import Logger from '../logger/Logger';
+import ResourceMonitor from '../resourcemonitor/ResourceMonitor';
+import AsyncScheduler from '../scheduler/AsyncScheduler';
+import IntervalScheduler from '../scheduler/IntervalScheduler';
+
+/**
+ * [[DefaultResourceMonitor]] uses an execution time of the setTimeout with a zero delay
+ * to estimate resource consumption. If the execution time exceeds the given threshold,
+ * this monitor outputs a log message.
+ * @internal
+ */
+export default class DefaultResourceMonitor implements ResourceMonitor {
+  private readonly intervalScheduler: IntervalScheduler;
+
+  private count: number = 0;
+  private timestampMs: number = 0;
+
+  constructor(
+    private readonly logger: Logger,
+    private readonly dataPoints: number = 3,
+    private readonly intervalMs: number = 5000,
+    private readonly thresholdMs: number = 100,
+    private readonly waitTimeMs: number = 30000
+  ) {
+    this.intervalScheduler = new IntervalScheduler(this.intervalMs);
+  }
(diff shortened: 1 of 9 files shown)


src/task/MonitorTask.ts
```diff
@@ -60,6 +63,7 @@ export default class MonitorTask
       { ...connectionHealthPolicyConfiguration },
       this.initialConnectionHealthData.clone()
     );
+    this.resourceMonitor = new DefaultResourceMonitor(context.logger);
   }
 
   removeObserver(): void {
```
ltrung, line 69: Just curious but is this called in `CleanStoppedSessionTask`?

src/resourcemonitor/DefaultResourceMonitor.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  start(): void {
+    this.intervalScheduler.start(() => {
+      if (document.visibilityState !== 'visible') {
+        return;
+      }
+
+      const start = performance.now();
+      new AsyncScheduler().start(() => {
```
ltrung, line 33: Can we put some comment about how the logic work here?  It would be nice to have a summary instead of going through code :D

src/resourcemonitor/DefaultResourceMonitor.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/** @internal */
+export default class DefaultResourceMonitor implements ResourceMonitor {
+  private readonly intervalScheduler: IntervalScheduler;
+
+  private count: number = 0;
+  private timestampMs: number = 0;
+
+  constructor(
+    private readonly logger: Logger,
```
ltrung, line 17: If builders want to override this value, how would they do it? Doesn't seem there is a way for them to pass in these values.

src/task/MonitorTask.ts
simmkyu: Yes, the `CleanStoppedSessionTask` calls the `removeObserver` of each task to clean up timers and resources.
https://github.com/aws/amazon-chime-sdk-js/blob/5071197d50eda528c8e99df4eb0fd6a7dc874965/src/task/CleanStoppedSessionTask.ts#L36-L38

src/resourcemonitor/DefaultResourceMonitor.ts
simmkyu: No, builders can't change this configuration in the internal component. After some tests, we can have a follow-up task:
- Expose a public audio-video observer for inefficient resources.
- We can also discuss whether to make these configurable.

src/task/MonitorTask.ts
simmkyu: I need to fix some tests that do not clean up the session. Even after the test completes, this resource interval keeps running to cause a missing `document` error.

PR #1517, opened 2021-08-17
Added Android WebView support

LeviSklaroff:
**Issue #:**
Implement Android Webview Support
**Description of changes:**
- Added Android WebView support
- Added a fix in the browser demo in cases where the browser does not support grabbing device labels. The key issue is that the browser demo will attempt to go into flow-need-permissions and flow-devices whenever enumerateDevices is called and returns empty. In Android WebView enumerateDevices does not work. More info can be found here: https://bugs.chromium.org/p/chromium/issues/detail?id=669492

**Testing**
Yes
- Build the sdk
- Build the browser demo
- Deploy the serverless demo and copy the given url for later
- Clone the AndroidWebview branch from https://github.com/aws-samples/amazon-chime-sdk
- Launch the newly added Android-Webview-sample folder in Android Studio 4.2.2
- In the AppConfig.kt file replace PLACEHOLDER with the url from the serverless demo
- Connect a physical android device that is in developer mode
- Launch the app on the physical device
- On the android device press the 'Go to WebView' button
- Fill in the fields on the page that appears
- Click continue
- Click continue on the next page
- Open the serverless url on another device and browser, then perform the last 3 steps
- Test for audio input on the android device
- Test for audio output on the android device
- Test for video input on the android device
Yes, can be tested using the serverless demos
Added a new browserbehavior function called doesNotSupportDevicesLabels,  this has been reviewed and approved.
No


CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ of videos.
   default. If the browser sends an incorrect packet rate, this will be logged
   as an error in the console.
 - Add a warning log in `InMemoryJSONEventBuffer`'s `send` function when retrying starts.
+- Added support for Chromium WebView
```
michhyun1, line 35: I think we can say Android WebView here, even though the user agent is chromium webview

src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -201,6 +203,10 @@ export default class DefaultBrowserBehavior implements BrowserBehavior, Extended
     return this.hasChromiumWebRTC() || this.isSafari();
   }
 
+  doesNotSupportMediaDeviceLabels(): boolean {
```
michhyun1, line 206: Can you leave a comment above this with the link to the chromium bug

michhyun1:
when you make changes - you can do this:

git add .
git commit --amend (--no-edit)
git push origin branch_name -f

this will make it so that you only have one commit in your PR, not 6.

ltrung:
> Added a fix in the browser demo in cases where the browser does not support grabbing device labels.
Can we put more details here? What does it mean not supporting device labels? Is there a Chromium bug for it?
Also if getUserMedia does not work where is the audio from?

ltrung:
Seem like you need to run npm run doc and commit the changes in FAQs.

LeviSklaroff:
> when you make changes - you can do this:
> 
> git add .
> git commit --amend (--no-edit)
> git push origin branch_name -f
> 
> this will make it so that you only have one commit in your PR, not 6.

Will do in the future!

CHANGELOG.md
LeviSklaroff: Changed!

src/browserbehavior/DefaultBrowserBehavior.ts
LeviSklaroff: Added!

ltrung:
> > when you make changes - you can do this:
> > git add .
> > git commit --amend (--no-edit)
> > git push origin branch_name -f
> > this will make it so that you only have one commit in your PR, not 6.
> 
> Will do in the future!

Actually you should try to do rebase and squash to collapse them into one commit.

LeviSklaroff:
> > Added a fix in the browser demo in cases where the browser does not support grabbing device labels.
> 
> Can we put more details here? What does it mean not supporting device labels? Is there a Chromium bug for it (I see one in FAQs but we can link it here too)?
> Also if getUserMedia does not work where is the audio from?

Added to the pull request! It's also not getUserMedia that fails, but rather enumerateDevices that fails and returns empty labels. This causes devices to have no name such (MacBook camera) however they still function with no issues, and the browser demo by default fills in labels with defaults.

LeviSklaroff:
> Seem like you forget to commit your changes in FAQs

Yeah, I had a misunderstanding on how the FAQs worked, and was editing the html file instead of the .md file and has now been fixed.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2162,17 +2162,22 @@ export class DemoMeetingApp
     // Also note that Firefox has its own device picker, which may be useful
     // for the first device selection. Subsequent device selections could use
     // a custom UX with a specific device id.
-    this.audioVideo.setDeviceLabelTrigger(
-      async (): Promise<MediaStream> => {
-        if (this.isRecorder() || this.isBroadcaster()) {
-          throw new Error('Recorder or Broadcaster does not need device labels');
+    if(!this.defaultBrowserBehaviour.doesNotSupportMediaDeviceLabels())
+    {
+      this.audioVideo.setDeviceLabelTrigger(
```
ltrung, line 2167: Hmm I am not sure I understand the logic here. Basically you do not call setDeviceLabelTrigger for AndroidWebview. However, I think in that case, we just use the default label trigger which is the same here (See https://github.com/aws/amazon-chime-sdk-js/blob/master/src/devicecontroller/DefaultDeviceController.ts#L139). Is that correct?

LeviSklaroff: That's correct, the key fix here is that it no longer sends the user back to flow-need-permission and flow-devices when the device labels are found to be empty after calling enumerate devices.

ltrung: Great detail. Can we update the PR description about the problem details then?

LeviSklaroff: Added into the description!

LeviSklaroff:
> > > when you make changes - you can do this:
> > > git add .
> > > git commit --amend (--no-edit)
> > > git push origin branch_name -f
> > > this will make it so that you only have one commit in your PR, not 6.
> > 
> > 
> > Will do in the future!
> 
> Actually you should try to do rebase and squash to collapse them into one commit.

Done!

michhyun1 (approved):

ltrung (approved):

PR #1520, opened 2021-08-18
Fix priority-based downlink policy to not unpaused tiles that are not paused by the policy

ltrung:
**Issue #:**
Currently with priority-based downlink policy enabled, clients cannot manually pause a remote video stream because the policy will unpause it (if it detects there is enough bandwidth to do so).

**Description of changes:**
We already maintain a set in the policy to keep track of which attendee's videos are paused by the policy so I just add a check to make sure the attendee is in the set before unpausing a video.

Also add a check in videoTileDidUpdate in the demo to prevent crash. Priority-based downlink policy can create an empty tile with no external user Id ([Code](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts#L768)) which will crash the demo.

**Testing**

    - Joined a meeting with priority based enabled
    - Tried to pause a video manually
    - Verified that the video is not unpaused.
    - Throttled the network
    - Verified remote videos are paused by the policy (with lightning bolt icon).
    - Unthrottled the network
    - Verified remote videos are unpaused.
    
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes, the meeting demo.


lzongkun (approved):
Thanks for fixing it!

lzongkun (approved):

PR #1528, opened 2021-08-21
Fix empty video tile with priority-based downlink policy

ltrung:
**Issue #:**
When there is a new video but there is not enough bandwidth, priority-based downlink policy will create an empty video tile because we do not want to subscribe to the new video stream then have to pause it anyway. When there is enough bandwidth, we will unpause the paused video tile and trigger resubscribe.
However, when we receive the video stream in peer connection, we have a logic to skip creating new tile and binding video stream if there is already a tile for that attendee. This does not check whether the existing tile has a bound video stream or not.  

**Description of changes:**
- Adding a logic in `CreatePeerConnectionTask` to bind video stream fixes the issue (with some refactoring)
**Testing**


lzongkun (approved):

CHANGELOG.md
```diff
@@ -8,14 +8,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-
 - Add `audioInputMuteStateChanged` to the `DeviceChangeObserver` interface. This is called whenever the device is changed or is muted or unmuted, allowing applications to adapt to OS-level mute state for input devices.
```
simmkyu, line 11: How about adding this info too?

- Add a new optional API `getVideoTileForAttendeeId` in `VideoTileController` and raise the `tileWillBePausedByDownlinkPolicy` event for empty video tiles.

simmkyu (approved):

devalevenkatesh (approved):

PR #1531, opened 2021-08-23
Fix default priority downlink policy to update default preference correctly

ltrung:
**Issue #1470:**
See https://github.com/aws/amazon-chime-sdk-js/pull/1470#issuecomment-903025231.
PR #1470 updated the priority-based default behavior but caused the default video preferences to no longer be updated after the first time ([Code](https://github.com/aws/amazon-chime-sdk-js/pull/1470/files?file-filters%5B%5D=.md&file-filters%5B%5D=.ts#diff-285a01f69c8c858f6996b49b347d75a82617659e6c8132be10ddd37f17d42e9eR161)). 
This will also affects `VideoAdaptiveProbePolicy` which extends `VideoPriorityBasedPolicy`. 

Another issue is we do not have a way to test out either `VideoAdaptiveProbePolicy` or the default behavior of `VideoPriorityBasedPolicy` so it is hard to catch these issues.

**Description of changes:**
- Make sure to not assign the default preference to the instance variable `videoPreferences`.
- Update the meeting demo:
   - Use  `VideoAdaptiveProbePolicy` when simulcast is enabled but not priority-based downlink policy.
   - Use the default priority-based policy behavior if pin functionality is disabled.
   
**Testing**

    - Case 1:
       - Joined a meeting with simulcast enabled.
       - Verified that you could see video from other attendees. 
    - Case 2:
       - Joined a meeting with priority-based downlink policy enabled but do not enable pin
       - Verified that you could see video from other attendees. 
    - Case 3:
       - Joined a meeting with priority-based downlink policy enabled but enable pin.
       - Verified that the pin functionality works.
    - Case 4: 
       - Joined meeting and verified that Pin button is only enabled if priority-based downlink policy and pin are enabled. 

### DIFF
--- a/src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
+++ b/src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
@@ -158,7 +158,6 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
     this.videoIndex = videoIndex;
     if (!this.videoPreferences) {
       this.updateDefaultVideoPreferences();
-      this.videoPreferences = this.defaultVideoPreferences;
     }
   }
 
@@ -275,7 +274,7 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
   protected calculateOptimalReceiveStreams(): void {
     const chosenStreams: VideoStreamDescription[] = [];
     const remoteInfos: VideoStreamDescription[] = this.videoIndex.remoteStreamDescriptions();
-    if (remoteInfos.length === 0 || !this.videoPreferences || this.videoPreferences.isEmpty()) {
+    if (remoteInfos.length === 0 || this.videoPreferences?.isEmpty()) {
       this.optimalReceiveStreams = [];
       return;
     }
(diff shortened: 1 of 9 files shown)


lzongkun (approved):
LGTM

lzongkun (approved):

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -275,7 +274,7 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
   protected calculateOptimalReceiveStreams(): void {
     const chosenStreams: VideoStreamDescription[] = [];
     const remoteInfos: VideoStreamDescription[] = this.videoIndex.remoteStreamDescriptions();
-    if (remoteInfos.length === 0 || !this.videoPreferences || this.videoPreferences.isEmpty()) {
+    if (remoteInfos.length === 0 || this.videoPreferences?.isEmpty()) {
```
devalevenkatesh, line 277: Checking, what happens when `this.videoPreferences` is null? Earlier it would go and set `this.optimalReceiveStreams` to `[]` but now it wont go in the `if` right?

ltrung: it should never be null and just be undefined. Previously it never go in because we always init it to an empty set.

simmkyu (approved):

PR #1545, opened 2021-08-26
Trace `getRTCPeerConnectionStats()` and `hasStartedLocalVideoTile()` with DEBUG level logs to reduce  INFO logging noise

devalevenkatesh:
**Issue #:**
- Info level logs print a lot of `getRTCPeerConnectionStats` -> `null` and similarly for `hasStartedLocalVideoTile()`.
- This creates a lot of noise in INFO level logging which is not needed.
- [Demo only] Local video `div` never closes even when local video is stopped.

**Description of changes:**
- Trace `getRTCPeerConnectionStats()` and `hasStartedLocalVideoTile()` with DEBUG level logs
- Use `availablelTileSize()` demo function to show stats only when videos
are turned on and available.
- Fix `videoTileDidUpdate()` in demo to correctly close local video tile. Currently, the `div` containing the local video tile never closes, the video closes but the `div` is still visible, this fixes that. This is due to `videoTileDidUpdate` sending us update like 4 times and we never hide the local video tile when the tile state becomes in-active. We call `stopLocalVideoTile` which does not hide the video tile. Only time we hide the tile currently in the demo is in `videoTileWasRemoved` AudioVideoObserver which is only called when one calls `removeLocalVideoTile()`. We do not call `removeLocalVideoTile()` in the demo.
- Optimize the WebRTC stats show in demo when at least 1 video is available.

**Testing**

- Join the demo browser meeting selecting only INFO level logs.
- Checked and see these traces do not appear now with INFO level logs.
- Join a meeting with DEBUG level logging, checked and see these `getRTCPeerConnectionStats()` and `hasStartedLocalVideoTile()` trace logs.
- Joined a meeting, turned a local video, turned it off and the `div` is not visible anymore.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes, the demo browser meeting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Clarify why not use default downlink policy with simulcast.
+- Trace `getRTCPeerConnectionStats()` and `hasStartedLocalVideoTile()` with DEBUG level logs instead of INFO to reduce logging noise.
 
 ### Removed
 
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -1380,6 +1380,9 @@ export class DemoMeetingApp
   }
 
   showVideoWebRTCStats(videoMetricReport: { [id: string]: { [id: string]: {} } }): void {
+    if (this.availablelTileSize() === 0) {
+      return;
+    }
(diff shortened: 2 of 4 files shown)


devalevenkatesh:
Closing for now as there is still changes need to be done on demo side if to use and implement `removeLocalVideoTile`. Need to test and work on video on-off in all cases such as pause / unpause or other places where the video tile is turned off in meetingV2 demo.

Issue #1568, opened 2021-09-03
Spurious error logged prior to devices being listed


richnew10:
Stack:

```
getGroupIdFromDeviceId (DefaultDeviceController.js?66c1:1016)
chooseInputIntrinsicDevice (DefaultDeviceController.js?66c1:1131)
acquireInputStream (DefaultDeviceController.js?66c1:1376)
acquireAudioInputStream (DefaultDeviceController.js?66c1:585)
run (ReceiveAudioInputTask.js?ae4c:26)
```

Choosing a device requires computing the group ID to populate the `DeviceSelection`. For some reason we do so by consulting the device list cache. If the cache doesn't exist because no devices have been listed, any attempt to select a device — even a stream or the null device — will cause this to be hit:

```typescript
    } else {
      this.logger.error(
        'Device cache is not populated. Please make sure to call list devices first'
      );
```

This was introduced in #1329.

Instead we should:

* Not use a group ID for the null device.
* Use the group ID from the settings of a media stream.
* Consider enumerating devices as needed in order to do the work.
* Probably not log an error, because applications seem to continue to work without this.

I have started to look at this, but will hand it off if I can't get it wrapped up.

ltrung:
@richnew10 I think we assume you always need to call `list<Audio/Video>InputDevice` first before calling the corresponding choose API and the current code would not work if device cache is not populated due to this check https://github.com/aws/amazon-chime-sdk-js/blob/master/src/devicecontroller/DefaultDeviceController.ts#L1388. This whole complication with device cache comes from that logic. In some browsers (I think Chrome), deviceId can be the same (`default`) so we need to rely on groupId to know if they are the same device.  However, since the input info does not contain groupId, we have to rely on the device cache. 
Also if this check fails, it will cause infinite cycle with `chooseAudioInputDevice` since that API will always call `chooseInputIntrinsicDevice` twice (https://github.com/aws/amazon-chime-sdk-js/blame/66cd9809ea156a66dbd988c866fef9b489ca7c20/src/devicecontroller/DefaultDeviceController.ts#L1747). 
Therefore, at least in the case with audio input, if you see that error message, you will probably see infinite cycle also and application would not work.

[...]

PR #1572, opened 2021-09-03
Rework how group IDs are computed.

richnew10:
**Issue #:**

#1568 

**Description of changes:**

This change now extracts the group ID from the stream, rather than using the device ID to navigate into the device info cache.

This will avoid an error being logged when choosing a null device, and in some other cases, when devices have not yet been enumerated.

Comments in the code should be explanatory.

**Testing**

In parts.

Unit testing. I trust the integration tests to catch regressions.

No: we always list devices first.

No.

No.

---


ltrung:
Close in favor of #1577

PR #1577, opened 2021-09-06
Do not try to set groupId  from cache for null device

ltrung:
**Issue #1568 :**

**Description of changes:**
Currently we requires to call list device API before calling choosing device API. Otherwise, the device cache is not populated and the error `Device cache is not populated. Please make sure to call list devices first` will be thrown. However, this does not make sense in the case of calling null device. 
Update the logic when choosing null device to just set groupId to ''  instead of looking the groupId from device cache.

**Testing**

2. How did you test these changes? Add unit tests to cover this case.
    1. In the meeting demo, open console to delete the device cache (`delete app.audioVideo.deviceController.deviceInfoCache`).
    2. Call `app.audioVideo.chooseAudioInputDevice(null)`
    3. Verify that there is no error message log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added Android WebView Sample UI test to workflow.
 - Add a new optional API `getVideoTileForAttendeeId` in `VideoTileController` and raise the `tileWillBePausedByDownlinkPolicy` event for empty video tiles.
 - Amazon Voice Focus will now trigger the `onCPUWarning` method on the `VoiceFocusTransformDeviceDelegate` when using the default inline execution mode. This will be called when the browser fails to schedule the worklet in a timely fashion (_e.g._, when the meeting code is running in an iframe /subframe) or when changes in the CPU environment (_e.g._, thermal throttling) cause the worklet to take too long for each audio render quantum.
-- The Amazon Voice Focus support check, `VoiceFocusDeviceTransformer.isSupported`, now warns to the logger when run in an iframe, and can be configured to fail in that case.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Clarify why not use default downlink policy with simulcast.
 - Corrected argument `isUnifiedPlan` in `withBandwidthRestriction` to `isFirefox`. Also marked as deprecated since we no longer use it.
 - Update data message limit in the API Overview guide.
+- Do not try to set groupId  from cache for null device.
```
richnew10, line 23: I suggest you steal the changelog entry from my PR. These entries are supposed to be useful for the customer, not a commit message.

test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -997,6 +997,58 @@ describe('DefaultDeviceController', () => {
...
+      expect(e.notCalled).to.be.true;
+      expect(i.called).to.be.true;
+      expect(w.notCalled).to.be.true;
+
+      e.restore();
+      i.restore();
+      w.restore();
+    });
+
+    it('Use groupdId from device cache to prevent inifite loop', async () => {
```
richnew10, line 1016: infinite

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1486,10 +1486,13 @@ export default class DefaultDeviceController
...
-        kind,
-        this.getMediaTrackSettings(newDevice.stream)?.deviceId || ''
-      );
+      newDevice.groupId =
+        device !== null
+          ? this.getGroupIdFromDeviceId(
+              kind,
+              this.getMediaTrackSettings(newDevice.stream)?.deviceId || ''
+            )
+          : '';
```
richnew10, line 1495: Just do `newDevice.groupId = '';` in the `null` section above, and then here you can do a simple conditional.

richnew10 (changes requested):

src/devicecontroller/DefaultDeviceController.ts
ltrung: The null section above is only for `audio` right?

test/devicecontroller/DefaultDeviceController.test.ts
ltrung: Fixed.

richnew10 (dismissed):

nainkunal933 (approved):

PR #1581, opened 2021-09-07
Do not trigger leave event for self attendee during reconnection

ltrung:
**Description of changes:**
Currently during reconnection, we can send out leaving attendee presence event for current attendee during reconnection https://github.com/aws/amazon-chime-sdk-js/blob/master/src/volumeindicatoradapter/DefaultVolumeIndicatorAdapter.ts#L128

However, there is always an audio info later after reconnection for the current attendee in that case so we should not send out the event during reconnection which does not make sense anyway since the fact we receive the audio info from backend means there is connection with the current client.

**Testing**

   1. Launch meeting demo and join with 2 attendees 
   2. Mute all of them
   3. Use this code in dev console to trigger reconnection for one attendee:
   ```
    app.audioVideo.audioVideoController.handleMeetingSessionStatus({
      statusCode: () => 17,
      isFailure: () => true,
      isTerminal: () => false,
      isAudioConnectionFailure: () => false,
     }, null);
   ```
   4. Verify that we do not see any presence = false event in the console log
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes, meeting demo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Clarify why not use default downlink policy with simulcast.
 - Corrected argument `isUnifiedPlan` in `withBandwidthRestriction` to `isFirefox`. Also marked as deprecated since we no longer use it.
 - Update data message limit in the API Overview guide.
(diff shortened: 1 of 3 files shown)


simmkyu (approved):

mklingb (approved):
LGTM

PR #1582, opened 2021-09-07
Updated SauceLabs Integration tests to the new mobile api

LeviSklaroff:
**Issue #:**
TestObject the saucelabs api we're currently using has been deprecated 
**Description of changes:**
Migrated to the new api.
**Testing**

Yes
Ran Data Message and Audio integration tests on MAC Chrome, Firefox, and Android Chrome.
No
No
No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update `startVideoPreviewForVideoInput` to support filters in the preview window.
 - Update browser demo to showcase preview filter capability.
 - Move `toLowerCasePropertyNames` inside `Utils.ts` and add test coverage.
+- Migrate SauceLabs mobile tests to new api. 
 
 ### Removed
 
--- a/integration/js/utils/WebdriverSauceLabs.js
+++ b/integration/js/utils/WebdriverSauceLabs.js
@@ -106,42 +106,34 @@ const getSauceLabsDomain = (platform)  => {
   const platformUpperCase = platform.toUpperCase();
   if (platformUpperCase === 'LINUX') {
     return 'us-east-1.saucelabs.com';
-  } else if (platformUpperCase === 'ANDROID' || platformUpperCase === 'IOS') {
-    return 'us1-manual.app.testobject.com';
-  }
-  return 'saucelabs.com';
+  } 
+  return 'us-west-1.saucelabs.com';
 };
 
 const getSafariIOSConfig = (capabilities) => {
   return {
-    platformName: 'iOS',
+    platformName: capabilities.platform,
     platformVersion: capabilities.version,
-    testobject_api_key: '059183B9A1A148D188B3AC5BA1ECBD52',
     deviceOrientation: 'portrait',
     name: capabilities.name,
   };
 };
 
 const getChromeAndroidConfig = capabilities => {
+
   return {
-    platformName: 'Android',
+    platformName: capabilities.platform,
     platformVersion: capabilities.version,
-    testobject_api_key: '059183B9A1A148D188B3AC5BA1ECBD52',
     deviceOrientation: 'portrait',
     chromeOptions: {
       'args': ['use-fake-device-for-media-stream', 'use-fake-ui-for-media-stream'],
       'w3c': false
     },
-    name: capabilities.name,
+    name: capabilities.name
   };
 };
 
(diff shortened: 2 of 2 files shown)


integration/js/utils/WebdriverSauceLabs.js
```diff
@@ -107,41 +107,33 @@ const getSauceLabsDomain = (platform)  => {
   if (platformUpperCase === 'LINUX') {
     return 'us-east-1.saucelabs.com';
   } else if (platformUpperCase === 'ANDROID' || platformUpperCase === 'IOS') {
-    return 'us1-manual.app.testobject.com';
+    return 'us-west-1.saucelabs.com';
```
ltrung, line 110: can it be just saucelabs.com which I think will resolve to us-west-1 anyway?

integration/js/utils/WebdriverSauceLabs.js
```diff
@@ -200,8 +192,17 @@ class SaucelabsSession {
 
   async getSessionId() {
     if (this.sessionId === undefined) {
-      const session = await this.driver.getSession();
-      this.sessionId = session.getId();
+      if(isMobileDomain(this.domain))
+      {
+        const cap = await this.driver.getCapabilities();
+        const url = cap.get('testobject_test_report_api_url')
```
ltrung, line 198: I thought testobject is deprecated?

integration/js/utils/WebdriverSauceLabs.js
harshitOberoi: Also is there a way to configure these values instead of hardcoding them in code.

LeviSklaroff: @ltrung I had thought this as well, but using simply saucelabs.com results in a capabilities not found error. Not entirely sure why this happens as the other tests being ran using the url seem to go to us-west-1, but it probably has to do with how saucelabs redirects requests. I don't think it would be a bad idea to change the default return to us-west-1 though that way we could get rid of the mobile check right there, what are your thoughts?

ltrung: Yeah if the other test with us-west-1 URL we can do that too :)

LeviSklaroff: @harshitOberoi The platform is already configurable, and as for the url this will cause more issues than merits. There are only two servers we use and specific platforms can only be ran on specific servers so making it configurable will introduce room for error in misconfiguration and result in more code in trying to configure each test.

integration/js/utils/WebdriverSauceLabs.js
LeviSklaroff: It's not exactly that testObject has been completely deprecated, but rather now it runs under the hood of the saucelabs api. The url it returns is the normal saucelabs test url https://app.saucelabs.com/tests/<JobID> this is the method provided by saucelabs support to grab the jobid.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   during reconnection.
 - Update `startVideoPreviewForVideoInput` to support filters in the preview window.
 - Update browser demo to showcase preview filter capability.
+- Migrated SauceLabs mobile tests to new api.
```
jing-chen1, line 30: Actually we use simple present in the ` CHANGELOG`, please use `Migrate`

LeviSklaroff: Changed!

integration/js/utils/WebdriverSauceLabs.js
LeviSklaroff: @ltrung Changed!

integration/js/utils/WebdriverSauceLabs.js
```diff
@@ -200,8 +188,17 @@ class SaucelabsSession {
 
   async getSessionId() {
     if (this.sessionId === undefined) {
-      const session = await this.driver.getSession();
-      this.sessionId = session.getId();
+      const cap = await this.driver.getCapabilities();
+      const url = cap.get('testobject_test_report_api_url')
+      if(url !== undefined)
```
ltrung, line 193: Can we add a comment here that say this is only for mobile test from the old test object framework?

LeviSklaroff: Definitely agree, added!

ltrung (approved):

PR #1589, opened 2021-09-10
Refactor toLowerCasePropertyNames and add test coverage

nainkunal933:
**Issue:** 
#1052 

**Description of changes:** Making the method static to resolve the `this` is undefined error in `toLowerCaseUpperNames`. 

**Testing**

2. How did you test these changes? I have added unit tests. Also, I checked the validity of these changes by updating the parameters of `MeetingSessionConfiguration` in `meetingV2.ts`. To test, replace the attendee response from JS SDK response `joinInfo.Attendee` to a response like this with metadata in it:
```json
{
	"Attendee": {
		"ExternalUserId": "User0",
		"AttendeeId": "ac121238-0bd1-cb38-a85a-a23d7fds1df9",
		"JoinToken": "YWMxMjEyMfsdfdsfdsfdsfsdw4LWE4NWEtYTIzZDdkYTgxZGY5OjY5NmM1MGI4LTkxOTItNGZkZS05NWJiLTcxNzFlYWZkN2U4YQ"
	},
	"@metadata": {
		"statusCode": 201,
		"effectiveUri": "https://chime.us-east-1.amazonaws.com/meetings/6fe13920-be68-42b3-8229-0276e5b62869/attendees",
		"headers": {
			"x-amzn-requestid": "fc876791-3efe-4943-895f-a54496cf111a",
			"content-type": "application/json",
			"content-length": "216",
			"date": "Thu, 11 Feb 2021 07:29:55 GMT"
		},
		"transferStats": {
			"http": [
				[]
			]
		}
	}
}
```
With the new changes, `toLowerCasePropertyNames` should be able to parse the data without any errors.
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? Yes browser demo


CHANGELOG.md
```diff
@@ -27,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   during reconnection.
 - Update `startVideoPreviewForVideoInput` to support filters in the preview window.
 - Update browser demo to showcase preview filter capability.
+- Move `toLowerCasePropertyNames` to a static method inside utils.js.
```
devalevenkatesh, line 31: Few suggestions:
- I think we can club these two changelogs together to learn about the scope of the change in one place. 
- It is not "a static method" right? 
- We do not have a `util.js`, so we change it to `Utils.ts`.

devalevenkatesh (approved):

test/utils/Utils.test.ts
```diff
@@ -3,18 +3,119 @@
...
+          AudioHostUrl: 'audio-host-url',
+          ScreenDataUrl: 'screen-data-url',
+          ScreenSharingUrl: 'screen-sharing-url',
+          ScreenViewingUrl: 'screen-viewing-url',
+          SignalingUrl: 'signaling-url',
+          TurnControlUrl: 'turn-control-url',
+        },
+        Http: [[]],
+      };
+
```
LeviSklaroff, line 63: I could be missing something, but shouldn't the input consist of upper case letters as we're trying to test converting upper case to lower case?

nainkunal933: The input on line 61 is testing for the scenario when we have two empty arrays. It is important I add that to the test as there is a if block in the method that was the root cause and I am testing against it. Also, the keys will be converted to lowercase and not the values. So, theoretically `Http` will become `http`.

LeviSklaroff: Ah, I thought it was changing the values to lower case, but yeah that makes sense.

PR #1593, opened 2021-09-14
Add csp setup in demo app and integration test

ldai1:
**Issue #:**

**Description of changes:**
Add csp setup in demo app and integration test
**Testing**

1. Have you successfully run `npm run build:release` locally? yes
2. How did you test these changes? integration test
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? meeting demo
4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? N/A
5. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved? N/A

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1594, opened 2021-09-14
Make video preferences easier to test by adding explicit configuration replacing non-functional pin button

hensmi-amazon:
**Issue #:** None, though this should help with testing RTP stream switching functionality.

**Description of changes:** The title plus an image explains it:
Screen Shot 2021-09-14 at 10 09 38 AM

Basically the 'pin' button did not actually pin a tile UI wise, it just set priority to high.  It was strange UX, and additionally did not let us easily test target resolution setting, and simulcast stream switches (which would require forcing some network event, now you can just use target resolution).  This change adds a config to remote videos which includes explicit settings.  I also moved the pause there.  I also retain usage of the default policy paths before buttons are set.

I will instruct QA how to use these buttons.

**Testing**
* Using priority policy, button shows up on remote videos but not local video.  Has pause + preference toggles.  Preference toggles work as expected and can be confirmed in logs.  Target resolution changes were tested in 3 person calls (when simulcast kicks in when turned on in a remote participant).
* Not using priority policy, button shows up on remote videos but not local video.  Has pause and nothing else.
* Confirmed pause still works in either case.
* Tested enabling video on either case + simulcast chrome sender.

1. Have you successfully run `npm run build:release` locally? y
2. How did you test these changes? see above
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? meetingdemo
4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? n
5. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved? n


demos/browser/app/meetingV2/meetingV2.html
```diff
@@ -492,224 +488,1063 @@ <h1 class="h3 mb-3 font-weight-normal text-center">Select devices</h1>
...
+            </div>
           </div>
           <div id="tile-1" class="video-tile">
             <video id="video-1" class="video-tile-video"></video>
             <div id="attendeeid-1" class="video-tile-attendeeid"></div>
             <div id="nameplate-1" class="video-tile-nameplate"></div>
             <div id="pause-state-1"></div>
             <button id="video-pause-1" class="video-tile-pause">Pause</button>
-            <button id="video-pin-1" class="video-tile-pin">Pin</button>
+            <div class="button-video-tile-config" role="group">
```
ltrung, line 529: nit: Can we use web component and move this to a reusable code instead of copy them in all 25 video elements :D

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -3219,34 +3219,81 @@ export class DemoMeetingApp
...
+  }
+
+  convertPriorityValue(value: string): number {
+    switch (value) {
+      case 'low':
+        return 10
+      case 'medium':
+        return 5
+      case 'high':
+        return 0
```
ltrung, line 3258: According to our [guide](https://github.com/aws/amazon-chime-sdk-js/pull/1593/files#diff-bf95afb6c6ffb26425700a8ed44789aa169e96cd3e93679a60122e2ca0a2e743R13) 1 is the highest priority so its probably should be 1 here.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -3501,9 +3569,6 @@ export class DemoMeetingApp
 
   remoteVideoSourcesDidChange(videoSources: VideoSource[]): void {
     this.log(`available remote video sources changed: ${JSON.stringify(videoSources)}`);
-    if (!this.enablePin) {
```
ltrung, line 3504: So originally I did this so that we have a way to test the default behavior for priority-based downlink (we accidentally broke that before because we have no way to test it in this demo). Maybe we can add a boolean flag here to only do this if users click on target size or priority first.

demos/browser/app/meetingV2/meetingV2.html
hensmi-amazon: Messaged you.  Totally agree, will likely do in a later PR :P .

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -582,6 +581,7 @@ export class DemoMeetingApp
       this.enableUnifiedPlanForChromiumBasedBrowsers = !(document.getElementById(
         'planB'
       ) as HTMLInputElement).checked;
+      this.usePriorityBasedDownlinkPolicy = (document.getElementById('priority-downlink-policy') as HTMLInputElement).checked;
```
ltrung, line 584: Do we need to do this? I thought it was already set above.

demos/browser/app/meetingV2/meetingV2.ts
hensmi-amazon: I have refactored it in a way where this is still possible as mentioned in desc.  Before you press any config, it will never call `chooseRemoteVideos` because it will short circuit if `this.attendeeIdToVideoPreference` is empty (tested this as well).

demos/browser/app/meetingV2/meetingV2.ts
hensmi-amazon: Firefox will save your check boxes.  Since this is default unchecked I added this for convenience.  Though there are probably other places we need to account for this as well.

demos/browser/app/meetingV2/meetingV2.ts
ltrung: Oh I missed that. Thanks

ltrung (dismissed):

ltrung (approved):

CHANGELOG.md
```diff
@@ -10,9 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add events `meetingReconnected`, `signalingDropped` and `receivingAudioDropped` to `eventDidReceive` by publishing them as stand alone events. Currently, these events were only included in the meeting history attribute when a meeting event is published. 
 
+### Removed
+
+### Fixed
 
 ### Changed
 - Move `toLowerCasePropertyNames` inside `Utils.ts` and add test coverage.
+- Modified browser demo to make video preferences easier to test by adding explicit configuration replacing non-functional pin button
```
richnew10, line 19: - The browser demo now offers a configuration menu on each video tile. This menu replaces the 'Pin' button, which previously set the priority to X. The new configuration menu allows the user to specify the desired video quality and priority, which will be respected by simulcast and priority downlink policies. This is useful for testing or to demonstrate the behavior of those policies.

demos/browser/app/meetingV2/meetingV2.html
richnew10: Please do it now, or generate the grid programmatically in `meetingV2.ts`. This is a big and unpleasant change.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -270,6 +270,8 @@ export class DemoMeetingApp
   static readonly DATA_MESSAGE_LIFETIME_MS: number = 300_000;
   // We reserve the last tile index for local video
   static readonly LOCAL_VIDEO_TILE_INDEX: number = DemoTileOrganizer.MAX_TILES;
+  static readonly DEFAULT_PRIORITY: number = 5;
```
richnew10, line 273: `DEFAULT_VIDEO_TILE_PRIORITY` or `VIDEO_TILE_DEFAULT_PRIORITY`.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -270,6 +270,8 @@ export class DemoMeetingApp
...
+  static readonly DEFAULT_TARGET_SIZE: TargetDisplaySize = TargetDisplaySize.High;
```
richnew10, line 274: Similar.

demos/browser/app/meetingV2/meetingV2.ts
richnew10: Thank you for doing this!

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -3222,34 +3222,81 @@ export class DemoMeetingApp
...
+  convertPriorityValue(value: string): number {
+    switch (value) {
+      case 'low':
+        return 10;
+      case 'medium':
+        return 5;
+      case 'high':
+        return 1;
+    }
+  }
```
richnew10, line 3263: Consider doing these conversion functions with typed data.

```
const PRIORITIES: {[Key in 'low' | 'medium' | 'high']: number} = {
  low: 10,
  medium: 5,
  high: 1,
};
```

rather than defining a whole function for them. Same for `convertTargetResolutionValue`.

PR #1596, opened 2021-09-15
Add safeguard for Nscale policy in case we increase to more than 25 videos

ltrung:
**Description of changes:**
Add safeguard for Nscale policy in case we increase to more than 25 videos to not run into exception when calculating target height.

**Testing**

2. How did you test these changes? Add unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 ### Fixed
+- Add safeguard for Nscale policy in case we increase to more than 25 videos.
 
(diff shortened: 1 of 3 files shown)


hensmi-amazon:
Would a function be more intuitive here?  Or is there a reason we use a table?

ltrung:
> Would a function be more intuitive here? Or is there a reason we use a table?

This is according to feedback from John D to provide more fine grain control for tweaking this in the future. Also one would not need to mess with a bunch of if statement if they want to change something.

hensmi-amazon (dismissed):

akshay-prabhu-github (approved):

hensmi-amazon (approved):

PR #1598, opened 2021-09-15
Update integration test setup documentation

nainkunal933:
**Issue #:**

**Description of changes:** I was setting up the integration test locally and found some missing information in the README. With the KITE version latest version of maven is not compatible and this will cause issues in setup. Maven version 3.6.3 or below are compatible with KITE. Added documentation to help the user get around this issue.

**Testing**

2. How did you test these changes? NA
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? NA
4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? NA

### DIFF
--- a/integration/js/README.md
+++ b/integration/js/README.md
@@ -9,6 +9,7 @@ To run integration tests you will need:
 - Node 10 or higher. Node 14 is supported.
 - npm 6.11 or higher. 6.14.8 is supported.
 - [KITE](https://github.com/webrtc/KITE).
+- Java runtime installed on the machine.
 
 ### Installing KITE on macOS
 
@@ -34,6 +35,10 @@ chmod +x configureMac.sh
 ```
 
 Choose 'y'.
+
+Note: If the `./configureMac.sh` script fails then make sure you are using Maven 3.6.3 or earlier. The KITE version in use does not work with the latest version of maven.
+For example, you can install maven 3.5 by calling `brew install maven@3.5`. If the issue persists, make sure the symlink is pointing to the right version of maven. Use `mvn -v` to check your maven version.
+If you have multiple installations of maven in homebrew then use `brew unlink maven` to unlink the latest version of maven and use `brew link maven@3.5` to use an older version.
 
 Now find the Terminal window that opened and complete interactive setup. A good configuration is:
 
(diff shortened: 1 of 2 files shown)


integration/js/README.md
```diff
@@ -33,7 +34,10 @@ chmod +x configureMac.sh
 ./configureMac.sh
 ```
 
-Choose 'y'.
+Choose 'y'. <br />
```
hensmi-amazon, line 37: is the `<br />`  necessary?

integration/js/README.md
```diff
@@ -33,7 +34,10 @@ chmod +x configureMac.sh
...
+Note: If the `./configureMac.sh` script fails then make sure you are using `maven@3.5`. We are not using the latest version of KITE and the setup does not seem to be compatible with the latest version of maven.<br />
```
ltrung, line 38: it is maven 3.6.3 as mentioned here https://github.com/aws/amazon-chime-sdk-js/blob/master/integration/js/script/install-kite#L12.
Also this can be shorter as "The KITE version in use does not work with later version of maven".

integration/js/README.md
ltrung: It should not be :)

nainkunal933: I just did it mainly for formatting. I want to make sure Note is on next line because Choose 'y' is part of the script above. Please let me know if you want me to remove that br. 

I can also use two spaces to add the line break if you prefer that. I chose `<br />` as that is easier to read.

ltrung:
Can you add the README to integration test watch list so we do not need to run integ tests when updating README here https://github.com/aws/amazon-chime-sdk-js/blob/master/.integration-watchlist

integration/js/README.md
hensmi-amazon: Let's use a line break since that is what the rest of the document does :).

nainkunal933: Done, thanks.

integration/js/README.md
nainkunal933: I have updated the documentation to recommend maven version 3.6.3 or earlier. I have marked the maven commands that I added as an example. Maven 3.6 is not available in homebrew.

I have reduced the length of the sentence. Thank you.

ltrung:
Need PR description :)

ltrung (approved):

PR #1603, opened 2021-09-16
Split content share integration test into two jobs

nainkunal933:
**Issue #:**

**Description of changes:** Currently the content share integration test is taking around 40 mins to run. In some cases, it is as high as 60 mins. I am splitting the content share integration test workflow into two jobs. The two jobs will be test suite one and test suite two. Since, jobs run in parallel this should should allow us to cut the runtime in half. This will consume one more sauce lab session as the process now will be in parallel.

**Testing**

2. How did you test these changes? NA
3. Can these changes be tested using the demo application? If yes, which demo application can be used to test it? NA


ltrung:
Can we group all smaller tests into one suite?

nainkunal933:
If we group all smaller tests into a smaller one than it does not have an impact on the performance. Please refer to this [PR](https://github.com/aws/amazon-chime-sdk-js/pull/1605) to check the performance. I do prefer to have tests in the same config file as that cleans up the code base. We should combine the configs at some point.

ltrung:
No i mean to save SauceLabs sessions

> If we group all smaller tests into a single config than it does not have an impact on the performance. Please refer to this [PR](https://github.com/aws/amazon-chime-sdk-js/pull/1605) to check the performance. I do prefer to have tests in the same config file as that cleans up the code base. We should combine the configs at some point.

No 

> If we group all smaller tests into a single config than it does not have an impact on the performance. Please refer to this [PR](https://github.com/aws/amazon-chime-sdk-js/pull/1605) to check the performance. I do prefer to have tests in the same config file as that cleans up the code base. We should combine the configs at some point.

No I mean to group them in run-test to save on SauceLab sessions :). They take 1 min each so doesn't need to parallel

nainkunal933:
@ltrung If you group the smaller tests in one config in `run-test` or you split them among two test suites, you will end up with two sauce lab sessions in each scenario. All of the tests inside a test suite / job are run sequentially and all of the jobs are run in parallel. The smaller tests are combined with the larger tests in test suites to get a reasonable average time. 

If you take the smaller ones in one config and put the two longer tests in other config then the test suite / job with longer tests will become the bottleneck.

nainkunal933:
> No i mean to save SauceLabs sessions
> 
> > If we group all smaller tests into a single config than it does not have an impact on the performance. Please refer to this [PR](https://github.com/aws/amazon-chime-sdk-js/pull/1605) to check the performance. I do prefer to have tests in the same config file as that cleans up the code base. We should combine the configs at some point.
> 
> No
> 
> > If we group all smaller tests into a single config than it does not have an impact on the performance. Please refer to this [PR](https://github.com/aws/amazon-chime-sdk-js/pull/1605) to check the performance. I do prefer to have tests in the same config file as that cleans up the code base. We should combine the configs at some point.
> 
> No I mean to group them in run-test to save on SauceLab sessions :). They take 1 min each so doesn't need to parallel

I see what you saying. I thought your question was in the context of content share integration test. If we merge smaller workflows then we will be able to reduce the number of sessions. I have merged two workflows and have updated this PR.

.github/workflows/miscellaneous-integration.yml
```diff
@@ -75,6 +74,12 @@
...
           run: npm ci
+        - name: Run Data Message Integration Test
+          if:  steps.test_needed.outputs.integ_test_required == 'true'
+          run: npm run test:integration-data-message
         - name: Run Meeting Readiness Checker Integration Test
           if:  steps.test_needed.outputs.integ_test_required == 'true'
-          run: npm run test:integration-meeting-readiness-checker
\ No newline at end of file
+          run: npm run test:integration-meeting-readiness-checker
+        - name: Run Meeting End Integration Test
```
ltrung, line 83: Can we also add messaging session at the end ?

nainkunal933: Talked offline about this, ignoring this comment.

ltrung (approved):

PR #1606, opened 2021-09-17
Add MediaCapture integration test

anuranduttaroy:
**Issue #:** NA

**Description of changes:**
- Add integration tests for media capture feature

**Testing**

    - Yes
    - Ran the integration tests manually on Saucelabs.
    - No. This is an integration test change. Note: This integration test is not being used currently for the PRs being opened.
    - NA
    - NA


achris (approved):

sinanliuAmazon (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add events `meetingReconnected`, `signalingDropped` and `receivingAudioDropped` to `eventDidReceive` by publishing them as stand alone events. Currently, these events were only included in the meeting history attribute when a meeting event is published. 
 - Add link to SIP Media Application examples in README.
+- Add integration test for MediaCapture features.
```
devalevenkatesh, line 13: nit: Add a space here after the log. Also, we do not enforce changelog for the integration tests or demo changes now.

integration/js/utils/SdkBaseTest.js
```diff
@@ -20,6 +20,10 @@ class SdkBaseTest extends KiteBaseTest {
       this.userArn = this.payload.userArn;
     }
 
+    if (testName === 'MediaCapture') {
+      this.region = this.payload.region;
```
devalevenkatesh, line 24: So, currently, we set the region directly through URL for some tests and other stages, so is there a way or should we set the `this.region` value to whatever is the selected one instead of empty when the test is not a 'MediaCapture' test?

anuranduttaroy: We are using the [region in the test](https://github.com/aws/amazon-chime-sdk-js/pull/1606/files#diff-bf5edd9d104e0847b257ff2fa050a7d5921edb0c5675f80fc9726ba9afbc59cbR28) to select the region as a step instead of using it directly from the URL before starting the meeting/adding the attendee. We are doing this because the `MediaCapture` feature is available in more than 18 regions and it's not required to maintain a separate URL for each region since we have the region param available in the test config.

This won't effect the other tests as it only does it for `MediaCapture` test and other test don't have the step to set the region. They just auto-populate the region from the URL and don't change it in the actual test.

devalevenkatesh: Yup, thanks for the explanation, this wont affect other test but since we are adding `region` field, checking if we can keep it updated with whatever region is selected or if you custom select a different one so that we have a valid region populated for all tests. @ltrung could you pleas help with any suggestions? I have no strong opinion but want to check once.

anuranduttaroy: Synced with you offline.

ltrung:
Do we use it only for canary? Should we add it to integration tests in GitHub action?

integration/configs/media_capture_test.config.json
```diff
@@ -0,0 +1,26 @@
+{
+    "`": 1,
```
ltrung, line 2: When do we use this config :D

anuranduttaroy: Currently it's not being used in the integration tests. I plan to add this to the integration test workflow for every PR.

CHANGELOG.md
```diff
@@ -11,7 +11,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add events `meetingReconnected`, `signalingDropped` and `receivingAudioDropped` to `eventDidReceive` by publishing them as stand alone events. Currently, these events were only included in the meeting history attribute when a meeting event is published. 
 - Added support for skipping full SDP renegotiations when switching simulcast streams.  This will result in less freezing when switching between layers in response to a network event as done in `VideoPriorityBasedPolicy`.  This will have no impact if not using simulcast.
 - Add link to SIP Media Application examples in README.
-
```
devalevenkatesh, line 14: nit: Need this extra line.

devalevenkatesh (approved):

ltrung (approved):

PR #1607, opened 2021-09-17
Add support for skipping SDP renegotiation for stream encoding change

bjason:
**Issue #:**

N/A

**Description of changes:**

When there are only bitrate changes in current implementation, it will trigger a resubscription. In this change, we avoid resubscribing when bitrate changes, instead just setting the values on the RTP sender. Though this option was not available when express was initially implemented the requisite APIs have been available across the board since 2018. 

**Testing**

yes
Smoke test 2-3 attendees call on Chromium and Safari with simulcast on/off. Used the network link conditioner to limit bandwidth to check resubscribe and skipping resubscribe behavior by checking console logs. 
Designed unit test.

yes
no
no


src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -747,13 +747,15 @@ export default class DefaultAudioVideoController
...
-    // If updating remote video without negotiation fails, fall back to renegotiation
+    // If updating local or remote video without negotiation fails, fall back to renegotiation
     needsRenegotiation ||= !this.updateRemoteVideosFromLastVideosToReceive();
+    needsRenegotiation ||= !this.updateLocalVideoFromPolicy();
     // `MeetingSessionContext.lastVideosToReceive` needs to be updated regardless
     this.meetingSessionContext.lastVideosToReceive = this.meetingSessionContext.videosToReceive;
     if (!needsRenegotiation) {
       this.logger.info('Update request does not require resubscribe');
-      return true; // Skip the subscribe!
+      this.actionFinishUpdating();
```
hensmi-amazon, line 757: Can you log why we need this?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -872,6 +874,17 @@ export default class DefaultAudioVideoController
     return true;
   }
 
+  updateLocalVideoFromPolicy(): boolean {
+    // Update bandwidth without renegotiation
+    this.logger.info('Updating local video from policy without renegotiation');
+    if (this.meetingSessionContext.enableSimulcast) {
```
hensmi-amazon, line 880: Can you log whats going on in this function?  Also this always returns true, which is not accurate for when we are switching simulcast streams.  I think i accounted for this logic in my branch, maybe it was mistakenly removed when you stripped out the RTP header extension stuff.

src/audiovideocontroller/DefaultAudioVideoController.ts
bjason: Added comment: update encoding in updateLocalVideoFromPolicy and apply the new bitrate to send here

src/audiovideocontroller/DefaultAudioVideoController.ts
bjason: Fixed! Thanks!

src/audiovideocontroller/DefaultAudioVideoController.ts
hensmi-amazon: Maybe can be more explicit that we are reusing the function.
```
// Call `actionFinishUpdating` to apply the new encoding parameters that may have been set in `updateLocalVideoFromPolicy`.
```

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -872,6 +876,25 @@ export default class DefaultAudioVideoController
...
+    if (this.mayNeedRenegotiationForSimulcastLayerChange) {
+      this.logger.info('Needs regenotiation for local video simulcast layer change');
+      this.mayNeedRenegotiationForSimulcastLayerChange = false;
+      return false;
+    }
+
+    // Update bandwidth without renegotiation
+    this.logger.info('Updating local video from policy without renegotiation');
+    if (this.meetingSessionContext.enableSimulcast) {
+      // AttachMediaInputTask will update simulcast streams encoding change
```
hensmi-amazon, line 889: Nit: Wrap symbols in `.  Also can you be more explicit ? like are we updating the tranceivers here?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -872,6 +876,25 @@ export default class DefaultAudioVideoController
...
+      new AttachMediaInputTask(this.meetingSessionContext).run();
+    } else {
+      // calling ReceiveVideoInputTask to update videoCaptureAndEncodeParameter from uplink policy
```
hensmi-amazon, line 892: Same here.  We are updating `videoCaptureAndEncodeParameter` on what?

hensmi-amazon (commented):
Looking good implementation wise.  Just need to tighten up comments.

hensmi-amazon (approved):
LGTM!

ltrung:
Please rebase and run npm run build:release and fix all the build issues (probably just lint and doc).

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -872,6 +876,26 @@ export default class DefaultAudioVideoController
...
+      this.logger.info('Needs regenotiation for local video simulcast layer change');
+      this.mayNeedRenegotiationForSimulcastLayerChange = false;
+      return false;
+    }
+
+    // Update bandwidth without renegotiation
+    this.logger.info('Updating local video from policy without renegotiation');
+    if (this.meetingSessionContext.enableSimulcast) {
+      // `AttachMediaInputTask` will update sender's simulcast streams encoding parameters of the local video transceiver
+      new AttachMediaInputTask(this.meetingSessionContext).run();
```
ltrung, line 890: Don't we also have some logic for simulcast in ReceiveVideoInputTask? 
https://github.com/aws/amazon-chime-sdk-js/blob/master/src/task/ReceiveVideoInputTask.ts#L50

hensmi-amazon: I don't think that logic matters, since the reason its integrated into the `VideoStreamIndex` is so we can advertise the new max bitrates in the subscribe message.  Since we intentionally avoid the subscribe it doesn't really matter (downlinks should be using avg bitrate).

ltrung (approved):

PR #1611, opened 2021-09-21
Fix stat info table in meeting demo

ltrung:
**Issue #:**
The video stat table is broken when there are more than 2 videos after we made the changes for video tile web component.
**Description of changes:**
- Remove the lookup logic and just add a new row of data to stat info table (since we create a new table everytime).
- Remove `getStatsForOutbound` as it is not used.
- Put the default video quality when simulcast is enabled into once place.

**Testing**

- Had a meeting with 3 videos on
- Clicked on stats button
- Verified that the stats are rendered correctly.

4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? N/A, demo only


hensmi-amazon (approved):
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

Issue #1524, opened 2021-08-19
Simulcast documentation feedback - downlink policy information


mark-vanr:
### What are you trying to do?

Using the documentation provided at https://aws.github.io/amazon-chime-sdk-js/modules/simulcast.html -  
I am trying to enable simulcast for users who are on the Chromium browsers which support it, and also have the `VideoAdaptiveProbePolicy` downlink policy enabled on browsers that do not support simulcast to take advantage of the quality controls it seems to offer.

### How can the documentation be improved to help your use case?
I am currently running into 2 difficulties: 

#### 1. Inaccuracy of default downlink policy statement
In the introductory paragraph of the article it states that 
> The VideoAdaptiveProbePolicy downlink policy adaptively subscribes to the best simulcast layer and is automatically selected if MeetingSessionConfiguration.enableUnifiedPlanForChromiumBasedBrowsers and MeetingSessionConfiguration.enableSimulcastForUnifiedPlanChromiumBasedBrowsers are set to true.

This doesn't seem to be accurate, unless it is happening somewhere I can't see. I was looking into the `DefaultAudioVideoController` (which I am using) and the [`if` block here ](https://github.com/aws/amazon-chime-sdk-js/blob/41d5a53d7fb9b7ed8591f645ced5bf7cebedcca1/src/audiovideocontroller/DefaultAudioVideoController.ts#L477) only changes the context's uplink bandwidth policy based on if `enableSimulcast` is enabled, but doesn't change the downlink. It appears that otherwise it uses the default downlink policy, which the [default MeetingSessionConfiguration sets to AllHighestVideoBandwidthPolicy here](https://github.com/aws/amazon-chime-sdk-js/blob/41d5a53d7fb9b7ed8591f645ced5bf7cebedcca1/src/meetingsession/MeetingSessionConfiguration.ts#L184) -- which also contains a comment implying both uplink and downlink policies will be overridden.

If I'm mistaken here then I'd be happy to be told so, it just looks as though the `VideoAdaptiveProbePolicy` is not in use - I also couldn't find any reference to the `VideoAdaptiveProbePolicy` constructor other than in test files.

#### 2. How to effectively use the VideoAdaptiveProbePolicy
Under the **Downlink Bandwidth Policy** segment, it states
> To enable it without simulcast set the MeetingSessionConfiguration.videoDownlinkBandwidthPolicy to VideoAdaptiveProbePolicy.

I tried to do that simply by calling 
```
[...]

PR #1614, opened 2021-09-22
Allow to pass in custom simulcast uplink policy

ltrung:
**Issue #1524:**
Previously, when simulcast was enabled, `DefaultSimulcastUplinkPolicy` is always used no matter the value passing from `MeetingConfiguration.videoUplinkBandwidthPolicy`. 

Also the default downlink policy for simulcast is `AllHighestVideoBandwidthPolicy` and not `VideoAdaptiveProbePolicy` as mentioned in the documentation.

**Description of changes:**
- Allow passing in video uplink bandwidth policy when simulcast is enabled as long as they implement the `SimulcastUplinkPolicy` interface.
- Update the default video downlink policy when simulcast to be `VideoAdaptiveProbePolicy`.
- Add an API ``isSimulcastSupported` so clients can check whether simulcast can be enabled and pass in their own uplink policy.

**Testing**

- Unit tests
- Joined meeting demo with simulcast enabled and check the value of video uplink and downlink policies when customize uplink and downlink are specified.

4. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? Yes


src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -484,27 +486,46 @@ export default class DefaultAudioVideoController
     this.meetingSessionContext.enableSimulcast = this.enableSimulcast;
 
     if (this.enableSimulcast) {
```
hensmi-amazon, line 488: Hmm, it'd be nice if builders didn't have to set `enableSimulcastForChromiumBasedBrowsers` if they were passing in a uplink policy that implemented `SimulcastUplinkPolicy`.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -484,27 +486,46 @@ export default class DefaultAudioVideoController
...
+
       this.meetingSessionContext.videoStreamIndex = new SimulcastVideoStreamIndex(this.logger);
     } else {
       this.meetingSessionContext.enableSimulcast = false;
       this.meetingSessionContext.videoStreamIndex = new DefaultVideoStreamIndex(this.logger);
-      if (!this.meetingSessionContext.videoDownlinkBandwidthPolicy) {
-        this.meetingSessionContext.videoDownlinkBandwidthPolicy = new AllHighestVideoBandwidthPolicy(
+
+      if (!this.meetingSessionContext.videoUplinkBandwidthPolicy) {
+        this.meetingSessionContext.videoUplinkBandwidthPolicy = new NScaleVideoUplinkBandwidthPolicy(
```
hensmi-amazon, line 519: I guess this means they need to check for browser support before picking the policy?  e.g. on FF if you just set SImulcastUplinkPolicy for everything, it would be completely broken given all the no-op functions that may be called depending on `enableSimulcast`?

src/audiovideocontroller/DefaultAudioVideoController.ts
rnewman: They shouldn’t need to; enableSimulcast should only be relevant if they don’t explicitly provide a policy.

src/audiovideocontroller/DefaultAudioVideoController.ts
rnewman: There are a few reasonable positions to take:

1. Simulcast policies seamlessly degrade to non-simulcast in unsupported browsers. 
2. Simulcast policy constructors throw if they won’t be able to function.
3. 2 plus exposing a way for builders to check (e.g., a static isSupported method on a simulcast policy class).

ltrung: Implement validation (option 1) in DefaultMeetingSession.

src/audiovideocontroller/DefaultAudioVideoController.ts
ltrung: Add enabling simulcast if passing in simulcast policy.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add events `meetingReconnected`, `signalingDropped` and `receivingAudioDropped` to `eventDidReceive` by publishing them as stand alone events. Currently, these events were only included in the meeting history attribute when a meeting event is published. 
 - Added support for skipping full SDP renegotiations when switching simulcast streams.  This will result in less freezing when switching between layers in response to a network event as done in `VideoPriorityBasedPolicy`.  This will have no impact if not using simulcast.
 - Add link to SIP Media Application examples in README.
+- Add API `isSimulcastSupported` so applications can check whether simulcast can be enabled and pass corresponding
```
hensmi-amazon, line 14: Seems like we are 50/50 on whether to manually split these onto new lines :P.

src/meetingsession/DefaultMeetingSession.ts
```diff
@@ -191,20 +193,47 @@ export default class DefaultMeetingSession implements MeetingSession, Destroyabl
...
+
+    if (enableSimulcast) {
       if (!this._configuration.enableUnifiedPlanForChromiumBasedBrowsers) {
-        this._configuration.enableSimulcastForUnifiedPlanChromiumBasedBrowsers = false;
         this.logger.info(
           'Simulcast requires enabling WebRTC Unified Plan for Chromium-based browsers'
         );
-      } else if (browserBehavior.hasChromiumWebRTC()) {
+        enableSimulcast = false;
+      } else if (this._configuration.videoUplinkBandwidthPolicy && !isSimulcastPolicy) {
```
hensmi-amazon, line 208: isn't the `this._configuration.videoUplinkBandwidthPolicy` check redundant with the body of `isSimulcastUplinkPolicy`?

src/meetingsession/DefaultMeetingSession.ts
```diff
@@ -191,20 +193,47 @@ export default class DefaultMeetingSession implements MeetingSession, Destroyabl
...
+          this._configuration.videoUplinkBandwidthPolicy = undefined;
+        }
+      } else {
+        this._configuration.enableSimulcastForUnifiedPlanChromiumBasedBrowsers = true;
       }
     }
   }
+
+  private isSimulcastUplinkPolicy(policy: VideoUplinkBandwidthPolicy | undefined): boolean {
+    return !!(policy && (policy as SimulcastUplinkPolicy).addObserver);
```
hensmi-amazon, line 237: Had to look up `!!`.  What a beautiful language...

src/meetingsession/DefaultMeetingSession.ts
```diff
@@ -191,20 +193,47 @@ export default class DefaultMeetingSession implements MeetingSession, Destroyabl
...
           'Simulcast requires WebRTC Unified Plan and is only supported on Chromium-based browsers'
         );
+        enableSimulcast = false;
+      }
+
+      if (!enableSimulcast) {
+        //Clean up if simulcast can't be enabled
+        this._configuration.enableSimulcastForUnifiedPlanChromiumBasedBrowsers = false;
+        if (isSimulcastPolicy) {
+          this._configuration.videoUplinkBandwidthPolicy = undefined;
```
hensmi-amazon, line 228: Can we add to the logs above to call out that we are removing the set policy?  I'm guessing builders may not look at code and realize what is happening.

src/audiovideocontroller/DefaultAudioVideoController.ts
hensmi-amazon: I guess its fine since at this point almost no builders are writing their own policies.  But technically this implementation requires builders with a custom non-simulcast policy to shim it (or check for support) on their own.

src/meetingsession/DefaultMeetingSession.ts
ltrung: isSimulcastPolicy is only set if there is a policy in configuration and it is a simulcast policy. This is the case when they pass in a non-simulcast policy but turn on Simulcast via `enableSimulcastForUnifiedPlanChromiumBasedBrowsers` flag.

richnew10: Are you sure that's what that means?

As a builder, I might want to enable simulcast _support_  in the SDK without turning on simulcast for a particular meeting, and if I specify a policy or a flag, I don't expect the SDK to override it.

I see the hierarchy as:

* If they pass a policy, use that policy. If the policy conflicts with browser support, either continue to work (simulcast degrades into unicast), or throw an error.
* If they don't pass a policy, use a flag like `enableSimulcast` to decide what to do. If their preference (`enableSimulcast`) conflicts with browser support, either continue to work (simulcast degrades into unicast), or throw an error.
* If they don't pass a policy or specify a preference, do the default thing, which might vary between browsers based on support.

CHANGELOG.md
richnew10: I always used to `qq` split them, but one-line-per-changelog works fine, so let's go with that.

guides/05_Simulcast.md
```diff
@@ -5,10 +5,7 @@ The uplink policy controls the configuration of the renditions through camera ca
...
 
-The [VideoAdaptiveProbePolicy](https://aws.github.io/amazon-chime-sdk-js/classes/videoadaptiveprobepolicy.html) 
-downlink policy adaptively subscribes to the best simulcast layer and should be used instead of the default downlink 
-policy. The default downlink policy [AllHighestDownlinkPolicy](https://aws.github.io/amazon-chime-sdk-js/classes/allhighestvideobandwidthpolicy.html) just subscribes to the highest video stream 
-available and does not choose optimal simulcast layer according to bandwidth condition.
+The [VideoAdaptiveProbePolicy](https://aws.github.io/amazon-chime-sdk-js/classes/videoadaptiveprobepolicy.html) downlink policy adaptively subscribes to the best simulcast layer and is automatically selected if [MeetingSessionConfiguration.enableUnifiedPlanForChromiumBasedBrowsers](https://aws.github.io/amazon-chime-sdk-js/classes/meetingsessionconfiguration.html#enableunifiedplanforchromiumbasedbrowsers) and [MeetingSessionConfiguration.enableSimulcastForUnifiedPlanChromiumBasedBrowsers](https://aws.github.io/amazon-chime-sdk-js/classes/meetingsessionconfiguration.html#enablesimulcastforunifiedplanchromiumbasedbrowsers) are set to true.
```
richnew10, line 8: Ah, in the meeting session configuration. Sure, yes, that makes sense. But add here "and no other uplink policy has been selected". And did you really want to make that breaking naming change?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -484,27 +486,40 @@ export default class DefaultAudioVideoController
     this.meetingSessionContext.enableSimulcast = this.enableSimulcast;
 
     if (this.enableSimulcast) {
-      const simulcastPolicy = new DefaultSimulcastUplinkPolicy(
-        this.configuration.credentials.attendeeId,
-        this.meetingSessionContext.logger
-      );
+      let simulcastPolicy = this.meetingSessionContext
+        .videoUplinkBandwidthPolicy as SimulcastUplinkPolicy;
```
richnew10, line 490: Just ignore the flag if they specified a policy. Don't overrule the customer's intent.

richnew10 (changes requested):

src/audiovideocontroller/DefaultAudioVideoController.ts
ltrung: We already handle that in DefaultMeetingSession.ts right? I set the flag if they pass in their own simulcast policy.

src/meetingsession/DefaultMeetingSession.ts
ltrung: I do not override the policy in this case right. This case is when they pass in a non-simulcast policy but has that simulcast flag on then we will clear out the simulcast flag.

guides/05_Simulcast.md
ltrung: Sorry what changes here is the breaking name change? :D

src/meetingsession/DefaultMeetingSession.ts
ltrung: Opt to throw error now and no longer remove policy :)

src/meetingsession/DefaultMeetingSession.ts
ltrung: Update the code to throw error when passing invalid values.

hensmi-amazon (approved):

richnew10 (approved):

PR #1617, opened 2021-09-24
Prevent crash in ReceiveVideoInputTask when there is no video track

ltrung:
**Description of changes:**
Add safeguard in `ReceivedVideoInputTask` to prevent crashing when video input stream does not contain any video track.
For example, this can happen when there is a race condition between `StartLocalVideoTile` and `StopLocalVideoTile`.

**Testing**

- In meeting demo, turn on and off video filter in rapid succession.
- Verify that there is no error

### DIFF
--- a/src/task/ReceiveVideoInputTask.ts
+++ b/src/task/ReceiveVideoInputTask.ts
@@ -56,6 +56,13 @@ export default class ReceiveVideoInputTask extends BaseTask {
     this.context.activeVideoInput = videoInput;
     if (videoInput) {
       const videoTracks = videoInput.getVideoTracks();
+      // There can be a race condition when there are several audioVideo.update calls (e.g., calling
+      // startLocalVideoTile and stopLocalVideoTile at the same time)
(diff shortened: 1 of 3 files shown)


michhyun1 (dismissed):

src/task/ReceiveVideoInputTask.ts
```diff
@@ -56,6 +56,9 @@ export default class ReceiveVideoInputTask extends BaseTask {
     this.context.activeVideoInput = videoInput;
     if (videoInput) {
       const videoTracks = videoInput.getVideoTracks();
+      if (!videoTracks || videoTracks.length === 0) {
```
hensmi-amazon, line 59: Can you mention the race condition here in a comment, since its kinda an unusual check (and most people don't like to `git blame` -> read PRs :P.  Also if that race condition occurs we will eventually recover right?  even if we no-opped here.

michhyun1 (approved):

PR #1624, opened 2021-09-28
Fix release script to add missing captureOutputPrefix param for SDK demo app

anuranduttaroy:
**Issue #:** NA

**Description of changes:**
Fix release script to add missing captureOutputPrefix param for SDK demo app. This was causing the demo app deployed for the release testing to not have the `captureOutputPrefix` and resulted in not creating S3 buckets and hence media capture did not start.

**Testing**

Yes.

This can be verified before the next release.

No. This is just a change in the release script. 

NA

NA


xuesichao (approved):
Thanks for your fixing!

PR #1625, opened 2021-09-28
Don't try to postMessage WebAssembly modules in Chrome 95 or later.

richnew10:
Chrome after 95 refuses to send WebAssembly modules between origins.

Instead, fall back to sending a buffer of data, which we can compile in
the destination origin.

I am following up with an additional layer of defense in the worker itself that can apply to older SDK releases.

https://www.chromestatus.com/feature/5650158039597056
https://developer.chrome.com/blog/wasm-module-sharing-restricted-to-same-origin/

---

**Testing**

No. This is a libs-only change.

Upstream. We will test this in the demo app at some point.

Yes: make sure that Voice Focus works in the demo app in Chrome 95 or later.

No.

Not really. This makes Chromium behave like Safari when messaging the Voice Focus worker.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,16 +4,27 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
 ## [Unreleased]
 
+### Added
+
+### Removed
+
 ### Fixed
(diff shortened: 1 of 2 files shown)


ltrung:
Can we add testing instruction here?

richnew10:
> Can we add testing instruction here?

It's literally "make sure other people can hear you with VF turned on".

ltrung:
> > Can we add testing instruction here?
> 
> It's literally "make sure other people can hear you with VF turned on".

Sorry. I totally missed your testing instruction :p

richnew10:
Rebased.

ltrung (approved):

ltrung (approved):

PR #1626, opened 2021-09-28
Fix issue when creating a Video Transform device in rapid succession.

michhyun1:
**Issue #:**
DOMException: The play() request was interrupted by a new load request. https://goo.gl/LdLk22

We can just drop this error thrown and allow the subsequent play() request to take over.

**Description of changes:**

Warp the play() method in a then/catch and drop the error - it means we already have another load request before the previous play() request comes in. 

**Testing**

Test in the demo app - rapidly turn on and off video filters. Verify that the video filters are working as expected.

JS SDK demo and React demo tested.

no

no


ltrung (dismissed):

ltrung (approved):

src/videoframeprocessor/DefaultVideoFrameProcessorPipeline.ts
```diff
@@ -149,7 +149,13 @@ export default class DefaultVideoFrameProcessorPipeline implements VideoFramePro
     this.sourceBuffers.push(canvasBuffer);
 
     this.videoInput.load();
-    await this.videoInput.play();
+    this.videoInput.play()
+      .then(() => {
+        this.logger.info('video element started playing.');
+      })
+      .catch(() => {
+        this.logger.warn('Video element play() overrided by another load().');
```
ldai1, line 157: Is this a overridden action or an interrupted action caused by another load?

ldai1:
There are conflicts in changelog.

src/videoframeprocessor/DefaultVideoFrameProcessorPipeline.ts
michhyun1: Another call to load() is being called before the previous play() could finish - we should just let the new load() command take over

ltrung (approved):

PR #1629, opened 2021-09-29
Remove extended debugging capability to reduce timeouts in saucelabs

michhyun1:
**Issue #:**

**Description of changes:**
According to saucelabs, this capability could "generate a lot of additional assets, which could be the cause for this intermittent issue."

**Testing**

2. How did you test these changes? this is an optional capability

### DIFF
--- a/integration/js/utils/WebdriverSauceLabs.js
+++ b/integration/js/utils/WebdriverSauceLabs.js
@@ -104,9 +104,10 @@ const getSauceLabsConfig = (capabilities) => {
     tags: [capabilities.name],
     seleniumVersion: '3.141.59',
     tunnelIdentifier: process.env.JOB_ID,
-    ...(capabilities.platform.toUpperCase() !== 'LINUX' && {
-      extendedDebugging: true
-    })
+    ...(capabilities.platform.toUpperCase() !== 'LINUX' && 
+      (capabilities.name).includes('ContentShare') && {
+        extendedDebugging: true
+      }),
   }
 };
 
--- a/integration/js/ContentShareOnlyAllowTwoTest.js
+++ b/integration/js/ContentShareOnlyAllowTwoTest.js
@@ -56,7 +56,7 @@ class ContentShareOnlyAllowTwoTest extends SdkBaseTest {
     await test_window_1.runCommands(async () => await ContentShareVideoCheck.executeStep(this, session, "ON", test_attendee_id_2));
     await test_window_3.runCommands(async () => await RosterCheck.executeStep(this, session, 5));
     await test_window_3.runCommands(async () => await ContentShareVideoCheck.executeStep(this, session, "ON", test_attendee_id_1));
-    await test_window_3.runCommands(async () => await ContentShareVideoCheck.executeStep(this, session, "ON", test_attendee_id_1));
+    await test_window_3.runCommands(async () => await ContentShareVideoCheck.executeStep(this, session, "ON", test_attendee_id_2));
 
     //Turn on Content Share for third participant
     await test_window_3.runCommands(async () => await ClickContentShareButton.executeStep(this, session, "ON"));


integration/js/utils/WebdriverSauceLabs.js
```diff
@@ -104,9 +104,7 @@ const getSauceLabsConfig = (capabilities) => {
     tags: [capabilities.name],
     seleniumVersion: '3.141.59',
     tunnelIdentifier: process.env.JOB_ID,
-    ...(capabilities.platform.toUpperCase() !== 'LINUX' && {
-      extendedDebugging: true
-    })
+    ...(capabilities.platform.toUpperCase() !== 'LINUX')
```
devalevenkatesh, line 107: You can remove this as well, I think the `extendedDebugging` was in place for NON linux platform hence the check.

ltrung: Last time we do this it does not solve the issue. Since this only limited to content share now can we only disable it for content share to see if it does any better?

integration/js/utils/SdkBaseTest.js
```diff
@@ -14,6 +14,7 @@ class SdkBaseTest extends KiteBaseTest {
     this.baseUrl = this.url;
     if (testName === 'ContentShareOnlyAllowTwoTest') {
       this.url = this.getTransformedURL(this.url, 'max-content-share', 'true');
+      this.capabilities['extendedDebugging'] = false;
```
devalevenkatesh, line 17: Checking if this should be done for all or at least `ContentShare` tests as all are failing or do you plan to only test `ContentShareOnlyAllowTwoTest`?

LeviSklaroff: We might wanna disable all ContentShare tests, it could be difficult to confirm if it actually fixes the bug if only 'ContentShareOnlyAllowTwoTest' is disabled rather then all of them.

michhyun1: ah, my bad- it should be for all.

integration/js/utils/SdkBaseTest.js
```diff
@@ -15,6 +15,9 @@ class SdkBaseTest extends KiteBaseTest {
     if (testName === 'ContentShareOnlyAllowTwoTest') {
       this.url = this.getTransformedURL(this.url, 'max-content-share', 'true');
     }
+    if (['ContentShare'].includes(testName)) {
```
devalevenkatesh, line 18: Should this be the other way?
`testName.includes('ContentShare')`

devalevenkatesh: Our testName can be 'ContentShareScreenCapture', 'ContentShareOnlyAllowTwoTest' or 'MeetingLeaveContentShareTest', so this check would not pass right?

integration/js/utils/SdkBaseTest.js
```diff
@@ -15,6 +15,9 @@ class SdkBaseTest extends KiteBaseTest {
     if (testName === 'ContentShareOnlyAllowTwoTest') {
       this.url = this.getTransformedURL(this.url, 'max-content-share', 'true');
     }
+    if (testName.includes('ContentShare')) {
+      this.capabilities['extendedDebugging'] = false;
+    }
```
ltrung, line 20: I don't think this worked as stated previously. This is to set the capabilities in the web driver where as extend-debugging is SauceLabs option

integration/js/utils/WebdriverSauceLabs.js
```diff
@@ -104,9 +104,10 @@ const getSauceLabsConfig = (capabilities) => {
     tags: [capabilities.name],
     seleniumVersion: '3.141.59',
     tunnelIdentifier: process.env.JOB_ID,
-    ...(capabilities.platform.toUpperCase() !== 'LINUX' && {
-      extendedDebugging: true
-    })
+    ...(capabilities.platform.toUpperCase() !== 'LINUX' && 
+      capabilities.extendedDebugging != false && {
```
devalevenkatesh, line 108: Checking if this can be kept as earlier. I think this is the first and starting place where we set the extendedDebugging to true so I am not sure if we have `capabilities.extendedDebugging` as valid. You can keep this as is I guess and the change in SDKBaseTest should be fine to override right?

devalevenkatesh: Request testing it once locally.

devalevenkatesh (approved):

PR #1634, opened 2021-09-30
Transcribe data model update

akashuc:
**Issue #:**

**Description of changes:**  Update transcribe data model on CHIME-JS-SDK. The changes include introducing PII entities and stabilization factor. This change also include adding fields to the protobuf model and adding decoding logic for the entity received.

**Testing:**

Yes MeetingV2 can be used to test this changes as well regression testing. The changes have been tested in gamma using the demo app.

Regression can be performed on the changes: 
1. Use meetingV2 demo app
2. Start and join the meeting
3. Enable Live Transcription 
4. Transcript should start coming as earlier.

**Checklist:**

2. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? no


src/screensignalingprotocol/ScreenSignalingProtocol.js
```diff
@@ -0,0 +1,199 @@
+/*eslint-disable block-scoped-var, id-length, no-control-regex, no-magic-numbers, no-prototype-builtins, no-redeclare, no-shadow, no-var, sort-vars*/
```
ltrung, line 1: I think this is generated by mistake (same as d.ts) as it is for a removed feature.

src/transcript/TranscriptAlternative.ts
```diff
@@ -1,9 +1,11 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import TranscriptEntity from './TranscriptEntity';
 import TranscriptItem from './TranscriptItem';
 
 export default class TranscriptAlternative {
   items: TranscriptItem[];
   transcript: string;
+  entities: TranscriptEntity[];
```
ltrung, line 10: Should this be optional?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add API `isSimulcastSupported` so applications can check whether simulcast can be enabled and pass corresponding
   policy.
+- Transcript Data Model Update for re:Invent.
```
ltrung, line 13: Instead of for re:Invent, can we state what the purpose of this feature is?

src/screensignalingprotocol/ScreenSignalingProtocol.js
haifengx: +1 can you uncommit the change?

src/screensignalingprotocol/ScreenSignalingProtocol.d.ts
```diff
@@ -0,0 +1,91 @@
+import * as Long from "long";
```
haifengx, line 1: uncommit changes to this file as well.

src/transcript/TranscriptEntity.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export default class TranscriptEntity {
```
haifengx, line 4: Do we need to update demo app to demonstrate how to use the additional data?

akashuc: Yes, the demo app has the changes and it has been tested locally. But since demo app needs AWS SDK to start the transcription, the changes has been tested using curl command since AWS SDK work is still in progress. The demo app can be used to demonstrate the PII/PHI data that are identified using some color coding and seeing the redacted values. The confidence demonstration is still under discussion.

haifengx: >  The demo app can be used to demonstrate the PII/PHI data that are identified using some color coding and seeing the redacted values.

Are we going to have a separate CR with these changes or should we include them in this CR?

src/screensignalingprotocol/ScreenSignalingProtocol.js
akashuc: Done!!

src/screensignalingprotocol/ScreenSignalingProtocol.d.ts
akashuc: Done!!

protocol/SignalingProtocol.proto
```diff
@@ -304,11 +304,23 @@ message SdkTranscriptItem {
...
+    optional bool stable = 9;
+}
+
+message SdkTranscriptEntity {
+    optional string category = 1;
+    optional double confidence = 2;
+    optional string content = 3;
+    optional int64 end_time = 4;
+    optional int64 start_time = 5;
+    optional string type = 6;
```
simmkyu, line 317: The `type` field in the `SdkTranscriptItem` message indicates the internal enum `Type`. Recommend using an enum here too if the `type` field is not arbitrary.

src/transcript/TranscriptEntity.ts
```diff
@@ -0,0 +1,11 @@
...
+  category: string;
+  confidence: number;
+  content: string;
+  endTime: number;
+  startTime: number;
+  type?: string;
```
simmkyu, line 10: Why is this only optional?

src/transcript/TranscriptEvent.ts
```diff
@@ -95,6 +96,14 @@ export class TranscriptEventConverter {
                 transcriptItem.vocabularyFilterMatch = item.vocabularyFilterMatch;
               }
 
+              if (item.stable) {
```
simmkyu, line 99: `transcriptItem.stable` will be `undefined` if `item.stable` is `false`.

You should do one of the following to assign the `false` value:

```
if (typeof item.stable === 'boolean') {
```

or

```
if (typeof item.stable !== 'undefined') {
```

src/transcript/TranscriptEvent.ts
```diff
@@ -95,6 +96,14 @@ export class TranscriptEventConverter {
...
+                transcriptItem.stable = item.stable;
+              }
+
+              if (item.confidence) {
```
simmkyu, line 103: Not sure what this string `item.confidence` can be. But if `item.confidence` is an empty string, `transcriptItem.confidence` remains `undefined`.

In JavaScript:

```
if ('') {
   // Not executed
}
```

src/transcript/TranscriptEvent.ts
```diff
@@ -107,6 +116,24 @@ export class TranscriptEventConverter {
               transcriptAlternative.items.push(transcriptItem);
             }
 
+            for (const entity of alternative.entities) {
+              if (!transcriptAlternative.entities) {
+                transcriptAlternative.entities = [];
```
simmkyu, line 121: How about setting `entities: []` when you create a TranscriptAlternative object at line 78?

src/transcript/TranscriptEvent.ts
```diff
@@ -107,6 +116,24 @@ export class TranscriptEventConverter {
...
+              }
+              const transcriptEntity: TranscriptEntity = {
+                category: entity.category,
+                confidence: entity.confidence,
+                content: entity.content,
+                startTime: entity.startTime as number,
+                endTime: entity.endTime as number,
+              };
+
+              if (entity.type) {
```
simmkyu, line 131: Ditto - an empty string will be ignored. Using the `Type` enum will help.

test/transcript/TranscriptEventTestDataHelper.ts
```diff
@@ -26,7 +27,7 @@ export const TRANSCRIPT_EVENT_TEST_VECTORS = {
     'CisKKQgBENKF2MwEGgt0ZXN0LXJlZ2lvbiISdGVzdC1jb25maWd1cmF0aW9uCisKKQgCENKF2MwEGgt0ZXN0LXJlZ2lvbiISdGVzdC1jb25maWd1cmF0aW9uCisKKQgDENKF2MwEGgt0ZXN0LXJlZ2lvbiISdGVzdC1jb25maWd1cmF0aW9uCisKKQgEENKF2MwEGgt0ZXN0LXJlZ2lvbiISdGVzdC1jb25maWd1cmF0aW9uCisKKQgFENKF2MwEGgt0ZXN0LXJlZ2lvbiISdGVzdC1jb25maWd1cmF0aW9u'
   ),
   TRANSCRIPT_SINGLE: decode(
-    'Cp0BEpoBCpcBCpQBCkcKBFRlc3QQ+eS4xbAvGhNzcGVha2VyLWF0dGVuZGVlLWlkIhhzcGVha2VyLWV4dGVybmFsLXVzZXItaWQoo924xbAvMAE4AQpCCgEuEIvluMWwLxoTc3BlYWtlci1hdHRlbmRlZS1pZCIYc3BlYWtlci1leHRlcm5hbC11c2VyLWlkKIrluMWwLzACEgVUZXN0Lg=='
+    'CpUCEpICCo8CCowCCkwKBFRlc3QQ++uqzMQvGhNzcGVha2VyLWF0dGVuZGVlLWlkIhhzcGVha2VyLWV4dGVybmFsLXVzZXItaWQopeSqzMQvMAE4AUIBMUgBCkIKAS4QjeyqzMQvGhNzcGVha2VyLWF0dGVuZGVlLWlkIhhzcGVha2VyLWV4dGVybmFsLXVzZXItaWQojOyqzMQvMAISBVRlc3QuGjwKA1BJSREAAAAAAADwPxoVQ29udGVudCBpcyBhIFBJSSBkYXRhII3sqszELyiO7KrMxC8yB0FkZHJlc3MaMwoDUElJEQAAAAAAAPA/GhVDb250ZW50IGlzIGEgUElJIGRhdGEgjeyqzMQvKI7sqszELw=='
```
simmkyu, line 30: Backward compatibility is a key in this PR. I recommend adding more data and tests instead of modifying the existing ones.

simmkyu (changes requested):

src/transcript/TranscriptEntity.ts
akashuc: We will include the demo app changes in the same PR.

protocol/SignalingProtocol.proto
```diff
@@ -304,11 +304,23 @@ message SdkTranscriptItem {
     }
     optional Type type = 6;
     optional bool vocabulary_filter_match = 7;
+    optional string confidence = 8;
```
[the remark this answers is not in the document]
akashuc: Nice catch!! The server side is having the same proto definition which was copied to the JS-SDK. Noah is working to change this and I will incorporate the change once he is done. It should be double on both item and entity.

protocol/SignalingProtocol.proto
akashuc: The type field in entity describes the type of PII data which is identified. For instance, type could be ADDRESS or PHONE or EMAIL or Credit/debit number. It has a wide range of values which could change periodically on the transcribe service. Considering this it would be a breaking change if we add this as enum. If any type values are removed or added to the transcribe entity type in the transcribe service would mean to have that change reflect in the client model as well. Also, the way the PII types are being sent to the transcribe api during start transcription is in form of list of strings so we have kept the type in the incoming stream as string only.

Please let me know if you have any suggestion to this.

src/transcript/TranscriptEntity.ts
akashuc: The type filed is only present in vanilla transcribe and not in medical transcribe. Rest of the fields are present for both cases. That is why type is only kept as optional.

src/transcript/TranscriptEvent.ts
akashuc: Right. I will make the necessary changes for this. Thanks!!

src/transcript/TranscriptEvent.ts
akashuc: I thought about that. But setting entities to line 78 would mean even if the entities are not present, still the alternative object will have an empty entity array. My thought was that if the builder has not requested for content identification or content redaction, the entities are not present in the alternative object that comes from server. Keeping this is mind I added it here so that the TranscriptAlternative will not have entity if it is not asked for.

Please let me know if you have some input on this.

test/transcript/TranscriptEventTestDataHelper.ts
akashuc: Sure I can do that. Definitely would be better that way.

protocol/SignalingProtocol.proto
simmkyu: Thanks for the explanation. If the `type` frequently changes in Amazon Transcribe, the `string` type makes sense.

src/transcript/TranscriptEvent.ts
simmkyu: Gotcha. `entities` will be an empty array if `alternative.entities` has at least one item.

protocol/SignalingProtocol.proto
```diff
@@ -304,11 +304,23 @@ message SdkTranscriptItem {
...
+    optional bool stable = 9;
+}
+
+message SdkTranscriptEntity {
+    optional string category = 1;
+    optional double confidence = 2;
```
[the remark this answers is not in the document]
akashuc: Yes. Keeping it aligned with the protobuf on the tincan side. If precision is the concern here, we don't really want a very high precision for this.

src/transcript/TranscriptEntity.ts
akashuc: The demo app has been included in this PR

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1211,30 +1221,96 @@ export class DemoMeetingApp
...
+        document.getElementById('engine-transcribe-partial-stabilization').classList.toggle('hidden', !engineTranscribeChecked);
+        document.getElementById('engine-transcribe-language-model').classList.toggle('hidden', !engineTranscribeChecked);
       });
     });
 
+    const contentidentificationCb = document.getElementById('contentidentification') as HTMLInputElement;
+    contentidentificationCb.addEventListener('click', () => {
+      (document.getElementById('contentredaction') as HTMLInputElement).disabled = contentidentificationCb.checked;
+      (document.getElementById('transcribe-entity-types') as HTMLInputElement).classList.toggle('hidden', !contentidentificationCb.checked);
+    })
```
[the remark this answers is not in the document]
akashuc: Done for similar comments as well!!

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1211,30 +1221,96 @@ export class DemoMeetingApp
...
         document.getElementById('engine-transcribe-medical-region').classList.toggle('hidden', engineTranscribeChecked);
+        document.getElementById('engine-transcribe-medical-content-identification').classList.toggle('hidden', engineTranscribeChecked);
+        document.getElementById('engine-transcribe-content-identification').classList.toggle('hidden', !engineTranscribeChecked);
+        document.getElementById('engine-transcribe-redaction').classList.toggle('hidden', !engineTranscribeChecked);
+        document.getElementById('engine-transcribe-partial-stabilization').classList.toggle('hidden', !engineTranscribeChecked);
+        document.getElementById('engine-transcribe-language-model').classList.toggle('hidden', !engineTranscribeChecked);
       });
     });
 
+    const contentidentificationCb = document.getElementById('contentidentification') as HTMLInputElement;
```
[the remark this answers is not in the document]
akashuc: Made the change to use camelCase and dash-case

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1211,30 +1221,96 @@ export class DemoMeetingApp
...
+      let partialStabilityFactor = '';
+      let languageModel = '';
+      const transcriptEntities:TranscriptEntity = {};
       if ((document.getElementById('engine-transcribe') as HTMLInputElement).checked) {
         engine = 'transcribe';
         languageCode = (document.getElementById('transcribe-language') as HTMLInputElement).value;
         region = (document.getElementById('transcribe-region') as HTMLInputElement).value;
+        contentIdentification = (document.getElementById('contentidentification') as HTMLInputElement).checked;
+        if (contentIdentification) {
+          transcriptEntities.contentIdentificationType = "PII";
```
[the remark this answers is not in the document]
akashuc: Done!!

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -423,6 +423,15 @@ interface TranscriptSegment {
   endTimeMs: number;
 }
 
+interface TranscriptEntity {
```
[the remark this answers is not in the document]
akashuc: Good idea to have a different name. Changed the name.

src/transcript/TranscriptEvent.ts
```diff
@@ -95,6 +96,14 @@ export class TranscriptEventConverter {
                 transcriptItem.vocabularyFilterMatch = item.vocabularyFilterMatch;
               }
 
+              if (typeof item.stable === 'boolean' && item.hasOwnProperty('stable')) {
```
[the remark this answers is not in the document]
akashuc: So it looks like the item.stable is never undefined. If stable field is not present in the data flowing in, then also referring item.stable returns false. So I am removing tem.stable check and just keeping hasOwnProperty check.

protocol/SignalingProtocol.proto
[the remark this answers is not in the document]
haifengx: Just for my understanding, are we keeping this as string type or going to change to double type?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for layers allocation negotiation in Chromium based browsers to avoid resubscribing to preemptively turn off simulcast streams or to switch layers.
 - Update browser compatibility doc for background blur
 - Add a doc to guide builders on managing video quality for different video layouts. See [guide](https://aws.github.io/amazon-chime-sdk-js/modules/videolayout.html).
+- Add ContentIdentification, ContentRedaction and transcript stabilization support for Amazon Transcribe and Amazon Transcribe Medical for live transcription.
```
haifengx, line 15: nitpick: Add PII identification and redaction, partial result stabilization support from Amazon Transcribe and Amazon Transcribe Medical for live transcription.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -423,6 +423,15 @@ interface TranscriptSegment {
   endTimeMs: number;
 }
 
+interface TranscriptEntityModel {
+  contentIdentificationType?: string;
+  contentRedactionType?: string;
+  enablePartialResultsStability?: boolean;
+  partialStabilityFactor?: string;
+  entityType?: string;
+  languageModel?: string;
```
haifengx, line 432: Are `enablePartialResultsStability`, `partialStabilityFactor`, `languageModel` related to TranscriptEntity`? Not sure whether we need to define the interface here.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2142,7 +2238,7 @@ export class DemoMeetingApp
           this.appendNewSpeakerTranscriptDiv(segment, speakerToTranscriptSpanMap);
         } else {
           const transcriptSpan = speakerToTranscriptSpanMap.get(newSpeakerId);
-          transcriptSpan.innerText = transcriptSpan.innerText + '\u00a0' + segment.content;
+          transcriptSpan.innerHTML = transcriptSpan.innerHTML + '\u00a0' + segment.content;
```
haifengx, line 2241: Any reason we need to set `innerHTML` here instead?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2153,6 +2249,14 @@ export class DemoMeetingApp
     let content = '';
     let attendee: Attendee = null;
     for (const item of result.alternatives[0].items) {
+      let transcriptWord = item.content;
+      if (item.hasOwnProperty('confidence') && !item.content.startsWith("[") && item.confidence < 0.5) {
+        item.content = `<span class='confidence-style'>${item.content}</span>`;
+      }
+      if (this.transcriptEntitySet.size > 0 && this.transcriptEntitySet.has(transcriptWord)) {
+        item.content = `<span class='entity-color'>${item.content}</span>`;
```
haifengx, line 2257: Keep the html elements and styling outside of TS code?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2202,7 +2306,7 @@ export class DemoMeetingApp
 
     const transcriptSpan = document.createElement('span') as HTMLSpanElement;
     transcriptSpan.classList.add('transcript-content');
-    transcriptSpan.innerText = segment.content;
+    transcriptSpan.innerHTML = segment.content;
```
haifengx, line 2309: Same here, any reason we need to set `innerHTML` instead?

demos/browser/server.js
```diff
@@ -190,7 +214,7 @@ function serve(host = '127.0.0.1:8080') {
         respond(response, 200, 'application/json', JSON.stringify(awsCredentials), true);
       } else if (request.method === 'GET' && requestUrl.pathname === '/audio_file') {
         const filePath = 'dist/speech.mp3';
-        fs.readFile(filePath, {encoding: 'base64'}, function(err, data) {
+        fs.readFile(filePath, { encoding: 'base64' }, function (err, data) {
```
haifengx, line 217: nit: unnecessary spaces

demos/browser/server.js
```diff
@@ -44,7 +44,7 @@ function serve(host = '127.0.0.1:8080') {
     log(`${request.method} ${request.url} BEGIN`);
     try {
       // Enable HTTP compression
-      compression({})(request, response, () => {});
+      compression({})(request, response, () => { });
```
haifengx, line 47: nit: unnecessary spaces

protocol/SignalingProtocol.proto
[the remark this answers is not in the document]
akashuc: We are changing it to double. Looks like the generated protocol had double but somehow this file was missed while committing. Made the changes to it.

CHANGELOG.md
akashuc: Made the changes.

demos/browser/app/meetingV2/meetingV2.ts
akashuc: Actually they are not. Should be named something like "TranscriptionStreamParams". It is more a collection of parameters that are provided to start transcription based on the user selection. The main idea for this was to reduce the actual number of parameters being passed to the method startLiveTranscription(https://github.com/aws/amazon-chime-sdk-js/blob/0a6f73b60c2e24e5b46cf42dfc3ffa415efa6719/demos/browser/app/meetingV2/meetingV2.ts#L1092-L1096) . Let me know if "TranscriptionStreamParams" looks ok to you.

src/transcript/TranscriptEvent.ts
```diff
@@ -107,6 +116,24 @@ export class TranscriptEventConverter {
...
+            for (const entity of alternative.entities) {
+              if (!transcriptAlternative.entities) {
+                transcriptAlternative.entities = [];
+              }
+              const transcriptEntity: TranscriptEntity = {
+                category: entity.category,
+                confidence: entity.confidence,
+                content: entity.content,
+                startTimeMs: entity.startTime as number,
+                endTimeMs: entity.endTime as number,
```
richnew10, line 128: Are `entity.startTime` and `endTime` numbers in the input? The `as` won't convert, it's just a TypeScript cast. If these come in as strings, you need `Number.parseInt(x, 10)`.

demos/browser/app/meetingV2/meetingV2.ts
akashuc: The content contains html rich text(some words in content based on whether they are PII identified or low confidence words). Using innerText will output string in plain text something like this: "My name is <span class='entity-color'>Akash</span>" instead of actually displaying the word Akash in color.

demos/browser/app/meetingV2/meetingV2.ts
akashuc: Same as below

demos/browser/app/meetingV2/meetingV2.ts
richnew10: @akashuc are you saying that a segment delivered over this push channel from Transcribe will include raw HTML, which we then inline directly into the page?

akashuc: @richnew10 Not sure if I understood your question here. But Transcribe is sending words in string format. We need to show the PII identified and low confidence words in demo app using color coding and underline respectively. The segment here is formed by concatenating those words that transcribe sends over. In this segment, we need to highlight/underline just those words that fits the criteria mentioned. So once we are forming the segment in demo app, we would be having a mix of plain text as well as html rich text.

richnew10: Oh, I see where in this file you're constructing the spans. You should probably do something a little more sophisticated here.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1918,6 +2014,14 @@ export class DemoMeetingApp
     let content = '';
     let attendee: Attendee = null;
     for (const item of result.alternatives[0].items) {
+      let transcriptWord = item.content;
+      if (item.hasOwnProperty('confidence') && !item.content.startsWith("[") && item.confidence < 0.5) {
+        item.content = `<span class='confidence-style'>${item.content}</span>`;
```
richnew10, line 2019: Ahem:

```
09:32:26.986 const item = { content: 'Jack & Jill' };
09:32:27.006 undefined
09:32:37.134 item.content = `<span class='confidence-style'>${item.content}</span>`;
09:32:37.148 "<span class='confidence-style'>Jack & Jill</span>"
```

You have some HTML escaping concerns here.

Rather than replacing strings in place, changing their 'type' from a string to HTML, you should separate these phases and use DOM operations.

src/transcript/TranscriptEvent.ts
akashuc: startTime is an int64 field which gets converted to number|long|null.
endTime?: (number|Long|null);

demos/browser/app/meetingV2/meetingV2.ts
haifengx: Can you check Amazon Transcribe console implementation. It seems that they are just identifying PII content in finalized results which can simplify the UI implementation. The package name is `WolverineConsole-Assets`.

demos/browser/app/meetingV2/meetingV2.ts
haifengx: Got it. Yeah that sounds more accurate. In that case do we need https://github.com/aws/amazon-chime-sdk-js/pull/1634/files#diff-e870111fa9bc964600df41f9dedbd7396bef0bd816432b2333d0e33da18c0b15R1031-R1035

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1896,6 +1994,7 @@ export class DemoMeetingApp
 
     const speakerToTranscriptSpanMap = new Map<string, HTMLSpanElement>();
     for (const segment of partialTranscriptSegments) {
+      console.log("segment from loop", segment);
```
haifengx, line 1997: remove?

haifengx (approved):
lgtm

haifengx (approved):

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -979,34 +989,111 @@ export class DemoMeetingApp
...
+        if (partialStabilityFactor) {
+          transcriptionStreamParams.partialStabilityFactor = partialStabilityFactor;
+        }
+        if (contentIdentification || contentRedaction) {
+          const selected = document.querySelectorAll('#transcribe-entity option:checked');
+          let values = '';
+          if (selected.length > 0) {
+            values = Array.from(selected).map(el => (el as HTMLInputElement).value).join(',');
+            if (values.startsWith(',')) {
+              values = values.substr(1);
```
richnew10, line 1066: This seems like a tricky way to phrase what you're doing here. Why are you not checking the first element before you `join`?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -979,34 +989,111 @@ export class DemoMeetingApp
...
+            transcriptionStreamParams.entityType = values;
+          }
+        }
+        let isCustomLanguageModelChecked = (document.getElementById('custom-language-model-checkbox') as HTMLInputElement).checked;
+          if (isCustomLanguageModelChecked) {
+            let languageModel = (document.getElementById('language-model-input-text') as HTMLInputElement).value;
+            if (languageModel) {
+              transcriptionStreamParams.languageModel = languageModel;
+          }
+        }
```
richnew10, line 1079: Something has gone wrong with your `{}` pairings here. Please investigate.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -979,34 +989,111 @@ export class DemoMeetingApp
...
       } else if ((document.getElementById('engine-transcribe-medical') as HTMLInputElement).checked) {
         engine = 'transcribe_medical';
         languageCode = (document.getElementById('transcribe-medical-language') as HTMLInputElement).value;
         region = (document.getElementById('transcribe-medical-region') as HTMLInputElement).value;
+        let contentIdentification = (document.getElementById('medical-content-identification-checkbox') as HTMLInputElement).checked;
+        if (contentIdentification) {
+          transcriptionStreamParams.contentIdentificationType = "PHI";
```
richnew10, line 1086: Single quotes, throughout.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1842,7 +1929,17 @@ export class DemoMeetingApp
       for (const result of transcriptEvent.results) {
         const resultId = result.resultId;
         const isPartial = result.isPartial;
-
+        if (!isPartial) {
+          if (result.alternatives[0].entities?.length > 0) {
+              for (const entity of result.alternatives[0].entities) {
+                //split the entity based on space
```
richnew10, line 1935: Indenting. Layout and formatting of comment line.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1842,7 +1929,17 @@ export class DemoMeetingApp
...
+                let contentArray = entity.content.split(" ");
```
richnew10, line 1936: Single quotes, `const`.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1907,45 +2005,57 @@ export class DemoMeetingApp
...
     for (const item of result.alternatives[0].items) {
+      const itemContentSpan = document.createElement('span') as HTMLSpanElement;
+      itemContentSpan.innerText = item.content;
+      itemContentSpan.classList.add('transcript-content');
+      if (item.hasOwnProperty('confidence') && !item.content.startsWith("[") && item.confidence < 0.5) {
+        itemContentSpan.classList.add('confidence-style');
+      }
+      if (this.transcriptEntitySet.size > 0 && this.transcriptEntitySet.has(item.content)) {
+        itemContentSpan.classList.add('entity-color');
+      }
```
richnew10, line 2028: This would really benefit from some examples to show what `item` might be in each case.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1907,45 +2005,57 @@ export class DemoMeetingApp
...
+          contentSpan.appendChild(itemContentSpan);
         }
       }
     }
 
     // Reached end of the result but there is no closing punctuation
     if (startTimeMs) {
       segments.push({
-        content: content,
+        contentSpan: contentSpan,
```
richnew10, line 2058: `contentSpan,` is sufficient.

protocol/SignalingProtocol.proto
```diff
@@ -304,11 +304,23 @@ message SdkTranscriptItem {
     }
     optional Type type = 6;
     optional bool vocabulary_filter_match = 7;
+    optional double confidence = 8;
+    optional bool stable = 9;
+}
+
+message SdkTranscriptEntity {
+    optional string category = 1;
```
richnew10, line 312: Is it possible that an older version of an SDK client receives these messages from the transcription service?

If so, what happens?

src/transcript/TranscriptEvent.ts
richnew10: I think you're still missing the point of my question. `entity` is a JavaScript object. `entity.startTime` is an attribute with some value. If that's always a `number`, you don't need `as number`. If it's not always a `number`, you need some more sophisticated transformation here. So why do you need the `as`?

src/transcript/TranscriptEntity.ts
richnew10, line 4: This whole thing needs a block comment and an example, and you probably need more restricted types than `string`. 

`endTimeMs` seems to be a Unix timestamp in milliseconds. 

`type` seems like it can be `null`, so you need ` | null` in there. It also seems to be a restricted set — maybe

```
'PUNCTUATION' | 'Address' | … | null
```

right?

src/signalingprotocol/SignalingProtocol.d.ts
```diff
@@ -3224,6 +3230,12 @@ export class SdkTranscriptItem implements ISdkTranscriptItem {
     /** SdkTranscriptItem vocabularyFilterMatch. */
     public vocabularyFilterMatch: boolean;
 
+    /** SdkTranscriptItem confidence. */
+    public confidence: number;
```
richnew10, line 3234: These need better descriptions somewhere. Is this 0-1, 0-100, or something else?

src/signalingprotocol/SignalingProtocol.d.ts
```diff
@@ -3192,6 +3192,12 @@ export interface ISdkTranscriptItem {
 
     /** SdkTranscriptItem vocabularyFilterMatch */
     vocabularyFilterMatch?: (boolean|null);
+
+    /** SdkTranscriptItem confidence */
+    confidence?: (number|null);
+
+    /** SdkTranscriptItem stable */
```
richnew10, line 3199: What does this mean?

richnew10 (changes requested):
This needs some thought to be given to the experience of a developer encountering this for the first time. Both developer docs (via block comments and included examples) and TypeScript help need attention.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -232,12 +232,21 @@ interface Toggle {
...
   attendee: Attendee;
   startTimeMs: number;
   endTimeMs: number;
 }
 
+interface TranscriptionStreamParams {
+  contentIdentificationType?: 'PII' | 'PHI';
+  contentRedactionType?: 'PII';
+  enablePartialResultsStability?: boolean;
+  partialStabilityFactor?: string;
```
noahmako, line 245: Please rename to `partialResultsStability` as per the official [Transcribe API](https://docs.aws.amazon.com/transcribe/latest/dg/API_streaming_StartStreamTranscription.html#API_streaming_StartStreamTranscription_RequestSyntax).

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -979,41 +989,117 @@ export class DemoMeetingApp
...
+    languageModelCb.addEventListener('click', () => {
+      (document.getElementById('language-model').classList.toggle('hidden', !languageModelCb.checked));
+    });
+
     const buttonStartTranscription = document.getElementById('button-start-transcription');
     buttonStartTranscription.addEventListener('click', async () => {
       let engine = '';
       let languageCode = '';
       let region = '';
+      const transcriptionStreamParams:TranscriptionStreamParams = {};
```
noahmako, line 1038: Missing space between colon and the type.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -979,41 +989,117 @@ export class DemoMeetingApp
...
+      await startLiveTranscription(engine, languageCode, region, transcriptionStreamParams);
     });
 
-    const startLiveTranscription = async (engine: string, languageCode: string, region: string) => {
-      const response = await fetch(`${DemoMeetingApp.BASE_URL}start_transcription?title=${encodeURIComponent(this.meeting)}&engine=${encodeURIComponent(engine)}&language=${encodeURIComponent(languageCode)}&region=${encodeURIComponent(region)}`, {
+    function isChecked(id: string): boolean {
+      return (document.getElementById(id) as HTMLInputElement).checked;
+    }
+
+    const startLiveTranscription = async (engine: string, languageCode: string, region: string, transcriptionStreamParams : TranscriptionStreamParams) => {
```
noahmako, line 1094: Delete extra space: `transcriptionStreamParams :` -> `transcriptionStreamParams:`

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1907,45 +2004,60 @@ export class DemoMeetingApp
...
-        content = '';
         startTimeMs = null;
         attendee = null;
       } else {
         if (this.noWordSeparatorForTranscription) {
-          content = content + item.content;
+          contentSpan.appendChild(itemContentSpan);
         } else {
-          content = content + ' ' + item.content;
+          itemContentSpan.innerText = ' ' + itemContentSpan.innerText
```
noahmako, line 2051: This will apply the entity/confidence styling to the space preceding the item content. Can we add two spans here, one for just the space without any classes on it and one for the item content?

demos/browser/server.js
```diff
@@ -145,6 +145,9 @@ function serve(host = '127.0.0.1:8080') {
         const languageCode = requestUrl.query.language;
         const region = requestUrl.query.region;
         let transcriptionConfiguration = {};
+        const transcriptionStreamParams = JSON.parse(requestUrl.query.transcriptionStreamParams);
```
noahmako, line 148: As we continue to add params, we should consider switching to use POST body rather than query params for everything.

docs/classes/videoadaptiveprobepolicy.html
```diff
@@ -165,7 +165,7 @@ <h3><span class="tsd-flag ts-flagProtected">Protected</span> default<wbr>Video<w
 					<aside class="tsd-sources">
 						<p>Inherited from <a href="videoprioritybasedpolicy.html">VideoPriorityBasedPolicy</a>.<a href="videoprioritybasedpolicy.html#defaultvideopreferences">defaultVideoPreferences</a></p>
 						<ul>
-							<li>Defined in <a href="https://github.com/aws/amazon-chime-sdk-js/blob/master/src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts#L73">src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts:73</a></li>
+							<li>Defined in <a href="https://github.com/aws/amazon-chime-sdk-js/blob/master/src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts#L74">src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts:74</a></li>
```
noahmako, line 168: Are these changes expected in this diff?

guides/11_Priority_Based_Downlink_Policy.md
```diff
@@ -6,6 +6,9 @@ Amazon Chime SDK for JavaScript allows builders to choose a downlink policy on e
 
 Amazon Chime SDK for JavaScript defines `VideoPriorityBasedPolicy` as the central class providing the ability to request remote video sources to receive and set their respective priorities. The policy ensures bandwidth is reserved for video sources with higher priorities. This policy can be used in conjunction with clients sending simulcast renditions of sources, as well as single stream or mixed environments.
 
+Since this policy uses webRTC's available downlink bandwidth estimation, which has not been supported in Firefox yet,
+the policy should be only used in Chrome and Safari.
```
noahmako, line 10: Are these changes expected in this diff?

demos/browser/app/meetingV2/meetingV2.html
```diff
@@ -354,6 +354,58 @@ <h1 class="h3 mb-3 font-weight-normal text-center">Select devices</h1>
...
+      </div>
+      <div id="engine-transcribe-stability">
+        <div id="engine-transcribe-partial-stabilization" class="custom-control custom-checkbox" style="text-align: left;">
+          <input type="checkbox" id="partial-stabilization-checkbox" class="custom-control-input">
+          <label for="partial-stabilization-checkbox" class="custom-control-label">Enable Partial Results Stabilization</label>
+        </div>
+        <div id="transcribe-partial-stability" class="hidden">
+          <label for="partial-stability" class="sr-only">Transcribe Entities</label>
+          <select id="partial-stability" class="custom-select"  style="width:100%">
+            <option value="" selected>--optional--</option>
```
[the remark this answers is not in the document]
noahmako: Also, should the value be set to something other than `""`? What happens if we actually pass that into the API? Or do we have logic which doesn't pass in the parameter at all if its empty?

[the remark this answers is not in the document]
akashuc: We are not passing anything to the parameter if it is empty.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -232,12 +232,21 @@ interface Toggle {
...
 
 interface TranscriptSegment {
-  content: string;
+  contentSpan: HTMLSpanElement,
   attendee: Attendee;
   startTimeMs: number;
   endTimeMs: number;
 }
 
+interface TranscriptionStreamParams {
```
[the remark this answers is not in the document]
akashuc: Left the existing code intact. We'll pick it up on next PR.

demos/browser/server.js
akashuc: Looks like a good idea. Can club this with combining parameters in a single interface

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -232,12 +232,21 @@ interface Toggle {
...
+  contentIdentificationType?: 'PII' | 'PHI';
+  contentRedactionType?: 'PII';
+  enablePartialResultsStability?: boolean;
+  partialResultsStability?: string;
```
noahmako, line 245: Nit, should `partialResultsStability` also be a literal type, e.g. `'high' | 'medium' | 'low'`?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1907,52 +2005,75 @@ export class DemoMeetingApp
...
-        content: content,
+        contentSpan: contentSpan,
         attendee: attendee,
         startTimeMs: startTimeMs,
         endTimeMs: result.endTimeMs,
       });
     }
   };
 
+  createEmptySpan(): HTMLSpanElement {
```
noahmako, line 2070: nit, might rename this to 'spaceSpan' instead of 'empty' since its actually holding some a whitespace.

script/generate-signaling-protocol
```diff
@@ -26,4 +26,5 @@ end
 
   # workaround for the Long type import issue. Remove when https://github.com/protobufjs/protobuf.js/pull/1166/files is released.
   File.write("src/#{dir}/#{protocol}.d.ts", "import * as Long from \"long\";\n" + File.read("src/#{dir}/#{protocol}.d.ts"))
+  File.write("src/#{dir}/#{protocol}.js", "$util.Long = undefined;\n$protobuf.configure();", mode:"a")
```
noahmako, line 29: Should we add this to the top of the file rather than very bottom, similar to line 28?

akashuc: We cannot add it to the top of the file, We can either add it to the bottom or we have to add it to a specific line once protobuf.util is intialialized. We can try adding it at a specific line if it makes a difference once we are changing the proto types from Long to number.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -979,41 +989,118 @@ export class DemoMeetingApp
...
+        if (partialResultsStability) {
+          transcriptionStreamParams.partialResultsStability = partialResultsStability;
+        }
+        if (isChecked('content-identification-checkbox') || isChecked('content-redaction-checkbox')) {
+          const selected = document.querySelectorAll('#transcribe-entity option:checked');
+          let values = '';
+          if (selected.length > 0) {
+            values = Array.from(selected).filter(node => (node as HTMLInputElement).value !== '').map(el => (el as HTMLInputElement).value).join(',');
+          } 
+          console.log("output", values);
```
noahmako, line 1066: Remove debug logs

noahmako (dismissed):

src/transcript/TranscriptEvent.ts
richnew10: So long as it's guarded by tests, that's fine.

src/transcript/TranscriptEntity.ts
richnew10: Anything that's an _input_ to Transcribe can be a restricted set; we can add to it so long as we document that it can grow.

Anything that's _returned_ from Transcribe and can change in the future needs to be an open value, and also needs to be heavily documented that it can change as Transcribe evolves!

src/transcript/TranscriptEvent.ts
akashuc: Got it. We do have tests for this now.

src/transcript/TranscriptEntity.ts
akashuc: Added extra documentation for entity types explaining different types that we support now.

noahmako (approved):

haifengx (dismissed):

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1907,52 +2004,75 @@ export class DemoMeetingApp
...
+          contentSpan.appendChild(itemContentSpan);
         }
       }
     }
 
     // Reached end of the result but there is no closing punctuation
     if (startTimeMs) {
       segments.push({
-        content: content,
+        contentSpan: contentSpan,
```
richnew10, line 2061: For future reference, you can just phrase this as 

```
  contentSpan,
```

richnew10 (approved):

noahmako (approved):

haifengx (approved):

test/transcript/TranscriptEvent.test.ts
```diff
@@ -227,4 +227,40 @@ describe('TranscriptEvent', () => {
...
+    const dataMessage = makeTranscriptDataMessageFrom(
+      TRANSCRIPT_EVENT_TEST_VECTORS.TRANSCRIPT_ENTITY
+    );
+    const actualEvents = TranscriptEventConverter.from(dataMessage);
+    expect(actualEvents.length).to.equal(1);
+    if (!(actualEvents[0] instanceof Transcript)) {
+      assert.fail();
+      return;
+    }
+    expect(actualEvents[0].results.length).to.eql(1);
```
simmkyu, line 245: nit: I meant you can `expect()` + `eql()` (deep comparison) in one line like the following:

```
expect(actualEvents).to.eql([
  {
     results: [
       {
          alternatives: [
              {
                  items: [
                       { confidence: 1, stable: true },
                       { stable: false },
                  ]
              }
          ]
       }
     ]
  }
]);

```

This helps:
- We can remove Array.length check
- We know an exact shape expected for the transcription data message

simmkyu (approved):

PR #1635, opened 2021-09-30
Update the default priority-based video downlink policy to adjust target size

ltrung:
**Description of changes:**
Update the default priority-based video downlink policy to adjust target size based on number of videos in the meeting.

**Testing:**

Add unit tests to verify the changes.

No, would need a custom simulcast policy that does not downscale the high resolution stream.

**Checklist:**

Yes

No

No


hensmi-amazon:
I'm assuming the answer is yes.  But give our current send parameters in NScale and simulcast policies this won't actually have any effect right?

hensmi-amazon (approved):

ltrung:
> I'm assuming the answer is yes. But give our current send parameters in NScale and simulcast policies this won't actually have any effect right?

Yes target size is only for Simulcast and this is only affect for the testing one I am working one where we always send 720p for high resolution stream. Also I made it so it is compatible with the current default simulcast where we downscale the high resolution stream to 360p after 4 videos.

PR #1652, opened 2021-10-06
Bind tile controller for VideoAdaptiveProbePolicy

ltrung:
**Issue #:**
Currently we only bind the video tile controller to  `VideoPriorityBasedPolicy` ([Code](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/audiovideocontroller/DefaultAudioVideoController.ts#L160)). That means we ignore any policies that extends `VideoPriorityBasedPolicy` such as `VideoAdaptiveProbePolicy` and builders would need to manually bind it.
So if the downlink policy is `VideoAdaptiveProbePolicy` - the current default downlink policy for simulcast, we will log a bunch of `tileController not found!` warnings. In addition, since there is no reference to video tile controller, the policy will not be able to pause video tiles due to bandwidth and will drop videos instead.

**Description of changes:**
- Add `bindToTileController` optional method to `VideoDownlinkBandwidthPolicy` so that we just autobind it for any downlink policy that implements that.
- Also move the code to start since we have default policy logic there which we would not know at the time of construct.

**Testing:**
- Add unit tests.
- Join a meeting in Chrome with Simulcast enabled and have several remote videos.
- Open Developer Console and verify that there is no `tileController not found!` warnings and there should be a `tileController bound` info log.
- Throttle the network to 3G and verify that the video tiles are paused and not dropping.

**Checklist:**

 Yes,

Yes


src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -157,9 +156,6 @@ export default class DefaultAudioVideoController
     this._audioMixController = new DefaultAudioMixController(this._logger);
     this.meetingSessionContext.logger = this._logger;
     this._eventController = new DefaultEventController(this, eventReporter);
-    if (configuration.videoDownlinkBandwidthPolicy instanceof VideoPriorityBasedPolicy) {
```
hensmi-amazon, line 160: Does `instanceof` not work for subclasses?

hensmi-amazon (approved):

src/audiovideocontroller/DefaultAudioVideoController.ts
ltrung: No it works for subclasses too but it did not work if someone implements `VideoDownlinkBandwidthPolicy` and wants to have access to tileController. So to fix this issue with Simulcast, I could just do an `instanceof` check here but I just expose the method to be more general.

PR #1654, opened 2021-10-07
Remove screen signaling protocol generation as it is no longer is use

akashuc:
**Issue #:**

**Description of changes:**
The screen signaling protocol is no longer in use so it should be removed from the generate-signaling-protocol script to avoid auto generating it again whenever the script runs.

**Testing:**

NA
Tested by running the script generate-signaling-protocol locally.

**Checklist:**

yes

No

No

### DIFF
--- a/script/generate-signaling-protocol
+++ b/script/generate-signaling-protocol
@@ -7,22 +7,23 @@ def verbose command
   system(command) || fail("Failed: #{command}")
 end
 
-protocol = 'SignalingProtocol'
-dir = protocol.downcase
+['SignalingProtocol'].each do |protocol|
+  dir = protocol.downcase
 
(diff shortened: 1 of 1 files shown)


script/generate-signaling-protocol
```diff
@@ -7,23 +7,22 @@ def verbose command
   system(command) || fail("Failed: #{command}")
 end
 
-['SignalingProtocol', 'ScreenSignalingProtocol'].each do |protocol|
```
haifengx, line 10: nit: I would just keep the array with a single element in case we need to add other protocol files in the future. But no strong preference.

haifengx (approved):

simmkyu (approved):

PR #1656, opened 2021-10-07
Do not pause sources that do not exist in priority downlink policy

hensmi-amazon:
**Issue #:** None

**Description of changes:** 
See title.  This would lead to a bug where native clients (e.g. mobile SDK in SDK world) enabling/disabling video would cause a pause to be signaled.  When re-enabling video the pause state would be stuck and the receiver would not receive any video.  This only effected native senders because (for better or worse) JS SDK clients change sending stream IDs every enable.

**Testing:**
Reproduced issue using SDK client with non-default preferences set + CLI clients that replicate native clients.  Post fix no longer sends out pause and hides video on re-enabling.

They can at least be sanity checked.  I'm not sure if possible to reproduce using Mobile SDK demos.

**Checklist:**

y

n

n

### DIFF
--- a/src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
+++ b/src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
@@ -752,7 +752,13 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
         ) as DefaultVideoTile;
         const paused = videoTile?.state().paused || false;
         if (!chosenStreams.some(stream => stream.attendeeId === preference.attendeeId)) {
-          if (videoTile) {
+          // We cannot rely on the existance of video tile to indicate that the source exists in the call
+          // because tiles will not be added or removed until after a full renegotiation (i.e. it will
+          // be behind the state provided by the index)
+          const streamExists = remoteInfos.some(
+            stream => stream.attendeeId === preference.attendeeId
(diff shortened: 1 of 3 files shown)


src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -752,7 +752,13 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
...
         const paused = videoTile?.state().paused || false;
         if (!chosenStreams.some(stream => stream.attendeeId === preference.attendeeId)) {
-          if (videoTile) {
+          // We cannot rely on the existance of video tile to indicate that the source exists in the call
+          // because tiles will not be added or removed until after a full renegotiation (i.e. it will
+          // be behind the state provided by the index)
+          const streamExists = remoteInfos.some(
+            stream => stream.attendeeId === preference.attendeeId
+          );
+          if (videoTile && streamExists) {
```
ltrung, line 761: Should we log an info log here that there is no stream exists.

hensmi-amazon: This is technically normal behavior and is hit everytime a video is removed.  Think a log may be overkill since we already log the full index.

XHatan (approved):

PR #1668, opened 2021-10-13
Content-Security-Policy changes

richnew10:
This PR:

* Fixes some errors in the CSP guide, which included untested CSP.
* Extends our CSP to reflect necessary changes to support Amazon Voice Focus in Chrome 95 and later.
* Augments the CSP and Voice Focus guides accordingly.
* Introduces generated CSP for the meeting demo.

Note that the Webpack config is forked to make this work: Webpack by default uses `eval` to load code in development mode, which naturally breaks completely when our CSP is enforced!

**Testing:**

Yes: load the demo app in any browser, check the Web Audio checkbox, and see if Voice Focus appears on the next screen. Check the Voice Focus box and make sure that speech still makes the green indicator fire.

**Checklist:**

No; demo-only change.

> 2. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved?

No.

> 3. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved?

No.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Amazon Voice Focus now works in Chrome 95 or later: WebAssembly policy changes required a change in how modules were loaded. This requires additional Content Security Policy changes, which are documented in the [CSP guide](https://aws.github.io/amazon-chime-sdk-js/modules/contentsecurity_policy.html) and the [Amazon Voice Focus guide](https://aws.github.io/amazon-chime-sdk-js/modules/amazonvoice_focus.html).
 - Add safeguard in `ReceivedVideoInputTask` to prevent crashing when video input stream does not contain any video track.
 - Add missing `captureOutputPrefix` param for SDK demo app in release script.
-- Amazon Voice Focus now works in Chrome 95 or later: WebAssembly policy changes required a change in how modules were loaded.
 - Add opt-in region `eu-south-1` to meetings demo in deploy-canary-demo script to support media capture canary.
 - Fix bug: DOMException: The play() request was interrupted by a new load request. https://goo.gl/LdLk22.
 - Fix `removeObserver` function in `DefaultVideoTransformDevice`.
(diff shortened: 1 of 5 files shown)


richnew10:
> Integration test failed in messaging with error: `Module not found: Error: Can't resolve 'util' in '/home/runner/work/amazon-chime-sdk-js/amazon-chime-sdk-js/demos/browser/node_modules/aws-sdk/lib'`

Yeah, the AWS SDK is causing problems again.

demos/browser/webpack.config.hot.js
```diff
@@ -0,0 +1,107 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * This file is a fork of webpack.config.js. 
+ * My apologies.
```
LeviSklaroff, line 6: We should probably note to make changes in both files when applicable here as well.

demos/browser/webpack.config.js
```diff
@@ -1,20 +1,57 @@
...
+
+  // Script hashes/nonces are not emitted for script-src-elem, so just add unsafe-inline.
+  'script-src-elem': "'self' https://*.sdkassets.chime.aws 'unsafe-inline'",
+  'worker-src': "'self' blob:",
+  'child-src': "'self' blob:",
+};
+
+// Modify our basic CSP to allow several things:
+// 1. Access to gamma assets for testing.
+csp['connect-src'] += ' https://*.sdkassets.gchime.aws';
```
LeviSklaroff, line 35: I don't think it's super necessary, but with the asset pipeline being used more often and gamma being blocked, it might be nice to be able to test with alpha as well.

richnew10: My hope is that this is sufficiently clear that devs can change it to their local test servers or to other CDNs as needed. This should be enough for routine QA.

demos/browser/webpack.config.hot.js
richnew10: Got another run, so done.

LeviSklaroff (approved):

PR #1689, opened 2021-10-18
Renamed appversionname and appversioncode

bijimm:
**Issue #:**

**Description of changes:**

Renamed appVersionName and appVersionCode fields to appName and appVersion respectively

**Testing:**

Locally ran demo

**Checklist:**

Yes

No

No


ltrung:
> @ltrung is this an internal enough change (not exposed, not used) that we can consider it non-breaking?

Yep, we do not use this field for now. Though I think @devalevenkatesh has plan to use it soon :)

devalevenkatesh:
> > @ltrung is this an internal enough change (not exposed, not used) that we can consider it non-breaking?
> 
> Yep, we do not use this field for now. Though I think @devalevenkatesh has plan to use it soon :)

Yes, that is correct, we have those fields but we do not populate it in yet, I will be making some changes once this and backend changes land to start using and sending these fields. (Will be an opt-in feature and nothing should change for builders if they do not want to use it.

PR #1693, opened 2021-10-19
Fixed integration tests failing on mobile

LeviSklaroff:
**Issue #:**
Integration tests do not work for mobile.

**Description of changes:**
- Removed call to missing function
- Changed integration tests to make use of javascript click function over selenium driver click function to fix issue with mobile test clicking at the wrong position.
- Removed testing of old Firefox versions for content share

**Testing:**
- Ran all meeting demo integration tests and confirmed no failures 
- Ran these integration mobile tests:
  - Transcription 
  - Voice Focus
  - VideoSimulcast
  - VideoProcessor
  - Video
  - ReconnectionCheck
  - MeetingLeaveVideoCheck
  - MeetingLeaveAudioCheck
  - MeetingEnd
  - EventReporting
  - DataMessageCheck
  - Audio
  - AppQuitVideoCheck
  - AppQuitAudioCheck
  
**Checklist:**

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -986,6 +986,10 @@ export class DemoMeetingApp
           break;
       }
       try {
+        if (this.chosenVideoTransformDevice) {
+          await this.chosenVideoTransformDevice.stop();
+          this.chosenVideoTransformDevice = null;
+        }
         await this.openVideoInputFromSelection(videoInput.value, true);
       } catch (err) {
         fatal(err);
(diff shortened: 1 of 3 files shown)


ltrung (commented):
There is also meeting_leave_content_share right?

integration/js/pages/AppPage.js
```diff
@@ -784,6 +784,9 @@ class AppPage {
         + "var sum = imageData.reduce(getSum);"
         + "return sum;"
   }
+  async clickElement(element){
+    await this.clickElement( element);
```
ltrung, line 788: Is this correct?
Also should this live in here? This seems like an ulti function that could apply to any page object.

LeviSklaroff: This is wrong, I missed a couple cases of the previous usage and did a replace all and didn't catch that I replaced this. Created a new util file to contain this function and updated.

PR #1738, opened 2021-10-29
Update similar log messages in DefaultMessagingSession and DefaultSig…

nainkunal933:
…nalingClient

**Issue #:**

**Description of changes:** The log messages were same and it would be helpful in debugging if the messages indicate the context under which no connection needs closing.

**Testing:**

**Checklist:**

### DIFF
--- a/src/messagingsession/DefaultMessagingSession.ts
+++ b/src/messagingsession/DefaultMessagingSession.ts
@@ -74,7 +74,7 @@ export default class DefaultMessagingSession implements MessagingSession {
       this.webSocket.close();
       CSPMonitor.removeLogger(this.logger);
     } else {
-      this.logger.info('no existing connection needs closing');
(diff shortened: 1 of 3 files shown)


simmkyu (approved):

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -74,7 +74,7 @@ export default class DefaultMessagingSession implements MessagingSession {
       this.webSocket.close();
       CSPMonitor.removeLogger(this.logger);
     } else {
-      this.logger.info('no existing connection needs closing');
+      this.logger.info('no existing messaging session needs closing');
```
ltrung, line 77: Can we keep the connection word? `messaging session connection` ? messaging session alone is misleading.

PR #1746, opened 2021-11-01
Add background filter option to reduce CPU utilization

davcolle:
https://github.com/aws/amazon-chime-sdk-js/compare/blur-cpu-improvement

**Issue #:**

**Description of changes:**

This change adds an option to background blur to enable reducing the frequency of segmenting the foreground from the background. The `filterCPUUtilization` option will set the threshold of percentage of time spent processing background blur. If the percentage of time processing blur exceeds the set threshold, segmentation will be skipped on some frames. This reduction in segmentations will reduce the CPU utilization with the tradeoff of adding a delay to updating the segmentation between the foreground and background.

```javascript
const processor = await BackgroundBlurVideoFrameProcessor.create(null, {filterCPUUtilization: 50});
```

**Testing:**

Background blur is integrated into the `browser` demo application. To try it out, launch the demo with `npm run start`, join the meeting, click on the camera icon to enable video, then select the filter drop down and select one of the `Background Blur X% CPU` options. You can select 10% - 40% cpu utilization from the drop down.

To manually test changing the `filterCPUUtilization` option go to the `demos/browser/app/meetingV2/meetingV2.ts` file and find the line where `BackgroundBlurVideoFrameProcessor.create` is called and update the code to add the options:

```javascript
this.bbprocessor = await BackgroundBlurVideoFrameProcessor.create(this.getBackgroundBlurSpec(), {filterCPUUtilization: 25});
```

And then run `npm run build` and `npm run start` to run the demo app.

Making this change will cause more skipping of background segmentation. The lower the value of CPU utilization the more likely the blur processor will skip segmentations.

**Checklist:**

Yes

Yes. I updated the `BackgroundBlurOptions` interface to add a new option. No existing fields were modified on the interface. This API change still needs to be reviewed.

No


src/backgroundblurprocessor/BackgroundBlurOptions.ts
```diff
@@ -10,9 +10,19 @@ export default interface BackgroundBlurOptions {
...
-  reportingPeriodMillis?: number;
+  /**
+   * The threshold CPU utilization to trigger skipping background filter updates which will reduce the amount of CPU
+   * used by background filtering. For example, If the reporting period is set to 1000 ms and 500 ms was dedicated to
+   * processing the background filter, then the CPU utilization for that reporting period is 50%. If
+   * {@link filterCPUUtilization} is set to 50 it will cause a `filterCPUUtilizationHigh` event to be fired from the
+   * `BackgroundBlurVideoFrameProcessorObserver`.
+   *
+   */
+  filterCPUUtilization?: number;
```
dimitarz, line 27: What's the unit here? Percent? I.e 0-100 or 0-1.

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
```diff
@@ -139,15 +146,27 @@ export default class BackgroundBlurProcessorProvided implements BackgroundBlurPr
...
     this.frameCounter = new BackgroundFilterFrameCounter(
       this.delegate,
       options.reportingPeriodMillis,
+      options.filterCPUUtilization,
       this.logger
     );
 
+    this.cpuMonitor = new BackgroundFilterMonitor(options.reportingPeriodMillis * 5, {
+      reduceCPUUtilization: () => {
+        this.updateVideoFramesPerFilterUpdate(Math.min(this.videoFramesPerFilterUpdate + 1, 5));
```
dimitarz, line 162: MAX_SKIPPED_SEGMENTATIONS = 5

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
```diff
@@ -139,15 +146,27 @@ export default class BackgroundBlurProcessorProvided implements BackgroundBlurPr
...
     this.setBlurStrength(options.blurStrength);
     this.delegate = new BackgroundBlurVideoFrameProcessorDelegate();
     this.frameCounter = new BackgroundFilterFrameCounter(
       this.delegate,
       options.reportingPeriodMillis,
+      options.filterCPUUtilization,
       this.logger
     );
 
+    this.cpuMonitor = new BackgroundFilterMonitor(options.reportingPeriodMillis * 5, {
```
dimitarz, line 160: why the *5?

src/backgroundblurprocessor/BackgroundBlurOptions.ts
davcolle: I'll update the name and comment to indicate that it is a percentage (0-100)

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
davcolle: I wanted to leave some time to adapt. By default the reporting period is 1 second. I think 5 seconds is a good amount of time to settle in to segmentation frequency change and the reevaluate the performance.

src/backgroundblurprocessor/BackgroundBlurVideoFrameProcessor.ts
```diff
@@ -193,6 +193,10 @@ export default class BackgroundBlurVideoFrameProcessor {
     if (!options.reportingPeriodMillis) {
       options.reportingPeriodMillis = 1000;
     }
+
+    if (!options.filterCPUUtilization) {
+      options.filterCPUUtilization = 50;
+    }
```
dimitarz, line 199: This is quite high isn't it? On average we see a 20-30% usage. Maybe 30%?

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
dimitarz: Ok this can get tricky if somebody sets the reporting period to 5s or 0.5s. Can we just set it to a flat 5s value?

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
ltrung, line 162: nit: Can we create constant for these value? Not sure what 5 or 1 mean

guides/15_Background_Filter_Video_Processor.md
```diff
@@ -116,9 +116,24 @@ let transformDevice = new DefaultVideoTransformDevice(logger, device, processors
...
+  reportingPeriodMillis?: number;
+  blurStrength?: number;
+  filterCPUUtilization?: number;
+}
+```
+
+* logger: A logger to which log output will be written.
+* reportingPeriodMillis: How often the video frame processor will report observed events. 
+* blurStrength: The amount of blur that will be applied to a video stream. 
+* filterCPUUtilization: The threshold CPU utilization to trigger skipping background filter updates which will reduce the amount of CPU used by background filtering. For example, If the reporting period is set to 1000 ms and 500 ms was dedicated to processing the background filter, then the CPU utilization for that reporting period is 50%. If `filterCPUUtilization` is set to 50 it will cause a `filterCPUUtilizationHigh` event to be fired from the `BackgroundBlurVideoFrameProcessorObserver`.
```
ltrung, line 135: Should builders do anything when receiving `filterCPUUtilizationHigh`?

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
```diff
@@ -415,3 +449,33 @@ export default class BackgroundBlurProcessorProvided implements BackgroundBlurPr
...
+    if (timestamp - this.lastEventTimestamp >= this.monitoringPeriodMillis) {
+      this.lastEventTimestamp = timestamp;
+      this.observer.reduceCPUUtilization();
+    }
+  }
+
+  frameReceived(): void {
+    const timestamp = Date.now();
+    // If a enough time has passed, reset the processor and continue to monitor
+    if (timestamp - this.lastEventTimestamp >= this.monitoringPeriodMillis * 2) {
```
ltrung, line 476: So we will always ramp up `videoFramesPerFilterUpdate` after 20s by default regardless of the CPU Utilization at that time?

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
```diff
@@ -303,8 +333,8 @@ export default class BackgroundBlurProcessorProvided implements BackgroundBlurPr
 
       if (this.scaledCanvas === undefined) {
         this.scaledCanvas = document.createElement('canvas');
-        this.scaledCanvas.width = inputCanvas.width * hscale;
-        this.scaledCanvas.height = inputCanvas.height * vscale;
+        this.scaledCanvas.width = this.spec.model.input.width;
```
ltrung, line 336: Is this change related? If not, could we document that in CHANGELOG :)

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
davcolle: will do.

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
davcolle: This will only be called if a high cpu event did not occur in the last 10 seconds. That would mean that the utilization is under the value set in the options and we have some capacity to speed up.

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
davcolle: This is unrelated. I will update in changelog.

guides/15_Background_Filter_Video_Processor.md
davcolle: I will add an example implementation and what action to take if this fires. Mainly you can disable background blur if the utilization is too high.

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
ltrung: Thanks. Also can we add it to the demo also?

guides/15_Background_Filter_Video_Processor.md
```diff
@@ -116,21 +116,40 @@ let transformDevice = new DefaultVideoTransformDevice(logger, device, processors
...
+
 ## Observer notifications
 
 You can optionally implement the `BackgroundBlurVideoFrameProcessorObserver` interface and use `addObserver` to receive callbacks when an event occurs:
 
 * `filterFrameDurationHigh`:  This event occurs when the amount of time it takes to apply the background blur is longer than expected. The measurement is taken from the time the process method starts to when it returns. For example, if the video is running at 15 frames per seconds and we are averaging more than 67 ms (1000 ms reporting period / 15 fps) to apply background blur, then a very large portion of each frame's maximum processing time is taken up by processing background blur.
+* `filterCPUUtilizationHigh`: This event occurs when the CPU utilization of background filtering exceeds the `filterCPUUtilization` defined in the `BackgroundBlurOptions`. This can be used as a trigger to disable background blur if CPU utilization is too high.
```
ltrung, line 142: Should we document that JS SDK also attempts to migrate this?

davcolle: yes. I'll update.

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
ltrung: Can we rename the variable `lastEventTimestamp` to indicate that it is the last timestamp from high cpu utilization event?

ltrung (dismissed):

xuesichao:
@davcolle Could you please resolve the conflict?

ltrung (approved):

PR #1769, opened 2021-11-09
Refactor some video demo components into classes, seperate files

hensmi-amazon:
**Issue #:** None

**Description of changes:** I was scrolling through meetingV2.ts and sprained my finger, so figured it would be helpful to break this up.  I'm not of the opinion that there is too much in this demo, but do think that in it's current form I wouldn't be super comfortable telling builders to just copy paste it as a start.  Its also a bit of a headache if you are just looking to see how one component was integrated.

So as a start this breaks out some video related components into their own classes, and moves them to their own files as well.  I can see the same being done for local video/audio configuration, and the chat messages.

**Testing:**

Smoke tested audio/video send and receive in a 3 attendee call.verified active speaker works.  verified correct stats overlays. verified pause, and video preference setting worked as expected.

There is no feature here.  Regression testing should suffice.

**Checklist:**

y

n

n


demos/browser/package-lock.json
```diff
@@ -1,12076 +1,12 @@
 {
   "name": "amazon-chime-sdk-js-demos",
   "version": "0.1.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
ltrung, line 4: Can you use Node 16+ ?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1957,16 +1717,22 @@ export class DemoMeetingApp
 
     // Hang on to this so we can unsubscribe later.
     this.activeSpeakerHandler = (attendeeIds: string[]): void => {
+      // First reset all roster active speaker information
       for (const attendeeId in this.roster) {
-        this.roster[attendeeId].active = false;
+          this.roster[attendeeId].active = false;
```
ltrung, line 1722: nit: Extra space?

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1135,7 +908,7 @@ export class DemoMeetingApp
           }
         } else {
           this.audioVideo.stopLocalVideoTile();
-          this.tileIndexToDemoVideoTile.get(DemoMeetingApp.LOCAL_VIDEO_TILE_INDEX).hide();
+        //   this.tileIndexToDemoVideoTile.get(DemoMeetingApp.LOCAL_VIDEO_TILE_INDEX).hide(); TODO
```
ltrung, line 911: what is the reason it has to be TODO :D

hensmi-amazon: Oops, this line of code isn't needed.  Switched to a comment.

ltrung (approved):

ltrung (approved):

PR #1783, opened 2021-11-11
Fix priority downlink policy bandwidth estimation metrics to work with Safari

ltrung:
**Issue #:**
The current priority-based downlink policy uses webRTC's available downlink bandwidth estimate to calculate whether it should pause video tiles to reserve bandwidth. However, the current implementation only checks stats from Chrome and not from other browser. 

**Description of changes:**
Update to read stats from Safari. Firefox did not support `availableOutgoingBitrate` at this point so update the guide to clarify that instead.

**Testing:**
Test priority-based downlink policy and check the bandwidth estimate to be correct (previously it is always null).

- Join a meeting with multiple videos in Safari with priority-based downlink policy
- Make sure if the bandwidth is good, it does not pause any video.
- Throttle the network and verify that video will be paused.

**Checklist:**


src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -191,23 +192,23 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
     this.prevDownlinkStats = this.downlinkStats;
     this.downlinkStats = new LinkMediaStats();
     const metricReport = clientMetricReport.getObservableMetrics();
-    this.downlinkStats.bandwidthEstimateKbps = metricReport.availableReceiveBandwidth / 1000;
+    if (!isNaN(metricReport.availableOutgoingBitrate)) {
```
hensmi-amazon, line 195: Comment?

ltrung: Added

hensmi-amazon: perfect

hensmi-amazon (approved):

hensmi-amazon (approved):

Issue #1784, opened 2021-11-11
Simulcast supported browsers list


paul365:
### What are you trying to do?
Hi, I'm using DefaultBrowserBehaviour.isSimulcastSupported() to check whether the browser supports simulcast or not, currently, it's returning false on Chrome Version 95.0.4638.69 (Official Build) (arm64) on M1 Mac, also on Safari Version 15.1 (17612.2.9.1.20). Possible to have a list of supported browsers/devices of simulcast? Or might this be a bug? Thanks

Update: the latest version of the SDK is returning true now for Chrome on M1 Mac. Was using 2.20.1 before.

### How can the documentation be improved to help your use case?
show list of browsers that supports simulcast, currently stating Currently, only Chrome 76 and above is supported.
*For example: Include a section on how to configure the meeting session.*

### What documentation have you looked at so far?
README.md, Amazon Chime SDK Developer Guide, https://aws.github.io/amazon-chime-sdk-js/modules/simulcast.html

devalevenkatesh:
Hi @paul365 ,

The simulcast is currently supported on Chromium based browsers, hence the Safari will return `false` as expected. I will create a PR to add a support table soon.

Please note that we already have the configuring meeting session with Simulcast explained at link below in the same guide:
https://aws.github.io/amazon-chime-sdk-js/modules/simulcast.html#creating-a-simulcast-enabled-meeting

paul365:
thank you @devalevenkatesh

georgezy-amzn:
Closing this due to inactivity. Please reopen or submit a support case if this is still an issue.

PR #1791, opened 2021-11-13
Add simulcast support table

devalevenkatesh:
**Issue #:**
#1784 

**Description of changes:**
- Documentation guide update.

**Testing:**
- Could not test webview but tested Samsung Internet in SauceLabs real device. video shows black but still could see no errors in devtools.

NA

**Checklist:**

NA.

NA.

NA.


guides/05_Simulcast.md
```diff
@@ -18,7 +18,15 @@ Simulcast support is built into the WebRTC library in the majority of browsers.
...
+The SimulcastUplinkPolicy configures RTCRtpSender to have three encoding layers, but only ever enables two of them.  Which two are enabled is based on a variety of factors. Experiments show that configuring RTCRtpSender with three encoding layers up front and dynamically enabling and disabling layers provides a better experience and reduces the burden of having to dynamically  manage capture resolution and encoding parameters.
+
+### Browser compatibility
+
+|Browser              |Supported versions  |
+|---                  |---                 |
+|Chrome               |76+                 |
+
+> Note: Simulcast in Amazon Chime SDK for JavaScript should also work in Chromium-based browsers on desktop like Chromium Edge and Opera; and on mobile like Chromium-Webview and Samsung Internet.
```
hensmi-amazon, line 29: Is it possible to get the actual version numbers of those browsers at the M76 point (maybe we can estimate too).

guides/05_Simulcast.md
```diff
@@ -18,7 +18,15 @@ Simulcast support is built into the WebRTC library in the majority of browsers.
...
 
-The SimulcastUplinkPolicy configures RTCRtpSender to have three encoding layers, but only ever enables two of them.  Which two are enabled is based on a variety of factors.   Experiments show that configuring RTCRtpSender with three encoding layers up front and dynamically enabling and disabling layers provides a better experience and reduces the burden of having to dynamically  manage capture resolution and encoding parameters.
+The SimulcastUplinkPolicy configures RTCRtpSender to have three encoding layers, but only ever enables two of them.  Which two are enabled is based on a variety of factors. Experiments show that configuring RTCRtpSender with three encoding layers up front and dynamically enabling and disabling layers provides a better experience and reduces the burden of having to dynamically  manage capture resolution and encoding parameters.
+
+### Browser compatibility
+
+|Browser              |Supported versions  |
+|---                  |---                 |
+|Chrome               |76+                 |
```
ltrung, line 27: If we want to add a support table, we want a full coverage of what Chromium browser we support (Chrome, Android Chrome, Edge, etc). :)

guides/05_Simulcast.md
richnew10: Our documentation should not routinely say "should work". Either we've tested it and it works (or doesn't), or it should be an uncommon case that most customers won't care about. I would argue that Edge is in the first set, not the second.

devalevenkatesh: Agree, I checked with internal QA around this and tested few but "should work" is not ideal, and will need more time to actually get all those browsers tested as mentioned here with the list supported version thanks! I have created an internal ticket to track this work and add support for Simulcast like we have in VF guide.

devalevenkatesh:
Closing PR and will re-open later as it needs more testing on specific browsers with versions supported by Amazon Chime SDK for JavaScript. Tracking in an internal ticket.

PR #1794, opened 2021-11-16
Update JS SDK Demos to be able to use Chime Meetings regional endpoints.

harshitOberoi:
**Issue #:**
Currently JS SDK Demo just use Chime Meetings Global Endpoint. So using the current setup cannot create sauce lab test to test chime meetings regional endpoints.

**Description of changes:**
Update JS SDK Browser and Serverless Demo to include ChimeMeetingsWrapper using which we can swtich between global and regional endpoints using controlRegion url parameter. 

**Testing:**

yes using both Browser and Serverless Demo.

STEPS:
1) Open Demo Url
2) Add url parameter : ?controlRegion=us-east-1

**Checklist:**

Yes

No

No


demos/browser/ChimeMeetingsWrapper.js
```diff
@@ -0,0 +1,52 @@
...
+const AWS = require('aws-sdk');
+
+module.exports = class ChimeMeetingsWrapper {
+
+    // Create an AWS SDK Chime object. Region 'us-east-1' is currently required.
+    // Use the MediaRegion property below in CreateMeeting to select the region
+    // the meeting is hosted in.
+    
+
+    constructor(currentRegion, controlRegion, endpoint){
```
aws-geoff, line 13: Can you clarify the functional difference between `currentRegion` and `controlRegion`?

demos/browser/ChimeMeetingsWrapper.js
```diff
@@ -0,0 +1,52 @@
...
+        this.chime = new AWS.Chime({ region: currentRegion });
+
+        // Set the AWS SDK Chime endpoint. The global endpoint is https://service.chime.aws.amazon.com
+        this.chime.endpoint = new AWS.Endpoint(endpoint);
+        this.controlRegion = (!!controlRegion) ? currentRegion : controlRegion;
+        console.log("Const this.controlRegion",this.controlRegion);
+        this.chimeMeetings = new AWS.ChimeSDKMeetings({ region: this.controlRegion});
+    }
+
+    getChimeInstance(glabal){
```
aws-geoff, line 24: 1. `global`
2. I wouldn't expect this to be a parameter to be present in each method, I'd expect it to be present in the constructor since we are attempting to abstract away this concept from the caller.

demos/browser/ChimeMeetingsWrapper.js
```diff
@@ -0,0 +1,52 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+const AWS = require('aws-sdk');
+
+module.exports = class ChimeMeetingsWrapper {
```
aws-geoff, line 6: Nit: Consider just `ChimeMeetings`

demos/browser/server.js
```diff
@@ -22,14 +23,16 @@ const indexPage = fs.readFileSync(indexPagePath);
...
+const sts = new AWS.STS({ region: currentRegion });
 
 // Set the AWS SDK Chime endpoint. The global endpoint is https://service.chime.aws.amazon.com.
 const endpoint = process.env.ENDPOINT || 'https://service.chime.aws.amazon.com';
-console.info('Using endpoint', endpoint);
 
-chime.endpoint = new AWS.Endpoint(endpoint);
+//Set useGlobalEndpoint to true to use global endpoint
+//Setting it false will use regional Chime Meetings Endpoints based upon region.
+const useGlobalEndpoint = false;
```
aws-geoff, line 34: The vars `endpoint` and `useGlobalEndpoint` are in semantic conflict since the default endpoint there is the global endpoint, despite `useGlobalEndpoint` is default `false`

demos/browser/server.js
```diff
@@ -22,14 +23,16 @@ const indexPage = fs.readFileSync(indexPagePath);
 // Create ans AWS SDK Chime object. Region 'us-east-1' is currently required.
 // Use the MediaRegion property below in CreateMeeting to select the region
 // the meeting is hosted in.
-const chime = new AWS.Chime({ region: 'us-east-1' });
-const sts = new AWS.STS({ region: 'us-east-1' })
+const currentRegion = 'us-east-1';
```
aws-geoff, line 26: What's the purpose of `currentRegion` here?

demos/serverless/src/ChimeMeetingsWrapper.js
```diff
@@ -0,0 +1,60 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+const AWS = require('aws-sdk');
+
+module.exports = class ChimeMeetingsWrapper {
```
aws-geoff, line 6: This looks like a dupe file.  Is there a way we can have both the browser and the serverless demos reference the same file?

demos/serverless/src/handlers.js
```diff
@@ -61,21 +70,28 @@ exports.join = async(event, context) => {
...
-      ]
     };
-    console.info('Creating new meeting: ' + JSON.stringify(request));
-    meeting = await chime.createMeeting(request).promise();
+
+    if (useGlobalEndpoint) {
+      request = {
+        ...request,
+        // Tags associated with the meeting. They can be used in cost allocation console
+        //Tags are not supported by new ChimeMeetingsSDK
```
aws-geoff, line 79: "Tags are not supported by the new ChimeSDKMeetings namespace"

aws-geoff (commented):
I think this is the right direction, my primary concerns are re:
1. Providing the `global` arg to each method call rather than to only the constructor, if possible I think we want the latter,
2. The functional difference between providing a `currentRegion` and a `controlRegion`,
3. The conflict between the default endpoint and the default `false` for `useGlobalEndpoint`, and
4. The dupe of the ChimeMeetings interface that wraps the Chime and ChimeSDKMeetings namespaces

demos/serverless/src/ChimeMeetingsWrapper.js
harshitOberoi: @aws-geoff Actually we need two separate files as these files need to access aws-sdk from node folders and both the browser and serverless demo has different package.json and thus sepearate node folders.

demos/serverless/src/handlers.js
harshitOberoi: yes but demo currently utilizes Tags for global endpoints so i have used tags based upon controlRegion.

demos/browser/server.js
harshitOberoi: In the Demo for aws.chime it is hardcorded to use 'us-east-1' as the defualt region also in the case where we do not pass in controlRegion we will set currentRegion as controlRegion

demos/browser/ChimeMeetingsWrapper.js
harshitOberoi: Current region is hardcorded to us-east-1 in either handler or server.js and controlRegion is being passed as url params. Also in the case where controlRegion is not passed in we set controlRegion = currentRegion

demos/browser/ChimeMeetings.js
```diff
@@ -0,0 +1,49 @@
...
+    
+
+    constructor(currentRegion, controlRegion, endpoint, useChimeGlobal){
+        console.log("const controlRegion",controlRegion);
+        this.chime = new AWS.Chime({ region: currentRegion });
+
+        // Set the AWS SDK Chime endpoint. The global endpoint is https://service.chime.aws.amazon.com
+        this.chime.endpoint = new AWS.Endpoint(endpoint);
+        this.controlRegion = (!!controlRegion) ? currentRegion : controlRegion;
+        console.log("Const this.controlRegion",this.controlRegion);
```
ltrung, line 20: Please remove console.log :)

demos/browser/ChimeMeetings.js
```diff
@@ -0,0 +1,49 @@
...
+        this.chimeMeetings = new AWS.ChimeSDKMeetings({ region: this.controlRegion});
```
ltrung, line 21: So do we need to test again global SDK or can we just test again the meeting SDK? Its a bit confusing to do it like this since its a demo so if a builder looks at this its confusing :)

demos/browser/server.js
```diff
@@ -22,14 +23,16 @@ const indexPage = fs.readFileSync(indexPagePath);
...
+const sts = new AWS.STS({ region: currentRegion });
 
 // Set the AWS SDK Chime endpoint. The global endpoint is https://service.chime.aws.amazon.com.
 const endpoint = process.env.ENDPOINT || 'https://service.chime.aws.amazon.com';
```
ltrung, line 30: Have we tested with gamma endpoint and non-us-east-1region passed in? Does it work?

demos/serverless/src/ChimeMeetings.js
```diff
@@ -0,0 +1,57 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+const AWS = require('aws-sdk');
+
+module.exports = class ChimeMeetings {
```
ltrung, line 6: Isn;t this the same one as the local server? Can we just have one?

ltrung (commented):
I assume there will be a follow PR to update test to use control region?

demos/browser/ChimeMeetingsWrapper.js
aws-geoff: Why is `currentRegion` necessary though?

demos/browser/server.js
aws-geoff: Can we remove `currentRegion`?

demos/serverless/src/handlers.js
aws-geoff: Sure, I'm saying the comment needs to change: the namespace is "ChimeSDKMeetings"

demos/browser/ChimeMeetings.js
aws-geoff: We want to continue to test the Chime namespace Meeting APIs (since they still exist and are available for public use) but we want it to be the non-default case.

demos/browser/ChimeMeetings.js
```diff
@@ -0,0 +1,49 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+const AWS = require('aws-sdk');
+
+module.exports = class ChimeMeetings {
+
+    // Create an AWS SDK Chime object. Region 'us-east-1' is currently required.
```
aws-geoff, line 8: Why is us-east-1 required?

demos/browser/ChimeMeetings.js
```diff
@@ -0,0 +1,49 @@
...
+    // Use the MediaRegion property below in CreateMeeting to select the region
+    // the meeting is hosted in.
+    
+
+    constructor(currentRegion, controlRegion, endpoint, useChimeGlobal){
```
aws-geoff, line 13: 1. Can we remove the `currentRegion` arg?
2. Consider `useChimeNamespace` instead of `useChimeGlobal` since "global" isn't referenced in any public docs

demos/browser/server.js
```diff
@@ -22,14 +23,16 @@ const indexPage = fs.readFileSync(indexPagePath);
...
-console.info('Using endpoint', endpoint);
 
-chime.endpoint = new AWS.Endpoint(endpoint);
+//Set useChimeGlobal to true to use global endpoint
+//Setting it false will use regional Chime Meetings Endpoints based upon region.
+const useChimeGlobal = false;
```
aws-geoff, line 34: We shouldn't use the term "global" since it's not present in our public docs and will be confusing to customers

demos/serverless/src/handlers.js
```diff
@@ -12,10 +13,14 @@ const ddb = new AWS.DynamoDB();
 // Create an AWS SDK Chime object. Region 'us-east-1' is currently required.
 // Use the MediaRegion property below in CreateMeeting to select the region
 // the meeting is hosted in.
-const chime = new AWS.Chime({ region: 'us-east-1' });
+const currentRegion = 'us-east-1';
+
+//Set useChimeGlobal to true to use global endpoint
```
aws-geoff, line 18: Refrain from using "global"

demos/serverless/src/handlers.js
harshitOberoi: ohh. Got it

demos/browser/server.js
harshitOberoi: no i have not tested the gamma endpoint but i can deploy and test.

demos/browser/ChimeMeetingsWrapper.js
harshitOberoi: updated to remove current region

demos/browser/ChimeMeetingsWrapper.js
harshitOberoi: removed

demos/browser/ChimeMeetingsWrapper.js
harshitOberoi: changed to ChimeMeetings

demos/browser/server.js
harshitOberoi: updated and removed useGlobalEndpoint

demos/serverless/src/handlers.js
harshitOberoi: updated

demos/browser/ChimeMeetings.js
harshitOberoi: removed

demos/browser/server.js
harshitOberoi: Updated the demo to be configurable for gamma endpoints and tested it by deploying it on gamma

demos/serverless/src/ChimeMeetings.js
harshitOberoi: I tried to move the ChimeMeetings file to demo level but it is not working to have a single file we will need to add webpack inside serverless demo which i think is out of scope for this ticket.

demos/browser/ChimeMeetings.js
harshitOberoi: I am not sure this comment was already present in handlers.js

demos/browser/ChimeMeetings.js
harshitOberoi: updated

demos/browser/server.js
harshitOberoi: removed

demos/serverless/src/handlers.js
harshitOberoi: removed

demos/browser/ChimeMeetings.js
dimitarz: This was here before when the SDK was using the global API.

demos/browser/server.js
```diff
@@ -44,11 +40,18 @@ if (captureS3Destination) {
...
-function getClientForMeeting(meeting) {
-  if (meeting && meeting.Meeting && meeting.Meeting.MeetingFeatures && meeting.Meeting.MeetingFeatures.Audio &&
+function getClientForMeeting(meeting, controlRegion) {
+  if (checkValidControlRegion(controlRegion)){
+    return (new AWS.ChimeSDKMeetings({ region: controlRegion }));
+  } 
+  else if (meeting && meeting.Meeting && meeting.Meeting.MeetingFeatures && meeting.Meeting.MeetingFeatures.Audio &&
     meeting.Meeting.MeetingFeatures.Audio.EchoReduction === 'AVAILABLE') {
-      return chimeRegional;
+      return (new AWS.ChimeSDKMeetings({ region: 'us-east-1' }));
```
aws-geoff, line 54: Is `us-east-1` the correct region here?  This looks like it should be
```
return new AWS.ChimeSDKMeetings({ region: controlRegion });
```

demos/serverless/deploy.js
```diff
@@ -12,6 +12,10 @@ let useEventBridge = false;
 let enableTerminationProtection = false;
 let disablePrintingLogs = false;
 let chimeEndpoint = 'https://service.chime.aws.amazon.com';
+let chimeMeetingsIADEndpoint = '-';
+let chimeMeetingsPDXEndpoint = '-';
+let chimeMeetingsFRAEndpoint = '-';
+let chimeMeetingsSINEndpoint = '-';
```
aws-geoff, line 18: Should these defaults be set to their actual values?

demos/browser/server.js
```diff
@@ -26,16 +26,12 @@ const chime = new AWS.Chime({ region: 'us-east-1' });
 // Set the AWS SDK Chime endpoint. The global endpoint is https://service.chime.aws.amazon.com.
```
aws-geoff, line 26: Remove references to "global"

demos/browser/server.js
```diff
@@ -26,16 +26,12 @@ const chime = new AWS.Chime({ region: 'us-east-1' });
...
 const endpoint = process.env.ENDPOINT || 'https://service.chime.aws.amazon.com';
 
-const chimeRegional = new AWS.ChimeSDKMeetings({ region: 'us-east-1' });
-const chimeRegionalEndpoint = process.env.REGIONAL_ENDPOINT || 'https://meetings-chime.us-east-1.amazonaws.com';
 
 const sts = new AWS.STS({ region: 'us-east-1' })
 
 console.info('Using global endpoint', endpoint);
```
aws-geoff, line 32: Remove references to "global"

demos/serverless/src/handlers.js
```diff
@@ -14,16 +14,44 @@ const ddb = new AWS.DynamoDB();
 // the meeting is hosted in.
 const chime = new AWS.Chime({ region: 'us-east-1' });
 
+const endpoint = process.env.CHIME_ENDPOINT;
+const globalEndpoint = "https://service.chime.aws.amazon.com";
```
aws-geoff, line 18: Remove references to "global"

demos/serverless/src/handlers.js
```diff
@@ -14,16 +14,44 @@ const ddb = new AWS.DynamoDB();
...
+chime.endpoint = new AWS.Endpoint(endpoint);
 
-const chimeRegional = new AWS.ChimeSDKMeetings({ region: 'us-east-1' });
+
+const chimeMeetingsRegionalEndPoints = getRegionalEndpoints();
+
+
+function checkValidControlRegion(controlRegion){
+  return (controlRegion == 'us-east-1' || controlRegion == 'us-west-2' || controlRegion == 'eu-central-1' || controlRegion == 'ap-southeast-1');
+}
```
aws-geoff, line 28: This function is duplicated in another file.  Can we put it in a common file so that this function can be de-duped?

demos/serverless/src/handlers.js
```diff
@@ -14,16 +14,44 @@ const ddb = new AWS.DynamoDB();
 // the meeting is hosted in.
 const chime = new AWS.Chime({ region: 'us-east-1' });
 
+const endpoint = process.env.CHIME_ENDPOINT;
+const globalEndpoint = "https://service.chime.aws.amazon.com";
 // Set the AWS SDK Chime endpoint. The global endpoint is https://service.chime.aws.amazon.com.
```
aws-geoff, line 19: Remove references to "global"

demos/serverless/src/handlers.js
```diff
@@ -14,16 +14,44 @@ const ddb = new AWS.DynamoDB();
...
+  return regionalEndpoints;
+}
 
 // return regional API just for Echo Reduction for now.
-function getClientForMeeting(meeting) {
-  if (meeting && meeting.Meeting && meeting.Meeting.MeetingFeatures && meeting.Meeting.MeetingFeatures.Audio &&
+function getClientForMeeting(meeting, controlRegion) {
+  if(checkValidControlRegion(controlRegion)){
+    const chimeRegional = new AWS.ChimeSDKMeetings({ region: controlRegion});
+    if(endpoint != globalEndpoint){
```
aws-geoff, line 44: Remove references to "global"

aws-geoff (commented):
If we can, guidance tells us that we want to refrain from using the terms "global" and "regional" and instead we should refer to the differences by their namespace, "Chime" and "ChimeSDKMeetings".

demos/serverless/deploy.js
```diff
@@ -41,6 +45,10 @@ function usage() {
   console.log(`  -a, --application                    Browser application to deploy, default '${app}'`);
   console.log(`  -e, --event-bridge                   Enable EventBridge integration, default is no integration`);
   console.log(`  -c, --chime-endpoint                 AWS SDK Chime endpoint, default is '${chimeEndpoint}'`);
+  console.log(`  -iad, --chime-iad-endpoint           AWS SDK Chime IAD endpoint, default is '${chimeMeetingsIADEndpoint}'`);
```
asnyatkov, line 48: I'm not sure I understand the design here. In order to make this extensible, you should have single argument `-rp/--regional-endpoint <value>`

demos/serverless/deploy.js
```diff
@@ -235,7 +255,7 @@ if (!fs.existsSync('build')) {
   fs.mkdirSync('build');
 }
 
-console.log(`Using region ${region}, bucket ${bucket}, stack ${stack}, endpoint ${chimeEndpoint}, enable-termination-protection ${enableTerminationProtection}, disable-printing-logs ${disablePrintingLogs}`);
+console.log(`Using region ${region}, bucket ${bucket}, stack ${stack}, endpoint ${chimeEndpoint}, ChimeMeetingsIADEndpoint=${chimeMeetingsIADEndpoint}, ChimeMeetingsPDXEndpoint=${chimeMeetingsPDXEndpoint}, ChimeMeetingsFRAEndpoint=${chimeMeetingsFRAEndpoint}, ChimeMeetingsSINEndpoint=${chimeMeetingsSINEndpoint}, enable-termination-protection ${enableTerminationProtection}, disable-printing-logs ${disablePrintingLogs}`);
```
asnyatkov, line 258: The same applies here: pass in single argument, not argument per region.

demos/serverless/src/handlers.js
```diff
@@ -14,16 +14,44 @@ const ddb = new AWS.DynamoDB();
...
+const chimeMeetingsRegionalEndPoints = getRegionalEndpoints();
+
+
+function checkValidControlRegion(controlRegion){
+  return (controlRegion == 'us-east-1' || controlRegion == 'us-west-2' || controlRegion == 'eu-central-1' || controlRegion == 'ap-southeast-1');
+}
+
+function getRegionalEndpoints(){
+  let regionalEndpoints = {
+    "us-east-1": process.env.CHIME_MEETINGS_IAD_ENDPOINT,
```
asnyatkov, line 32: You need single environment variable CHIME_MEETINGS_ENDPOINT set to a specific value.

asnyatkov (changes requested):

demos/serverless/deploy.js
harshitOberoi: As this code is public github repository we cannot put the actual values.

demos/browser/server.js
harshitOberoi: Actually no as it would come to this condition only if controlRegion is not valid.

demos/serverless/deploy.js
harshitOberoi: We need all of them as we need to switch between them based upon url parameter `controlRegion`.

demos/serverless/deploy.js
harshitOberoi: We need all of them as we need to switch between them based upon url parameter `controlRegion`.

demos/serverless/src/handlers.js
harshitOberoi: We need all of them as we need to switch between them based upon url parameter `controlRegion`. So in case controlRegion is us-east-1 we need IAD endpoint and if controlRegion is us-west-2 we need pdx endpoint

demos/serverless/deploy.js
aws-geoff: We can't put the public prod endpoints here?

demos/serverless/src/handlers.js
harshitOberoi: We can not have a common file as they have two different project folders and to have a single common file we need to set a webpack inside each of these folders.

harshitOberoi: as for the same reason "getClientForMeeting" was dupilicated between both the files.

demos/serverless/deploy.js
harshitOberoi: Actually we do not want the default values to be prod values, As in case when we provide endpoint = gammaEndpoint and we do not provide the regional endpoints it will use prod regional endpoints which would be incorrect. I have put in a condition in handlers.js at line 44 `const regionalEndpoint = (chimeMeetingsRegionalEndPoints[controlRegion] === '-') ? endpoint : chimeMeetingsRegionalEndPoints[controlRegion];`  so in case if regional endpoint is '-' we would use endpoint provided to make it backward compatible.

harshitOberoi:
Closing this Pull Request will open a new one with finalized changes

Issue #1707, opened 2021-10-22
Add a do-not-pause priority for priority-based downlink policy


mark-vanr:
# Community Note
* Please vote for this issue by adding a 👍 reaction to the issue. This will help the community and maintainers from the Amazon Chime SDK team to prioritize this request.
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.

# Tell us about your request
## What do you want us to build?
Which Amazon Chime SDK or feature area is this request for?
[amazon-chime-sdk-js](https://github.com/aws/amazon-chime-sdk-js)

## Tell us about the problem you are trying to solve and why is it hard?
Using simulcast and the existing priority-based downlink policy, we are currently setting content streams to priority 1 if it is there, and other attendees to a lower priority. When sharing content, we'd like to be able to maintain the view of the content that is being shared, even if the downlink bandwidth of the attendee is very poor - even if it's a frozen frame as packets are missed, it will at least keep the context in front of the meeting attendee. In the current policy, even priority 1 streams will be paused in low-bandwidth scenarios. 

If possible, it would be great to have a priority 0 or a non-pausing priority so that we can always keep that stream open with a reduced experience, as that is (in our opinion) a better experience in the case of content sharing than having the large, sharing video tile go 'blank' as the stream is paused

## How are you currently solving a problem?
Currently setting content to priority 1 in the downlink policy preferences, but it will still be paused if the attendee's downlink is detected to be too low.

ltrung:
@mark-vanr and @dathacky, we just added an option `keepLastFrameWhenPaused` in the meeting session configuration to allow you to keep the last frame when video is paused. Let me know if this meets your requirement.

hensmi-amazon:
Closing as mitigation is likely provided. Please reopen if the feature mentioned by Trung does not solve your issue.

PR #1798, opened 2021-11-17
Add bool flag to disable network pause on priority based policy

hensmi-amazon:
**Issue #:** https://github.com/aws/amazon-chime-sdk-js/issues/1707 + internally

**Description of changes:** See docstring.  Note this is not ideal given we *should* have good reason to pause, but is a bit of a bandaid on how the current tile controller works in that it will blank out videos, and also that sometimes the BWE is just too sensitive.  Fortunately this API is not on the facade.

In a V3 this may not be necessary if we can manage to not blank out tiles when we pause them.

**Testing:**

Added toggles to demo (+100 to any CSS wizard who can get those to always show on top).  bools work as expected.

Disabling network pause should cause tiles to not be paused

**Checklist:**

1. Have you successfully run `npm run build:release` locally? y

n

n


dchen-aws:
I am unsure about this change. disabledownlinkpolicypause may conflict with the priority system already in place. Would it be making more sense to add a way to configure that builder has to acknowledge the pause callback before we actually pausing it?

ltrung:
I thought we want to add it as a parameter? We already have [shouldPauseTiles](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts#L75), can we juist make it a configurable parameter?

hensmi-amazon:
> I am unsure about this change. disabledownlinkpolicypause may conflict with the priority system already in place. Would it be making more sense to add a way to configure that builder has to acknowledge the pause callback before we actually pausing it?

I don't really specifically see how this conflicts with the priority system.  It still fills up bitrates on high priority, then medium priority, etc.  its just that when it runs out of BWE the current logic will start pausing tiles, this change allows ignoring that logic for specific videos.  If we did a callback and builders never wanted to pause a tile it would be  (psudo code)
```
onVideoTileWantsPause(attendeeId) {
  if (dontWantPaused(attendeeId)) {
    return false;
   }
  return true
}  
```

instead of
```
videoPreferenceForAttendeeWeDontWantPaused.disabledownlinkpolicypause = true
```

and would have identical end behavior (unless you are seeing something different).  I think its fine to allow explicit setting, and is much more easier for builder to understand (it's even what was requested in the github issue).

hensmi-amazon:
> I thought we want to add it as a parameter? We already have [shouldPauseTiles](https://github.com/aws/amazon-chime-sdk-js/blob/master/src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts#L75), can we juist make it a configurable parameter?

I actually initially implemented it this way in response to internal ticket (including unit tests :P). But:
* It's too blunt.  The whole point of pausing is because without it, it is (theoretically) likely that more videos will be frozen on network event for longer total time then paused videos (i.e. `sum([timeFrozen for attendee in remoteAttendee]) > sum([timePaused for attendee in remoteAttendee])`...more psudo code :D).  So recommending that this be limited to very specific use cases like content is only possible if we allow the fine grained control.  You can still achieve the parameter on config by just setting it to true for all videos.
* The linked github issue which has a good amount of +1s requests fine grain control for specific use cases like content.

dchen-aws:
> > I am unsure about this change. disabledownlinkpolicypause may conflict with the priority system already in place. Would it be making more sense to add a way to configure that builder has to acknowledge the pause callback before we actually pausing it?
> 
> I don't really specifically see how this conflicts with the priority system. It still fills up bitrates on high priority, then medium priority, etc. its just that when it runs out of BWE the current logic will start pausing tiles, this change allows ignoring that logic for specific videos. If we did a callback and builders never wanted to pause a tile it would be (psudo code)
> 
> ```
> onVideoTileWantsPause(attendeeId) {
>   if (dontWantPaused(attendeeId)) {
>     return false;
>    }
>   return true
> }  
> ```
> 
> instead of
> 
> ```
> videoPreferenceForAttendeeWeDontWantPaused.disabledownlinkpolicypause = true
> ```
> 
> and would have identical end behavior (unless you are seeing something different). I think its fine to allow explicit setting, and is much more easier for builder to understand (it's even what was requested in the github issue).

The difference here is that, prior to this, giving a video tile a higher priority means it will be prioritized for both availability and quality over others. Now with this change, a video with lowest priority but with disabledownlinkpolicypause == true would not be paused and may cause impact on video with higher priority.

And the difference between the config and the callback is that the earlier is a preset the other is reactive action to take after events happen, which is a more conscious choice and can take in other factors.

hensmi-amazon:
> > > I am unsure about this change. disabledownlinkpolicypause may conflict with the priority system already in place. Would it be making more sense to add a way to configure that builder has to acknowledge the pause callback before we actually pausing it?
> > 
> > 
> > I don't really specifically see how this conflicts with the priority system. It still fills up bitrates on high priority, then medium priority, etc. its just that when it runs out of BWE the current logic will start pausing tiles, this change allows ignoring that logic for specific videos. If we did a callback and builders never wanted to pause a tile it would be (psudo code)
> > ```
> > onVideoTileWantsPause(attendeeId) {
> >   if (dontWantPaused(attendeeId)) {
> >     return false;
> >    }
> >   return true
> > }  
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead of
> > ```
> > videoPreferenceForAttendeeWeDontWantPaused.disabledownlinkpolicypause = true
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and would have identical end behavior (unless you are seeing something different). I think its fine to allow explicit setting, and is much more easier for builder to understand (it's even what was requested in the github issue).
> 
> The conflict is that, prior to this, giving a video tile a higher priority means it will be prioritized for both availability and quality over others. Now with this change, a video with lowest priority but with disabledownlinkpolicypause == true would not be paused and may cause impact (video freeze) on video with higher priority.
> 
> And the difference between the config and the callback is that the earlier is a preset, to the builders meaning it's set based on static factors, e.g. business logic, roles and layouts, the later is reactive action to take after events happen, which is a more conscious choice and can take in other factors.

Chatted with Daren and he's more onboard with this now.  Working with trung on naming, though the current naming IMO is sufficiently explicit as it implies (I don't want this tile to pause -> Give it bool to not pause) rather then (I don't want this to pause -> Give it 'super'-priority -> Implies that its not paused_, etc.

hensmi-amazon:
Closing for now, customer seems good with keeping last tile.

PR #1814, opened 2021-11-19
Fixing mobile sdk canary failure

akashuc:
**Issue #:**
Mobile-SDK and JS-SDK demo apps share same chime endpoint to start Live Transcription. The introduction of new parameters on js sdk demo app serverless deployment is changing the endpoint defintions which is making mobile-sdk canaries to fail. This fail is due to the additional parameters being mandatory instead of optional in handle.js file of demo app. 

So when JS-SDK is starting Live Transcription, it is running fine because the JS-SDK demo app passes the parameters that the endpoint needs. But the mobile-sdk fails since they don't know about the new parameters. 

This change makes the new parameters introduced to be optional, so that mobile-sdk canaries can call them without explicitly adding new parameters to their code.

**Description of changes:**
Handled the scenario when mobile-sdk/js-sdk demo app is not sending the new parameters.
**Testing:**

**Checklist:**

### DIFF
--- a/demos/browser/server.js
+++ b/demos/browser/server.js
@@ -145,7 +145,10 @@ function serve(host = '127.0.0.1:8080') {
         const languageCode = requestUrl.query.language;
         const region = requestUrl.query.region;
         let transcriptionConfiguration = {};
-        const transcriptionStreamParams = JSON.parse(requestUrl.query.transcriptionStreamParams);
+        let transcriptionStreamParams = {};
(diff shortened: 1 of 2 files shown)


ltrung:
Need to give more detailed description. What is the shared endpoint here? and what is the failure?

ziyiz-amzn (approved):

ltrung (approved):

PR #1816, opened 2021-11-20
Add Amazon Chime Echo Reduction support in Voice Focus.

dimitarz:
**Issue #:** 

**Description of changes:** 
Adding Echo Reduction improvements to Voice Focus. Echo reduction helps keep echoes—sounds from a user’s loudspeaker that get picked up by their microphone—from circulating back into meeting audio and bringing discussions to a standstill. The echo reduction feature is available in conjunction with the noise reduction provided by Amazon Voice Focus. 

**Testing:**

Added an option to enable this via the Demo app. To enable it, start the demo app and click on 'Enable Echo Reduction (new meetings only)'. This will create a new meeting with the correct attributes such that clients which connect can enable Echo Reduction.

To enable the feature, one must create the meeting as follows:
```typescript
// Create meeting 
const meetingInfo = await chime.createMeeting({
  ...
  MeetingFeatures: {
    Audio: {
      EchoReduction: 'AVAILABLE' 
    }
  } 
}).promise();

// Add attendee 
const attendeeInfo = await chime.createAttendee({...});
const joinInfo = { 
  JoinInfo: {
    Meeting: meetingInfo,
    Attendee: attendeeInfo,
  }
}
```
The client then has to explicitly turn on Echo Reduction as follows:
```typescript
// Select the Echo Reduction model
const spec: VoiceFocusSpec = {
  name: 'ns_es',
  ...
};

// Create the Voice Focus device
const transformer = VoiceFocusDeviceTransformer.create(spec, { logger }, joinInfo);

this.audioVideo = this.meetingSession.audioVideo;
const vfDevice = await transformer.createTransformDevice(chosenAudioInput);

// Enable Echo Reduction on this client
await vfDevice.observeMeetingAudio(this.audioVideo);
```

**Checklist:**

2. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? Yes.

3. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved? No and no.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license. Yes


demos/browser/package.json
```diff
@@ -32,7 +32,7 @@
   },
   "dependencies": {
     "amazon-chime-sdk-js": "file:../..",
-    "aws-sdk": "^2.1032.0",
+    "aws-sdk": "file:/Users/dimitarz/Downloads/aws-sdk",
```
ltrung, line 35: Need to update this before merge.

demos/browser/server.js
```diff
@@ -19,25 +19,39 @@ console.info('Using index path', indexPagePath);
...
 const sts = new AWS.STS({ region: 'us-east-1' })
 
 // Set the AWS SDK Chime endpoint. The global endpoint is https://service.chime.aws.amazon.com.
 const endpoint = process.env.ENDPOINT || 'https://service.chime.aws.amazon.com';
-console.info('Using endpoint', endpoint);
-
+console.info('Using global endpoint', endpoint);
 chime.endpoint = new AWS.Endpoint(endpoint);
 
+if (process.env.ENDPOINT) {
```
ltrung, line 34: I don't think we need this for now right? This is mainly to test with gamma endpoint.

demos/browser/server.js
```diff
@@ -19,25 +19,39 @@ console.info('Using index path', indexPagePath);
 
 const indexPage = fs.readFileSync(indexPagePath);
 
-// Create ans AWS SDK Chime object. Region 'us-east-1' is currently required.
+// Create ans AWS SDK Chime object. Region 'us-east-1' is globally available..
 // Use the MediaRegion property below in CreateMeeting to select the region
 // the meeting is hosted in.
 const chime = new AWS.Chime({ region: 'us-east-1' });
+const chimeRegional = new AWS.ChimeSDKMeetings({ region: 'us-east-1' });
```
ltrung, line 26: Do you think we could just use ChimeSDKMeeting but just hard code it the region to `us-east-1` like this?

demos/serverless/deploy.js
```diff
@@ -11,8 +11,8 @@ let app = `meetingV2`;
 let useEventBridge = false;
 let enableTerminationProtection = false;
 let disablePrintingLogs = false;
-let chimeEndpoint = 'https://service.chime.aws.amazon.com';
-let chimeServicePrincipal = 'chime.amazonaws.com'
+let chimeEndpoint = 'https://meetings-chime.us-east-1.amazonaws.com ';
```
ltrung, line 14: Do we need to make this change? ChimeServicePrincipal is used by media capture i think.

demos/browser/server.js
dimitarz: It's pre-existing functionality for the global endpoint, I just added for the regional one as well. See line #30.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1243,7 +1243,7 @@ export default class DefaultDeviceController
     streamId = proposedConstraints?.video.streamId;
     /* istanbul ignore next */
     // @ts-ignore
-    return !!streamId && streamId === selection.constraints?.video?.streamId;
+    return !!streamId && streamId === selection?.constraints?.video?.streamId;
```
ltrung, line 1246: When can selection be null?

src/voicefocus/VoiceFocusDeviceTransformer.ts
```diff
@@ -133,8 +133,28 @@ export class VoiceFocusDeviceTransformer {
   static async create(
     spec: VoiceFocusSpec = {},
     options: VoiceFocusDeviceOptions = {},
-    config?: VoiceFocusConfig
+    config?: VoiceFocusConfig,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    createMeetingResponse?: any,
+    // eslint-disable-next-line
+    createAttendeeResponse?: any
```
ltrung, line 140: Is this used?

src/voicefocus/VoiceFocusDeviceTransformer.ts
```diff
@@ -133,8 +133,28 @@ export class VoiceFocusDeviceTransformer {
...
   ): Promise<VoiceFocusDeviceTransformer> {
+    if (createMeetingResponse) {
```
ltrung, line 142: Can we just pass in `createMeetingResponse?.Meeting?.MeetingFeatures?.Audio?.EchoReduction === 'AVAILABLE';` as bool | undefined instead of the whole response?

demos/browser/server.js
dimitarz: It was functionality that was there before (i.e. allowing setting of the ENDPOINT in the demo app).

PR #1819, opened 2021-11-21
Switched `DefaultSimulcastUplinkPolicy` non-simulcast mode's enabled stream

hensmi-amazon:
**Issue #:** N/a (may look around for one).

**Description of changes:** This impacts only the non-simulcast mode of the simulcast policy when is when there are 2 or less participants (not 2 or less senders). Copying comment:
```
        // See comment above `shouldDisableSimulcast` for usage.
        //
        // The value of `newActiveStreams` is somewhat irrelevant since this single
        // stream will adapt based on both sender and receiver network conditions.
        //
        // We use the middle stream here to work around a bug in Chromium where
        // it seems when a transceiver is created when BWE is low (e.g. on a reconnection),
        // it will never reset the encoder even when `setParameters` is called.  WebRTC bug
        // #12788 seems to call a similar issue out as fixed for VP8, it's not clear if this
        // is the same issue for H.264. Additionally we are not able to force a keyframe
        // request from the backend since it will only be sending padding (which also
        // don't have MID due to #10822). Since we don't scale when simulcast is disabled
        // this doesn't have any end-user effect.
```

I had fixed a backend issue where padding was not included in the BWE module. But then it still seemed like Chrome was ignoring the values set via `setParams`:
```
SDK - simulcast: current encoding parameters 
rid=low maxBitrate=0 active=false 
rid=mid maxBitrate=0 active=false 
rid=hi maxBitrate=1200000 active=true 
```
and then (browser logs)
```
Simulcast streams:
0: 320x180 min_kbps: 30 target_kbps: 99 max_kbps: 133 max_fps: 60 max_qp: 56 num_tl: 2 active: false
1: 640x360 min_kbps: 150 target_kbps: 500 max_kbps: 700 max_fps: 60 max_qp: 56 num_tl: 2 active: false
2: 1280x720 min_kbps: 30 target_kbps: 1200 max_kbps: 1200 max_fps: 60 max_qp: 56 num_tl: 2 active: true
```
but on reconnection…
```
SDK - simulcast: current encoding parameters 
rid=low maxBitrate=0 active=false 
rid=mid maxBitrate=0 active=false 
rid=hi maxBitrate=1200000 active=true 
```
…
```
Simulcast streams:
0: 320x192 min_kbps: 36 target_kbps: 107 max_kbps: 142 max_fps: 60 max_qp: 56 num_tl: 2 active: true
1: 640x384 min_kbps: 160 target_kbps: 537 max_kbps: 726 max_fps: 60 max_qp: 56 num_tl: 2 active: true
2: 1280x768 min_kbps: 610 target_kbps: 1200 max_kbps: 1200 max_fps: 60 max_qp: 56 num_tl: 2 active: true
```
no matter how many times set params is called.  However I noticed that if not using the top stream it worked fine, hence the fix.

**Testing:**
Smoke tested 1, 2, 3 participant calls (i.e. to move in and out of simulcast disabled mode).  All resolutions and send streams were as expected.  Disconnecting and reconnecting a client now immediately starts sending non-padding packets, and recovered BWE normally.

* Connect a Chrome client with simulcast enabled, and enable send.
* Connect any receiver to be able to inspect video.
* Disable Wifi on sender. Wait 15 seconds for it to enter reconnect loop.
* Re-enable Wifi and reconnect.  Video should immediately be visible.

**Checklist:**

y

n

n


demos/browser/package-lock.json
```diff
@@ -1,11121 +1,28 @@
 {
   "name": "amazon-chime-sdk-js-demos",
   "version": "0.1.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
ltrung, line 4: Wrong version :p

src/videouplinkbandwidthpolicy/DefaultSimulcastUplinkPolicy.ts
```diff
@@ -316,14 +332,14 @@ export default class DefaultSimulcastUplinkPolicy implements SimulcastUplinkPoli
     const toBps = 1000;
     const nameArr = SimulcastTransceiverController.NAME_ARR_ASCENDING;
     const bitrateArr = bitratesKbps;
-
-    let scale = 4;
+    // Don't scale the single simulcast stream regardless of its layer.å
```
ltrung, line 335: nit: Weird character at the end

ltrung:
Can we clarify that this issue is with 2 attendee calls only where simulcast is not enabled in the PR description?

src/videouplinkbandwidthpolicy/DefaultSimulcastUplinkPolicy.ts
```diff
@@ -128,12 +130,24 @@ export default class DefaultSimulcastUplinkPolicy implements SimulcastUplinkPoli
...
+        // #12788 seems to call a similar issue out as fixed for VP8, it's not clear if this
+        // is the same issue for H.264. Additionally we are not able to force a keyframe
+        // request from the backend since it will only be sending padding (which also
+        // don't have MID due to #10822). Since we don't scale when simulcast is disabled
+        // this doesn't have any end-user effect.
         this.newActiveStreams = ActiveStreams.kHi;
         newBitrates[0].maxBitrateKbps = 0;
-        newBitrates[1].maxBitrateKbps = 0;
-        newBitrates[2].maxBitrateKbps = 1200;
+        newBitrates[1].maxBitrateKbps = 1200;
```
ltrung, line 149: Curious. Can't it be the 1st (index 0) stream?

ltrung (dismissed):

ltrung (approved):

src/videouplinkbandwidthpolicy/DefaultSimulcastUplinkPolicy.ts
franko-franicevich: With the recent fixes, we're switching back from our fork to the canonical AWS repo, but keeping our own uplink policy (and merging in your changes). With that in mind, I have a question to help me understand:

The comment above says: 
"  // The value of `newActiveStreams` is somewhat irrelevant since this single
   // stream will adapt based on both sender and receiver network conditions."
       
So this means it's OK to leave 'newActiveStreams' to kHi, while only populating the medium stream index[1]?

hensmi-amazon: I actually re-read that comment and also thought it was a bit confusing so i rewrote it in https://github.com/aws/amazon-chime-sdk-js/pull/1900 (also note I added another comment to the bottom of that block).

But to answer your question, yes, and it only effects the `encodingSimulcastLayersDidChange` callback to your application on the sender side (which currently is purely informational).

PR #1824, opened 2021-11-23
Add browser compatibility test

nainkunal933:
**Issue #:**

**Description of changes:** For adding more visibility on browser support for the various features of JS SDK, these browser compatibility tests are being added. They will run everyday and test the features against several different browser and OS combinations.

**Testing:** There should be a new GitHub action that will run everyday at 12 am. 

*Can these tested using a demo application? Please provide reproducible step-by-step instructions.*

**Checklist:**

1. Have you successfully run `npm run build:release` locally? Yes

2. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? No

3. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved? No

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1827, opened 2021-11-23
Added support for non-audio and non-video IOS tests

LeviSklaroff:
**Issue #:**
We have no IOS support with our integration tests.
**Description of changes:**
Added support for non-audio and non-video IOS tests. More changes are needed for other tests.
**Testing:**
Ran Data Message and Meeting End Tests on IOS 14.7/15, Android Chrome, FireFox Latest, and Chrome Latest

**Checklist:**

Yes

No

No

### DIFF
--- a/integration/js/pages/AppPage.js
+++ b/integration/js/pages/AppPage.js
@@ -33,6 +33,7 @@ function findAllElements() {
     switchToSipFlow: By.id('to-sip-flow'),
 
     authenticationFlow: By.id('flow-authenticate'),
+    permisssionsFlow: By.id('flow-need-permission'),
     deviceFlow: By.id('flow-devices'),
     meetingFlow: By.id('flow-meeting'),
     sipAuthenticateFlow: By.id('flow-sip-authenticate'),
@@ -91,6 +92,7 @@ class AppPage {
 
   async open(stepInfo) {
     await TestUtils.open(stepInfo);
+    await this.waitForBrowserDemoToLoad();
   }
 
   async close(stepInfo) {
@@ -312,20 +314,37 @@ class AppPage {
     }
     return 'failed'
   }
-
-  async waitForAuthentication() {
-    let timeout = 10;
-    let i = 0;
-    let authenticating = true;
-    while (authenticating && i < timeout) {
-      authenticating = await this.isAuthenticating();
-      if (authenticating === false) {
-        return 'done'
-      }
-      i++;
-      await TestUtils.waitAround(1000);
+  
+  async waitForElement(element, timeout = 10000) {
+    try {
+      await this.driver.wait(until.elementLocated(element), timeout);
+      await this.driver.wait(until.elementIsVisible(this.driver.findElement(element)), timeout);
+      return 'done';
     }
-    return 'failed'
+    catch(e) {
+      return 'failed';
+    }
+  }
+
+  async waitForElementToNotBeVisible(element, timeout = 10000) {
+    try {
+      await this.driver.wait(until.elementIsNotVisible(this.driver.findElement(element)), timeout);
+      return 'done';
+    }
+    catch(e) {
+      return 'failed';
+    }
+  }
+
+  async waitForDeviceFlow() {
+    if(await this.waitForElementToNotBeVisible(elements.authenticationFlow) === 'failed') {
+      return 'failed';
+    }
+    return await this.waitForElementToNotBeVisible(elements.permisssionsFlow);
+  }
+
+  async waitForBrowserDemoToLoad() {
+    return await this.waitForElement(elements.authenticationFlow, 30000)
   }
 
   async waitForSipAuthentication() {
@@ -344,18 +363,7 @@ class AppPage {
   }
 
   async waitToJoinTheMeeting() {
-    let timeout = 20;
-    let i = 0;
-    let joining = true;
-    while (joining && i < timeout) {
-      joining = await this.isJoiningMeeting();
-      if (joining === false) {
-        return 'done'
-      }
-      i++;
-      await TestUtils.waitAround(1000);
-    }
-    return 'failed'
+    return await this.waitForElement(elements.meetingFlow, 20000)
   }
 
   async isLiveTranscriptionPresentInDeviceMenu() {
@@ -517,14 +525,6 @@ class AppPage {
   async enableVoiceFocusInLobby() {
     const elem = await this.driver.findElement(elements.addVoiceFocusInput);
     return clickElement(this.driver, elem);
-  }
-
-  async isJoiningMeeting() {
-    return await this.driver.findElement(elements.deviceFlow).isDisplayed();
-  }
-
-  async isAuthenticating() {
-    return await this.driver.findElement(elements.authenticationFlow).isDisplayed();
   }
 
   async isMeetingEnding() {
--- a/integration/js/pages/TestAppPage.js
+++ b/integration/js/pages/TestAppPage.js
@@ -89,7 +89,7 @@ class TestAppPage {
     await joinMeetingButton.click();
   }
 
-  async waitForAuthentication() {
+  async waitForDeviceFlow() {
     await TestUtils.waitAround(5000);
   }
 
--- a/integration/js/steps/AuthenticateUserStep.js
+++ b/integration/js/steps/AuthenticateUserStep.js
@@ -69,7 +69,7 @@ class AuthenticateUserStep extends AppTestStep {
     }
     await this.page.authenticate();
     this.logger("waiting to authenticate");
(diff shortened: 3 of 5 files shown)


nainkunal933:
@LeviSklaroff If the update in utilities ran successfully against Data Message and Meeting End test then why don't update the integration testing config for those tests?

If need be, please create a `mobileIntegrationTest` folder under configs. Please refer to #1824 on why I am asking for separate folder under `configs`.

integration/js/pages/AppPage.js
```diff
@@ -551,7 +567,11 @@ class AppPage {
   }
 
   async isAuthenticating() {
-    return await this.driver.findElement(elements.authenticationFlow).isDisplayed();
+    return await this.driver.findElement(elements.authenticationFlow).isDisplayed() || await this.driver.findElement(elements.permisssionsFlow).isDisplayed();
+  }
+  
+  async isWaitingForPageToLoad() {
+    return await this.driver.findElements(elements.authenticationFlow).length > 0;
   }
```
nainkunal933, line 575: Please add a method description here. This method is testing whether the page is loaded via checking HTML element length. In the description, please explain why a `length > 0` means the page is loaded? It is not very intuitive for users who are not familiar with the mobile stack.

integration/js/pages/AppPage.js
```diff
@@ -328,6 +329,21 @@ class AppPage {
     return 'failed'
   }
 
+  async waitForPageToLoad() {
+    let timeout = 10;
+    let i = 0;
+    let loading = false;
```
nainkunal933, line 335: The variable should be name something like: `finishedLoading`. `loading` indicates a running process as false from the start which is the opposite of what the code is doing.

integration/js/steps/OpenAppStep.js
```diff
@@ -17,6 +17,7 @@ class OpenAppStep extends AppTestStep {
   async run() {
     this.logger(`Opening : ${this.url}`);
     await this.page.open(this);
+    await this.page.waitForPageToLoad();
```
nainkunal933, line 20: If `waitForPageToLoad` is only needed for mobile testing then will it add a delay for integration tests as well? I think not but want to know if you have tested that.

integration/js/pages/AppPage.js
```diff
@@ -551,7 +567,11 @@ class AppPage {
   }
 
   async isAuthenticating() {
-    return await this.driver.findElement(elements.authenticationFlow).isDisplayed();
+    return await this.driver.findElement(elements.authenticationFlow).isDisplayed() || await this.driver.findElement(elements.permisssionsFlow).isDisplayed();
+  }
```
nainkunal933, line 571: I think the check for permissions should be a separate method based on how it is used. 

If you prefer only one method then the method name needs to change. The method is now checking for authentication or device permissions dialog.

integration/js/pages/AppPage.js
LeviSklaroff: Will change

integration/js/steps/OpenAppStep.js
ltrung: Why do we need to do this? Is this something specific to iOS?

integration/js/pages/AppPage.js
ltrung: I think just a better method name would be great. Can we specify which page this is (e.g., isiWatingForAuthPageToLoad)?
Also why can't we use isDisplayed() here instead of checking for length?

LeviSklaroff: Will add. It's not actually checking the length of the element or anything with the mobile stack, it's just checking if a element from our browser demo exists so it won't try to run the rest of the test while it's on the default page, only a problem with IOS Safari though.

integration/js/pages/AppPage.js
```diff
@@ -328,6 +329,21 @@ class AppPage {
     return 'failed'
   }
 
+  async waitForPageToLoad() {
+    let timeout = 10;
+    let i = 0;
+    let loading = false;
+    while (!loading && i < timeout) {
```
ltrung, line 336: We reuse this pattern a lot in this code base. Can we abstract it out to a helper wait function? :D

LeviSklaroff: Yeah I'll go ahead and add that

integration/js/pages/AppPage.js
LeviSklaroff: I'll go ahead and change the method name to isLoadingDeviceFlow. I'd make them separate methods, but we only use the function in one place so not much reason to separate them.

integration/js/pages/AppPage.js
LeviSklaroff: @ltrung Yeah I'll change to isWaitingForAuthPageToLoad, can't use isDisplayed because it's not actually finding the element it's checking if it can find the element. findElement produces a error if it can't find the element and there's no selenium function for doesElementExist so to find if the element exists I use findElements and see if it found any that exist.

integration/js/steps/OpenAppStep.js
LeviSklaroff: This technically will add a very slight delay to all tests by having to call the method and go through the first few steps but it won't actually do any waiting unless the page hasn't been loaded for some reason which in that case we want to wait a moment. At the moment this is a IOS Safari specific problem, I'm not entirely sure why the open function returns before the page is actually ready but I've even tried calling selenium driver waitforpage and it says it's completed before it actually finishes.

LeviSklaroff:
> @LeviSklaroff If the update in utilities ran successfully against Data Message and Meeting End test then why don't update the integration testing config for those tests?
> 
> If need be, please create a `mobileIntegrationTest` folder under configs. Please refer to #1824 on why I am asking for separate folder under `configs`.

The main purpose of these config files are for our Github Actions/Integration Tests, I wasn't planning on adding any actions for IOS.

integration/js/pages/AppPage.js
```diff
@@ -312,20 +314,40 @@ class AppPage {
...
+      return 'done';
+    }
+    catch(e)
+    {
+      return 'failed';
+    }
+  }
+
+  async waitForDeviceFlow() {
+    if(await this.waitForElementToNotBeVisible(elements.authenticationFlow) === 'failed')
```
ltrung, line 344: nit: space

ltrung: Actually no {} ?

integration/js/pages/AppPage.js
```diff
@@ -312,20 +314,40 @@ class AppPage {
...
-      authenticating = await this.isAuthenticating();
-      if (authenticating === false) {
-        return 'done'
-      }
-      i++;
-      await TestUtils.waitAround(1000);
+  async waitForElement(element, timeout = 10000)
+  {
+    try
+    {
```
ltrung, line 320: nit: try {

integration/js/pages/AppPage.js
```diff
@@ -312,20 +314,40 @@ class AppPage {
...
-    let authenticating = true;
-    while (authenticating && i < timeout) {
-      authenticating = await this.isAuthenticating();
-      if (authenticating === false) {
-        return 'done'
-      }
-      i++;
-      await TestUtils.waitAround(1000);
+  async waitForElement(element, timeout = 10000)
+  {
```
ltrung, line 318: nit: try {

integration/js/pages/AppPage.js
```diff
@@ -312,20 +314,40 @@ class AppPage {
...
+    }
+  }
+  async waitForElementToNotBeVisible(element, timeout = 10000)
+  {
+    try
+    {
+      await this.driver.wait(until.elementIsNotVisible(this.driver.findElement(element)), timeout);
+      return 'done';
+    }
+    catch(e)
```
ltrung, line 337: nit: spacing

LeviSklaroff: Fixed

integration/js/pages/AppPage.js
LeviSklaroff: Fixed

integration/js/pages/AppPage.js
LeviSklaroff: Fixed

integration/js/pages/AppPage.js
LeviSklaroff: Fixed

integration/js/pages/AppPage.js
ltrung: Can we also fix waitForElement(...) { ?

LeviSklaroff: Oops, fixed

ltrung (approved):

PR #1830, opened 2021-11-24
Add simple pagination to demo if priority based policy is used

hensmi-amazon:
**Issue #:** None, but i was looking into https://github.com/aws/amazon-chime-sdk-js/issues/1807 & https://github.com/aws/amazon-chime-sdk-js/issues/1805 and needed this functionality.

**Description of changes:** Using `chooseRemoteVideoSources` to only subscribe to a subset of videos is a critical API that should be easily smoke testable in our demo, as it not only is a complex API use, but also a common one.  Therefore this change adds simple pagination to demo if priority based policy is used and a checkbox is hit.  This will additionally be a necessity if we ever increase video send ability past the receive limit.

Note the following screenshot was taken when i forced the page limit to 1.  But it was additionally tested at 3.

Screen Shot 2021-11-24 at 10 20 42 AM

**Testing:** Smoke tested with pagination enabled and disabled. Pagination works as expected.

*Can these tested using a demo application? Please provide reproducible step-by-step instructions.*
Enabled pagination and 5+ sends in a conference.  Press the buttons.  Self video and content will be shown on every page.

**Checklist:**

1. Have you successfully run `npm run build:release` locally?
y

2. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved?
n

3. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved?
n

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1807, opened 2021-11-18
Wrong stream for a split second after calling chooseRemoteVideoSources on Firefox


andreisergiu98:
### What happened and what did you expect to happen?

We use priority based downlink policy to unsubscribe to streams that are not in view. As shown in the video below, after `chooseRemoteVideoSources` is called Firefox will render a wrong video stream for a moment, afterwards the correct stream is rendered.
There is also a similar issue where we choose remote video sources some of the streams will go blank for a couple of milliseconds.

Mentions:
Simulcast is off.

https://user-images.githubusercontent.com/11740700/142455751-22179977-3509-4e52-b802-39ddc66f510f.mp4

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

See video.

### Amazon Chime SDK for JavaScript version

2.21.1

### What browsers are you seeing the problem on?

Firefox

### Browser version

94.0.1

### Browser console logs

[meeting-logs.txt](https://github.com/aws/amazon-chime-sdk-js/files/7566396/meeting-logs.txt)

hensmi-amazon:
Thanks for the report.  For our investigation, could you describe how you are binding the tiles? Are you reusing Video HTML elements?

hensmi-amazon:
I believe I have an easy repro, and a likely fix in https://github.com/aws/amazon-chime-sdk-js/pull/1831 .  If you have a chance could you try that branch?  It will hopefully be a part of next release.

andreisergiu98:
Hi! Thank you for your response!
We tried out your PR and while streams are rendered correctly we encountered some disconnections when we call `chooseRemoteVideoSources` multiple times in rapid succession (both on firefox and chrome).

```
2021-11-25T14:26:58.909Z [WARN] SDK - SubscribeAndReceiveSubscribeAckTask connection was terminated with code 1006 and reason:  console.js:107:15
[...]

Issue #1805, opened 2021-11-18
Streams can get mixed up or blank when using simulcast


andreisergiu98:
### What happened and what did you expect to happen?

![image](https://user-images.githubusercontent.com/11740700/142432375-0fefa845-43d9-4770-8c98-d3a1bdaa93ca.png)

Attendee 1 renders stream of attendee 2 and attendee 2 appears blank.

The issue is random, happens only locally for some attendee(s), does not revert to its initial state no matter what change (except browser refresh) any attendee makes (whether local or remote), and only happens when simulcast is active.

Upon further testing, it seems that the issue does not happen with simulcast disabled.

Mentions:
Default attendee broadcast quality: 180p at max/100Kbps
When made speaker/spotlight attendees will have their local stream stopped and then re-started in order to broadcast with a higher quality: 960p. Quality is reverted to 180p when an attendee is no longer in spotlight.
Screen-sharing is not affected and remains the only stream that is broadcast at 1080p without bandwidth restrictions
Streams are rendered using RemoteVideo from `amazon-chime-sdk-component-library-react`

What we'd like to know if this mixup can happen because of this issue https://github.com/aws/amazon-chime-sdk-js/issues/1448.

Also we tried applying the patch from the comment and while it did improve things, it seems that the mixup can still happen.

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

Join meeting, start/stop cameras or just wait.

### Amazon Chime SDK for JavaScript version

2.21.1

### What browsers are you seeing the problem on?

Chrome

### Browser version

96.0.4664.45

### Browser console logs

We were not able to reproduce the issue with logs enabled.

hensmi-amazon:
Thanks for the report. Do you have any reproductions with logs enabled?  If not do you have a call ID with an approximate timestamp?  I don't think that issue linked would caused swapped streams, but I have been looking into logic in the transceiver controller to see if a certain order of adding and removal could cause the mapping to get out of order.

Lastly, wanted to confirm:
* If you are using the priority based policy?
[...]

PR #1831, opened 2021-11-24
No longer try to reuse inactive transcievers to fix stale frame on FF

hensmi-amazon:
**Issue #:** https://github.com/aws/amazon-chime-sdk-js/issues/1807 (& *possibly* https://github.com/aws/amazon-chime-sdk-js/issues/1805, but hard to tell without logs).

**Description of changes:** On FF there is the easily reproducible (using https://github.com/aws/amazon-chime-sdk-js/pull/1830) issue of the HTML video element taking the last couple frames from the transceiver.  In addition the application team also seems to have fixed a general wrong tile issue that was also caused by the logic (though their fix is a bit different, and seems to have accidentally caused the intentional change of this PR).

AFAIK there is no significant benefit to reusing transceivers (*as is*) besides maybe saving on some SDP size.  So this change should not have any impact.

Note that initially this change had issues with ever growing SDPs when simply marking them as inactive, as browsers do not like to remove inactive sections that aren't at the end of the list.  I then ran into the issue of Firefox randomly putting inactive sections at the end of the SDP, which is why I added functionality to simply ensure that our streams were lined up correctly (in the future we should just be using Maps of MID to source, like we do for simulcast stream switches).

**Testing:**

Smoke tested many to many calls with toggles.  Additionally ran steps below.

* Start up 4 sending clients, and one receiver with priority and pagination enabled.
* Toggle a few of the senders off and on.
* Paginate and look for stale videos.
* Repeat steps 2 and 3.

**Checklist:**

y

n

n


test/transceivercontroller/SimulcastTransceiverController.test.ts
```diff
@@ -419,19 +419,22 @@ describe('SimulcastTransceiverController', () => {
...
         tracks: [new SdkTrackMapping({ streamId: 7, trackLabel: 'v_7' })],
       });
       videoStreamIndex.integrateSubscribeAckFrame(subackFrame);
       setTransceiverStreamId(videoStreamIndex, videosToReceive);
+
       videosToReceive = new DefaultVideoStreamIdSet([7, 8]);
       videoSubscriptions = tc.updateVideoTransceivers(videoStreamIndex, videosToReceive);
-      expect(videoSubscriptions).to.deep.equal([0, 7, 8]);
-      verifyTransceiverDirections(['recvonly', 'recvonly']);
+      expect(videoSubscriptions).to.deep.equal([0, 0, 0, 7, 8]);
```
hensmi-amazon, line 432: Note this won't grow endlessly because the browser will clean up and remove inactive transceivers.

hensmi-amazon:
Pause this while I look into possible issues with it not cleaning up transceivers.

src/task/SubscribeAndReceiveSubscribeAckTask.ts
```diff
@@ -94,6 +99,42 @@ export default class SubscribeAndReceiveSubscribeAckTask extends BaseTask {
...
+  private fixUpSubscriptionOrder(sdp: string, videoSubscriptions: number[]): number[] {
+    const subscriptionsWithoutZero = videoSubscriptions.filter((value: number) => value !== 0);
+    let subscriptionsWithoutZeroIndex = 0;
+
+    const directions = new DefaultSDP(sdp).videoSectionDirections();
+    const newSubscriptions: number[] = [];
+    for (const direction of directions) {
+      if (direction === 'recvonly') {
+        if (subscriptionsWithoutZeroIndex >= subscriptionsWithoutZero.length) {
+          this.context.logger.warn(
```
ltrung, line 118: Can this actually happen? If it is unexpected, should it be at least error log?

src/task/SubscribeAndReceiveSubscribeAckTask.ts
```diff
@@ -69,6 +69,11 @@ export default class SubscribeAndReceiveSubscribeAckTask extends BaseTask {
 
     this.context.videoStreamIndex.subscribeFrameSent();
 
+    // See comment above `fixUpSubscriptionOrder`
+    const videoSubscriptions = this.context.browserBehavior.requiresUnifiedPlan()
```
ltrung, line 73: Do we still need to do this? I thought we no longer support planB?

src/transceivercontroller/DefaultTransceiverController.ts
```diff
@@ -217,20 +217,127 @@ export default class DefaultTransceiverController implements TransceiverControll
     videosToReceive: VideoStreamIdSet
   ): void {
     const videosRemaining = videosToReceive.array();
+    if (transceivers.length !== 0 && !transceivers[0].stop) {
```
ltrung, line 220: Can we add this to V3 for consideration? Seem like a straightforward low hanging fruit change :D

src/task/SubscribeAndReceiveSubscribeAckTask.ts
hensmi-amazon: Safari 12 ayy

src/transceivercontroller/DefaultTransceiverController.ts
hensmi-amazon: Yup, maybe we can just do a grep for version numbers/webrtc function checks in the codebase since I'm sure there are other checks like this.  We can also raise limits based off usage.

src/task/SubscribeAndReceiveSubscribeAckTask.ts
hensmi-amazon: With browsers, i don't want to ever trust that the SDP will look how I expect, so I'm just trying to be safe here. Its not a hard failures (in fact there probably wouldn't be any impact), so I think a warning is sufficient.

ltrung:
>NOTE: This should not be merged as is because there is a backend bug in response to the fact that when Firefox tosses a section at the end of the SDP it will for some reason be VP8 only, which confuses the logic there. A fix is being deployed but it will not complete until next week. For testing one can use https://github.com/aws/amazon-chime-sdk-js/tree/no-reuse-transceivo-with-munge which shoves H.264 back in that section. 

Should this be removed?

CHANGELOG.md
```diff
@@ -5,29 +5,20 @@ All notable changes to this project will be documented in this file.
...
-## [2.26.0] - 2022-01-14
-    
-### Added
-    
-### Removed
-    
 ### Changed
 - Made `SimulcastUplinkObserver.encodingSimulcastLayersDidChange` (*not* `AudioVideoObserver.encodingSimulcastLayersDidChange`) synchronous.
-    
-### Fixed
```
ltrung, line 29: So the order right now (as automated added by script) is Added, Removed, Changed, Fixed so can we swap the Changed and Fixed part to be consistence :D

ltrung (dismissed):

ltrung (approved):

haofiltered:
Hi there,
On the latest chime SDK version (2.26.0),  there is an issue with FF that remote video tile will display as black screen sometimes when first entering. It's not reproduce-able every time, but definitely can be reproduced. We had simulcast off, and the issue was gone after downgrading to 2.25.0.

haofiltered:
@hensmi-amazon 
Just two remote tiles. And they randomly go black. If the remote tile joins after the FF user joined, there will be no issue. The issue only appears when a remote user joins first, then the FF user joins after. 

Here is the log with one remote tile that shows black. [log.txt](https://github.com/aws/amazon-chime-sdk-js/files/8043319/log.txt)

hensmi-amazon:
So i was able to reproduce with Firefox logs, and it seems that they have poor unsignalled SSRC handling, which occurs when SDPs are received before UDP packets (which should technically be supported). I can even easily reproduce when i apply an artificial 100ms delay to the signaling channel. 

I think this being seen more (I think i've actually seen this issue for years) was not due to this change in particular but indirectly related to https://github.com/aws/amazon-chime-sdk-js/pull/1937 . I am able to see recovery (with the same issue reproducing at the same rate occasionally) on additional subscribes, and before that change the call would often start off with a ton of unnecessary subscribes.

Luckillly...I am able to get the issue to stop reproducing if I don't negotiated MID/RID/RRID on receive sections (a bit complicated lol). However that needs to be deployed to our backend so it will take maybe a week. I'll update this comment chain when its deployed (though in the future I recommend just cutting a new github issue :), it will usually make its way to me eventually).

haofiltered:
@hensmi-amazon @ltrung 
Hi Henry and Trung,
The same issue happened again, and we are using the latest `2.27.0` version. 
Here is the log: [logs.txt](https://github.com/aws/amazon-chime-sdk-js/files/8162571/logs.txt)

hensmi-amazon:
Moving to issue, will ask question there

Issue #1609, opened 2021-09-21
chooseVideoInputDevice fails without throwing any error on getting called continuously with different device id


shriyajuneja:
### What happened and what did you expect to happen?

I'm implementing flip camera feature, which calls chooseVideoInputDevice with two different device ids alternately
`await meetingManager.audioVideo?.chooseVideoInputDevice(deviceId);`
If this function is called continuously (3-4 times), chooseVideoInputDevice fails internally but doesn't throw any error and logs `No video device chosen, stopping local video tile`. 

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

switch video input device ids continuously

### Amazon Chime SDK for JavaScript version

2.17.0

### What browsers are you seeing the problem on?

Chrome, Firefox, Samsung Internet

### Browser version

Chrome - 91.0.4472

### Meeting and Attendee ID Information.

Meeting Id - "5deee7e2-8cf3-4501-babe-d31bda610706"
Attendee Id - "84066e51-124f-c659-60ce-2581e470ab4a"

### Browser console logs

Logs -
2021-09-21T21:27:21.173Z [ERROR] SDK - no video device chosen, stopping local video tile
2021-09-21T21:27:21.175Z [ERROR] Logger 5148be7a-cb27-4512-82e1-2772cb080706 - no video device chosen, stopping local video tile
2021-09-21T21:27:21.176Z [INFO] Logger 5148be7a-cb27-4512-82e1-2772cb080706 - deferring transition from Updating with Update
2021-09-21T21:27:21.180Z [WARN] Logger 5148be7a-cb27-4512-82e1-2772cb080706 - could not acquire video input from current device

[1632259711638.log](https://github.com/aws/amazon-chime-sdk-js/files/7206398/30d1-122-176-144-233.ngrok.io-1632259711638.log)

ltrung:
@shriyajuneja Thanks for reporting this. Can you share us a code snippet of your flip camera function?

shriyajuneja:
```
[...]

PR #1841, opened 2021-11-30
Avoid race condition on asynchronous update tasks

hensmi-amazon:
**Issue #:** https://github.com/aws/amazon-chime-sdk-js/issues/1448 & https://github.com/aws/amazon-chime-sdk-js/issues/1805

**Description of changes:** https://github.com/aws/amazon-chime-sdk-js/issues/1448 explains issue (see my comment at bottom).  This change simply defers ingestion of index/resubscribe check until after the update completes.

**Testing:** With the pagination demo I can repro this occaisonally if I join a meeting with 3 videos, and 2 per page, local backend.  The resubscribe spam in beginning of call seem to lead to one video being missing.  I don't see the issue after change.  Its also trivial to inspect logs to verify this works as expected.

Not really, without looking at logs and trying to spam resubscribes.

**Checklist:**

y

n

n


src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -1040,6 +1046,13 @@ export default class DefaultAudioVideoController
   }
 
   private async actionUpdateWithRenegotiation(notify: boolean): Promise<void> {
+    // `actionUpdateWithRenegotiation` contains a significant portion of asynchronous tasks, so we need
+    // to explicitly defer any task operation which may be performed on the event queue that may modify
+    // mutable state needed to be consistent over the course of the update. The operations above do not
+    // need this protection because they are synchronous.
+    this.monitorTask.pauseResubscribeCheck();
+    this.receiveIndexTask.pauseIngestion();
+
```
ltrung, line 1055: race condition can still happen in other task like `ReceiveVideoInputTask` right?

hensmi-amazon: With `integrateUplinkPolicyDecision` ? I'm not too concerned about that impact since we are mostly relying on avg bitrate anyways.

ltrung: No, it can cause video to fail when calling choose video input multiple when the video is on (https://github.com/aws/amazon-chime-sdk-js/blob/master/src/task/ReceiveVideoInputTask.ts#L32). See this issue https://github.com/aws/amazon-chime-sdk-js/issues/1609

ltrung (approved):

PR #1844, opened 2021-12-01
Improve transcription integration test and run as Github workflow action

noahmako:
**Issue #:** N/A.

**Description of changes:** This change improves the stability of the transcription integration test and integrate it into the miscellaneous integration test GitHub workflow.

**Testing:** Tests pass with local selenium grid.

*Can these tested using a demo application? Please provide reproducible step-by-step instructions.* The change can be confirmed by clicking on the "Pre-recorded speech" option in the demo app and observing that the audio file plays exactly once and does not loop.

**Checklist:**

Yes.

No, this PR only touches integration tests & demo app functionality.

No, this PR only touches integration tests & demo app functionality.

### DIFF
--- a/.github/workflows/transcription-integration.yml
+++ b/.github/workflows/transcription-integration.yml
@@ -1,4 +1,4 @@
-  name: Miscellaneous Integration Workflow
+  name: Transcription Integration Workflow
 
   on:
     pull_request:
@@ -15,7 +15,7 @@
 
   jobs:
     integ-data-message:
-      name: Miscellaneous Integration Test
+      name: Transcription Integration Test
       runs-on: ubuntu-latest
 
       steps:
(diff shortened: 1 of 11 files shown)


ziyiz-amzn (approved):

integration/configs/transcription_test.config.json
```diff
@@ -10,13 +10,10 @@
         "noOfThreads": 1,
         "testImpl": "TranscriptionTest.js",
         "payload": {
-          "url": "http://localhost:8080/",
+          "url": "http://localhost:8080/?region=us-east-1",
+          "region": "us-east-1",
```
ltrung, line 14: Don't think you need this if you specify it in the URL already.

noahmako: The metrics get the region string from the payload rather than the query parameter.

ltrung: Then do you need the query parameter?

noahmako: Updated to only use the payload parameter and wired it up through the test.

.github/workflows/transcription-integration.yml
```diff
@@ -0,0 +1,79 @@
+  name: Transcription Integration Workflow
+
+  on:
+    pull_request:
+      branches:
+        - master
+        - release-1.x
```
ltrung, line 7: You don't need to run on release-1.x branch

.github/workflows/transcription-integration.yml
```diff
@@ -0,0 +1,79 @@
...
+    integ-data-message:
+      name: Transcription Integration Test
+      runs-on: ubuntu-latest
+
+      steps:
+        - name: Checkout Package
+          uses: actions/checkout@v2
+          with:
+            fetch-depth: 0
+        - name: Pack the Chime SDK and install the tarball into the Demo
```
ltrung, line 26: This should not be needed.

ltrung (approved):

PR #1865, opened 2021-12-09
Fix type comparison

hokyungh:
**Issue #:**

**Description of changes:**
`useChimeSDKMeetings` is getting from environment, so it is going to be string
We should compare against string `'true'`.
**Testing:**
Deployed my own instance and confirm that false will not use chime sdk meeting.

**Checklist:**

Y

N

N

### DIFF
--- a/demos/serverless/src/handlers.js
+++ b/demos/serverless/src/handlers.js
@@ -30,7 +30,7 @@ if(endpoint != 'https://service.chime.aws.amazon.com'){
 
 // return chime meetings SDK client just for Echo Reduction for now.
 function getClientForMeeting(meeting) {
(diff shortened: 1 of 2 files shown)


demos/serverless/src/handlers.js
```diff
@@ -30,7 +30,7 @@ if(endpoint != 'https://service.chime.aws.amazon.com'){
 
 // return chime meetings SDK client just for Echo Reduction for now.
 function getClientForMeeting(meeting) {
-  if(useChimeSDKMeetings){
```
harshitOberoi, line 33: can you also make the same change in demos/browser/server.js line 52

harshitOberoi (approved):

PR #1869, opened 2021-12-10
Upgrade the latest macOS version to Monterey for Safari test to test Safari 15

xuesichao:
**Issue #:**

The Sauce Labs releases the support of macOS Monterey recently. https://saucelabs.com/blog/macOS-12-with-safari-15-now-available-on-sauce-labs. We are now able to test our SDK on Safari 15 now. However currently we are testing it on macOS Big Sur with Safari 14.

**Description of changes:**
Update the `latest` version of macOS from `macOS 11.00` to `macOS 12` when the testing browser is Safari.

**Testing:**

N/A

**Checklist:**

N/A

No

No


integration/js/utils/WebdriverSauceLabs.js
```diff
@@ -9,7 +9,7 @@ const getPlatformName = capabilities => {
   switch (platform) {
     case 'MAC':
       if (browserName === 'safari') {
-        return version === 'latest' ? 'macOS 11.00' : 'macOS 10.13';
+        return version === 'latest' ? 'macOS 12' : 'macOS 10.13';
       }
       return 'macOS 10.14';
```
ltrung, line 14: can we also update this value to something else oter than 10.14 which is very old?

jing-chen1 (dismissed):

integration/js/utils/WebdriverSauceLabs.js
xuesichao: I'm not sure which one to use. I think this should be related to the OS we supported.

xuesichao: @ltrung Since the latest version is bumped up for one version, I bumped up this for one as well.

xuesichao: Tried to upgrade the default version to `macOS 10.15` or `macOS 11`. But either of them failed to pass the GitHub Action. So have to keep the default Mac version unchanged.

ltrung: I think we want to update from macOS 10.14 to BigSur. 10.14 is pretty old :)

jing-chen1 (approved):

ltrung (changes requested):
Can we update the default mac version too?

ltrung (approved):

PR #1889, opened 2021-12-17
Temporarily disable layers extension SDP munging for Chrome 97

hensmi-amazon:
**Issue #:** None

**Description of changes:** Just copying here:

So the issue occurs with JS SDK releases including[ this commit](https://github.com/aws/amazon-chime-sdk-js/commit/4076a6c8f337b7c7ba299ff314ac69db58fed46b) which added SDP munging for a specific header extension, and was released with 2.21.0 on 2021-11-01. The existing behavior with chrome was:

1. We add the header extension to the active camera section so that it could be negotiated with backend. Use of SDP munging for extensions is supported and recommended by libwebrtc maintainers for lack of explicit API support.
2. When the video is disabled, it seems that even though we don't re-add the line because it is an inactive section, chrome adds it anyways, but uses a different ID.
3. When video is re-enabled, we add the extension again.

This all works fine on < M97. However M97 includes [this change](https://webrtc.googlesource.com/src/+/b62ee8ce94e5f10e0a94d6f112e715cc4d0cd9dc) to fix a crash for use of Janus clients (though original fix is not really relevant to us). This change makes it a hard failure to try to change header extension IDs. So when chromium adds it with the new ID in the inactive section it causes a failure.

**Testing:** Smoke tested on Chrome 97 to confirm there is no longer any issue.

Only regression tests.

**Checklist:**

y

n

n


ltrung:
In the PR description, can we say that our fix right now is temporarily remove this logic?

ltrung:
So do we have a future plan to re-add this logic?

hensmi-amazon:
It's in the commit/title :).  And yes, but again it was for quality improvement but not activated. Wording is intentional to make clear its not going to lead to regressions.

And darn somehow removing code broke coverage...will fix

ltrung:
>This all works fine on < M97. However M98 includes this change to fix a crash for use of Janus clients (though original fix is not really relevant to us). This change makes it a hard failure to try to change header extension IDs. So when chromium adds it with the new ID in the inactive section it causes a failure.

Is this change for M97 or M98?

hensmi-amazon:
M97, just a typo in description.

simmkyu (approved):

fippo:
> Use of SDP munging for extensions is supported and recommended by libwebrtc maintainers for lack of explicit API support.

Do you have a reference for this? If such a statement exists anywhere in the documentation that documentation needs to be updated.

In general if you can get chrome/libwebrtc into a state where is attempts to reuse a header extension in a way that isn't valid (and thus prevented by @alvestrand's fix) filing an issue over at crbug.com would be much appreciated.

hensmi-amazon:
Hi @fippo :), pardon my quick words, I wasn't referring to any official documentation, just meant that in my time skimming webrtc bugs I hadn't seen any recommendations against munging/munging extensions and in libwebrtc it seems reasonably supported. It was definitely a bug on our end, we should have handled adding the extension consistently on media sections across negotiations.

It was a bit strange that upon munging the layers allocation extension, it would start readding it to inactive sections with a different ID (though stuff we should have handled in our code). I will look for a limited repro and post a bug when i get a chance.

fippo:
@hensmi-amazon weird-sdp-ids-misbehaving are my jam so this caught my attention.

You might want to express interesting in exposing proper APIs instead of having to rely on munging in https://bugs.chromium.org/p/chromium/issues/detail?id=1051821

PR #1921, opened 2022-01-10
Priority based downlink policy will not drop video instantly

bjason:
**Issue #:**

When using `VideoPriorityBasedPolicyConfig` in priority based downlink policy, the `networkIssueResponseDelayFactor` and `networkIssueRecoveryDelayFactor` are sometimes ignored and the stream is paused/unpaused immediately instead when a network event occurs. 

**Description of changes:**

No public API change. 

Changed the dependency of the network issue delays from the last subscribe time to last network event happen time, which will now reflect the fluctuation of network bandwidth.  

See change shown in the following videos (both using the default preset):

https://user-images.githubusercontent.com/14038310/152218645-d3287ff4-f0f3-44d8-b830-123cbb2ffcd9.mp4

_Previous logic, you could see the video turns off immediately after network goes bad in 0:17._

https://user-images.githubusercontent.com/14038310/152218704-be596789-0394-46e1-a048-11223b9228a5.mp4

_New logic, similar operation would not cause the same issue._

**Testing:**

Only regression tests.

**Checklist:**

yes

no

no


ltrung:
>When using the response_delay parameter in priority based downlink policy Instead, the response delay will be ignored and the stream is paused immediately instead during start up period.

What is `response_delay`? Did not find it in the code?

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -353,6 +353,14 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
         NetworkEvent.Decrease,
         numberOfParticipants
       );
+
+      // to prevent from instant drop due to the first-time dip
+      if (
+        this.startupPeriod &&
```
hensmi-amazon, line 359: Doesn't this still run into the issue where when you are out of the startup period, it will instantly drop down if there hasn't been a subscribe in a while?

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -337,29 +341,10 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
...
-        numberOfParticipants
-      );
-    } else if (currEstimated < prevEstimated) {
-      // if bw decreases, we use response delay
-      this.timeBeforeAllowSubscribeMs = this.getSubscribeDelay(
-        NetworkEvent.Decrease,
-        numberOfParticipants
-      );
-    }
+    let networkEventStartTimestamp = this.getNetworkEventStartTimestamp();
```
ltrung, line 344: nit: const instead of let

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -419,6 +404,41 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
     this.optimalReceiveStreams = chosenStreams.slice();
   }
 
+  private getNetworkEventStartTimestamp() {
+    const numberOfParticipants = this.subscribedReceiveSet.size();
+    const prevEstimated = this.prevDownlinkStats.bandwidthEstimateKbps;
```
hensmi-amazon, line 409: I don't think this creates the logic we want either since `prevDownlinkStats` and `downlinkStats` are continually updated. For example estimations [1000, 1200, 1200, 1200, ...] will lead to BWE never increases. [100, 10000, 10001, 10000 ...] will additionally be stuck at 100. I'm also confused with how if we see e.g. BWE going down, we are resetting the `decrease` timestamp, but setting the network event timestamp  to the increase timestamp.

I think there is a lot of state here and it can be pretty confusing. I also think this is becoming hard to test. I recommend making an `/* internal */` class which composes the logic we want, with its own unit tests, and clear comments.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -31,6 +31,101 @@ class LinkMediaStats {
...
+      networkEventStartTimestamp = this.networkIncreaseTimestamp;
+    } else {
+      // if bw is stable, we want to look back to find when it starts to become the current state
+      // e.g. [400, 500, 400, 400, 300, *300*], we are at the last 300. We know the bw is decreasing since point 500, which is stored in networkEventChangeTimestamp
+      networkEventStartTimestamp = this.networkEventChangeTimestamp;
+    }
+
+    if (
+      (previousNetworkEvent &&
+        previousNetworkEvent !== this.currentNetworkEvent) /* network state is changing */ ||
```
simmkyu, line 89: optional (don't need to modify): I don't think some comments are unnecessary as the code is self-explaining well.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -31,6 +31,101 @@ class LinkMediaStats {
...
+      this.networkEventChangeTimestamp === 0 /* running the first time */
+    ) {
+      // set networkEventChangeTimestamp to now since network state is changing
+      this.networkEventChangeTimestamp = Date.now();
+    }
+
+    // if networkEventStartTimestamp is not defined, it means the bw is increasing/decreasing ever since the beginning.
+    // e.g. [300, 400, *500*], we are at 500. We set networkEventStartTimestamp to networkDecreaseTimestamp (undefined) above.
+    // So here, we set it to the beginning, which is networkEventChangeTimestamp (explanation above)
+    if (!networkEventStartTimestamp) networkEventStartTimestamp = this.networkEventChangeTimestamp;
```
simmkyu, line 99: When `networkEventStartTimestamp` is `0` (else case), this condition will also be `true`.

```
if (!0) {
   // Executed!
}
```

If this is intentional, update the comment, e.g., `... is 0 or undefined`. If not, explicitly check for `undefined`.

```
if (typeof networkEventStartTimestamp !== 'undefined') {
```

test/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.test.ts
```diff
@@ -1220,7 +1245,7 @@ describe('VideoPriorityBasedPolicy', () => {
       resub = policy.wantsResubscribe();
       expect(resub).to.equal(false);
 
-      incrementTime(5100);
+      incrementTime(8100);
```
simmkyu, line 1248: I recommend keeping existing tests to ensure that your change is backward compatible.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -336,29 +304,13 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
...
-        numberOfParticipants
-      );
-    }
+    const currentEstimated = this.downlinkStats.bandwidthEstimateKbps;
 
     // If no major changes then don't allow subscribes for the allowed amount of time
     if (
       noMajorChange &&
-      Date.now() - this.lastSubscribeTimestamp < this.timeBeforeAllowSubscribeMs
+      !this.videoPriorityBasedPolicyConfig.allowSubscribe(numberOfParticipants, currentEstimated)
```
hensmi-amazon, line 313: Cool! Any time we are able to abstract away a good chunk of code and state (technically didn't *remove* state but we probably would have needed to add it) which results in a diff like this is a good win. 

As a first time reader, one now doesn't need to clutter their attempts to initially understand the `VideoPriorityBasedPolicy` with the supplementary adaption delay configuration (which they can save for another time :).

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicyConfig.ts
```diff
@@ -1,6 +1,13 @@
...
 
+/** @internal */
+const enum NetworkEvent {
+  Stable,
+  Decrease,
+  Increase,
+}
+
 /**
  * [[VideoPriorityBasedPolicyConfig]] contains the network issue response delay and network issue recovery delay.
```
hensmi-amazon, line 12: This probably should be expanded to note the additional behavior added.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicyConfig.ts
```diff
@@ -20,6 +27,9 @@ export default class VideoPriorityBasedPolicyConfig {
     public networkIssueResponseDelayFactor: number = 0,
     public networkIssueRecoveryDelayFactor: number = 0
   ) {
+    this.currentNetworkEvent = NetworkEvent.Stable;
```
hensmi-amazon, line 30: This can be set on the same line as declaration.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicyConfig.ts
```diff
@@ -34,4 +44,80 @@ export default class VideoPriorityBasedPolicyConfig {
     }
     this.networkIssueRecoveryDelayFactor = networkIssueRecoveryDelayFactor;
   }
+
+  private static readonly MINIMUM_DELAY = 2000;
```
hensmi-amazon, line 48: Nit: I understand why this is put down here but lets just put all these constants and members above the constructor for consistency with code base.

hensmi-amazon (commented):
Looks good! just a few nits

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -336,33 +301,20 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
...
     if (
       noMajorChange &&
-      Date.now() - this.lastSubscribeTimestamp < this.timeBeforeAllowSubscribeMs
+      this.probeFailed &&
+      !this.videoPriorityBasedPolicyConfig.allowSubscribe(numberOfParticipants, currentEstimated)
     ) {
       return;
     }
 
+    this.probeFailed = false;
```
hensmi-amazon, line 316: Nit: This needs a comment. Sometimes things are self-explanatory, sometimes not. But also in a class with this much mutable state I always recommend going overkill with the comments.

test/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.test.ts
hensmi-amazon: Worth noting since release we have considered small behavior modifications to the priority policy to be acceptable since we label that behavior may change. But I do think that when you have to adjust an existing test we should leave a PR comment explaining why it needed to change :).

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
hensmi-amazon: i.e. when I see this in this diff my thought is immediately 'why' :P

test/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.test.ts
bjason: I see. Since we are changing from comparing to the last subscribe timestamp to comparing a reference timestamp, the behavior will be changed. Other than the description in this PR, do we need to state the change somewhere?

hensmi-amazon: In PR is fine as its specific to the diff. e.g. my PRs almost always have 1+ self comments on them immediately on post :).

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -336,33 +301,22 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
...
       noMajorChange &&
-      Date.now() - this.lastSubscribeTimestamp < this.timeBeforeAllowSubscribeMs
+      this.probeFailed &&
+      !this.videoPriorityBasedPolicyConfig.allowSubscribe(numberOfParticipants, currentEstimated)
     ) {
       return;
     }
 
+    // when probe failed, we set timeBeforeAllowSubscribeMs to 3x longer
+    // ince we have passed the window now, we will try to probe again
```
ltrung, line 317: nit: typo `ince` -> `since`

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -301,14 +269,11 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
     // subscribe interval will be changed if probe failed, will take this temporary interval into account
     if (
       noMajorChange &&
-      this.probeFailed &&
       Date.now() - this.lastSubscribeTimestamp < this.timeBeforeAllowSubscribeMs
```
ltrung, line 272: Is `this.timeBeforeAllowSubscribeMs` always be `VideoPriorityBasedPolicy.MIN_TIME_BETWEEN_SUBSCRIBE_MS;` now?

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -301,14 +269,11 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
     // subscribe interval will be changed if probe failed, will take this temporary interval into account
```
ltrung, line 269: Need to update this comment right? It mention probe failed but you remove it.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -336,33 +301,22 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
...
-    } else if (currEstimated < prevEstimated) {
-      // if bw decreases, we use response delay
-      this.timeBeforeAllowSubscribeMs = this.getSubscribeDelay(
-        NetworkEvent.Decrease,
-        numberOfParticipants
-      );
-    }
+    const currentEstimated = this.downlinkStats.bandwidthEstimateKbps;
 
     // If no major changes then don't allow subscribes for the allowed amount of time
```
ltrung, line 307: Can we update this comment regarding probeFailed?

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
bjason: `timeBeforeAllowSubscribeMs` will be used in probing: if probe failed, this time window will be extended.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
bjason: Thanks!

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
bjason: I took a second look on the comments and updated some. Thanks for bringing it up.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicyConfig.ts
```diff
@@ -1,10 +1,20 @@
...
+  Stable,
+  Decrease,
+  Increase,
+}
+
 /**
  * [[VideoPriorityBasedPolicyConfig]] contains the network issue response delay and network issue recovery delay.
  */
 export default class VideoPriorityBasedPolicyConfig {
+  private static readonly MINIMUM_DELAY = 2000;
```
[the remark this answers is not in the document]
simmkyu: NVM. Addressed in the latest revision.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicyConfig.ts
```diff
@@ -34,4 +49,69 @@ export default class VideoPriorityBasedPolicyConfig {
...
+      );
+
+      if (previousNetworkEvent !== NetworkEvent.Decrease) {
+        this.bandwidthDecreaseTimestamp = Date.now();
+      } else if (Date.now() - this.bandwidthDecreaseTimestamp > timeBeforeAllowSubscribeMs) {
+        this.referenceBitrate = currentEstimated;
+        return true;
+      }
+      return false;
+    } else {
```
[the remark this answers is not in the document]
bjason: yes, it returns the same bandwidth consecutively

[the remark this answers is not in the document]
simmkyu: Thanks for checking.

hensmi-amazon (approved):
🚀🚀

PR #1924, opened 2022-01-11
Fix deploy-canary-demo script to add media capture bucket prefix

nehtambe:
**Issue #:**
Current JS SDK serverless demo apps were not deployed with -o parameter (media capture bucket prefix). In order to test media capture, we need to pass this parameter, which creates S3 buckets to store the capture artifacts. 

**Description of changes:**

1.  Add --o parameter required to use media capture.

**Testing:**

Yes, deployed demo app with new parameter -o from local to personal account. 

**Checklist:**

Yes

No

No


demos/serverless/deploy.js
```diff
@@ -192,39 +191,50 @@ function createCaptureS3Buckets(bucketPrefix, regions) {
...
-          ]
-        }
-      }]
-    });
-
-    fs.writeFileSync('build/bucket_policy.json', bucketPolicy, {encoding: 'utf8', flag: 'w'});
+  if (JSON.parse(useChimeSDKMeetings)) {
+    console.log(`Creating S3 buckets for media capture pipelines artifacts.  Bucket prefix: ${bucketPrefix} Regions:[${region}, "us-east-1"]`);
+    createS3Bucket(bucketPrefix, region);
+    createS3Bucket(bucketPrefix, "us-east-1");
```
ltrung, line 197: Instead of hard-coding, can we just pass in us-east-1 as a command parameter?

nehtambe: I think instead of passing as a parameter, we can create a constant - defaultMediaCaptureRegion= "us-east-1"

ltrung: But why does it has to be us-east-1 though ?
>Create media capture S3 buckets only in us-east-1 and in the respective region, instead of creating in all 14 regions. We need to test only in these two for new ChimeSDK meetings.

This can't be the reason. This is a demo code and we should refrain from editing the code for a specific testing purpose.

nehtambe: Creating bucket in us-east-1 region too because Gamma Chime SDK meetings can not be placed in other regions than us-east-1. So, if we want to run media capture tests with Demo app Urls pointing to Gamma Chime SDK meetings, we need to place meeting in us-east-1 and create media capture bucket in us-east-1 region.

ltrung: All of that seems like can be controls through commands? From an external perspective, there is no such thing as a gamma endpoint.

nehtambe: Yes, right. Good point. From external perspective, we should be able to start media capture in any media region. I will revert the code.

ltrung (dismissed):

script/deploy-canary-demo
```diff
@@ -26,9 +26,9 @@ ChimeSDKMeetingsBeta)
...
-    npm run deploy -- -r ap-southeast-1 -b chime-sdk-meetings-demo-beta-canary-ap-southeast-1 -s chime-sdk-meetings-demo-beta-canary-ap-southeast-1 -p $GAMMA_CHIME_SERVICE_PRINCIPAL -c $GAMMA_CHIME_ENDPOINT_AP_SOUTHEAST_1 -t -l
+    npm run deploy -- -r us-east-1 -b chime-sdk-meetings-demo-beta-canary-us-east-1 -s chime-sdk-meetings-demo-beta-canary-us-east-1 -o chime-sdk-meetings-iad-demo-beta-canary -p $GAMMA_CHIME_SERVICE_PRINCIPAL -c $GAMMA_CHIME_ENDPOINT_US_EAST_1 -t -l
+    npm run deploy -- -r eu-central-1 -b chime-sdk-meetings-demo-beta-canary-eu-central-1 -s chime-sdk-meetings-demo-beta-canary-eu-central-1 -o chime-sdk-meetings-fra-demo-beta-canary -p $GAMMA_CHIME_SERVICE_PRINCIPAL -c $GAMMA_CHIME_ENDPOINT_EU_CENTRAL_1 -t -l
+    npm run deploy -- -r ap-southeast-1 -b chime-sdk-meetings-demo-beta-canary-ap-southeast-1 -s chime-sdk-meetings-demo-beta-canary-ap-southeast-1 -o chime-sdk-meetings-sin-demo-beta-canary -p $GAMMA_CHIME_SERVICE_PRINCIPAL -c $GAMMA_CHIME_ENDPOINT_AP_SOUTHEAST_1 -t -l
```
michhyun1, line 31: can you take a look at the `-l` option? Maybe we might want to hide the output of the s3 buckets endpoints when running these on github actions.

nehtambe: Yes, disabled printing logs for S3 bucket creation.

michhyun1 (approved):

PR #1934, opened 2022-01-14
Fix delay on simulcast uplink stream switches

hensmi-amazon:
**Issue #:** None

**Description of changes:** 

`updateLocalVideoFromPolicy` wasn't really working as expected for two reasons:
* `encodingSimulcastLayersDidChange` is not called until `chooseEncodingParameters` so it doesn't make sense to call before. Reversed the ordering here.
* `SimulcastUplinkObserver.encodingSimulcastLayersDidChange` is async so it would likely not be updated in time. Made this synchronous.
* `AttachMediaInputTask` and `ReceiveVideoInputTask` are run async (async on async!) so `chooseEncodingParameters` would also not be called in time.

**Testing:**
Smoke tested with three simulcast chrome senders with priority. Smoke tested with three non-simulcast chrome senders. Removed the third sender and there is no longer a long video freeze. There is a small one but that will not likely be fixable until we wire up the layers allocation extension again. This only occurs on participant number changes < 8 so its not very frequent.

* 3 chrome senders using simulcast.
* One leaves conference
* No freezes on other applications

**Checklist:**

y

y

y

### DIFF
--- a/test/audiovideocontroller/DefaultAudioVideoController.test.ts
+++ b/test/audiovideocontroller/DefaultAudioVideoController.test.ts
@@ -1490,7 +1490,7 @@ describe('DefaultAudioVideoController', () => {
         .false;
     });
 
-    it('will skip renegotiation if we are only completing simulcast stream encoding change', async () => {
+    it('will skip renegotiation if simulcast streams do not change', async () => {
       const logger = new NoOpDebugLogger();
       const loggerSpy = sinon.spy(logger, 'info');
       configuration.enableUnifiedPlanForChromiumBasedBrowsers = true;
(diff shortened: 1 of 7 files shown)


test/audiovideocontroller/DefaultAudioVideoController.test.ts
```diff
@@ -1539,17 +1529,14 @@ describe('DefaultAudioVideoController', () => {
...
       audioVideoController.meetingSessionContext.transceiverController = new TestTransceiverController();
-      // @ts-ignore
-      audioVideoController.meetingSessionContext.signalingClient = new TestSignalingClient(
-        webSocketAdapter,
-        new NoOpDebugLogger()
-      );
 
+      // One that requires subscribe
+      audioVideoController.update({ needsRenegotiation: false });
+      // One that doesn't
```
ltrung, line 1535: this code confuses me :)

ltrung (approved):

test/audiovideocontroller/DefaultAudioVideoController.test.ts
hensmi-amazon: The unit testing war against async code (and components with many dependencies) will likely never be won. But this unit test was originally confirming the bug in the code :P ...

ltrung: So the confusing part is `needsRenegotiation` is both set to false but the comment says one requires subcribe and one doesn't.  :D

hensmi-amazon: `needsRenegotiation` could also have been named `forceRenegotiation` :)

Issue #1946, opened 2022-01-20
The video bitrate beyond the selected value?


dathacky:
I noticed that the bitrate value during the meeting exceeds the bitrate value previously selected, is this a bug? If not please explain bitrate to help me. Thank you!

<img width="395" alt="Screen Shot 2022-01-20 at 16 42 56" src="https://user-images.githubusercontent.com/37299929/150314800-ef2c9c4b-0eb8-4260-b656-a0401d0c9563.png">
<img width="758" alt="Screen Shot 2022-01-20 at 16 45 07" src="https://user-images.githubusercontent.com/37299929/150314900-7b442253-32e4-4514-a8e0-52a849a16ad5.png">

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

Join meeting and select resolution 360p 15fps 600 kbps max

### Amazon Chime SDK for JavaScript version

2.25.0

### What browsers are you seeing the problem on?

Chrome 

### Browser version

Version 97.0.4692.71 (Official Build) (x86_64)

### Browser console logs

No

ltrung:
@dathacky Hi, thanks for reporting. We are currently looking at this issue. Could you let us know how you reproduced it in our meeting demo?

dathacky:
> @dathacky Hi, thanks for reporting. We are currently looking at this issue. Could you let us know how you reproduced it in our meeting demo?

I joined the meeting and select resolution 360p 15fps 600 kbps max

ltrung:
@dathacky Can you still reproduce this issue? We pushed out some fixes in 2.26 version.

dathacky:
it's fixed, thank you

PR #1949, opened 2022-01-21
Fix the bug that the max bandwidth is ignored

simmkyu:
**Issue #:**
https://github.com/aws/amazon-chime-sdk-js/issues/1946

**Description of changes:**

When retrying the connection, the SDK resets the transceiver controller, which is responsible for setting the max bandwidth on `RTCRtpTransceiver`.
https://github.com/aws/amazon-chime-sdk-js/blob/4d779e41136d927da7816e38c5c8cd6535f40897/src/task/CleanRestartedSessionTask.ts#L18

At the end of reconnection, the SDK calls `NScaleVideoUplinkBandwidthPolicy.updateTransceiverController` to update the max bandwidth on a new transceiver controller. However, `NScaleVideoUplinkBandwidthPolicy` sees that the max bandwidth remains the same, so it does not update the max bandwidth.
https://github.com/aws/amazon-chime-sdk-js/blob/4d779e41136d927da7816e38c5c8cd6535f40897/src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts#L172-L180

**Testing:**

1. Open the Chime SDK serverless demo, type any meeting info, and navigate to the **Select devices** page.
1. Open the browser console and run the following to set the maxBandwidth to 700 Kbps.
    ```
    app.audioVideo.chooseVideoInputQuality(640, 360, 15, 700)
    ```
1. Choose **Join** to join a meeting and enable the WebRTC stats button.
1. Turn on your video and hover your mouse over a video tile. You can see the **Bitrate (bps)** is around 700,000.
1. In the browser console, run the following to force reconnecting the session.
    ```
    app.audioVideo.audioVideoController.handleMeetingSessionStatus({ isFailure: () => true, isTerminal: () => false, statusCode: () => 17, isAudioConnectionFailure: () => false, toString: () => 'Error' });
    ```
1. **The **Bitrate (bps)** should not go higher.**

**Checklist:**

### DIFF
--- a/src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
+++ b/src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
@@ -167,27 +167,21 @@ export default class NScaleVideoUplinkBandwidthPolicy implements VideoUplinkBand
     if (!settings) {
       return;
     }
-
     const encodingParams: RTCRtpEncodingParameters = this.calculateEncodingParameters(settings);
-    if (
-      !this.encodingParametersHaveSameBitrateAndScaling(
-        encodingParams,
-        this.encodingParamMap.get(NScaleVideoUplinkBandwidthPolicy.encodingMapKey)
-      )
-    ) {
+    if (this.shouldUpdateEndcodingParameters(encodingParams)) {
       this.encodingParamMap.set(NScaleVideoUplinkBandwidthPolicy.encodingMapKey, encodingParams);
       this.transceiverController.setEncodingParameters(this.encodingParamMap);
     }
   }
 
-  private encodingParametersHaveSameBitrateAndScaling(
-    encoding1: RTCRtpEncodingParameters,
-    encoding2: RTCRtpEncodingParameters
-  ): boolean {
+  private shouldUpdateEndcodingParameters(encoding: RTCRtpEncodingParameters): boolean {
+    /* istanbul ignore next: sender, getParameters, and encodings are optional */
+    const transceiverEncoding = this.transceiverController
+      ?.localVideoTransceiver()
+      ?.sender?.getParameters()?.encodings?.[0];
     return (
-      encoding1 === encoding2 ||
-      (encoding1.maxBitrate === encoding2.maxBitrate &&
-        encoding1.scaleResolutionDownBy === encoding2.scaleResolutionDownBy)
+      encoding.maxBitrate !== transceiverEncoding.maxBitrate ||
+      encoding.scaleResolutionDownBy !== transceiverEncoding.scaleResolutionDownBy
     );
   }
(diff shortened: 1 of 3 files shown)


src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts
```diff
@@ -167,27 +167,21 @@ export default class NScaleVideoUplinkBandwidthPolicy implements VideoUplinkBand
...
     }
   }
 
-  private encodingParametersHaveSameBitrateAndScaling(
-    encoding1: RTCRtpEncodingParameters,
-    encoding2: RTCRtpEncodingParameters
-  ): boolean {
+  private shouldUpdateEndcodingParameters(encoding: RTCRtpEncodingParameters): boolean {
+    /* istanbul ignore next: sender, getParameters, and encodings are optional */
+    const transceiverEncoding = this.transceiverController
```
ltrung, line 179: It is the same thing but can we loop through the encoding array and check the key since we do the same in transceiver controller.

simmkyu: Could you elaborate more on how to check the key?

https://github.com/aws/amazon-chime-sdk-js/blob/4d779e41136d927da7816e38c5c8cd6535f40897/src/transceivercontroller/DefaultTransceiverController.ts#L37-L56

ltrung: I think you can just check if `encoding[i].rid === encodingMapKey`.

simmkyu: The value of `encodingMapKey` is `video`, but `rid` is not available.

```
active: true
adaptivePtime: false
maxBitrate: 601000
networkPriority: "low"
priority: "low"
scaleResolutionDownBy: 1
```

1. NScale passes a map of <encodingMapKey, object>.
   https://github.com/aws/amazon-chime-sdk-js/blob/4d779e41136d927da7816e38c5c8cd6535f40897/src/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.ts#L179
2. The transceiver controller only uses the values but not `encodingMapKey`.
   https://github.com/aws/amazon-chime-sdk-js/blob/4d779e41136d927da7816e38c5c8cd6535f40897/src/transceivercontroller/DefaultTransceiverController.ts#L31

ltrung: My bad. We only use that as a set in the video policy itself, the encoding params does not have it.

ltrung (approved):

PR #1951, opened 2022-01-21
Fix the bug that the max bandwidth is ignored if set before starting a meeting

simmkyu:
**Issue #:**
https://github.com/aws/amazon-chime-sdk-js/issues/1946

**Description of changes:**
- If the builder calls `chooseVideoInputQuality` API before starting a meeting, the SDK does not apply the max bandwidth to a local video.

**Testing:**

I created a minimal Chime SDK app to test the scenario.

**Checklist:**

Yes

No

No

### DIFF
--- a/test/audiovideocontroller/DefaultAudioVideoController.test.ts
+++ b/test/audiovideocontroller/DefaultAudioVideoController.test.ts
@@ -805,7 +805,8 @@ describe('DefaultAudioVideoController', () => {
       );
 
       // use this opportunity to verify that these can be called before start
-      audioVideoController.setVideoMaxBandwidthKbps(100);
+      const MAX_BANDWIDTH_KBPS = 100;
+      audioVideoController.setVideoMaxBandwidthKbps(MAX_BANDWIDTH_KBPS);
       audioVideoController.handleHasBandwidthPriority(false);
 
       let sessionStarted = false;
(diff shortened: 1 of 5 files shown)


ltrung:
Can we link to this issue https://github.com/aws/amazon-chime-sdk-js/issues/1946

test/audiovideocontroller/DefaultAudioVideoController.test.ts
```diff
@@ -825,12 +826,15 @@ describe('DefaultAudioVideoController', () => {
       await delay(defaultDelay);
       expect(sessionStarted).to.be.true;
       expect(sessionConnecting).to.be.true;
-      audioVideoController.setVideoMaxBandwidthKbps(100);
       audioVideoController.handleHasBandwidthPriority(false);
       audioVideoController.handleHasBandwidthPriority(true);
       audioVideoController.handleHasBandwidthPriority(true);
```
xuesichao, line 831: This is not related to this PR. But I'm curious why we need two of this?

simmkyu: Looking at the `expect` lines. I don't think calling `handleHasBandwidthPriority` multiple times is not required. It might be for idempotence check.

Issue #1964, opened 2022-01-27
Amazon Chime SDK JS V3 Plan


ltrung:
The Amazon Chime SDK JS team is planning for the next major version v3.0.0 to:
- Improve device workflow by decoupling the device controller from audio video controller, better high level API to handle audio/video permission mode when joining. 
- Improve our client metrics and audio/video bandwidth events.
- Decouple event controller so that it can be used out of the meeting context. 
- Improve audioVideo.update to make sure it is idempotent and prevent race condition when switching between different video inputs.
- Other minor clean up code.

We will update here when we have more details to share. Please let us know if you are looking forward to any of the upcoming changes by commenting or adding a 👍 reaction to the issue.

Please refer to the [v3 Migration guide](https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/17_Migration_to_3_0.md) on how to migrate to v3.

If you want to try out the beta version, please use `npm install amazon-chime-sdk-js@beta` to get the latest beta version.  

# Community Note
* Please vote for this issue by adding a 👍 reaction to the issue. This will help the community and maintainers from the Amazon Chime SDK team to prioritize this request.
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.

ltrung:
[amazon-chime-sdk-js@v3.0.0-beta.0](https://www.npmjs.com/package/amazon-chime-sdk-js/v/3.0.0-beta.0) has been released. You can find more info from [CHANGELOG.md](https://github.com/aws/amazon-chime-sdk-js/blob/v3.0.0-beta.0/CHANGELOG.md#300-beta0---2022-02-08).
We also start a [migration doc](https://github.com/aws/amazon-chime-sdk-js/blob/v3.0.0-beta.0/guides/17_Migration_to_3_0.md) from v2 to v3 that will be updated with all the changes that are not backward compatible. 
The main changes in the first beta build is:
- Decouple event controller from audio video controller so that it does not depend on the audio video lifecycle.
- Clean up a lot of unused/deprecated code. Major thing is remove support for Safari 12 as well as planB.

ltrung:
[amazon-chime-sdk-js@v3.0.0-beta.1](https://www.npmjs.com/package/amazon-chime-sdk-js/v/3.0.0-beta.1) has been released. The main change in this beta build is:
[...]

PR #1969, opened 2022-01-29
Decoupled EventController from AudioVideo and MeetingSession

LeviSklaroff:
**Issue #:**
Currently EventController is dependent on AudioVideoController, which prevents it from being used after Meeting has ended and limits what the EventController can be used for.

**Description of changes:**
- Removed AudioVideoController as dependency from EventController
- Removed eventDidRecieve from AudioVideoObserver and put it into a new EventObserver interface
- Updated DeviceController to take in EventController as an optional dependency 
- Updated all usage of EventController

**Testing:**
- Tested that events are properly received in the meeting demo
- Tested edge cases such as giving or not giving certain parameters using unit tests

- Deploy meeting demo
- Open developer console
- Create/Join a meeting
- Confirm that events are being received in the console.

**Checklist:**

Yes

Yes, and yes

Yes


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 ### Changed
+- Decoupled EventController from AudioVideo and MeetingSession
```
hensmi-amazon, line 15: I'm assuming we are going to have an in depth transition guide so its not a big deal to not go into details here.

docs/interfaces/meetingsession.html
```diff
@@ -91,7 +91,7 @@ <h3>Properties</h3>
 								<li class="tsd-kind-property tsd-parent-kind-interface"><a href="meetingsession.html#configuration" class="tsd-kind-icon">configuration</a></li>
 								<li class="tsd-kind-property tsd-parent-kind-interface"><a href="meetingsession.html#contentshare" class="tsd-kind-icon">content<wbr>Share</a></li>
 								<li class="tsd-kind-property tsd-parent-kind-interface"><a href="meetingsession.html#devicecontroller" class="tsd-kind-icon">device<wbr>Controller</a></li>
-								<li class="tsd-kind-property tsd-parent-kind-interface"><a href="meetingsession.html#eventreporter" class="tsd-kind-icon">event<wbr>Reporter</a></li>
```
hensmi-amazon, line 94: @ltrung totally random but i just wanted to check....is there any other way to do docs that doesn't require committing the regenerated along with every commit? :P

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -217,6 +216,9 @@ export default class DefaultAudioVideoController
   get mediaStreamBroker(): MediaStreamBroker {
     return this._mediaStreamBroker;
   }
+  get eventController(): EventController {
```
hensmi-amazon, line 219: nit: whitespace

src/eventcontroller/EventController.ts
```diff
@@ -1,16 +1,34 @@
...
+import EventReporter from '../eventreporter/EventReporter';
 import AudioVideoEventAttributes from './AudioVideoEventAttributes';
 import DeviceEventAttributes from './DeviceEventAttributes';
 import EventName from './EventName';
 import MeetingHistoryState from './MeetingHistoryState';
 
 /**
- * [[EventController]] keeps track of events and notifies audio-video observers.
- * An event describes the success and failure conditions for the meeting session.
+ * [[EventController]] keeps track of a list of event observers and notifies them of SDK events.
```
hensmi-amazon, line 12: We probably can call out that it can also, call out to an external service with the events since we do that in our default implementation.

src/eventcontroller/EventController.ts
```diff
@@ -1,16 +1,34 @@
...
+
+  /**
+   * Removes an observer for event published to this controller.
+   */
+  removeObserver(observer: EventObserver): void;
+
+  /**
+   * Passes each observer of this controller to a given function.
+   */
+  forEachObserver(observerFunc: (observer: EventObserver) => void): void;
```
hensmi-amazon, line 30: Does this need to be public?

src/eventcontroller/EventController.ts
```diff
@@ -23,4 +41,5 @@ export default interface EventController {
    * Adds a state to the meeting history stack.
```
hensmi-amazon, line 41: Can we improve this docstring? i.e. I don't know what this does.

src/eventcontroller/EventController.ts
```diff
@@ -1,16 +1,34 @@
...
 import AudioVideoEventAttributes from './AudioVideoEventAttributes';
 import DeviceEventAttributes from './DeviceEventAttributes';
 import EventName from './EventName';
 import MeetingHistoryState from './MeetingHistoryState';
 
 /**
- * [[EventController]] keeps track of events and notifies audio-video observers.
- * An event describes the success and failure conditions for the meeting session.
+ * [[EventController]] keeps track of a list of event observers and notifies them of SDK events.
+ * An event can describe success or failure of SDK components that make use of this controller.
```
hensmi-amazon, line 13: I feel like this info makes more sense on `publishEvent`.

src/eventcontroller/EventController.ts
```diff
@@ -1,16 +1,34 @@
...
+ * Example: The success and failure for a meeting session.
```
hensmi-amazon, line 14: As we have chosen (I think it makes sense to, but not super familiar), to make this injectable, can we call out why or why not a builder may want to inject their own or more likely just use the default implementation?

src/eventcontroller/EventController.ts
```diff
@@ -23,4 +41,5 @@ export default interface EventController {
    * Adds a state to the meeting history stack.
    */
   pushMeetingState(state: MeetingHistoryState): Promise<void>;
+  readonly eventReporter?: EventReporter;
```
hensmi-amazon, line 44: Docstring plz

src/eventobserver/EventObserver.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import EventAttributes from '../eventcontroller/EventAttributes';
+import EventName from '../eventcontroller/EventName';
+
+export default interface EventObserver {
+  /**
+   * Called when specific events occur during the meeting and includes attributes of the event. This can be used to
+   * create analytics around meeting metric.
```
hensmi-amazon, line 10: Is 'metrics' the right word here? We don't want to confuse builders with that callback.

src/eventobserver/EventObserver.ts
```diff
@@ -0,0 +1,13 @@
...
+   */
+  eventDidReceive?(name: EventName, attributes: EventAttributes): void;
```
hensmi-amazon, line 12: Lol, if this is the only function...probably doesn't have to be optional right? @simmkyu

src/meetingsession/DefaultMeetingSession.ts
```diff
@@ -49,9 +46,14 @@ export default class DefaultMeetingSession implements MeetingSession, Destroyabl
...
-
-    this.setupEventReporter(configuration, logger, eventReporter);
-    this._deviceController = deviceController;
+    if (eventController) {
+      this._eventController = eventController;
+    } else {
+      this._eventController = new DefaultEventController(configuration, logger);
+    }
+    if (!deviceController.getEventController()) {
+      deviceController.setEventController(this.eventController);
```
hensmi-amazon, line 55: This seems odd, and I'm not sure its totally necessary given how few builders are going to be injecting these (would be nice to get rid of the getter/setter on the device controller). But maybe this will look different when we decouple device controller so I won't think about it too much. But i wouldn't mind a `TODO: Consider whether this is necessary`.

src/eventcontroller/DefaultEventController.ts
```diff
@@ -15,7 +22,7 @@ import EventName from './EventName';
 import flattenEventAttributes from './flattenEventAttributes';
 import MeetingHistoryState from './MeetingHistoryState';
 
-export default class DefaultEventController implements EventController {
+export default class DefaultEventController implements EventController, Destroyable {
```
hensmi-amazon, line 25: Can we call out a few sentences of implementation details? In particular the usage of `EventReporter`

hensmi-amazon (commented):
Nice, simple change with immediate builder/dev benefit. Great stuff, just have some nits on documentation I think we should improve while we are looking at it :).

CHANGELOG.md
LeviSklaroff: Yes, although that actually is supposed to be part of this commit, I've gone ahead and added it as `guides/17_Migration_to_3_0.md`

src/audiovideocontroller/DefaultAudioVideoController.ts
LeviSklaroff: Fixed

src/eventcontroller/EventController.ts
LeviSklaroff: That's a good idea, added.

src/eventcontroller/EventController.ts
LeviSklaroff: Perhaps not, but because the same function on `AudioVideoController` is public builder might have been calling their own observerFunction that uses `eventDidRecieve` so it's best to make it public to avoid difficult migration where they would have to then make their own `EventController`

src/eventcontroller/EventController.ts
LeviSklaroff: Added

src/eventobserver/EventObserver.ts
LeviSklaroff: I think `metrics` is the right word, but I do think we can be a bit more explicit that analytics are just one of the different use cases and that it's a example so I updated the comment

src/eventobserver/EventObserver.ts
LeviSklaroff: Lol yeah you're right, really no need to make it optional.

src/meetingsession/DefaultMeetingSession.ts
LeviSklaroff: This is a bit of a odd case, but because we want to decouple the `DeviceController` from the `MeetingSession` we need it to be able to have it's own `EventController`, but we also want for the `DeviceController` to be able to easily use the same `EventController` as the `MeetingSession` meaning a builder shouldn't have to create their own `EventController` for this to be possible so there's a need to be able to set the controller after creation. However, there is also a possibility that a builder might want to be able to have different `EventController`'s for `DeviceController` and `MeetingSession` so we need to also check to make sure there isn't already one on `DeviceController`

src/eventcontroller/DefaultEventController.ts
LeviSklaroff: Added a few comments, let me know if you think there's any parts that are confusing that I should clarify.

src/eventcontroller/EventController.ts
LeviSklaroff: Updated

src/eventcontroller/EventController.ts
LeviSklaroff: I think it makes more sense to be on the controller to help contextualize better what the controller is actually doing, if I'm thinking about wanting to use the `EventController` I'd look at the documentation for the interface and see that I can give observers that will get notified by events but I don't really know what those events are unless I looked at the documentation on publishing events but typically builders will be observing rather then publishing.

src/eventcontroller/EventController.ts
LeviSklaroff: Hmm, I think that might be better for a blog post or something similar. We don't typically document why a builder might want to inject their own version of an interface.

hensmi-amazon: I think many builders would like to live off doc strings rather then possibly out of date blog posts. I'm also just thinking a few sentences. If we don't want builders injecting we should call that out too. @ltrung @simmkyu what do you think?

> We don't typically document why a builder might want to inject their own version of an interface.

'We don't currently do it' !== 'Good practice' lol.

src/meetingsession/DefaultMeetingSession.ts
hensmi-amazon: It seems to me like those may be phantom use cases, or a case where we are trying a bit too hard to save builders who are already doing complicated things (like injecting) a single line of code or requirement, but I will still wait until the device controller stuff before nagging more, since I'm not really familiar with how that will look.

src/eventcontroller/EventController.ts
hensmi-amazon: Fair enough. Maybe makes sense to just link the guide here too if we are trying to maximize contextualization :).

src/eventobserver/EventObserver.ts
hensmi-amazon: With current documentation at a first glance I immediately think that this function has some synergy with `AudioVideoObserver.metricsDidReceive` (which it doesn't) if my comment wasn't clear. I still have that thought but maybe its only me :P.

guides/17_Migration_to_3_0.md
```diff
@@ -0,0 +1,125 @@
...
+```shell
+npm install --save amazon-chime-sdk-js@3
+```
+
+## Interface changes
+
+Version 3 of the Amazon Chime SDK for JavaScript makes a number of interface
+changes.
+
+In many cases you should not need to adjust your application code at all. This will be the case if:
```
hensmi-amazon, line 16: Love this section. I wouldn't even mind putting it in bold.

guides/17_Migration_to_3_0.md
```diff
@@ -0,0 +1,125 @@
...
+constructor(
+    configuration: MeetingSessionConfiguration,
+    logger: Logger,
+    eventReporter?: EventReporter
+  )
+```
+
+### Updating `EventReporter` use cases
+
+The `DefaultMeetingSession` constructor no longer takes in a `EventReporter` as it is no longer in charge constructing a reporter.
```
hensmi-amazon, line 52: `in charge *of* constructing a reporter.`

Also i prefer `is no longer responsible for`

guides/17_Migration_to_3_0.md
```diff
@@ -0,0 +1,125 @@
...
+const eventReporter = new EventReporter(...)
+…
+const deviceController = new DefaultDeviceController(logger);
+const eventController = new DefaultEventController(configuration, logger, eventReporter)
+this.meetingSession = new DefaultMeetingSession(configuration, logger, deviceController, eventController);
+})
+```
+
+### Updating `eventDidReceive` use cases
+The `eventDidReceive` function that was part of `AudioVideoObserver` has been moved to `EventObserver` which is a observer that the `EventController` now handles. Addtionnaly because of this if you were
```
hensmi-amazon, line 77: spelling. Also i think we decided on just letting markdown do the word wrap.

guides/17_Migration_to_3_0.md
LeviSklaroff: It seems a bit much to bold the entire section, how about just the first two lines?

guides/17_Migration_to_3_0.md
LeviSklaroff: Yeah I can just remove it, I wanted to add a bit of context as to why we're removing it from the constructor, but with the context from decoupling the event controller it is a bit redundant.

guides/17_Migration_to_3_0.md
```diff
@@ -0,0 +1,125 @@
...
+    logger: Logger,
+    eventReporter?: EventReporter
+  )
+```
+
+### Updating `EventReporter` use cases
+
+The `DefaultMeetingSession` constructor no longer takes in a `EventReporter` as it is no longer in charge constructing a reporter.
+
+If your code looked like this:
```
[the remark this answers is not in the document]
LeviSklaroff: That's fair, I'll update it to be `Before in 2.0` and `After in 3.0`

src/meetingsession/DefaultMeetingSession.ts
LeviSklaroff: Yeah we can discuss more about this once we see what the `DeviceController` looks like decoupled.

src/eventcontroller/EventController.ts
LeviSklaroff: I don't think I disagree, but I feel it would feel out of place if we were to document it on some interfaces but not others. I'll wait to hear what @ltrung  and @simmkyu think.

src/eventcontroller/EventController.ts
LeviSklaroff: That's fair, I'll add that in.

src/devicecontroller/NoOpDeviceController.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { EventController } from '..';
```
devalevenkatesh, line 4: Should this import come from an absolute path like others instead of relative import from `src/index.ts`?

src/devicecontroller/NoOpDeviceController.ts
```diff
@@ -69,6 +70,23 @@ export default class NoOpDeviceController
   getVideoInputQualitySettings(): VideoQualitySettings | null {
     return null;
   }
+  getEventController(): EventController {
```
devalevenkatesh, line 73: Could you please let me know why we need ` getEventController` and `setEventController` when we are adding the `eventController` dependency through the constructor? Also, the `eventController` is not private in the `NoOpDeviceControllerWithEventController` class.

devalevenkatesh, line 73: To follow the file pattern, could you please add a extra line between variable names and methods?

src/eventcontroller/DefaultEventController.ts
```diff
@@ -96,7 +134,36 @@ export default class DefaultEventController implements EventController {
       await this.eventReporter?.reportEvent(timestampMs, name, flattenedAttributes);
```
devalevenkatesh, line 134: This would not work since the variable is now changed to `this._eventReporter`, so you have to update the usage accordingly.

src/eventcontroller/DefaultEventController.ts
```diff
@@ -38,16 +54,20 @@ export default class DefaultEventController implements EventController {
   private deviceName: string;
 
   constructor(
-    private audioVideoController: AudioVideoController,
-    private eventReporter?: EventReporter
+    configuration: MeetingSessionConfiguration,
+    logger: Logger,
+    eventReporter?: EventReporter
```
devalevenkatesh, line 59: Can we keep `private eventReporter?: EventReporter` and then use `this.eventReporter`? Do we need `private _eventReporter: EventReporter`?

src/eventcontroller/DefaultEventController.ts
```diff
@@ -108,28 +175,47 @@ export default class DefaultEventController implements EventController {
...
+  }
+
+  /**
+   * Clean up this instance and resources that it created.
+   *
+   * After calling `destroy`, internal fields like `eventReporter` will be unavailable.
+   */
+  async destroy(): Promise<void> {
+    if (isDestroyable(this.eventReporter)) {
+      await this.eventReporter.destroy();
```
devalevenkatesh, line 215: Based on earlier comments, update `eventReporter` usage to `this._eventReporter`. Or use `private eventReporter?: EventReporter` in constructor and then just use `eventReporter`.

src/meetingsession/DefaultMeetingSession.ts
```diff
@@ -49,9 +46,14 @@ export default class DefaultMeetingSession implements MeetingSession, Destroyabl
 
     CSPMonitor.addLogger(this._logger);
     CSPMonitor.register();
-
-    this.setupEventReporter(configuration, logger, eventReporter);
-    this._deviceController = deviceController;
+    if (eventController) {
+      this._eventController = eventController;
+    } else {
+      this._eventController = new DefaultEventController(configuration, logger);
```
devalevenkatesh, line 52: We are creating multiple `DefaultEventController` objects, once here and one in `DefaultAudioVideoController`, this will lead to creating multiple `DefaultMeetingEventReporter` objects. Should we avoid this and use a single `DefaultEventController` across multiple classes in JS SDK?

devalevenkatesh (commented):
Did an initial review on implementation, will read through design doc to see whether I am missing any changed context on implementation approach. Thanks! Looking good.

src/devicecontroller/NoOpDeviceController.ts
LeviSklaroff: Yes this should be, nice catch!

src/devicecontroller/NoOpDeviceController.ts
LeviSklaroff: It was unintentional that the `eventController` was a required dependency in the constructor and that it was public, so I updated it. The main purpose of the `NoOpDeviceControllerWithEventController` as well as the getter and setters is to test to make sure `MeetingSession` does not overwrite the `EventController` in the `DeviceController` and that it will set one if there isn't already one.

LeviSklaroff: Done!

src/eventcontroller/DefaultEventController.ts
LeviSklaroff: This works cause there's a getter for the EventReporter called eventReporter.

src/eventcontroller/DefaultEventController.ts
LeviSklaroff: It was changed to allow for a getter. The main use of the getter outside of the EventController is testing purposes, but as we exposed the eventReporter in the MeetingSession previously it would be best to continue to expose the EventReporter inside the EventController.

src/eventcontroller/DefaultEventController.ts
LeviSklaroff: As mentioned before this works because there is a getter.

src/meetingsession/DefaultMeetingSession.ts
LeviSklaroff: This actually won't create multiple because the meeting session will pass its eventController to the audioVideoController, but it's probably better to get rid of the default construction inside AudioVideo anyways, as it's a bit redundant.

guides/17_Migration_to_3_0.md
```diff
@@ -0,0 +1,120 @@
+# Migration from SDK v1 to SDK v2
```
ltrung, line 1: Should be V2 to V3

src/devicecontroller/NoOpDeviceController.ts
```diff
@@ -70,6 +71,21 @@ export default class NoOpDeviceController
...
   }
 }
+export class NoOpDeviceControllerWithEventController extends NoOpDeviceController {
+  private eventController: EventController;
+  constructor(eventController?: EventController) {
+    super();
+    this.eventController = eventController;
+  }
+
+  getEventController(): EventController {
```
xuesichao, line 81: Do we need to add the `undefined` to the type signature here? `this.eventController` might be `undefined` here if I'm not wrong

```ts
getEventController(): EventController | undefined {};
```

guides/17_Migration_to_3_0.md
```diff
@@ -0,0 +1,120 @@
...
+    logger: Logger,
+    eventReporter?: EventReporter
+  )
+```
+
+### Updating `EventReporter` use cases
+
+The `DefaultMeetingSession` constructor no longer takes in a `EventReporter`.
+
+Before in 2.0:
```
ltrung, line 51: Should it be 2.x?

guides/17_Migration_to_3_0.md
```diff
@@ -0,0 +1,120 @@
...
+constructor(
+    configuration: MeetingSessionConfiguration,
+    logger: Logger,
+    eventReporter?: EventReporter
+  )
+```
+
+### Updating `EventReporter` use cases
+
+The `DefaultMeetingSession` constructor no longer takes in a `EventReporter`.
```
ltrung, line 49: In this case, it is more like we take event controller instead of reporter right?

guides/17_Migration_to_3_0.md
```diff
@@ -0,0 +1,120 @@
...
+
+```typescript
+const configuration = new MeetingSessionConfiguration(…);
+…
+const deviceController = new DefaultDeviceController(logger);
+this.meetingSession = new DefaultMeetingSession(configuration, logger, deviceController);
+this.audioVideo = this.meetingSession.audioVideo;
+this.eventController = this.meetingSession.eventController;
+this.audioVideo.addObserver(...)
+this.eventController.eventController.addObserver({
```
ltrung, line 107: Why are there two eventController here?

guides/17_Migration_to_3_0.md
```diff
@@ -0,0 +1,120 @@
...
+          eventDidReceive(name: EventName, attributes: EventAttributes): void {
+            ...
+          }
+});
+this.audioVideo.forEachObserver((observer: AudioVideoObserver) => {
```
ltrung, line 112: forEachObserver is more like an internal API. Wonder if we should just mark it internal. What do you think @simmkyu @devalevenkatesh

src/eventcontroller/EventController.ts
ltrung: I think we try to document more on the interface if possible now. At least @richnew10 used to get on me all the time for that :)

src/task/MonitorTask.ts
```diff
@@ -436,12 +436,9 @@ export default class MonitorTask
     if (attendeeId === presentAttendeeId && present && !this.presenceHandlerCalled) {
       this.presenceHandlerCalled = true;
       this.context.attendeePresenceDurationMs = Date.now() - this.context.startAudioVideoTimestamp;
-      /* istanbul ignore else */
-      if (this.context.eventController) {
-        this.context.eventController.publishEvent('attendeePresenceReceived', {
-          attendeePresenceDurationMs: this.context.attendeePresenceDurationMs,
-        });
-      }
+      this.context.eventController?.publishEvent('attendeePresenceReceived', {
```
ltrung, line 439: This is not available before Node 14. Could we update the README to replace the Node 12 or higher.

LeviSklaroff: Still works for Node 12 when you install through NPM, as Kyu mentioned in Slack

guides/17_Migration_to_3_0.md
LeviSklaroff: Typo, updated

guides/17_Migration_to_3_0.md
LeviSklaroff: Yeah that's more accurate, updated

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -218,6 +212,10 @@ export default class DefaultAudioVideoController
     return this._mediaStreamBroker;
   }
 
+  get eventController(): EventController {
```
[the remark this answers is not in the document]
LeviSklaroff: Yeah this should have a undefined signature.

src/devicecontroller/NoOpDeviceController.ts
LeviSklaroff: Ah yeah, before I had the constructor require a eventController so that wasn't a option, but yeah it should have undefined now.

test/eventcontroller/DefaultEventController.test.ts
```diff
@@ -92,28 +116,55 @@ describe('DefaultEventController', () => {
...
           expect(attributes.videoInputErrorMessage).to.equal(videoInputErrorMessage);
+          eventController.destroy();
           done();
         },
       });
       eventController.publishEvent(eventName, {
         videoInputErrorMessage,
       });
     });
+    it('only gets events when observing', async () => {
```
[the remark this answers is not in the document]
LeviSklaroff: Added

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
     
 ### Changed
+- Decoupled EventController from AudioVideo and MeetingSession
```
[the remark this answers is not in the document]
LeviSklaroff: Added

src/eventcontroller/EventController.ts
LeviSklaroff: Yeah we can try just changing it to private for now, I'll say the functionality has been deprecated for eventDidRecieve but if a builder wants the functionality they can request it to be made public again.

guides/17_Migration_to_3_0.md
devalevenkatesh: Yes agree. The builder should get examples around the observer notification and have not to call the observers with the updates in their apps explicitly.

src/meetingsession/DefaultMeetingSession.ts
```diff
@@ -32,15 +29,15 @@ export default class DefaultMeetingSession implements MeetingSession, Destroyabl
...
   private _deviceController: DeviceController;
   private audioVideoFacade: AudioVideoFacade;
-  private _eventReporter: EventReporter;
 
   constructor(
     configuration: MeetingSessionConfiguration,
     logger: Logger,
     deviceController: DeviceControllerBasedMediaStreamBroker,
-    eventReporter?: EventReporter
+    eventController?: EventController
```
ltrung, line 40: wonder if we could just do private readonly here? Would it be the same as _eventController and a getter?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -141,7 +139,7 @@ export default class DefaultAudioVideoController
     webSocketAdapter: WebSocketAdapter,
     mediaStreamBroker: MediaStreamBroker,
     reconnectController: ReconnectController,
-    eventReporter?: EventReporter
+    eventController?: EventController
```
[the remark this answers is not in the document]
LeviSklaroff: Why do we need it to always be able to be published? We currently don't have any part of the SDK that relies on events being published, obviously it wouldn't be able to publish events without it, but the SDK will work fine without it so I'm not sure why it needs to be a hard dependency.

src/mediastreambroker/DeviceControllerBasedMediaStreamBroker.ts
```diff
@@ -2,10 +2,21 @@
...
 
 export default interface DeviceControllerBasedMediaStreamBroker
   extends DeviceController,
     MediaStreamBroker {
   // This interface combines a device controller with a media stream broker
+
+  /**
+   * Set the current event controller for publishing events
+   */
+  setEventController?(eventController: EventController): void;
```
ltrung, line 16: Why are these get set an optional APIs?

src/eventcontroller/DefaultEventController.ts
```diff
@@ -60,6 +80,26 @@ export default class DefaultEventController implements EventController {
...
+  addObserver(observer: EventObserver): void {
+    this.observerQueue.add(observer);
+  }
+
+  removeObserver(observer: EventObserver): void {
+    this.observerQueue.delete(observer);
+  }
+
+  /** @internal */
+  private forEachObserver(observerFunc: (observer: EventObserver) => void): void {
```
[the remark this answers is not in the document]
ltrung: Yeah the current pattern is public API but marked as internal :)

src/eventcontroller/DefaultEventController.ts
```diff
@@ -96,7 +135,38 @@ export default class DefaultEventController implements EventController {
...
     } catch (error) {
       /* istanbul ignore next */
-      this.audioVideoController.logger.error(`Error reporting event ${error}`);
+      this.logger.error(`Error reporting event ${error}`);
+    }
+  }
+
+  /** @internal */
+  private setupEventReporter(
+    configuration: MeetingSessionConfiguration,
```
ltrung, line 144: Since this is to decouple event controller from meeting in general, should we pass in a meeting session configuration object at all? Can we event client configuration? We can have a factory method in event client configuration to create it from a meeting session configuration.
@devalevenkatesh what do you think

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -170,7 +171,8 @@ export default class DefaultDeviceController
   constructor(
     private logger: Logger,
     options?: { enableWebAudio?: boolean },
-    private browserBehavior: ExtendedBrowserBehavior = new DefaultBrowserBehavior()
+    private browserBehavior: ExtendedBrowserBehavior = new DefaultBrowserBehavior(),
+    private eventController?: EventController
```
ltrung, line 175: is this it weird to have a private property with get & set? Can this just be public? You can define this property in the interface too so no need for get and setter.

src/audiovideocontroller/DefaultAudioVideoController.ts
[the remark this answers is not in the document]
ltrung: Good point. So it should not be optional in device controller either?

src/mediastreambroker/DeviceControllerBasedMediaStreamBroker.ts
LeviSklaroff: We could make it required. I was thinking if a builder has their own DeviceController that they don't want to use the EventController for they could just not implement these.

src/meetingsession/DefaultMeetingSession.ts
LeviSklaroff: We could do private here, but we can't do readonly because we can't set it to undefined upon destruction.

test/messagingsession/DefaultMessagingSession.test.ts
```diff
@@ -58,7 +58,7 @@ describe('DefaultMessagingSession', () => {
       hostname: string,
       _path: string,
       _payload: string,
-      _queryParams: Map<string, string[]>
+      _queryParams: Map<string, string[]> | null
```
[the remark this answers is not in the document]
LeviSklaroff: I could have sworn I got rid of this, not supposed to be a part of this.

src/eventcontroller/DefaultEventController.ts
LeviSklaroff: We could for the setupEventReporter, but I don't think we'd be able to get the meta data we sent as part of our events from event client configuration

src/devicecontroller/DefaultDeviceController.ts
LeviSklaroff: Yeah that makes it a lot cleaner.

guides/17_Migration_to_3_0.md
```diff
@@ -0,0 +1,75 @@
...
+const configuration = new MeetingSessionConfiguration(…);
+const eventReporter = new EventReporter(...)
+…
+const deviceController = new DefaultDeviceController(logger);
+const eventController = new DefaultEventController(configuration, logger, eventReporter)
+this.meetingSession = new DefaultMeetingSession(configuration, logger, deviceController, eventController);
+})
+```
+
+### `forEachObserver` use cases
```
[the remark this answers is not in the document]
LeviSklaroff: I don't think it would be good to drop since it was possible to use it as listed in a capacity in 2.x,

src/eventcontroller/EventController.ts
```diff
@@ -20,7 +37,13 @@ export default interface EventController {
   ): Promise<void>;
 
   /**
-   * Adds a state to the meeting history stack.
+   * A function that can be called to push a meeting state to the meeting history i.e. push a meeting event to the event
```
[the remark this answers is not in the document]
LeviSklaroff: Updated

src/mediastreambroker/DeviceControllerBasedMediaStreamBroker.ts
```diff
@@ -2,10 +2,21 @@
...
+import EventController from '../eventcontroller/EventController';
 import MediaStreamBroker from '../mediastreambroker/MediaStreamBroker';
 
 export default interface DeviceControllerBasedMediaStreamBroker
   extends DeviceController,
     MediaStreamBroker {
   // This interface combines a device controller with a media stream broker
+
+  /**
+   * Set the current event controller for publishing events
```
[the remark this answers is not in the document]
LeviSklaroff: Updated

test/eventcontroller/DefaultEventController.test.ts
```diff
@@ -33,37 +51,42 @@ describe('DefaultEventController', () => {
 
   describe('construction', () => {
     it('can create with the user agent', () => {
-      eventController = new DefaultEventController(audioVideoController);
+      eventController = new DefaultEventController(emptyConfiguration, logger);
       expect(eventController).to.exist;
+      eventController.destroy();
```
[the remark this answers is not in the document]
LeviSklaroff: Updated, and added a destruction test.

src/audiovideocontroller/DefaultAudioVideoController.ts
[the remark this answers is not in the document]
devalevenkatesh: > Why do we need it to always be able to be published? We currently don't have any part of the SDK that relies on events being published.

Yes, we need always publish for the ingestion service and SDK is not dependent but the ingestion service is. I meant that we have to publish meeting events by default and thats a must for reporting events out currently for event ingestion so any benefits in keeping it optional? Generally it would be good to avoid optional parameters on methods. I am ok with keeping it optional.

src/eventcontroller/DefaultEventController.ts
```diff
@@ -28,6 +35,15 @@ export default class DefaultEventController implements EventController {
   private browserMajorVersion: string;
   /** @internal */
   private meetingHistoryStates: { name: MeetingHistoryState; timestampMs: number }[] = [];
+  /** @internal */
+  private observerQueue: Set<EventObserver> = new Set<EventObserver>();
```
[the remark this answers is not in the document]
LeviSklaroff: Updated to `observerSet`. I actually copied the naming from `AudioVideoController` so we should probably also update it in `AudioVideoController` at some point.

src/eventobserver/EventObserver.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import EventAttributes from '../eventcontroller/EventAttributes';
+import EventName from '../eventcontroller/EventName';
+
+export default interface EventObserver {
+  /**
+   * Called when specific events occur during the SDK session and includes attributes of the event. Example: This can be used to
```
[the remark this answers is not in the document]
LeviSklaroff: Updated

src/eventcontroller/DefaultEventController.ts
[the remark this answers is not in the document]
LeviSklaroff: forEachObserver was changed to private, but I did mistakenly mark it as internal

src/eventcontroller/DefaultEventController.ts
```diff
@@ -96,7 +135,38 @@ export default class DefaultEventController implements EventController {
       await this.eventReporter?.reportEvent(timestampMs, name, flattenedAttributes);
     } catch (error) {
       /* istanbul ignore next */
-      this.audioVideoController.logger.error(`Error reporting event ${error}`);
+      this.logger.error(`Error reporting event ${error}`);
+    }
+  }
+
+  /** @internal */
+  private setupEventReporter(
```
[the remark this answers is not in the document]
LeviSklaroff: I totally misunderstood what the internal tag was for because there were some private variables marked internal. Removed all of the internal tags from private variables/functions

guides/17_Migration_to_3_0.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Updating `EventController` use cases
+The `EventController` has been decoupled from the `AudioVideoController` which has added addtional required functions to the interface.
+```typescript
+addObserver(observer: EventObserver): void;
+removeObserver(observer: EventObserver): void;
+forEachObserver(observerFunc: (observer: EventObserver) => void): void;
+```
+
+Additionally because of this the constructor of `DefaultEventController` has been updated
```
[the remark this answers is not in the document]
LeviSklaroff: Updated

guides/17_Migration_to_3_0.md
```diff
@@ -0,0 +1,75 @@
...
+__In many cases you should not need to adjust your application code at all. This will be the case if:__
+
+* You do not implement your own `EventController` or construct `DefaultEventController` yourself.
+* You do not pass a `EventReporter` yourself or use the reporter through `MeetingSession`.
+* You do not explicitly call `addObserver` on any instances of `AudioVideoController` using a `AudioVideoObserver` that implements `eventDidReceive`.
+
+If your application does not meet all three criteria, read on.
+
+### Updating `EventController` use cases
+The `EventController` has been decoupled from the `AudioVideoController` which has added addtional required functions to the interface.
```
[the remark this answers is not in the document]
LeviSklaroff: Updated, although I only list the new additions to the interface

guides/17_Migration_to_3_0.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Interface changes
+
+__Version 3 of the Amazon Chime SDK for JavaScript makes a number of interface changes.__
+
+__In many cases you should not need to adjust your application code at all. This will be the case if:__
+
+* You do not implement your own `EventController` or construct `DefaultEventController` yourself.
+* You do not pass a `EventReporter` yourself or use the reporter through `MeetingSession`.
+* You do not explicitly call `addObserver` on any instances of `AudioVideoController` using a `AudioVideoObserver` that implements `eventDidReceive`.
```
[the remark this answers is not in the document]
LeviSklaroff: Updated, reworded a little bit to try and cover all bases

guides/17_Migration_to_3_0.md
```diff
@@ -53,4 +58,68 @@ handle them please remove.
 - AudioCallEnded
 - TURNMeetingEnded
 - StateMachineTransitionFailed
-- AudioDeviceSwitched
\ No newline at end of file
+- AudioDeviceSwitched
+
+### Updating `EventController` use cases
+We have de-coupled the `EventController` from `AudioVideoController`. Check below for the new addtions.
```
devalevenkatesh, line 64: nit: addtions -> additions

guides/17_Migration_to_3_0.md
```diff
@@ -53,4 +58,68 @@ handle them please remove.
...
+```typescript
+interface EventController {
+  // Adds an observer for event published to this controller.
```
devalevenkatesh, line 67: Sorry, this feels a bit confusing to me. Should we simply say: "Adds an observer to receive events." ? Same for removeObserver.

guides/17_Migration_to_3_0.md
```diff
@@ -53,4 +58,68 @@ handle them please remove.
...
+Use case 2: Update creation of `EventController`
+```typescript
+// Before in 2.x:
+const configuration = new MeetingSessionConfiguration(…);
+const logger = new Logger(…);
+const eventReporter = new EventReporter(...)
+…
+const audioVideoController = new DefaultAudioVideoController(configuration, logger, ..., eventReporter);
+const eventController = new DefaultEventController(audioVideoController, eventReporter)
+// After in 3.x:
```
devalevenkatesh, line 87: nit: Add a new line before 3.x.

guides/17_Migration_to_3_0.md
```diff
@@ -53,4 +58,68 @@ handle them please remove.
...
+const logger = new Logger(…);
+const eventReporter = new EventReporter(...)
+…
+const eventController = new DefaultEventController(configuration, logger, eventReporter)
+const audioVideoController = new DefaultAudioVideoController(configuration, logger, ..., eventController);
+```
+
+### Updating `EventReporter` use cases
+
+The `DefaultMeetingSession` constructor no longer takes in a `EventReporter` and instead optionally takes in an `EventController`.
```
devalevenkatesh, line 98: You may want to call out if nothing is passed whether any default implementation be created internally.

src/audiovideocontroller/AudioVideoController.ts
```diff
@@ -117,5 +117,5 @@ export default interface AudioVideoController extends AudioVideoControllerFacade
   /**
    * Returns the [[EventController]] for this audio-video controller.
    */
-  readonly eventController?: EventController;
+  readonly eventController: EventController;
```
devalevenkatesh, line 120: Since, eventController is optional through constructor for DAV, should this still be optional even though MeetingSession will always send one?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -170,7 +171,8 @@ export default class DefaultDeviceController
   constructor(
     private logger: Logger,
     options?: { enableWebAudio?: boolean },
-    private browserBehavior: ExtendedBrowserBehavior = new DefaultBrowserBehavior()
+    private browserBehavior: ExtendedBrowserBehavior = new DefaultBrowserBehavior(),
+    public eventController?: EventController | undefined
```
devalevenkatesh, line 175: For learning: Does this need `| undefined` as it is optional ?

src/eventcontroller/DefaultEventController.ts
```diff
@@ -15,39 +22,42 @@ import EventName from './EventName';
...
   // We can consider replacing "detect-browser" in DefaultBrowserBehavior.
-  /** @internal */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private parserResult: any;
-  /** @internal */
   private browserMajorVersion: string;
-  /** @internal */
   private meetingHistoryStates: { name: MeetingHistoryState; timestampMs: number }[] = [];
+  private observerSet: Set<EventObserver> = new Set<EventObserver>();
+
```
devalevenkatesh, line 35: nit: Remove extra line.

src/eventcontroller/DefaultEventController.ts
```diff
@@ -15,39 +22,42 @@ import EventName from './EventName';
...
+  private logger: Logger;
+  private configuration: MeetingSessionConfiguration;
+  private _eventReporter: EventReporter;
 
   // Compute these once so we're not doing work on each event.
-  /** @internal */
```
devalevenkatesh, line 33: Sorry do not remember why this was added? Can you once check if Ok to remove this with Kyu?

src/eventcontroller/EventController.ts
```diff
@@ -1,16 +1,34 @@
...
 import EventName from './EventName';
 import MeetingHistoryState from './MeetingHistoryState';
 
 /**
- * [[EventController]] keeps track of events and notifies audio-video observers.
- * An event describes the success and failure conditions for the meeting session.
+ * [[EventController]] keeps track of a list of event observers and notifies them of SDK events.
+ * An event can describe success or failure of SDK components that make use of this controller.
+ * Example: The success and failure for a meeting session.
+ * If EventIngestionUrl is available in MeetingSessionConfiguration, then the generated meeting events will
```
devalevenkatesh, line 15: nit: Would be great to backtick these coding names. EventIngestionUrl, MeetingSessionConfiguration and so on. Can skip if not followed elsewhere to get this in in-time.

src/eventcontroller/EventController.ts
```diff
@@ -1,16 +1,34 @@
...
+ * be reported to Amazon Chime's backend using the EventReporter.
+ *
+ * In most cases you will want to use the default EventController, but you may want to implement your
+ * own if you wish to keep track of more meta data or wish to implement your own custom events.
  */
 export default interface EventController {
+  /**
+   * Adds an observer for event published to this controller.
```
devalevenkatesh, line 23: Please check documentation comment for the same.

devalevenkatesh: https://github.com/aws/amazon-chime-sdk-js/pull/1969#discussion_r812462378

src/eventcontroller/EventController.ts
```diff
@@ -20,7 +38,13 @@ export default interface EventController {
   ): Promise<void>;
 
   /**
-   * Adds a state to the meeting history stack.
+   * A function that can be called to push a meeting state to the meeting history i.e. push a meeting event to the event
+   * history. This can be used to update the history without notifying EventObservers. It will also report to the Amazon Chime backend without notifying event observers.
+   */
+  pushMeetingState?(state: MeetingHistoryState): Promise<void>;
```
ltrung, line 44: Why is this optional?

src/mediastreambroker/DeviceControllerBasedMediaStreamBroker.ts
```diff
@@ -2,10 +2,16 @@
...
 
 export default interface DeviceControllerBasedMediaStreamBroker
   extends DeviceController,
     MediaStreamBroker {
   // This interface combines a device controller with a media stream broker
+
+  /**
+   * EventController for publishing events.
+   */
+  eventController?: EventController | undefined;
```
ltrung, line 16: Why does this have to be in the interface level?

guides/17_Migration_to_3_0.md
LeviSklaroff: Updated

guides/17_Migration_to_3_0.md
LeviSklaroff: Yeah I had some weird wording here, updated

guides/17_Migration_to_3_0.md
LeviSklaroff: Updated

guides/17_Migration_to_3_0.md
LeviSklaroff: Yeah probably a good idea to call out, updated

src/audiovideocontroller/AudioVideoController.ts
LeviSklaroff: Yes this should be.

src/devicecontroller/DefaultDeviceController.ts
LeviSklaroff: No this doesn’t, I think I put it there for testing at earlier and forgot to rid of it.

src/eventcontroller/DefaultEventController.ts
LeviSklaroff: Updated

src/eventcontroller/DefaultEventController.ts
LeviSklaroff: Checked with Kyu, previously it was needed so the doc won't have it, but checked local doc and seems to be fine without it. Will check live version once it's updated

src/eventcontroller/EventController.ts
LeviSklaroff: I didn't know you could in docstring, updated

src/eventcontroller/EventController.ts
LeviSklaroff: Updated

src/eventcontroller/EventController.ts
LeviSklaroff: Not really core to the interface, currently we don't use it so if a builder doesn't wanna implmen

LeviSklaroff: > Not really core to the interface, currently we don't use it so if a builder doesn't want to implement it, there's no loss of functionality.

src/mediastreambroker/DeviceControllerBasedMediaStreamBroker.ts
LeviSklaroff: Otherwise if a builder implements their own `DeviceController` that has an `EventController` we won't be able to use it unless it's defined in the interface

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -287,7 +288,7 @@ export class DemoMeetingApp
 
   // feature flags
   enableWebAudio = false;
-  logLevel = LogLevel.INFO;
+  logLevel = LogLevel.DEBUG;
```
devalevenkatesh, line 291: If done with testing, change this back to INFO :) ?

src/eventcontroller/EventController.ts
ltrung: Should we just remove it then if it is not used? @devalevenkatesh

demos/browser/app/meetingV2/meetingV2.ts
LeviSklaroff: Updated :)

src/mediastreambroker/DeviceControllerBasedMediaStreamBroker.ts
ltrung: Do we need access to it outside of the device controller class?

src/eventcontroller/EventController.ts
devalevenkatesh: I think based on ingestion service request we changed or updated events to use `publishEvent` over `pushMeetingState`. I see earlier there were reference to this:
- https://github.com/aws/amazon-chime-sdk-js/blob/v2.12.0/src/audiovideocontroller/DefaultAudioVideoController.ts#L998
- https://github.com/aws/amazon-chime-sdk-js/blob/v2.12.0/src/task/MonitorTask.ts#L348

Would it be good to keep it now and audit later if we still need it?

ltrung: If it is not used. It is deadcode. We can always bring it back later if needed :D

guides/17_Migration_to_3_0.md
```diff
@@ -53,4 +58,69 @@ handle them please remove.
...
+
+  // Removes an observer for event published to this controller.
+  removeObserver(observer: EventObserver): void;
+
+  // EventReporter that the EventController uses to send events to the Amazon Chime backend.
+  readonly eventReporter?: EventReporter;
+}
+```
+
+Use case 2: Update creation of `EventController`
```
devalevenkatesh, line 78: This says Use case 2, could not find use case 1. Also, could not find update to:
https://github.com/aws/amazon-chime-sdk-js/pull/1969#discussion_r812463235

src/eventcontroller/EventController.ts
devalevenkatesh: Ok to remove and agree we can add it back if required.

src/mediastreambroker/DeviceControllerBasedMediaStreamBroker.ts
LeviSklaroff: We use it in order to keep the EventController between the MeetingSession and DeviceController the same. https://github.com/aws/amazon-chime-sdk-js/blob/acd66ec4783114c0452162b77d8dab479e1c2f18/src/meetingsession/DefaultMeetingSession.ts#L52

src/eventcontroller/EventController.ts
LeviSklaroff: Will remove then

devalevenkatesh (commented):
1. Few changes required on other dependent guide documentation:
https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/06_Meeting_Events.md#meeting-events

Change the first codeblock to showcase adding `eventDidReceive` observer to `eventController` and not `audioVideo`.

```
const observer = {
  eventDidReceive(name, attributes) {
    // Handle a meeting event.
  }
}
                      // Was audioVideo here
meetingSession.eventController.addObserver(observer);
```

2. Check if any ingestion guide code needs update: https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/12_Client_Event_Ingestion.md#opt-out-of-event-ingestion 
```

File: src/eventcontroller/EventController.ts (hunk 61 above)
Comment (devalevenkatesh) [reply src/eventcontroller/EventController.ts:44]: Yes, lets remove, found the PR below which added all events to `publishEvent` from `pushMeetingState`:
https://github.com/aws/amazon-chime-sdk-js/pull/1580

File: src/mediastreambroker/DeviceControllerBasedMediaStreamBroker.ts (hunk 62 above)
Comment (ltrung) [reply src/mediastreambroker/DeviceControllerBasedMediaStreamBroker.ts:16]: Should it be in the DeviceController interface then?

**Comment by LeviSklaroff** [comment]
> 1. Few changes required on other dependent guide documentation:
>    https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/06_Meeting_Events.md#meeting-events
> 
> Change the first codeblock to showcase adding `eventDidReceive` observer to `eventController` and not `audioVideo`.
> 
> ```
> const observer = {
>   eventDidReceive(name, attributes) {
>     // Handle a meeting event.
>   }
> }
>                       // Was audioVideo here
> meetingSession.eventController.addObserver(observer);
> ```
> 
> 2. Check if any ingestion guide code needs update: https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/12_Client_Event_Ingestion.md#opt-out-of-event-ingestion

1. Already caught, and has been changed
2. Updated

File: guides/17_Migration_to_3_0.md (hunk 64 above)
Comment (LeviSklaroff) [reply guides/17_Migration_to_3_0.md:82]: Added Use Case 1, and line 101 had "or creates one if none is given." added at the end.

File: guides/17_Migration_to_3_0.md [hunk 65]
```diff
@@ -53,4 +58,73 @@ handle them please remove.
...
+  // Adds an observer for event published to this controller.
+  addObserver(observer: EventObserver): void;
+
+  // Removes an observer for event published to this controller.
+  removeObserver(observer: EventObserver): void;
+
+  // EventReporter that the EventController uses to send events to the Amazon Chime backend.
+  readonly eventReporter?: EventReporter;
+
+  // pushMeetingState has been deprecated
```
devalevenkatesh, line 78: Since we are removing we can say pushMeetingState removed over deprecated as SDK was not internally using it.

LeviSklaroff: While it was meant to be internal it was never marked as internal meaning some builder might have been using it, so I'd say it's safer to say deprecated

src/mediastreambroker/DeviceControllerBasedMediaStreamBroker.ts
LeviSklaroff: Yes we can move it to the DeviceController interface

devalevenkatesh (approved):

xuesichao:
Link to https://github.com/aws/amazon-chime-sdk-js/issues/1964.

PR #1980, opened 2022-02-01
Updated Github actions to test on previous major versions

LeviSklaroff:
**Issue #:**
Github actions are required to be updated to test old release builds whenever we create a new major version
**Description of changes:**
Updated actions to automatically test on old versions
**Testing:**

**Checklist:**


ltrung:
>Updated github workflows to test on old major versions

nit: can we change it to previous major version? 2.x is not really old :D

.github/workflows/browser-compatibility-test-previous-major-version.yml
```diff
@@ -0,0 +1,309 @@
...
+    steps:
+      - name: Checkout Package
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Get Version
+        id: get-version
+        run: |
+          current_version=$(.github/script/get-current-version)
+          echo "::set-output name=version-number::$(($current_version - 1))"
```
ltrung, line 33: Does this syntax work? I tried `echo "$((3.0.0-1))"` and it threw an arithmetic error for me.

LeviSklaroff: Ah I'll have to edit it then, I thought it just outputted the first number

ltrung:
Will this run against release-1.x once merged? Note that a lot of tests are not available in 1.x branch so you may want to disable it (background blur, vf, checker, etc.)

LeviSklaroff:
> Will this run against release-1.x once merged? Note that a lot of tests are not available in 1.x branch so you may want to disable it for now (background blur, vf, checker, etc.) or we can wait till we have v3 branch.

This is supposed to run 1.x at the moment, probably best to just not merge it till v3 and I'll just expect them to fail while testing the workflow.

ltrung:
> > Will this run against release-1.x once merged? Note that a lot of tests are not available in 1.x branch so you may want to disable it for now (background blur, vf, checker, etc.) or we can wait till we have v3 branch.
> 
> This is supposed to run 1.x at the moment, probably best to just not merge it till v3 and I'll just expect them to fail while testing the workflow.

If so can we split this into 2 PR. The other GitHub actions for CI can be merged now :)

LeviSklaroff:
> > > Will this run against release-1.x once merged? Note that a lot of tests are not available in 1.x branch so you may want to disable it for now (background blur, vf, checker, etc.) or we can wait till we have v3 branch.
> > 
> > 
> > This is supposed to run 1.x at the moment, probably best to just not merge it till v3 and I'll just expect them to fail while testing the workflow.
> 
> If so can we split this into 2 PR. The other GitHub actions for CI can be merged now :)

That's a good idea! I'll do that.

nainkunal933:
Why are you not testing previous major versions for `audio-video-transform-integration.yml` https://github.com/aws/amazon-chime-sdk-js/blob/main/.github/workflows/audio-video-transform-integration.yml

.github/workflows/audio-integration.yml
```diff
@@ -4,7 +4,7 @@
     pull_request:
       branches:
         - main
-        - release-1.x
+        - 'release-**.x'
```
nainkunal933, line 7: This would test against any of the older major releases for e.g. if that is 1.x, 2.x or 3.x. How would I exclude a previous major release in case of a very old version that is deprecated?

nainkunal933 (changes requested):
Rather than ** which is a wild character can you please use [] to match the previous major versions. [] will give us more flexibility. Ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

For example, release-[1-2].x

.github/workflows/audio-integration.yml
ltrung: We can delete the branch basically. If we deprecate a version there is no reason to keep the branch.

ltrung (changes requested):
What about `audio-video-transform-integration.yml` ?

ltrung:
> Rather than ** which is a wild character can you please use [] to match the previous major versions. [] will give us more flexibility. Ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
> 
> For example, release-[1-2].x

Prefer ** for now. We can always change it later.

nainkunal933:
> > Rather than ** which is a wild character can you please use [] to match the previous major versions. [] will give us more flexibility. Ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
> > For example, release-[1-2].x
> 
> Prefer ** for now. We can always change it later.

Any reasons why you prefer ** over []? I like [] because I can exclude a branch without having to delete it.

ltrung:
> > > Rather than ** which is a wild character can you please use [] to match the previous major versions. [] will give us more flexibility. Ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
> > > For example, release-[1-2].x
> > 
> > 
> > Prefer ** for now. We can always change it later.
> 
> Any reasons why you prefer ** over []? I like [] because I can exclude a branch without having to delete it.

Because we don't need those flexibility. We will delete release-1.x next week.

LeviSklaroff:
> Why are you not testing previous major versions for `audio-video-transform-integration.yml` https://github.com/aws/amazon-chime-sdk-js/blob/main/.github/workflows/audio-video-transform-integration.yml

I'll add this, I just missed it.

.github/workflows/audio-integration.yml
LeviSklaroff: There's also syntax to exclude branches if we want to, if we put in !release-1.x that will exclude that branch from being ran on.

ltrung (approved):

PR #1982, opened 2022-02-02
Reroll "No longer try to reuse inactive transcievers to fix stale frame on FF"

hensmi-amazon:
**Issue #:** See https://github.com/aws/amazon-chime-sdk-js/pull/1831 which links #1807 (& possibly #1805, but hard to tell without logs).

**Description of changes:** So it turns out that not only does FF randomly put inactive sections at the end of the SDP. It will also just straight up not have transceivers in the same order as media sections (*technically* nothing in the spec that requires that but it sure would be nice) which causes issues with our slightly fragile way of trying to subscribing to videos by having stream IDs in the same order as media sections. `fixUpSubscriptionOrder` now uses the transceiver controller to make sure the MIDs match up to the stream IDS. 

**Testing:**
Smoke tested 4 firefox senders toggling video (see exact repro of original issue below). Also verified simulcast stream switching on chrome sending for good measure. SDPs did not grow indefinitely.

1. Join 3 firefox clients 1, 2, 3
2. Enable video on 1
3. Enable video on 2
4. Disable video on 1
5. Enable video on 1
6. Confirm both videos are shown on 3

**Checklist:**

y

n

n


src/transceivercontroller/DefaultTransceiverController.ts
```diff
@@ -217,20 +217,113 @@ export default class DefaultTransceiverController implements TransceiverControll
     videosToReceive: VideoStreamIdSet
   ): void {
     const videosRemaining = videosToReceive.array();
+    if (transceivers.length !== 0 && !transceivers[0].stop) {
```
hensmi-amazon, line 220: This file's changes are identical to previous PR

src/task/SubscribeAndReceiveSubscribeAckTask.ts
```diff
@@ -94,6 +99,59 @@ export default class SubscribeAndReceiveSubscribeAckTask extends BaseTask {
...
+  //
+  // Firefox occasionally tosses stopped transceivers at the end of the SDP without reason
+  // and in general we don't want to be at the mercy of SDP sections not being in the same
+  // order as `getTransceivers`, so we simply recalculate the array here to enforce that
+  // expected invarient until we refactor our signaling to simply take a mapping of MID to
+  // subscription.
+  //
+  // This only works on Unified Plan SDPs
+  private fixUpSubscriptionOrder(sdp: string, videoSubscriptions: number[]): number[] {
+    if (this.context.transceiverController.getMidForStreamId === undefined) {
```
hensmi-amazon, line 113: @xuesichao an example of why the required interface/default implementation is annoying :D. Do we really want to support custom transceiver controllers? not really for reasons like this....

src/sdp/SDP.ts
```diff
@@ -84,4 +85,9 @@ export default interface SDP {
    * Removes H.264 from the send section.
    */
   removeH264SupportFromSendSection(): SDP;
+
+  /**
+   * List of parsed media sections sections in order they occur on SDP.
+   */
+  mediaSections(): SDPMediaSection[];
```
ltrung, line 92: So this is a stretch as I doubt anyone would implement their own SDP but shouldn't this be optional :D

src/sdp/DefaultSDP.ts
```diff
@@ -53,6 +54,29 @@ export default class DefaultSDP implements SDP {
...
 
+  private static mediaType(sdpLine: string): 'audio' | 'video' | null {
+    const match = /m=(audio|video)/g.exec(sdpLine);
+    if (match === null) {
+      return null;
+    }
+    return match[1] as 'audio' | 'video';
+  }
+
+  private static mid(sdpLine: string): string | null {
```
ltrung, line 65: nit: Can we change the return type to string | undefined?

src/sdp/DefaultSDP.ts
```diff
@@ -619,4 +643,36 @@ export default class DefaultSDP implements SDP {
...
+      return [];
+    }
+
+    const parsedSections: SDPMediaSection[] = [];
+    for (let i = 1; i < sections.length; i++) {
+      const section = new SDPMediaSection();
+      const lines = DefaultSDP.splitLines(sections[i]);
+      for (const line of lines) {
+        const mediaType = DefaultSDP.mediaType(line);
+        if (mediaType) {
```
ltrung, line 659: If it is not media, should we just break from this? Any reason to continue?

src/sdp/DefaultSDP.ts
```diff
@@ -619,4 +643,36 @@ export default class DefaultSDP implements SDP {
...
     return new DefaultSDP(newSDP);
   }
+
+  mediaSections(): SDPMediaSection[] {
+    const sections = DefaultSDP.splitSections(this.sdp);
+    if (sections.length < 2) {
+      return [];
+    }
+
+    const parsedSections: SDPMediaSection[] = [];
```
ltrung, line 653: nit: Feel like the code is easier to read if we name this parsedMediaSections :p

src/sdp/DefaultSDP.ts
hensmi-amazon: There is no sections that are not media sections after they begin, and we start from index 1 to avoid that non-media section.

src/task/SubscribeAndReceiveSubscribeAckTask.ts
```diff
@@ -94,6 +99,59 @@ export default class SubscribeAndReceiveSubscribeAckTask extends BaseTask {
...
+      // of the transceivers should have `mid` set by now (see comment above `getMidForStreamId`)
+      const mid = this.context.transceiverController.getMidForStreamId(streamId);
+      if (mid === undefined) {
+        this.context.logger.warn(`Could not find MID for stream ID: ${streamId}`);
+        continue;
+      }
+      midsToStreamIds.set(mid, streamId);
+    }
+
+    const sections = new DefaultSDP(sdp).mediaSections();
```
ltrung, line 129: Wonder what's the point of interface if we hard coded SDP here :))

ltrung (approved):

src/task/SubscribeAndReceiveSubscribeAckTask.ts
hensmi-amazon: The world may never know

ltrung (approved):

PR #1986, opened 2022-02-02
Update deploy action for release 2.x

nainkunal933:
**Issue #:**
Deploy script should handle multiple branches and metrics should not send all metrics to the same namespace. 
**Description of changes:**
The deploy action needs to be updated to deploy demos against 2.x branch along with the latest branch. Metrics need to account for previous major release tests. 
**Testing:**

No. These are integration test changes which were tested locally with KITE. 

**Checklist:**

Yes

No

No


.github/workflows/deploy.yml
```diff
@@ -25,6 +25,7 @@ jobs:
       - name: Checkout Package
         uses: actions/checkout@v2
         with:
+          ref: ${{ github.ref_name }}
```
LeviSklaroff, line 28: I believe these are redundant as the default value should be the SHA of the commit triggering this workflow. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

xuesichao, line 28: Could tell me which branch is this `ref_name` referring to? and where we could access this variable?

integration/js/utils/SdkBaseTest.js
```diff
@@ -165,11 +165,11 @@ class SdkBaseTest extends KiteBaseTest {
         console.log(e);
         await this.updateSeleniumTestResult(false);
         await this.quitSeleniumSessions();
-        await emitMetric('Common', this.capabilities, 'SeleniumInit', 0);
+        await emitMetric('Common', this.capabilities, 'SeleniumInit', 0, this.payload?.isPreviousMajorVersion);
```
xuesichao, line 168: Shall we use a global variable `this.isPreviousMajorVersion` to replace `this.paylaod?.isPreviousMajorVersion`?

.github/workflows/deploy.yml
nainkunal933: The `ref_name` will change based on which branch this workflow was triggered on. You can find more information about it under GitHub context here: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

Documentation:


github.ref_name | string | The branch or tag name that triggered the workflow run.
-- | -- | --

integration/js/utils/SdkBaseTest.js
nainkunal933: We can use a global variable for `isPreviousMajorVersion` but there is no real benefit of doing it. 

This variable is being sent to KITE via the `payload` property. Some tests will have this property set to true while others will not have this property. We can make it a global property and set it to false by default and overwrite it whenever we do get it from the payload but not sure if its worth it.

.github/workflows/deploy.yml
nainkunal933: The default value would be the branch name that triggered the workflow so I do not think this is redundant.

xuesichao: Yeah according to the [doc](https://github.com/actions/checkout#usage):
> 
```
- uses: actions/checkout@v2
  with:
    # Repository name with owner. For example, actions/checkout
    # Default: ${{ github.repository }}
    repository: ''

    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
    ref: ''
    ...
```

It is default to the main branch and will change based on which event triggers this workflow. Do we need to declare it explictly?

integration/js/utils/SdkBaseTest.js
xuesichao: I think currently we are assigning the payload data to global variables in the constructor. If there's no specific reason of not doing this, I would prefer to follow the convention. There's no specific benefit of doing this in this case. But it is a good practice to follow. The constructor could helps ensure the variable is in the good states before we use it. 
Although `this.payload?.isPreviousMajorVersion)` could handle the same logic here, but constructor could better manager the states.

https://github.com/aws/amazon-chime-sdk-js/blob/fca27b4f6e45160b7fbbde9115b16c826ed472bf/integration/js/utils/SdkBaseTest.js#L19-L25

xuesichao:
Split this PR into two as discussed earlier:
PR to update the metric reporting for release 2.x: https://github.com/aws/amazon-chime-sdk-js/pull/1987/files

LeviSklaroff:
Successful run: https://github.com/aws/amazon-chime-sdk-js/runs/5060868746?check_suite_focus=true
Was ran by editing required workflow params to run on this branch after a pull_request

LeviSklaroff:
> Successful run: https://github.com/aws/amazon-chime-sdk-js/runs/5060868746?check_suite_focus=true Was ran by editing required workflow params to run on this branch after a pull_request

Correction, some jobs failed. Will investigate.

LeviSklaroff:
https://github.com/aws/amazon-chime-sdk-js/runs/5071164404?check_suite_focus=true last step is expected to fail at this time

.github/workflows/deploy.yml
```diff
@@ -53,7 +54,13 @@ jobs:
       - name: Install SAM CLI
         run: script/github-action-awscli-installation
       - name: Run Deployment Script
-        run: script/deploy-canary-demo
+        if: ${{ github.ref == 'refs/heads/main' }}
+        run: script/deploy-canary-demo 
+      - name: Run Deployment Script Previous Major Version
+        if: ${{ github.ref != 'refs/heads/main' }}
```
ltrung, line 60: This is nice but does that mean I have to update the demo URLs in canary config for prev major version (from 2.x to 3.x) everytime we have a new major version?
It is not as nice but should we just hard code to prev-major version so we don't have to update config.

.github/workflows/deploy.yml
```diff
@@ -53,7 +54,13 @@ jobs:
...
+        run: |
+          current_version=$(.github/script/get-current-version)
+          script/deploy-canary-demo -${current_version%%.*}x
```
ltrung, line 63: Nice way doing this :))

ltrung:
@LeviSklaroff Can we rename the PR title and description to remove all the metric part ?

.github/workflows/deploy.yml
LeviSklaroff: That's how it would work right now, but we would have to update the configs anyways for their meta data anyways, so that it's not updating 2x metrics when we're on 3x, this also allows us the option to support more then just the last version, but I do see your point and I wouldn't say I disagree with just hard coding it. We could make it pretty easy to update though by just making it a single variable.

.github/workflows/deploy.yml
LeviSklaroff: Thanks!

LeviSklaroff:
> @LeviSklaroff Can we rename the PR title and description to remove all the metric part ?

Oops lol, done.

.github/workflows/deploy.yml
ltrung: I think for metrics it is already prev_version or something hard-coded.  There is no 2.x metrics :D

LeviSklaroff: Ah ok, then I'll make it prev-major-version if we ever want to update it to be version number won't be a hard change.

LeviSklaroff:
Run after changing to hard coded 'prev-major-version': https://github.com/aws/amazon-chime-sdk-js/runs/5072728101?check_suite_focus=true

LeviSklaroff:
> Run after changing to hard coded 'prev-major-version': https://github.com/aws/amazon-chime-sdk-js/runs/5072728101?check_suite_focus=true

Looks like we reached the bucket character limit, I'll change it to be "-prev-major"  instead

script/deploy-canary-demo
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -euxo pipefail
+versionInfix=${1+$1}
```
xuesichao, line 3: Could we rename this `versionInfix` to `version`, I think it is not `Infix` in this context.

xuesichao: Just want to clarify a little bit. 
If I'm right,`${1+$1}` returns `$1` only when `$1` exists.
So the goal you want to achieve is:
* By default the version should be empty for latest Canary
* If there's a version number, append it at the end of the demo name

script/deploy-canary-demo
```diff
@@ -10,43 +11,43 @@ ChimeAlpha)
...
-    npm run deploy -- -r ap-southeast-1 -b chime-sdk-meetings-demo-gamma-canary-ap-southeast-1 -s chime-sdk-meetings-demo-gamma-canary-ap-southeast-1 -o chime-sdk-meetings-sin-demo-gamma-canary -t -l
+    npm run deploy -- -r us-east-1 -b chime-sdk-meetings-demo-gamma-canary-us-east-1$versionInfix -s chime-sdk-meetings-demo-gamma-canary-us-east-1$versionInfix -o chime-sdk-meetings-iad-demo-gamma-canary$versionInfix -t -l
+    npm run deploy -- -r us-west-2 -b chime-sdk-meetings-demo-gamma-canary-us-west-2$versionInfix -s chime-sdk-meetings-demo-gamma-canary-us-west-2$versionInfix -o chime-sdk-meetings-pdx-demo-gamma-canary$versionInfix -t -l
+    npm run deploy -- -r eu-central-1 -b chime-sdk-meetings-demo-gamma-canary-eu-central-1$versionInfix -s chime-sdk-meetings-demo-gamma-canary-eu-central-1$versionInfix -o chime-sdk-meetings-fra-demo-gamma-canary$versionInfix -t -l
+    npm run deploy -- -r ap-southeast-1 -b chime-sdk-meetings-demo-gamma-canary-ap-southeast-1$versionInfix -s chime-sdk-meetings-demo-gamma-canary-ap-southeast-1$versionInfix -o chime-sdk-meetings-sin-demo-gamma-canary$versionInfix -t -l
     ;;
 
 esac
```
xuesichao, line 53: nit: add new line end of the file

LeviSklaroff: Added

script/deploy-canary-demo
LeviSklaroff: Sure, changed

LeviSklaroff: > Just want to clarify a little bit. If I'm right,`${1+$1}` returns `$1` only when `$1` exists. So the goal you want to achieve is:
> 
> * By default the version should be empty for latest Canary
> * If there's a version number, append it at the end of the demo name

That's correct

xuesichao (dismissed):
Looks good to me, but you might want to get approval from Trung as well before merging.

script/deploy-canary-demo
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -euxo pipefail
+version=${1+$1}
```
ltrung, line 3: Kinda like the infix or suffix name earlier. This prev-major is not a version :p.
Also I never remember any bash syntax but I assume this is a just a fancy way of saying if $1 is set then version = $1? Is it any different from just version = $1?

LeviSklaroff: How about `canarySufix`? Also it is effectively version=$1, but the main difference is that when I tested giving no param to the script with version=$1 it gave a error because 1 was never set, but this syntax is safer because it effectively checks if $1 exists before setting it to $1, effectively equates to this:
`
if($1)
  version=$1
else
  version=nothing
`

xuesichao: `canarySuffix` looks good to me.

xuesichao: oh right, could you also add a comment for this line?
> if($1) version=$1 else version=nothing
this is good enough

ltrung: Yeah canarySuffix sounds good enough

LeviSklaroff: Updated, although deploying for previous version won't work until our s3 bucket limit is increased, but based on output there isn't anything wrong with the code itself and we can merge. https://github.com/aws/amazon-chime-sdk-js/runs/5073025117?check_suite_focus=true

ltrung (approved):

PR #1987, opened 2022-02-03
Update metrics reporting for release 2.x

xuesichao:
**Issue #:**

**Description of changes:**
* Add the ability to set the `namespaceInfix` in the `payload` of `*_test.config.json` files to differentiate the metrics reports for different Canaries.
  * The latest Canary CloudWatch Metric has namespace in format like this`Canary/Common` (unchanged).
  * The other Canary CloudWatch Metrics have namespace `Canary/${namespaceInfix}/Common`.
* Format the code a little bit.

**Testing:**
Update the `*_test.config.json` and run the test on SauceLabs, the metrics are reported correctly to the CloudWatch.

*Can these tested using a demo application? Please provide reproducible step-by-step instructions.*
N/A

**Checklist:**

1. Have you successfully run `npm run build:release` locally?
N/A

2. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved?
N/A

3. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved?
N/A

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1988, opened 2022-02-04
Remove optional chaining operation to support node version below 14

xuesichao:
**Issue #1987 :**
In PR #1987, we use optional chaining operation for integration test which does not work with node 12 system.

**Description of changes:**
Replace the optional chaining with native JavaScript syntax.

**Testing:**
Test the replaced logic in JavaScript compiler and the result are exactly same as before.

N/A

**Checklist:**

N/A

N/A

N/A

### DIFF
--- a/integration/js/utils/AppTestStep.js
+++ b/integration/js/utils/AppTestStep.js
@@ -18,7 +18,7 @@ class AppTestStep extends TestStep {
       this.driver = sessionInfo.driver;
     }
 
-    this.namespaceInfix = kiteBaseTest.payload?.namespaceInfix || '';
+    this.namespaceInfix = kiteBaseTest.payload.namespaceInfix || '';
   }
 
   async step() {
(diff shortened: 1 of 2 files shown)


integration/js/utils/AppTestStep.js
```diff
@@ -18,7 +18,7 @@ class AppTestStep extends TestStep {
       this.driver = sessionInfo.driver;
     }
 
-    this.namespaceInfix = kiteBaseTest.payload?.namespaceInfix || '';
+    this.namespaceInfix = kiteBaseTest.payload && kiteBaseTest.payload.namespaceInfix || '';
```
ltrung, line 21: This is kinda hard to read, can we just do a normal if else?

xuesichao: Remove the guard, which is not very necessary. Cause we need `url` in `payload`, which means `payload` should not be `undefined` anyway. And the reset of code don't have this kind of guard.

ltrung (approved):

Issue #1985, opened 2022-02-02
Reconnection timeout on VPN


simmkyu:
### What happened and what did you expect to happen?

The Chime SDK for JavaScript reconnection timed out when turning WiFI on/off on the VPN connection.

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

1. Connect to VPN.
1. Join a meeting.
1. Turn your WiFi off.
1.  Wait several seconds.
1. Turn your WiFi on and wait for your VPN to reconnect.
1. SDK attempts to retry the connection.

### Amazon Chime SDK for JavaScript version

2.26.0

### What browsers are you seeing the problem on?

Chrome and Chromium-based environments like Electron

### Browser version

Chrome 97.0.4692.99

### Meeting and Attendee ID Information.

Not required

## Timeout

SDK does not publish any event after `WebSocketClosing`.

```
00:00:00.097 adding connection request to queue: wss://__chime_signaling_server_url__
00:00:00.097 notifying event: WebSocketClosing
00:00:15.087 timeout reached for task AudioVideoReconnect/__your_meeting_id__/__your_attendee_id__/Timeout15000ms
```

## Success

SDK should publish `WebSocketConnecting` and `WebSocketOpen` events immediately after `WebSocketClosing`.

```
00:00:00.504Z adding connection request to queue: wss://__chime_signaling_server_url__
00:00:00.504Z notifying event: WebSocketClosing
00:00:00.752Z opening connection to wss://__chime_signaling_server_url__
00:00:00.753Z notifying event: WebSocketConnecting
00:00:00.898Z notifying event: WebSocketOpen
00:00:00.898Z Opening signaling connection took 393 ms
```

simmkyu:
## Root cause

Suppose that VPN is enabled.  The root cause is that an existing WebSocket object does not receive the `close` event when calling the `WebSocket.close` API. The Chime SDK for JavaScript opens a new connection only when receiving the `WebSocket.close` event during retries.

## Reproducible steps without using the Chime SDK for JavaScript

[...]

PR #1993, opened 2022-02-07
Fix vpn reconnecting issue

simmkyu:
**Issue #:**
https://github.com/aws/amazon-chime-sdk-js/issues/1985

**Description of changes:**
- During retries, the Chime SDK for JavaScript calls the `WebSocket.close` API on the previous WebSocket object, waits for the WebSocket `close` event, and creates a new WebSocket object. The issue was that the `close` event did not arrive in Chrome and Chromium environments when toggling WiFi on a VPN connection. (This behavior is reproducible in a minimal WebSocket app in Chrome 97.0.4692.99.)

  With this PR, SDK will continue creating a WebSocket if the WebSocket `close` event does not arrive for two seconds.

**Testing:**

**Testing scenario 1: Reconnection when the WebSocket connectivity is stable**

1. Open the [Chime SDK serverless demo](https://github.com/aws/amazon-chime-sdk-js/tree/main/demos/serverless) and join a meeting.
1. Open the browser console and run the following script to force retrying the connection.
    ```
    app.audioVideo.audioVideoController.handleMeetingSessionStatus({
      isFailure: () => true,
      isTerminal: () => false,
      statusCode: () => 17,
      isAudioConnectionFailure: () => false,
      toString: () => 'Error',
    });
    ```
1. Ensure that the attendee is present in the meeting.

⠀

**Testing scenario 2: Reconnection when the WebSocket connectivity is not stable**

1. Connect to VPN.
1. Open the [Chime SDK serverless demo](https://github.com/aws/amazon-chime-sdk-js/tree/main/demos/serverless) and join a meeting.
1. Open the browser console and filter the console by **transition**. For example, you can type **transition** in the top input of the Chrome console.
   image
1. Turn your WiFi off.
1. Wait until SDK outputs the **Reconnect** message to the console. SDK starts to retry the connection due to no internet access.
   ```
   [INFO] SDK - transitioning from Connected to Connecting with Reconnect
   ```
1. Turn your WiFi on. Ensure that your VPN is online again.
1. Ensure that the attendee becomes present soon after VPN is online. Before this PR, SDK timed out and failed to reconnect to the session.

**Checklist:**

Yes

No

No


src/websocketadapter/WebSocketAdapter.ts
```diff
@@ -41,6 +41,14 @@ export default interface WebSocketAdapter {
    */
   addEventListener(handler: string, eventListener: EventListener): void;
 
+  /**
+   * Remove an event listener for one of the handler types.
+   *
+   * @param handler The handler being assigned.
+   * @param eventListener The event listener to add.
+   */
+  removeEventListener(handler: string, eventListener: EventListener): void;
```
ltrung, line 50: unnecessary probably but shouldn't this be optional ?

simmkyu: A good catch!

simmkyu: Fixed.

src/signalingclient/DefaultSignalingClient.ts
```diff
@@ -211,6 +214,25 @@ export default class DefaultSignalingClient implements SignalingClient {
...
+        DefaultSignalingClient.CLOSE_EVENT_TIMEOUT_MS
+      );
+      const handler = (event: CloseEvent): void => {
+        this.webSocket?.removeEventListener('close', handler);
+        scheduler.stop();
+        this.closeEventHandler(event);
+      };
+
+      // Remove the existing close handler to prevent SDK from opening a new connection.
+      this.webSocket?.removeEventListener('close', this.closeEventHandler);
```
ltrung, line 229: Do we need to check for undefined/null webSocket here (as the check above in the if condition seems to assume it exists and same as the add below).

src/signalingclient/DefaultSignalingClient.ts
```diff
@@ -211,6 +214,25 @@ export default class DefaultSignalingClient implements SignalingClient {
...
+      this.webSocket.addEventListener('close', handler);
+      scheduler.start(() => {
+        handler(
+          new CloseEvent('close', { wasClean: false, code: 1006, reason: '', bubbles: false })
```
ltrung, line 233: Can we document what close code 1006 is and why we choose it here?

src/signalingclient/DefaultSignalingClient.ts
simmkyu: oops. Not `this.webSocket?` but I should've checked the existence of `removeEventListener`.

simmkyu: I'll add the `/* istanbul ignore next */` since there's no point of testing a WebSocket object without `removeEventListener`. Builders cannot pass the custom WebSocket object.

ltrung: We should just mark these classes internal in next major version :)

src/signalingclient/DefaultSignalingClient.ts
simmkyu: Added a comment

test/signalingclient/DefaultSignalingClient.test.ts
```diff
@@ -647,6 +648,83 @@ describe('DefaultSignalingClient', () => {
...
+          } else if (event.type === SignalingClientEventType.WebSocketClosed) {
+            done();
+          }
+        }
+      }
+      testObjects.signalingClient.registerObserver(new TestObserver());
+      testObjects.signalingClient.openConnection(testObjects.request);
+    });
+
+    it('opens a new connection after closing the first one', done => {
```
simmkyu, line 707: Added one more test to test opening a connection twice without delay.

ltrung (approved):

PR #1996, opened 2022-02-08
Parallelize deployment of prod canary demo apps

nehtambe:
**Issue #:**
Adding a sequential deployment of ChimeSDKMeetings specific demos to postrelease script would increase the average deployment time to ~ 1 hour. 

**Description of changes:**

1. Update publish.yml to add the following job to deploy the serverless demo apps in parallel. 

- deploy_chime_prod_demo

2. Add a step to job 'deploy_chime_prod_demo' which checks every minute if the new version is published to npm or not before starting next steps. The maximum time for verification is set to 5 minutes.
3. Remove the current 'postrelease' script, instead use deploy_canary_demo script to deploy serverless demo to prod canary regions

**Testing:**

No

Tested publish.yaml script in parts using a private repo. Verified jobs are running in parallel.

**Checklist:**

Yes

No

No


.github/workflows/publish.yml
```diff
@@ -37,3 +37,97 @@ jobs:
...
+          echo "Current demo version:" $demo_current_version
+          if [[ *$demo_current_version* != "^#${{ steps.npm_version.outputs.npm_version }}" ]];
+          then
+            echo "Demo npm version is not the same as the current released version"
+          else
+            npm run build
+            npm run build --app=meetingReadinessChecker
+            script/github-action-awscli-installation
+            script/postrelease
+          fi
```
michhyun1, line 86: Im wondering if it might be simpler just to add this to the matrix below. Itd make this file simpler. Anyways, we're still going to deploy all of the demos at the same time, I dont think it really matters that it needs its own job

nehtambe: Yes, initially did that way, then thought its better to keep separate so whoever goes through it have an idea about which Meetings client is getting deployed. Also, before merging the CR we need to - 
1. Configure PROD_AWS_ACCESS_KEY and PROD_AWS_SECRET_ACCESS_KEY in Github secrets. 
2. Increase S3 bucket limit in Prod canary account.

.github/workflows/publish.yml
```diff
@@ -37,3 +37,95 @@ jobs:
         run: npm publish --tag ${{ steps.npm_tag.outputs.npm_tag }}
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
+      - name: Wait for 300 seconds
```
ltrung, line 40: instead of hard-coding a time like this, can we just actively check whether the version is available every minute?

.github/workflows/publish.yml
```diff
@@ -37,3 +37,95 @@ jobs:
...
+          npm install amazon-chime-sdk-js@${{ steps.npm_version.outputs.npm_version }}
+      - name: Verify npm version
+        run: |
+          demo_current_version=$(.github/script/get-demo-browser-current-version)
+          echo "Current demo version:" $demo_current_version
+          if [[ *$demo_current_version* != "^#${{ steps.npm_version.outputs.npm_version }}" ]];
+          then
+            echo "Demo npm version is not the same as the current released version"
+          else
+            npm run build
```
ltrung, line 81: You dont have to run build. The deployment script will take care of that.

.github/workflows/publish.yml
```diff
@@ -37,3 +37,95 @@ jobs:
...
+      - name: Install SDK and verify npm version
+        run: |
+          cd demos/browser
+          npm uninstall amazon-chime-sdk-js
+          npm install amazon-chime-sdk-js@${{ steps.npm_version.outputs.npm_version }}
+      - name: Verify npm version
+        run: |
+          demo_current_version=$(.github/script/get-demo-browser-current-version)
+          echo "Current demo version:" $demo_current_version
+          if [[ *$demo_current_version* != "^#${{ steps.npm_version.outputs.npm_version }}" ]];
```
ltrung, line 77: Not sure the value of this check here. I dont think i ever see the installation failed to install the correct version. If you want to check for it, better to exit 1 to mark the job to fail. Otherwise, noone would know.

.github/workflows/publish.yml
```diff
@@ -37,3 +37,95 @@ jobs:
...
+      AWS_DEFAULT_OUTPUT: text
+      NAME: ${{ matrix.name }}
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.PROD_AWS_ACCESS_KEY }}
+          aws-secret-access-key: ${{ secrets.PROD_AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+      - name: Checkout Package
```
ltrung, line 104: There is a lot of duplicate steps here. Can you create a common composite action similar to this one https://github.com/aws/amazon-chime-sdk-js/blob/main/.github/actions/setup-integration-test/action.yml

script/postrelease
```diff
@@ -1,42 +1,45 @@
...
-end
-
-verbose('npm run build')
-verbose('npm run build --app=meetingReadinessChecker')
-
-Dir.chdir('../serverless/')
-verbose('npm run deploy -- -b chime-sdk-demo-prod-canary -o chime-sdk-demo-prod-canary -s chime-sdk-demo-prod-canary -u false -t -l')
-verbose('npm run deploy -- -b chime-sdk-meeting-readiness-checker-prod-canary -s chime-sdk-meeting-readiness-checker-prod-canary -a meetingReadinessChecker -u false -t -l')
+#!/bin/bash
+set -euxo pipefail
```
ltrung, line 2: Well this is basically the same as the deploy-canary-demo script now. Can we just delete this and move what you added to that script?

.github/workflows/publish.yml
nehtambe: Sure, will remove it

.github/workflows/publish.yml
nehtambe: It was in the old script, so added here too, I am not sure why this was added. If this is needed, will add Exit 1

.github/workflows/publish.yml
nehtambe: Let me add everything under one job. This will avoid duplication of code.

script/postrelease
nehtambe: Yes, we can do that will remove postrelease file

.github/workflows/publish.yml
nehtambe: Yes, we can do it. I will this step to the new job.

.github/workflows/publish.yml
nehtambe: Updated, just creating a single job

ltrung (approved):

PR #1997, opened 2022-02-09
Fixed prev major version deployment reaching max character limit for …

LeviSklaroff:
…S3 bucket

**Issue #:**
Prev major version deployment would fail on meeting demo because the s3 bucket character limit was reached
**Description of changes:**
Changed suffix from prev-major to PM
**Testing:**

No

**Checklist:**

Yes

No

No


.github/workflows/deploy.yml
```diff
@@ -58,7 +58,7 @@ jobs:
         run: script/deploy-canary-demo 
       - name: Run Deployment Script Previous Major Version
         if: ${{ github.ref != 'refs/heads/main' }}
-        run: script/deploy-canary-demo -prev-major
+        run: script/deploy-canary-demo -PM
```
xuesichao, line 61: Not sure if we are able to change `PM` to `prev`? `PM` looks like product manager, lol...
We may want to rename it. but it's not a big deal.

LeviSklaroff: I don't disagree, but I wasn't sure what else to name it, we could do pmv for previous-major-version, but I don’t think thats to intuitive either.

xuesichao: The S3 bucket name has a limit length of [3, 63], just curious which demo name exceeds this limitation.
The longest one I could find is: `chime-sdk-meetings-demo-gamma-canary-ap-southeast-1`, with suffix `-prev-major' the length is exactly 63.

LeviSklaroff: There's ones that adds on iad so it becomes `chime-sdk-meetings-iad-demo-gamma-canary-prev-major-ap-southeast-1` which is 67 characters, we do have some wiggle room to not make it longer then 4 characters but ideally it would be better to try and keep it shorter incase we ever wanted to add any other suffix.

xuesichao: Got it, yeah, we might need to consider reducing the demo name in the future, it's quite redundant.
But the `pm` is still confusing in this context:
`chime-sdk-meetings-iad-demo-gamma-canary-pm-ap-northeast-1`
how about using `prev` instead:
`chime-sdk-meetings-iad-demo-gamma-canary-prev-ap-northeast-1`
or
`chime-sdk-meetings-iad-demo-gamma-prev-canary-ap-northeast-1`

LeviSklaroff: Yeah I can change it to `prev` that works better. I'll update it

jing-chen1 (approved):

.github/workflows/deploy.yml
LeviSklaroff: Ok updated, had to clear out buckets to make sure it worked properly.

xuesichao: Thank you!🥇

xuesichao (approved):

PR #1999, opened 2022-02-09
Adjust the recovery behavior of `VideoPriorityBasedPolicy`

hensmi-amazon:
**Issue #:** None

**Description of changes:** In application logs, internal testing, and simulation running, we found a handful of edges in the current policy where it was recovering much slower then intended. This is not a comprehensive fix of every issue, but instead a focused effort to find areas where the policy wasn't behaving how it was originally meant to (i.e. moreso bug fixes/mitigations to browser BWE behavior in certain scenarios, then widespread behavior changes which largely would require backend changes).

In order of impact:
* We now override the BWE with the actual used bitrate if no packet loss exists and we are in probing state.
  * Before, the BWE would heavily lag behind receive bitrate so e.g. if there were two senders with [300, 1200] streams, BWE was 300, and you were subscribing just to 1 300kbps stream, probing would start, upgrade to the 1200 stream, but BWE would need to reach 1200 to mark the probe as complete. In this case (especially for BWE < 1000), it could take 30-60 seconds just to reach that value, even though it was receiving that amount without issue. See https://github.com/aws/amazon-chime-sdk-js/pull/1999#issue-1129081724 for notes on related code changes.
* We now use a separate value to track bitrate differentials to keep subscriptions equal, other then `this.prevTargetRate`.
  * Before, if BWE was gradually increasing it would hit the if statement every cycle, even if it had eventually grown, e.g. 1 MB.
* We raise the threshold for packets lost (note, not a percentage) to 2 to cancel a probe. It seems that occasionally on upgrade, or initial subscribe  there is a single packet 'lost' (moreso likely just a sequence number skipped). This probably shouldn't happen on backend but this is such a low threshold the change should be otherwise benign. Otherwise probes would randomly fail in unconstrained networks.

This diff also adjusts the max probe wait time to 30 seconds, as the previous max of 60 is quite high, and can be reached in just 1 minute of bad network conditions, so having to wait 60 seconds just to try recovery is a bit slow. In an ideal world we would be able to use GCC type probing from the backend instead of having to tune this wait time.

**Testing:**
I have tested these changes with simulcast on/off, and additionally with another policy that always sends 300/1200 kbps with 5 senders (who needs load test client when your M1 can encode 10 streams without even turning the fan on!) where some of these issues are more prevalent. I've tested with one receive stream having higher priority. I've additionally run some local simulations against these changes to confirm behavior is as expected.

* Join a meeting with multiple senders. Some with simulcast some without.
* On a receiver, vary the priority of the incoming videos. 
* Constrain receiver network for 30 seconds using network link conditioner set to around 1000kbps downlink.
* Unconstrain and wait a bit for recovery.

**Checklist:**

y

y

y


src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -427,7 +440,24 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
...
+          this.rateProbeState === RateProbeState.Probing &&
+          this.downlinkStats.usedBandwidthKbps > this.downlinkStats.bandwidthEstimateKbps &&
+          this.downlinkStats.packetsLost < VideoPriorityBasedPolicy.SPURIOUS_PACKET_LOST_THRESHOLD
+        ) {
+          this.logger.info(
+            `bwe: In probe state, overriding estimate ${this.downlinkStats.bandwidthEstimateKbps} with actual receive bitrate ${this.downlinkStats.usedBandwidthKbps}`
+          );
+          targetBitrate =
+            this.downlinkStats.usedBandwidthKbps +
+            VideoPriorityBasedPolicy.USED_BANDWIDTH_OVERRIDE_BUFFER_KBPS;
```
ltrung, line 457: How do you determine this buffer value?

hensmi-amazon: Totally arbitrary. Just has to be above what the average bitrates are. Since its still pretty small it shouldn't make it too aggressive or anything.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -552,31 +582,33 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
...
+        return UseReceiveSet.PreProbe;
+      }
     }
     const subscribedRate = this.calculateSubscribeRate(this.optimalReceiveStreams);
     if (this.chosenStreamsSameAsLast(chosenStreams) || targetDownlinkBitrate > subscribedRate) {
-      let avgRate = 0;
-      for (const chosenStream of chosenStreams) {
-        avgRate += chosenStream.avgBitrateKbps;
-      }
-      if (targetDownlinkBitrate > avgRate) {
```
ltrung, line 572: Is this logic unnecessary? It doesn't seem to be mentioned in the PR description.

ltrung:
>Before, if BWE was gradually increasing it would hit the if statement every cycle, even if it had eventually grown, e.g. 1 MB.

A snippet of the if statement would be nice :)

ltrung (dismissed):

ltrung (approved):

PR #2004, opened 2022-02-10
Update variable name for previous release metrics

nainkunal933:
**Issue #:**

**Description of changes:** Changing the variable name as the older one was not very clear.

**Testing:**

**Checklist:**

Yes

No

No

### DIFF
--- a/integration/js/utils/AppTestStep.js
+++ b/integration/js/utils/AppTestStep.js
@@ -18,7 +18,7 @@ class AppTestStep extends TestStep {
       this.driver = sessionInfo.driver;
     }
 
-    this.namespaceInfix = kiteBaseTest.payload.namespaceInfix || '';
+    this.cwNamespaceInfix = kiteBaseTest.payload.cwNamespaceInfix || '';
   }
 
   async step() {
@@ -60,9 +60,9 @@ class AppTestStep extends TestStep {
       return;
     }
     if (this.emitMetricToCommonNamespace()) {
-      emitMetric('Common', this.test.capabilities, this.metricName(), value, this.namespaceInfix);
+      emitMetric('Common', this.test.capabilities, this.metricName(), value, this.cwNamespaceInfix);
     }
-    emitMetric(this.test.testName, this.test.capabilities, this.metricName(), value, this.namespaceInfix);
+    emitMetric(this.test.testName, this.test.capabilities, this.metricName(), value, this.cwNamespaceInfix);
   }
 
   finished(message, data) {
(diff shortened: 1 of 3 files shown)


ltrung (dismissed):

integration/js/utils/AppTestStep.js
```diff
@@ -18,7 +18,7 @@ class AppTestStep extends TestStep {
       this.driver = sessionInfo.driver;
     }
 
-    this.namespaceInfix = kiteBaseTest.payload.namespaceInfix || '';
+    this.cwNameSpaceInfix = kiteBaseTest.payload.cwNameSpaceInfix || '';
```
LeviSklaroff, line 21: It should be `cwNamespaceInfix` because `Namespace` is one word

integration/js/utils/CloudWatch.js
```diff
@@ -4,13 +4,13 @@ var cloudWatch = new AWS.CloudWatch({
   apiVersion: '2010-08-01'
 });
 
-module.exports.emitMetric = async (namespace, capabilities, metric_name, value, namespaceInfix) => {
+module.exports.emitMetric = async (namespace, capabilities, metric_name, value, cwNameSpaceInfix) => {
```
xuesichao, line 7: It's kinda confusing to me that one argument is named as `namespace` and the other is named as `cwNameSpaceInfix`.

Because it's feel like there are two different namespaces. One is `namespace`, the other is `cwNamespace`.
But what `cwNameSpaceInfix` does is to add an string segment into `namespace`.

LeviSklaroff: I agree with this, we should probably also rename `namespace` to `cwNamespace`

integration/js/utils/CloudWatch.js
```diff
@@ -4,13 +4,16 @@ var cloudWatch = new AWS.CloudWatch({
   apiVersion: '2010-08-01'
 });
 
-module.exports.emitMetric = async (namespace, capabilities, metric_name, value, namespaceInfix) => {
+// cwNamespaceInfix is a param that is used to update the namespace of CloudWatch metrics.
+// This infix will be added when this variable is passed to the KITE test from the payload.
+// It is used to differentiate metrics between the previous major release and the latest release.
```
LeviSklaroff, line 9: Hmm, I think the cwNamespaceInfix is self explanatory, but the larger problem is it's just not clear what the emitMetric function is supposed to do in the the first place. So if you could change the comment to be something more like `This function will create a metric in Cloudwatch using the given params in this format: TestType/cwNamespaceInfix/namespace/metric_name` helps a lot with understanding what the function is actually doing.

LeviSklaroff (approved):

integration/js/utils/CloudWatch.js
nainkunal933: Done

Issue #2017, opened 2022-02-14
Blur Processor Worker Does Not Close


MitchTalmadge:
### What happened and what did you expect to happen?

After calling `BackgroundBlurVideoFrameProcessor.create()` followed by `processor.destroy()` once the processor has been created, the Dedicated Worker does not close and the message `unknown operation close` is emitted by the worker.

Here I have repeated the example 10 times by use of this [codepen](https://codepen.io/mitchtalmadge/pen/podWbox), and you can see that the workers are still alive in the Chrome task manager:

![image](https://user-images.githubusercontent.com/8848080/153934507-dad9a78f-c887-46de-ac8b-f4250841852f.png)

![image](https://user-images.githubusercontent.com/8848080/153934782-d8098c14-3499-4dee-a30c-e234747009ca.png)

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

1. Create a `BackgroundBlurVideoFrameProcessor`
2. Destroy it.
3. Check the Chrome task manager and see that the background worker is still there.

Try my example: https://codepen.io/mitchtalmadge/pen/podWbox (you need to use the F12 console, not the codepen console, or you won't see the message.)

### Amazon Chime SDK for JavaScript version

2.26.1

### What browsers are you seeing the problem on?

Google Chrome

### Browser version

98.0.4758.82

### Browser console logs

```
iframeConsoleRunner-d0f3648046d2aaca07bd0037b9e061a26c74a8a999b75672ad6a638cca641472.js:1 2022-02-14T19:44:24.937Z [INFO] BackgroundBlurProcessor - BackgroundBlur processor successfully created
iframeConsoleRunner-d0f3648046d2aaca07bd0037b9e061a26c74a8a999b75672ad6a638cca641472.js:1 2022-02-14T19:44:24.937Z [INFO] BackgroundBlurProcessor - BackgroundBlur spec: {
  "paths": {
[...]

PR #2020, opened 2022-02-15
Fix worker leak for background blur and replacement (#2017)

richhx:
**Issue #:** #2017

**Description of changes:**
This fixes a worker leak for background blur and replacement by changing the `destroy` message to `stop`. Previously, when the wrong message was sent, the worker would not be properly shut down and would still run, thereby consuming unnecessary resources. 

This change is for the 3.x release branch. A separate one will be created to hotfix this in 2.x.

**Testing:**
Added a unit test to `BackgroundBlurProcessor.test.ts` and verified the reproduction case as mentioned in the initial issue.

No.

**Checklist:**

Yes.

No.

No.

Yes.


michhyun1:
Hey Richard, thanks for the PR - perhaps you can open a PR from a remote branch rather than from a forked repo. That should resolve these failing CI workflow jobs.

PR #2022, opened 2022-02-15
Fix worker leak for background blur and replacement (#2017)

richhx:
**Issue #:** #2017

**Description of changes:**
This fixes a worker leak for background blur and replacement by changing the `destroy` message to `stop`. Previously, when the wrong message was sent, the worker would not be properly shut down and would still run, thereby consuming unnecessary resources. Same as #2020 but not forked.

This change is for the 3.x release branch. A separate one will be created to hotfix this in 2.x.

**Testing:**
Added a unit test to `BackgroundBlurProcessor.test.ts` and verified the reproduction case as mentioned in the initial issue.

To verify the changes when using background blur in the SDK, toggling background blur on and off (e.g. calling `BackgroundBlurVideoFrameProcessor.create()` and then `processor.destroy()`) should not leak web workers anymore. This is verified by running the demo application (or the simple demo linked in the issue), switching between background blur 10% and 20%, and using the chrome task manager to see that the number of workers keeps increasing without getting deleted.

*Can these tested using a demo application? Please provide reproducible step-by-step instructions.*
Yes. 
1. Launch the demo application.
2. Create a meeting
3. Enable your camera
4. Join the meeting
5. Select video on by clicking the video icon.
6. Select Background Blur 10% by clicking the circle half filled white and black.
7. Select Background Blur 20%.
8. Keep alternating between 10% and 20% and notice that the number of "Dedicated Worker:"s keep increasing in the Chrome task manager (sort by CPU).

**Checklist:**

1. Have you successfully run `npm run build:release` locally?
Yes.

2. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved?
No.

3. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved?
No.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
Yes.

PR #2023, opened 2022-02-15
Fix worker leak for background blur and replacement (#2017)

richhx:
**Issue #:** #2017

**Description of changes:**
This is the same as #2022 except for the 2.x release branch.

This fixes a worker leak for background blur and replacement by changing the destroy message to stop. Previously, when the wrong message was sent, the worker would not be properly shut down and would still run, thereby consuming unnecessary resources.

**Testing:**
Added a unit test to BackgroundBlurProcessor.test.ts and verified the reproduction case as mentioned in the initial issue.

To verify the changes when using background blur in the SDK, toggling background blur on and off (e.g. calling `BackgroundBlurVideoFrameProcessor.create()` and then `processor.destroy()`) should not leak web workers anymore. This is verified by running the demo application (or the simple demo linked in the issue), switching between background blur 10% and 20%, and using the chrome task manager to see that the number of workers keeps increasing without getting deleted.

Yes. 
1. Launch the demo application.
2. Create a meeting
3. Enable your camera
4. Join the meeting
5. Select video on by clicking the video icon.
6. Select Background Blur 10% by clicking the circle half filled white and black.
7. Select Background Blur 20%.
8. Keep alternating between 10% and 20% and notice that the number of "Dedicated Worker:"s keep increasing in the Chrome task manager (sort by CPU).

**Checklist:**

Yes.

No.

No.

Yes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.27.1] - 2022-02-17
+
+### Added
+
+### Removed
+
+### Fixed
+- Fix a worker resource leak with `BackgroundBlurProcessor` and `BackgroundReplacementProcessor`.
+
 ## [2.27.0] - 2022-01-27
     
 ### Added
--- a/src/backgroundfilter/BackgroundFilterProcessor.ts
+++ b/src/backgroundfilter/BackgroundFilterProcessor.ts
@@ -398,7 +398,7 @@ export default abstract class BackgroundFilterProcessor {
     this.delegate.removeObserver(this.cpuMonitor);
     this.canvasVideoFrameBuffer.destroy();
     this.worker?.postMessage({ msg: 'destroy' });
-    this.worker?.postMessage({ msg: 'close' });
+    this.worker?.postMessage({ msg: 'stop' });
     this.targetCanvas?.remove();
     this.targetCanvas = undefined;
     this.scaledCanvas?.remove();
--- a/test/backgroundprocessor/BackgroundBlurProcessor.test.ts
+++ b/test/backgroundprocessor/BackgroundBlurProcessor.test.ts
@@ -919,16 +919,33 @@ describe('BackgroundBlurProcessor', () => {
     });
 
     it('can destroy', async () => {
-      backgroundFilterCommon.stubInit({ initPayload: 2, loadModelPayload: 2 });
+      const worker = sandbox.spy(
+        backgroundFilterCommon.stubInit({ initPayload: 2, loadModelPayload: 2 }),
+        'postMessage'
+      );
       stubSupported(true, false);
 
       let bbprocessor = (await BackgroundBlurVideoFrameProcessor.create()) as BackgroundBlurProcessorBuiltIn;
       bbprocessor['blurCanvas'] = null;
       await bbprocessor.destroy();
 
+      // Verify `stop` is called for BackgroundBlurProcessorBuiltIn.
+      // Workers loaded for background blur must maintain API contract
+      // of `stop` command existing. Integration tests should exist on
+      // the worker code to ensure that it does not change. As a result,
+      // a unit test is fine here to ensure that `stop` contract is
+      // fulfilled on the client side.
+      expect(worker.callCount).to.be.equal(4);
+      expect(worker.calledWith({ msg: 'stop' })).to.be.true;
+      worker.resetHistory();
+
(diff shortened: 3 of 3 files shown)


src/backgroundfilter/BackgroundFilterProcessor.ts
```diff
@@ -398,7 +398,7 @@ export default abstract class BackgroundFilterProcessor {
     this.delegate.removeObserver(this.cpuMonitor);
     this.canvasVideoFrameBuffer.destroy();
     this.worker?.postMessage({ msg: 'destroy' });
-    this.worker?.postMessage({ msg: 'close' });
+    this.worker?.postMessage({ msg: 'stop' });
```
devalevenkatesh, line 401: For learning: Should the `stop` be before `destroy`?

richhx: Good question. `destroy` should go first because `destroy` is meant to release the internal state of the blur process within the worker whereas `stop` actually ends the worker by calling `self.close() (https://developer.mozilla.org/en-US/docs/Web/API/DedicatedWorkerGlobalScope/close). Ideally, the command should've been named `close` but the API contract is already made, so changing it now isn't worth the effort.

Released worker code for reference https://static.sdkassets.chime.aws/bgblur/workers/worker.js. I can also ping you the internal package if you'd prefer to look at that instead.

devalevenkatesh: Ok makes sense, thanks for the explanation :)

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.27.1] - 2022-02-XX
```
devalevenkatesh, line 8: I am a bit behind catching on the actual release process and the naming convention in this CHANGELOG so tagging Trung to check once. @ltrung could you please check this one?

devalevenkatesh (changes requested):
You also need check the test coverage as it is not fulfilled, below is the report from the failed build task.
ERROR: Coverage for branches (99.98%) does not meet global threshold (100%)

Statements   : 100% ( 10049/10049 )
Branches     : 99.98% ( 4555/4556 )
Functions    : 100% ( 1876/1876 )
Lines        : 100% ( 9931/9931 )

CHANGELOG.md
devalevenkatesh: Rest looks good to me.

devalevenkatesh (changes requested):
Under Testing, can we mention how to check for the worker resources through browser devtools so while testing one can verify this change by using the background blur feature.

package.json
```diff
@@ -34,7 +34,7 @@
     "doc": "ruby script/generate-docs",
     "postdoc": "node script/update-typedoc-link.js",
     "test": "npm run test:fast",
-    "test:glob": "cross-env TS_NODE_PROJECT=test/tsconfig.json TS_NODE_CACHE=false nyc mocha --stack-trace-limit=1000  --async-stack-traces -v -r esm -rv ts-node/register -- ",
+    "test:glob": "cross-env TS_NODE_PROJECT=test/tsconfig.json TS_NODE_CACHE=false nyc --reporter=text mocha --stack-trace-limit=1000  --async-stack-traces -v -r esm -rv ts-node/register -- ",
```
richhx, line 37: I added the `--reporter=text` here to make it easier to find out why coverage is missing.

richhx:
> You also need check the test coverage as it is not fulfilled, below is the report from the failed build task:
> 
> ```
> ERROR: Coverage for branches (99.98%) does not meet global threshold (100%)
> 
> =============================== Coverage summary ===============================
> Statements   : 100% ( 10049/10049 )
> Branches     : 99.98% ( 4555/4556 )
> Functions    : 100% ( 1876/1876 )
> Lines        : 100% ( 9931/9931 )
> ================================================================================
> ```

Thanks, didn't see that >_> I added `--reporter=text` to the test to make it easier to find the offending coverage lines.

CHANGELOG.md
devalevenkatesh: Could you fix the date here? So instead of 2022-02-XX, it should be the day the next version open for development which would be todays as this will get in today :)

package.json
ltrung: I dont think it is very useful. I prefer to see the summary over this. This disable the local html report in local which is very useful to view it (instead of text).

richhx: I found the summary not very helpful to find where the missing branch was. If you prefer to use local html report, that's fine.

CHANGELOG.md
richhx: Changed it to 2022-02-17

devalevenkatesh (approved):

PR #2032, opened 2022-02-17
Removed upward logic in VideoPriorityBasedPolicyConfig

bjason:
**Issue #:**

The content share may fail to resume after turning off network link conditioner. 

**Description of changes:**

Removed upward logic in `VideoPriorityBasedPolicyConfig`.

**Testing:**

Smoke tested locally. Used network link conditioner to control bandwidth to validate video tile will resume after bandwidth recovers.

no

**Checklist:**

yes

no

no

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     
 ### Changed
 - Add support for pre-release in Versioning.
-- Removed upward logic in VideoPriorityBasedPolicyConfig.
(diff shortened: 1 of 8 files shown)


ltrung:
@bjason Can we put more details on the PR description about the issue? Why we want to remove the upward logic (maybe explain what that is also)? Otherwise, noone would understand why we do this several months from now.

hensmi-amazon (dismissed):
For context, this was causing issues to recovery time, and its initial focus was towards slowing downturns in BWE when the network is actually stable anyways. We may come back to configuring the recovery delay another time.

vidya-mohan (changes requested):
Can you update the changelog with more details on the fix - right now its notclear what the fix is doing

hensmi-amazon (approved):

vidya-mohan (approved):

PR #2039, opened 2022-02-18
Fix media capture

nehtambe:
**Issue #:**
Media capture is not working in non-prod demo apps (Beta stage) because we are replacing the endpoint value of Chime client with Chime SDK meetings endpoint.

**Description of changes:**
To fix the above issue - 
1. Added a new optional parameter -m to pass Chime SDK meetings endpoint. Whenever an internal team wants to use ChimeSDKMeetings to create regional meetings, they should pass SDK meetings endpoint under this parameter not under -c (chime-endpoint).
2. Update deploy-canary-demo script to pass both Chime and Chime SDK meetings endpoint.

**Testing:**

Yes, using the following commands.

npm run deploy -- -r us-east-1 -b chime-sdk-meetings-demo-us-east-1-test -s chime-sdk-meetings-demo-us-east-1-test -o chime-sdk-meetings-iad-demo-test -p chime.aws.internal -c <chime_gamma_endpoint> -m <chime_sdk_meetings_iad_gamma_endpoint> -t -l

npm run deploy -- -r eu-central-1  -b chime-sdk-meetings-demo-eu-central-1 -test -s chime-sdk-meetings-demo-eu-central-1 -test -o chime-sdk-meetings-fra-demo-test -p chime.aws.internal -c <chime_gamma_endpoint> -m <chime_sdk_meetings_fra_gamma_endpoint> -t -l

npm run deploy -- -r ap-southeast-1 -b chime-sdk-meetings-demo-ap-southeast-1-test -s chime-sdk-meetings-demo-ap-southeast-1-test -o chime-sdk-meetings-sin-demo-test -p chime.aws.internal -c <chime_gamma_endpoint> -m <chime_sdk_meetings_sin_gamma_endpoint> -t -l

**Checklist:**

Yes

No

No


demos/serverless/deploy.js
```diff
@@ -13,6 +13,7 @@ let useEventBridge = false;
 let enableTerminationProtection = false;
 let disablePrintingLogs = false;
 let chimeEndpoint = 'https://service.chime.aws.amazon.com';
+let chimeSDKMeetingsEndpoint = 'https://service.chime.aws.amazon.com';
```
devalevenkatesh, line 16: There is a `chimeEndpoint` with same value already, any reasons we need a new one?

demos/serverless/deploy.js
```diff
@@ -48,6 +49,7 @@ function usage() {
   console.log(`  -l, --disable-printing-logs          Disable printing logs`);
   console.log(`  -o, --capture-output-prefix          Prefix for S3 bucket name`);
   console.log(`  -i, --opt-in-regions                 Comma separated list of additional opt-in regions to enable for media capture`);
+  console.log(`  -m, --chime-sdk-meetings-endpoint    AWS SDK Chime Meetings endpoint`);
```
devalevenkatesh, line 52: Suggest changing the name:
AWS SDK Chime Meetings -> AWS Chime SDK Meetings endpoint

devalevenkatesh (commented):
Request to check with someone from meetings SDK once.

demos/serverless/deploy.js
nehtambe: Followed the similar pattern as "--chime-endpoint AWS SDK Chime endpoint"

nehtambe:
> Request to check with someone from meetings SDK once.

Yes, I have asked Harshit to check the PR

harshitOberoi (changes requested):

demos/serverless/deploy.js
```diff
@@ -251,7 +256,7 @@ spawnOrFail('sam', ['package', '--s3-bucket', `${bucket}`,
                     `--output-template-file`, `build/packaged.yaml`,
                     '--region',  `${region}`]);
 console.log('Deploying serverless application');
-let parameterOverrides = `Region=${region} UseChimeSDKMeetings=${useChimeSDKMeetings} UseEventBridge=${useEventBridge} ChimeEndpoint=${chimeEndpoint} ChimeServicePrincipal=${chimeServicePrincipal}`
+let parameterOverrides = `Region=${region} UseChimeSDKMeetings=${useChimeSDKMeetings} UseEventBridge=${useEventBridge} ChimeEndpoint=${chimeEndpoint} ChimeServicePrincipal=${chimeServicePrincipal} ChimeSDKMeetingsEndpoint=${chimeSDKMeetingsEndpoint}`
```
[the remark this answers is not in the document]
nehtambe: We are using this at other places too, should we replace the conditions using 'ChimeSDKMeetingsEndpoint'?

demos/serverless/src/handlers.js
```diff
@@ -13,6 +13,7 @@ const ddb = new AWS.DynamoDB();
 const endpoint = process.env.CHIME_ENDPOINT;
 const currentRegion = process.env.REGION;
 const useChimeSDKMeetings = process.env.USE_CHIME_SDK_MEETINGS;
```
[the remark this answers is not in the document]
nehtambe: Didn’t understand the first part. This is a boolean variable and it will always have a default value.

demos/serverless/deploy.js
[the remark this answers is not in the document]
nehtambe: I was checking the code, we are using it to determine which client to use. For prod demo app we will not pass ChimeSDKMeetingsEndpoint, so I think we will need "useChimeSDKMeetings" to determine the client.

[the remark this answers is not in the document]
harshitOberoi: so for prod instead of paasing -u  we should pass ChimeSDKMeetingsEndpoint if we want to use meetingsSDK

demos/serverless/src/handlers.js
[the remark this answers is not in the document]
harshitOberoi: useChimeSDKMeetings  should be true if we pass USE_CHIME_SDK_MEETINGS otherwise it should be false. SO instead of passing -u flag to determine whar chime client to use we should use USE_CHIME_SDK_MEETINGS parameter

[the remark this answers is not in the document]
nehtambe: I think you meant set value of constant useChimeSDKMeetings to false, but when CHIME_SDK_MEETINGS_ENDPOINT is passed, set the value to true

demos/serverless/deploy.js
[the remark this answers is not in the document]
nehtambe: Okay yes we can do that, in that case it will have same value as CHIME_ENDPOINT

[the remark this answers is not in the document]
harshitOberoi: I think i am fine with that parameter. Just handle a case if -u is true and ChimeSDKMeetingsEndpoint is null use chimeEndpoint.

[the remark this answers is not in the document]
nehtambe: ChimeSDKMeetingsEndpoint will always have default value - 'https://service.chime.aws.amazon.com', we will have to remove the default value in order to check for if the endpoint is passed or not. If we do this, everyone, including external and internal users, will have to pass the ChimeSDKMeetingsEndpoint (-m param) in the deployment command

demos/serverless/src/handlers.js
```diff
@@ -23,8 +24,8 @@ const chime = new AWS.Chime({ region: 'us-east-1' });
 chime.endpoint = new AWS.Endpoint(endpoint);
 
 const chimeSDKMeetings = new AWS.ChimeSDKMeetings({ region: currentRegion });
-if(endpoint != 'https://service.chime.aws.amazon.com'){
-  chimeSDKMeetings.endpoint = new AWS.Endpoint(endpoint);
+if(chimeSDKMeetingsEndpoint != 'https://service.chime.aws.amazon.com' && useChimeSDKMeetings === 'true'){
+  chimeSDKMeetings.endpoint = new AWS.Endpoint(chimeSDKMeetingsEndpoint);
```
devalevenkatesh, line 28: If this is not satisfied what is the `chimeSDKMeetings` endpoint? Is there a default value of `https://service.chime.aws.amazon.com` to it?

script/deploy-canary-demo
```diff
@@ -28,9 +28,9 @@ ChimeSDKMeetingsBeta)
...
-    npm run deploy -- -r eu-central-1 -b chime-sdk-meetings-demo-beta-canary-eu-central-1$canarySuffix -s chime-sdk-meetings-demo-beta-canary-eu-central-1$canarySuffix -o chime-sdk-meetings-fra-demo-beta-canary$canarySuffix -p $GAMMA_CHIME_SERVICE_PRINCIPAL -c $GAMMA_CHIME_ENDPOINT_EU_CENTRAL_1 -t -l
-    npm run deploy -- -r ap-southeast-1 -b chime-sdk-meetings-demo-beta-canary-ap-southeast-1$canarySuffix -s chime-sdk-meetings-demo-beta-canary-ap-southeast-1$canarySuffix -o chime-sdk-meetings-sin-demo-beta-canary$canarySuffix -p $GAMMA_CHIME_SERVICE_PRINCIPAL -c $GAMMA_CHIME_ENDPOINT_AP_SOUTHEAST_1 -t -l
+    npm run deploy -- -r us-east-1 -b chime-sdk-meetings-demo-beta-canary-us-east-1$canarySuffix -s chime-sdk-meetings-demo-beta-canary-us-east-1$canarySuffix -o chime-sdk-meetings-iad-demo-beta-canary$canarySuffix -p $GAMMA_CHIME_SERVICE_PRINCIPAL -c $GAMMA_CHIME_ENDPOINT -m $GAMMA_CHIME_ENDPOINT_US_EAST_1 -t -l
```
devalevenkatesh, line 31: So this is for supporting media capture tests using ChimeSDKMeetings endpoint for creating meeting and Chime's gamma endpoint for calling media capture APIs right?

demos/serverless/src/handlers.js
devalevenkatesh: Also, would it makes sense to just remove the `useChimeSDKMeetings` now that we are adding an explicit `chimeSDKMeetingsEndpoint` URL as new parameter?

nehtambe: Yes, right, it will be pointing to Prod endpoint

script/deploy-canary-demo
nehtambe: Yes, correct. This is for handling the scenario where we want to create regional meetings using  ChimeSDKMeetings and test a capture using those regional meetings. This is especially for Non-Prod regions, since we have different endpoints.

nehtambe: So, if someone wants to create regional meetings for non-prod regions, they have to pass the ChimeSDKMeetings endpoint in new parameter ChimeSDKMeetingsEndpoint

devalevenkatesh (commented):
LGTM! I will approve once mobile team checks since they use the same demo and hence checking if they need to update something on their side before final approval.

demos/serverless/src/handlers.js
devalevenkatesh: So `useChimeSDKMeetings` is basically true by default and says to use the `ChimeMeetingSDK` for regional endpoints and the `chimeSDKMeetingsEndpoint` actually specifies the regional `endpoint`. Be default, all deployments use the `ChimeMeetingSDK` regional endpoint leaving the canary test demos as we are deploying new demos to test the regional endpoints already.

Discussed offline but the clarification is ^.

demos/serverless/src/handlers.js
```diff
@@ -23,8 +24,8 @@ const chime = new AWS.Chime({ region: 'us-east-1' });
 chime.endpoint = new AWS.Endpoint(endpoint);
 
 const chimeSDKMeetings = new AWS.ChimeSDKMeetings({ region: currentRegion });
-if(endpoint != 'https://service.chime.aws.amazon.com'){
-  chimeSDKMeetings.endpoint = new AWS.Endpoint(endpoint);
+if(chimeSDKMeetingsEndpoint != 'https://service.chime.aws.amazon.com' && useChimeSDKMeetings === 'true'){
```
ltrung, line 27: if `useChimeSDKMeetings` is false but `chimeSDKMeetingsEndpoint` is set then what does that mean? Can we just always respect `chimeSDKMeetingsEndpoint` and if it is set then automatically set `useChimeSDKMeetings` to true?

nehtambe: In that case, we will be using Global Meetings. Also, we can not do that because we are defaulting value of chimeSDKMeetingsEndpoint to https://service.chime.aws.amazon.com, and we will not have any way to distinguish between when to use Global Meetings Client and Chime SDK meetings client for Prod.

hokyungh (approved):
Did sanity test with new url. No issue found LGTM

ltrung (approved):

harshitOberoi (approved):

PR #2054, opened 2022-02-23
Remove sdp interface

akshay-prabhu-github:
**Issue #:**
We always ever use a DefaultSDP and there is no real need for a SDP interface.

**Description of changes:**
Removing the SDP interface.

**Testing:**

Unit tests succeeded.

No

**Checklist:**

Y

N

N


src/sdp/DefaultSDP.ts
```diff
@@ -1,21 +1,23 @@
...
-import SDP from './SDP';
 import SDPCandidateType from './SDPCandidateType';
 import SDPMediaSection from './SDPMediaSection';
 
 /**
- * Implements [[SDP]]. [[SDP]] also includes a few helper functions for parsing string.
+ * [[DefaultSDP]] includes a few helper functions for parsing sdp string.
  */
-export default class DefaultSDP implements SDP {
+export default class DefaultSDP {
```
ltrung, line 10: Since there is only one implementation of SDP now, I dont think we should call it default anymore and just call it SDP :)

hensmi-amazon (dismissed):
woot

src/sdp/SDP.ts
```diff
@@ -5,66 +5,567 @@ import SDPCandidateType from './SDPCandidateType';
 import SDPMediaSection from './SDPMediaSection';
 
 /**
- * [[SDP]] manages and helps munge an SDP during negotiation.
+ * [[SDP]] includes a few helper functions for parsing sdp string.
  */
-export default interface SDP {
+export default class SDP {
```
ltrung, line 10: Since we do not need to expose this to builders, can we just use the tag `/** @internal */` above this and do not use export keyword so we do not export it?

akshay-prabhu-github: I tried adding the tag `/** @internal */` and it started throwing errors as there are some external api files that make use of the SDP class.

ltrung (approved):

Issue #1660, opened 2021-10-08
DeviceController audio memory leak


Nanosync:
### What happened and what did you expect to happen?

I was profiling a simple React component that only imports the chime JS sdk. It simply creates a new DefaultDeviceController when mounted, and calls the `.destroy()` method on unmount.

It seems that `AudioBufferSourceNode` persists and a new one is created every time the component is mounted.

This was discovered using heap snapshots in Chrome.

Heap snapshot of initially unmounted component that is mounted and unmounted, repeated twice (Unmounted, Mounted, Unmounted, Mounted, Unmounted):
![heap](https://user-images.githubusercontent.com/10258208/137267297-23702cfd-1554-41b5-b846-b0fe7789e147.png)

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

1. Create a new `DefaultDeviceController`
2. Call `.destroy()` on the created instance
3. Create a new instance of `DefaultDeviceController` again
4. Memory leak is observed with an additional `AudioBufferSourceNode`

### Amazon Chime SDK for JavaScript version

2.17.0

### What browsers are you seeing the problem on?

Chrome

### Browser version

94.0.4606.71

### Browser console logs

N/A

emopti-jrufer:
I am running into the same issue. I am not using using the React components but rather using the SDK in an Angular app. What I discovered when trying to figure out why the memory in our app was increasing after every meeting is that the "AudioBufferSourceNode" objects were being created during the teardown of the meeting. Initially followed the convention used in the examples for tearing down by calling `MeetingSession.audioVideo.chooseXXXDevice(null)`. The `chooseAudioInputDevice(null)` is creating a the empty audio device. Next tried calling `MeetingSession.destroy()`, which is not documented anywhere, instead of calling `chooseAudioInputDevice(null)`. This resulted in the same behavior because it is doing the same thing.

Overall there needs to be a short circuit when calling `chooseAudioInputDevice(null)` during `destroy` so it does not create new resources that are never cleaned up. The following reference is what is creating the empty device during destroy. 

[...]

PR #2062, opened 2022-02-24
Fix DefaultDeviceController memory leaks issue

xuesichao:
**Issue #1660 :**

#### Root Cause
Whenever `synthesizeAudioDevice(0)` is called, a new `AudioBuffer` is created, preserves in memory and is never garbage collected and thus causing memory leak issue.
https://github.com/aws/amazon-chime-sdk-js/blob/0792b712b272c951ea967fe5f9ab0bf908c26689/src/devicecontroller/DefaultDeviceController.ts#L1042-L1058

#### Solution
The `AudioBuffer` can be shared by different `AudioBufferSourceNode`. So instead of creating a new `AudioBuffer` each time, add a new property `defaultAudioBuffer` to preserve a `AudioBuffer` which can be reused when creating empty audio device and solve the memory leak issue.

**Description of changes:**
* Add a static property `defaultAudioBuffer` to `DefaultDeviceController` to store the `AudioBuffer`.
* In the `destroy()` method of `DefaultDeviceController`, use cleaning up active audio device logic instead of calling `chooseAudioInputDevice(null)`.

**Testing:**
#### Conclusion
The memory only increases once no matter how many `DefaultDeviceController` instance we initialized and how many times we call `chooseAudioInputDevice(null)` method.

1. Replace `meetingReadinessChecker.ts` with below code:
    ```js
    import { ConsoleLogger, LogLevel, DefaultDeviceController } from "../../../../build";
    
    window.addEventListener('load', async () => {
      const test = async () => {
        let logger = new ConsoleLogger('SDK', LogLevel.DEBUG);
    
        let controller = new DefaultDeviceController(logger);
        await controller.chooseAudioInputDevice(null);
        await controller.chooseAudioInputDevice(null);
        await controller.destroy();
        controller = undefined;
    
        controller = new DefaultDeviceController(logger);
        await controller.chooseAudioInputDevice(null);
        await controller.chooseAudioInputDevice(null);
        await controller.destroy();
        controller = undefined;
      };
      await test();
      console.debug('test');
    });
    ```
2. Running MeetingReadinessChecker demo with Chrome
3. In debugger tool, add breakpoint whenever:
   * initializing a new `DefaultDeviceController`
   * calling `destroy()` method
   * calling `chooseAudioInputDevice(null)` method
4. Use heap snapshot to scan the memory usage everytime when debugger stops at breakpoint and compare the heap snapshots.

image

**Checklist:**

Yes

6. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved?
No

7. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved?
No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add compression support when sending and receiving sdp messages.
+- Add a static property `defaultAudioBuffer` to `DefaultDeviceController` to store the `AudioBuffer` used to create empty audio devices. When an empty audio device is created, the method reuses this `defaultAudioBuffer` instead of creating a new `AudioBuffer` each time and causing memory leak. ([Issue #1660](https://github.com/aws/amazon-chime-sdk-js/issues/1660)).
 
 ### Removed
 
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -122,6 +122,7 @@ export default class DefaultDeviceController
   private static defaultSampleRate = 48000;
   private static defaultSampleSize = 16;
   private static defaultChannelCount = 1;
+  private static defaultAudioBuffer: AudioBuffer | undefined;
   private static audioContext: AudioContext | null = null;
 
   private deviceInfoCache: MediaDeviceInfo[] | null = null;
@@ -265,7 +266,11 @@ export default class DefaultDeviceController
     // Deselect any audio input devices and throw away the streams.
     // Discard the current video device, if there is one.
     // Discard any audio or video transforms.
-    await this.chooseAudioInputDevice(null);
+    const active = this.activeDevices['audio'];
+    if (active) {
+      this.releaseActiveDevice(active);
+      delete this.activeDevices['audio'];
+    }
     await this.chooseVideoInputDevice(null);
 
     // Tear down any Web Audio infrastructure we have hanging around.
@@ -1039,27 +1044,30 @@ export default class DefaultDeviceController
       // The AudioContext object uses the sample rate of the default output device
       // if not specified. Creating an AudioBuffer object with the output device's
       // sample rate fails in some browsers, e.g. Safari with a Bluetooth headphone.
-      try {
-        source.buffer = audioContext.createBuffer(
-          1,
-          audioContext.sampleRate * 5,
-          audioContext.sampleRate
-        );
-      } catch (error) {
-        if (error && error.name === 'NotSupportedError') {
-          source.buffer = audioContext.createBuffer(
+      if (!DefaultDeviceController.defaultAudioBuffer) {
+        try {
+          DefaultDeviceController.defaultAudioBuffer = audioContext.createBuffer(
             1,
-            DefaultDeviceController.defaultSampleRate * 5,
-            DefaultDeviceController.defaultSampleRate
+            audioContext.sampleRate * 5,
+            audioContext.sampleRate
           );
-        } else {
-          throw error;
+        } catch (error) {
+          if (error && error.name === 'NotSupportedError') {
+            DefaultDeviceController.defaultAudioBuffer = audioContext.createBuffer(
+              1,
+              DefaultDeviceController.defaultSampleRate * 5,
+              DefaultDeviceController.defaultSampleRate
+            );
+          } else {
+            throw error;
+          }
         }
       }
 
       // Some browsers will not play audio out the MediaStreamDestination
       // unless there is actually audio to play, so we add a small amount of
       // noise here to ensure that audio is played out.
+      source.buffer = DefaultDeviceController.defaultAudioBuffer;
       source.buffer.getChannelData(0)[0] = 0.0003;
       source.loop = true;
       source.connect(outputNode);
--- a/test/devicecontroller/DefaultDeviceController.test.ts
+++ b/test/devicecontroller/DefaultDeviceController.test.ts
@@ -2780,6 +2780,9 @@ describe('DefaultDeviceController', () => {
       domMockBehavior.mediaDeviceHasSupportedConstraints = false;
       domMockBehavior.audioContextDefaultSampleRate = Infinity;
       try {
+        // @ts-ignore
+        DefaultDeviceController.defaultAudioBuffer = undefined;
+
         DefaultDeviceController.synthesizeAudioDevice(0);
       } catch (error) {
         throw new Error('This line should not be reached.');
@@ -2789,6 +2792,9 @@ describe('DefaultDeviceController', () => {
     it('fails if the create buffer throws a non-NotSupportedError error', async () => {
       domMockBehavior.audioContextCreateBufferSucceeds = false;
       try {
+        // @ts-ignore
+        DefaultDeviceController.defaultAudioBuffer = undefined;
+
         DefaultDeviceController.synthesizeAudioDevice(0);
         throw new Error('This line should not be reached.');
       } catch (error) {
--- a/test/dommock/DOMMockBuilder.ts
+++ b/test/dommock/DOMMockBuilder.ts
@@ -1243,8 +1243,9 @@ export default class DOMMockBuilder {
 
     GlobalAny.AudioBufferSourceNode = class MockAudioBufferSourceNode {
       start(_when?: number, _offset?: number, _duration?: number): void {}
-
-      connect(_destinationParam: AudioParam, _output?: number): void {}
+      stop(_when?: number): void {}
+      connect(_destination: AudioNode | AudioParam, _output?: number): void {}
+      disconnect(_destination?: AudioNode | AudioParam, _output?: number, _input?: number): void {}
     };
 
     GlobalAny.AudioBuffer = class MockAudioBuffer {
--- a/docs/classes/defaultdevicecontroller.html
+++ b/docs/classes/defaultdevicecontroller.html
@@ -147,7 +147,7 @@
 						<li class="tsd-description">
 							<aside class="tsd-sources">
 								<ul>
-									<li>Defined in <a href="https://github.com/aws/amazon-chime-sdk-js/blob/main/src/devicecontroller/DefaultDeviceController.ts#L169">src/devicecontroller/DefaultDeviceController.ts:169</a></li>
+									<li>Defined in <a href="https://github.com/aws/amazon-chime-sdk-js/blob/main/src/devicecontroller/DefaultDeviceController.ts#L170">src/devicecontroller/DefaultDeviceController.ts:170</a></li>
 								</ul>
 							</aside>
 							<h4 class="tsd-parameters-title">Parameters</h4>
@@ -184,7 +184,7 @@
 					<aside class="tsd-sources">
 						<p>Implementation of <a href="../interfaces/devicecontrollerbasedmediastreambroker.html">DeviceControllerBasedMediaStreamBroker</a>.<a href="../interfaces/devicecontrollerbasedmediastreambroker.html#eventcontroller">eventController</a></p>
 						<ul>
-							<li>Defined in <a href="https://github.com/aws/amazon-chime-sdk-js/blob/main/src/devicecontroller/DefaultDeviceController.ts#L175">src/devicecontroller/DefaultDeviceController.ts:175</a></li>
+							<li>Defined in <a href="https://github.com/aws/amazon-chime-sdk-js/blob/main/src/devicecontroller/DefaultDeviceController.ts#L176">src/devicecontroller/DefaultDeviceController.ts:176</a></li>
 						</ul>
 					</aside>
 				</section>
@@ -202,7 +202,7 @@
 							<aside class="tsd-sources">
 								<p>Implementation of <a href="../interfaces/devicecontrollerbasedmediastreambroker.html">DeviceControllerBasedMediaStreamBroker</a>.<a href="../interfaces/devicecontrollerbasedmediastreambroker.html#acquireaudioinputstream">acquireAudioInputStream</a></p>
 								<ul>
-									<li>Defined in <a href="https://github.com/aws/amazon-chime-sdk-js/blob/main/src/devicecontroller/DefaultDeviceController.ts#L731">src/devicecontroller/DefaultDeviceController.ts:731</a></li>
+									<li>Defined in <a href="https://github.com/aws/amazon-chime-sdk-js/blob/main/src/devicecontroller/DefaultDeviceController.ts#L736">src/devicecontroller/DefaultDeviceController.ts:736</a></li>
 								</ul>
 							</aside>
 							<h4 class="tsd-returns-title">Returns <span class="tsd-signature-type">Promise</span><span class="tsd-signature-symbol">&lt;</span><span class="tsd-signature-type">MediaStream</span><span class="tsd-signature-symbol">&gt;</span></h4>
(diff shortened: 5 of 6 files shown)


test/devicecontroller/DefaultDeviceController.test.ts
```diff
@@ -208,6 +208,18 @@ describe('DefaultDeviceController', () => {
       // Calling it again is safe.
       await deviceController.destroy();
     });
+
+    // it('the reference to audioBufferSourceNode should be undefined when disposed', async () => {
```
LeviSklaroff, line 212: Why is this unit test commented out?

CHANGELOG.md
```diff
@@ -6,40 +6,39 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 - Add support for aws-sdk js v3 for messaging session.
-    
+
 ### Fixed
 - Fix a worker resource leak with `BackgroundBlurProcessor` and `BackgroundReplacementProcessor`.
-    
+
 ## [2.27.0] - 2022-01-27
-    
+
```
LeviSklaroff, line 26: Why is there a line here? It looks like below we have no space between Added and version.

test/devicecontroller/DefaultDeviceController.test.ts
xuesichao: Forgot to remove this, will push a new commit to fix this. Thanks for catching!

xuesichao: Updated

CHANGELOG.md
xuesichao: It is best to have the `Heading`s surrounded by blank lines according to markdown lint. But this is not mandatory.
This change is to remove the unnecessary spaces/tabs.

LeviSklaroff: Gotcha, that makes sense. I just noticed it wasn't consistent but looks like you updated the other ones now.

LeviSklaroff (approved):

CHANGELOG.md
xuesichao: Yes, I usually clean up when I raise a PR. But there are still a lot of small issues, and they are increasing all the time. I can not always update them, but it's better than not doing anything.

CHANGELOG.md
```diff
@@ -6,60 +6,68 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [3.0.0-beta.1] - 2022-02-23
-    
+
```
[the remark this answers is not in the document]
xuesichao: Format stuff.
1. remove unnecessary spaces/tabs
2. fix several typos
3. make sure the `Heading`s are surrounded by blank lines.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -268,7 +269,18 @@ export default class DefaultDeviceController
...
+    // The active video device has already been cleaned up when `chooseVideoInputDevice(null)` is called above.
+    let active = this.activeDevices['audio'];
+    this.releaseActiveDevice(active);
+    active = undefined;
+    delete this.activeDevices['audio'];
+
     // Tear down any Web Audio infrastructure we have hanging around.
+    DefaultDeviceController.audioBufferSourceNode.stop();
+    DefaultDeviceController.audioBufferSourceNode.disconnect();
+    DefaultDeviceController.audioBufferSourceNode.buffer = undefined;
```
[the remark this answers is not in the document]
xuesichao: https://developer.mozilla.org/en-US/docs/Web/API/AudioBufferSourceNode
>The AudioBufferSourceNode interface is an [AudioScheduledSourceNode](https://developer.mozilla.org/en-US/docs/Web/API/AudioScheduledSourceNode) which represents an audio source consisting of in-memory audio data, stored in an [AudioBuffer](https://developer.mozilla.org/en-US/docs/Web/API/AudioBuffer).

So the idea is that the `audioBufferSourceNode.buffer` stores the actual audio source data. And since the `AudioBufferSourceNode` is not garbage collected. This audio source data preserve in memory and cause the leak.

This line manually cleans up this audio source data, so it fix the memory leak issue.
```
DefaultDeviceController.audioBufferSourceNode.buffer = undefined;
```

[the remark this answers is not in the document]
xuesichao: So here's the case:
Ideally if we can clean up the `audioBufferSourceNode` itself, the `audioBufferSourceNode.buffer` should be garbage collected as well. But somehow it preserve in the memory and I tried so many ways to clean it up but failed.

So I have to clean up the `audioBufferSourceNode.buffer` instead. Technically with only `    DefaultDeviceController.audioBufferSourceNode.buffer = undefined;` could fix this memory leak issue. 

But it's best practice to clean up the audio node like this:
```
 DefaultDeviceController.audioBufferSourceNode.stop();
 DefaultDeviceController.audioBufferSourceNode.disconnect();
```
So I keep these even they don't work now.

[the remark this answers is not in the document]
xuesichao: > Did not see we do it for VF? Or we have issue with VF too?

For `emptyAudioInputDevice` it has no actual input so we need to create an `AudioBuffer` to serve as audio input source.

For VF, the audio input source are from the real devices. so it does not have `AudioBuffer`, thus no need to clean up.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -1060,10 +1072,10 @@ export default class DefaultDeviceController
       // Some browsers will not play audio out the MediaStreamDestination
       // unless there is actually audio to play, so we add a small amount of
       // noise here to ensure that audio is played out.
-      source.buffer.getChannelData(0)[0] = 0.0003;
-      source.loop = true;
-      source.connect(outputNode);
-      source.start();
+      this.audioBufferSourceNode.buffer.getChannelData(0)[0] = 0.0003;
```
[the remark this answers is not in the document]
xuesichao: >Should this be a class variable or instance variable? Technically you can have multiple device controllers that call this right?

Yes, as we discussed earlier in the standup, it's better to be a private property. However the `createEmptyAudioDevice` is a static method, it can not access the private property. If we update `createEmptyAudioDevice` to private method, it's a breaking change.

>Also can't we just set sourceNode then assign it at the end before return to audioBufferSourceNode. Not sure why we need to replace everywhere like this.

Good point. Updated!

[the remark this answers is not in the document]
xuesichao: Yes, `this` is confusing. Updated to:
```js
const source = audioContext.createBufferSource();

// Store the reference to AudioBufferSourceNode for later cleanup.
DefaultDeviceController.audioBufferSourceNode = source;
```

CHANGELOG.md
ltrung: > It is best to have the Headings surrounded by blank lines according to markdown lint. But this is not mandatory.
This change is to remove the unnecessary spaces/tabs.

According to? Anyway, its better to fix it in the script that generates it https://github.com/aws/amazon-chime-sdk-js/blob/main/script/version-utils.js#L64.

src/devicecontroller/DefaultDeviceController.ts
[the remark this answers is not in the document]
ltrung: Nice. just want to check

CHANGELOG.md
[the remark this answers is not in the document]
xuesichao: >3. make sure the Headings are surrounded by blank lines.

This is not mandatory, but I saw this rule in `markdown lint` and think it's a good practice to follow and can make the style of our CHANGELOG raw code consistent, so I usually do a clean up once a while when I raise a PR.

src/devicecontroller/DefaultDeviceController.ts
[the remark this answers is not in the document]
ltrung: You can just return the node itself instead of the media stream right? Then you can set it to an instance variable.
```
static synthesizeAudioDevice(toneHz: number): audioBufferSourceNode {
```

I think you can create a new method with same name but return node instead.

CHANGELOG.md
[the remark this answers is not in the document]
ltrung: How do you fix it? Is there a way to automate it? So you don't have to do it everytime :)

CHANGELOG.md
xuesichao: I saw this rule in markdown lint and think it's a good practice to follow and can make the style of our CHANGELOG raw code consistent, so I usually do a clean up once a while when I raise a PR. ([markdown lint](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md022))

It's just a VSCode plugin, so it's not that official. But I observe we did this at the beginning when creating the CHANGLOG, so just want to follow it.

I will push a new commit to update that script.

CHANGELOG.md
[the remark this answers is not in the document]
xuesichao: The `markdownlint` plugin can fix this blank line issue automatically, so it's simple.
The problem is we don't have any tools in SDK to validate the CHANGELOG format for now. So we might need to pay attention to this when reviewing the PR?
Not sure it worth the effort to make it automated.

src/devicecontroller/DefaultDeviceController.ts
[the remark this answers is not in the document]
xuesichao: >You can just return the node itself instead of the media stream right?

I don't think so. When creating an empty device, the stream it returns is from `MediaStreamAudioDestinationNode`.
```
AudioBufferSourceNode ---connect--> MediaStreamAudioDestinationNode --> stream (returned).
```

If we return `audioBufferSourceNode` directly, then we need create the MediaStreamAudioDestinationNode and connect AudioBufferSourceNode to MediaStreamAudioDestinationNode and get/return the stream. Feel like it kinda break the change.

CHANGELOG.md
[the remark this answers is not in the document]
ltrung: We can ask everyone to install the same plugin to make sure we stay consistent :)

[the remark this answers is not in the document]
xuesichao: https://www.npmjs.com/package/markdownlint
Sure. I can try to make a POC of this package later.

test/dommock/DOMMockBuilder.ts
```diff
@@ -1243,8 +1243,9 @@ export default class DOMMockBuilder {
 
     GlobalAny.AudioBufferSourceNode = class MockAudioBufferSourceNode {
       start(_when?: number, _offset?: number, _duration?: number): void {}
-
-      connect(_destinationParam: AudioParam, _output?: number): void {}
+      stop(_when?: number): void {}
+      connect(_destination: AudioNode | AudioParam, _output?: number): void {}
+      disconnect(_destination?: AudioNode | AudioParam, _output?: number, _input?: number): void {}
```
xuesichao, line 1248: This is not used now but we might need this in the future, so I keep this change.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -265,7 +266,11 @@ export default class DefaultDeviceController
     // Deselect any audio input devices and throw away the streams.
     // Discard the current video device, if there is one.
     // Discard any audio or video transforms.
-    await this.chooseAudioInputDevice(null);
+    const active = this.activeDevices['audio'];
```
ltrung, line 269: nit: could we put a comment explain not to call `chooseAudioInputDevice(null)` to clean up audio either here or maybe better above the `chooseAudioInputDevice` method?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -122,6 +122,7 @@ export default class DefaultDeviceController
   private static defaultSampleRate = 48000;
   private static defaultSampleSize = 16;
   private static defaultChannelCount = 1;
+  private static defaultAudioBuffer: AudioBuffer | undefined;
```
ltrung, line 125: When do we clear out this? Doesn't that mean we will always have audio buffer in memory until the class itself is removed?

xuesichao: I made a POC and the conclusion is that an `AudioBuffer` will be GC'd once there is no reference to this object.

When creating empty audio device, the `AudioBufferSourceNode` refers to this `defaultAudioBuffer`:

https://github.com/aws/amazon-chime-sdk-js/blob/b3b463c522479d7590cd2b9bed5d55bc14d7fa9c/src/devicecontroller/DefaultDeviceController.ts#L1070

 which means even if we set `DefaultDeviceControl.defaultAudioBuffer = undefined`, there're still references to `defaultAudioBuffer` and thus the memory will not be released.

So if we really want to clear this out, we need an array to track each `AudioBufferSourceNode`s and clean up the `buffer` property of them and  later clean up the `DefaultDeviceControl.defaultAudioBuffer`. I don't think we want to do that.

> Doesn't that mean we will always have audio buffer in memory until the class itself is removed?

Yes, the `DefaultDeviceControl.defaultAudioBuffer` takes up to 0.96MB in the memory and I think it is acceptable.
This size can be reduce if we reduce the `duration` of `AudioBuffer`. Currently it's 5s and we can reduce it to 1s or 2s.And it's a static property, it makes sense to preserve until the class is removed. We basically never clean up the `AudioContext` as well.

ltrung: Could we just make this instance variable and clear it out when we dont need it (Basically make it part of release audio stream)?

xuesichao: 1. I don't think it should be instance variable. These `synthesize*Device()` method are more like utility, the best solution I think is to have another class to store them as static method and variables. It should not be part of `DefaultDeviceController` and should be decoupled.

2. If we change it to instance variable, it will break current design for those have no dependency to `DefaultDeviceController`, for example, there is no way to access to `DefaultDeviceController` instance here:
https://github.com/aws/amazon-chime-sdk-js/blob/0792b712b272c951ea967fe5f9ab0bf908c26689/src/contentsharecontroller/ContentShareMediaStreamBroker.ts#L24-L30

src/devicecontroller/DefaultDeviceController.ts
xuesichao: The issue is there's no API to clean up audio devices like what `chooseVideoInputDevice(null)` does. Do we need to create a method and expose it to builder? Maybe something like this:
```js
clearAudioInputDevice(){
    const active = this.activeDevices['audio'];
    if (active) {
      this.releaseActiveDevice(active);
      delete this.activeDevices['audio'];
    }
}
```

In my opinion the effect of `chooseVideoInputDevice(null)` and `chooseAudioInputDevice(null)` should be similar. Current API is quite confusing. We might want to improve this in the future major version.

src/devicecontroller/DefaultDeviceController.ts
ltrung: So let's say if I have multiple `DefaultDeviceController` each of them call `chooseAudioInputDevice(null)`. What is the size of the buffer? Will it increase for each source node (which we do not clean up)?

xuesichao: No. I put the result in the description.

simmkyu: > So let's say if I have multiple `DefaultDeviceController` each of them call `chooseAudioInputDevice(null)`. What is the size of the buffer? Will it increase for each source node (which we do not clean up)?


If you make multiple `chooseAudioInputDevice(null)` calls, shouldn't they use a single `DefaultDeviceController.defaultAudioBuffer` static object?

src/devicecontroller/DefaultDeviceController.ts
simmkyu: If we do not call `await this.chooseAudioInputDevice(null)` to send a fake audio stream, is it possible that the Chime backend considers the attendee dropped, e.g., in Safari?

src/devicecontroller/DefaultDeviceController.ts
xuesichao: >If you make multiple chooseAudioInputDevice(null) calls, shouldn't they use a single DefaultDeviceController.defaultAudioBuffer static object?

Yes, they use the single `DefaultDeviceController.defaultAudioBuffer` and thus the memory increases only the first time we call `chooseAudioInputDevice(null)` where the `DefaultDeviceController.defaultAudioBuffer` is initialized.

src/devicecontroller/DefaultDeviceController.ts
xuesichao: > If we do not call `await this.chooseAudioInputDevice(null)` to send a fake audio stream, is it possible that the Chime backend considers the attendee dropped, e.g., in Safari?

Good point, I never think of this case. Need to do some testing and confirm it.

If this is the reason why there's a difference between `chooseVideoInputDevice(null)` and `chooseAudioInputDevice(null)`, it makes a lot sense.

xuesichao:
In `DefaultDeviceController` there are two kinds of memory leak issue:
* Major: `AudioBuffer` leaks, 0.96MB data accumulates in memory per `synthesizeAudioDevice(0)` call
* Minor: `GainNode`, `MediaStreamAudioDestinationNode`, `OscillatorNode` and `AudioBufferSourceNode` leak. No more than 1KB accumulates in memory per `synthesizeAudioDevice()` call.

This PR only fix the major device memory leak issue. Discussed with team that we don't want to fix this in V2 as it's not critical. We want to fix Major and Minor together in V3.

So close this PR for now.

PR #2072, opened 2022-02-28
Video orientation change issue

jhaarpan:
**Issue #:**

**Description of changes:**
- Added an event listener to detect change of orientation and change the target canvas size when Background Blur/ Background Replacement is enabled.

**Testing:**
- Ran npm run build:release locally
- Tested the change manually by changing orientation on Android Chrome

- Deploy the serverless demo of amazon-chime-sdk-js v2.25.0
- Start a meeting from PC
- Join the meeting from an Android device
- Set BackgroundBlur to video filter on the Android device
- Change the orientation of the Android device from portrait to landscape
- The height and width of the video should change on the basis of change of orientation

**Checklist:**


src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
```diff
@@ -47,6 +47,20 @@ export default class BackgroundBlurProcessorProvided
     this.logger.info('BackgroundBlur processor successfully created');
     this.logger.info(`BackgroundBlur spec: ${this.stringify(this.spec)}`);
     this.logger.info(`BackgroundBlur options: ${this.stringify(options)}`);
+
+    screen.orientation.addEventListener('change', _event => {
```
hensmi-amazon, line 51: Can this be in `BackgroundFilterProcessor`?

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
```diff
@@ -47,6 +47,20 @@ export default class BackgroundBlurProcessorProvided
...
+      if (
```
hensmi-amazon, line 52: Does this event listener need to be removed?

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
```diff
@@ -47,6 +47,20 @@ export default class BackgroundBlurProcessorProvided
...
+        (screen.orientation.type.startsWith('portrait') &&
+          this.canvasCtx.canvas.height < this.canvasCtx.canvas.width) ||
```
hensmi-amazon, line 54: Can you put a comment that explains what this is doing in plain english? It's a bit tricky at first glance even if its doing something simple.

jhaarpan: Sure, I can add a comment

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
jhaarpan: Yes, I will do that

src/backgroundblurprocessor/BackgroundBlurProcessorProvided.ts
hensmi-amazon: What was the issue?

jhaarpan: It was not changing the orientation

hensmi-amazon: Is the listener not called? I don't see how it wouldn't work. At the very least can you put the listener function in shared code?

jhaarpan: The listener is called, but sometimes there is a delay which causes the canvas context to not change and hence causing the video to be garbled

src/backgroundfilter/BackgroundFilterProcessor.ts
```diff
@@ -404,6 +404,7 @@ export default abstract class BackgroundFilterProcessor {
     this.scaledCanvas?.remove();
     this.scaledCanvas = undefined;
     this.logger.info(`${this.filterType} frame process destroyed`);
+    screen.orientation.onchange = null;
```
hensmi-amazon, line 407: You need to call `removeEventListener`

ltrung (approved):

PR #2077, opened 2022-03-01
Fix backward compatibility check

devalevenkatesh:
**Issue #:**

**Description of changes:**

**Testing:**

**Checklist:**

### DIFF
--- a/.github/workflows/release-backwards-compatiblity.yml
+++ b/.github/workflows/release-backwards-compatiblity.yml
@@ -7,8 +7,9 @@
 name: Backwards Compatibility Test
 
 on:
-  push:
+  pull:
     branches:
+      - 'main'
       - 'release-**'
       - '!release-**.x'
 
@@ -21,10 +22,27 @@ env:
   MESSAGING_USER_ARN: ${{secrets.MESSAGING_USER_ARN}}
 
 jobs:
+  check-backward-compatibility-run:
+    name: Check if backward compatibility check is needed for the version we want to release
+    runs-on: ubuntu-latest
+    outputs:
+      is-beta: ${{ steps.check-if-beta-version.outputs.is-beta }}
+    steps:
+      - name: Check if beta version
+        id: check-if-beta-version
+        run: |
+          current_version=$(.github/script/get-current-version)
+          if [[ "$current_version" == *"beta"* ]]; then
+            echo "::set-output name=is-beta::yes"
+          else
+            echo "::set-output name=is-beta::no"
+          if
+
   release-integ-test:
     name: Run Backwards Compatible Integration Tests with Release Tarball
     runs-on: ubuntu-latest
-
+    needs: check-backward-compatibility-run
+    if: ${{ needs.check-backward-compatibility-run.outputs.is-beta == 'yes' }}
     steps:
       - name: Setup Node.js - 16.x
         uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/release-backwards-compatiblity.yml
```diff
@@ -21,10 +22,33 @@ env:
...
+      is-beta: ${{ steps.check-if-beta-version.outputs.is-beta }}
+    steps:
+      - name: Checkout Package
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Check if beta version
+        id: check-if-beta-version
+        run: |
+          current_version=$(.github/script/get-current-version)
```
ltrung, line 38: I dont think this is entirely correct. We dont want to run this for the first official version either (e.g., v3.0.0 as no prev major exist).

.github/workflows/release-backwards-compatiblity.yml
```diff
@@ -21,10 +22,33 @@ env:
...
+            echo "::set-output name=is-beta::no"
+          fi
+          
+
   release-integ-test:
     name: Run Backwards Compatible Integration Tests with Release Tarball
     runs-on: ubuntu-latest
-
+    needs: check-backward-compatibility-run
+    if: ${{ needs.check-backward-compatibility-run.outputs.is-beta == 'no' }}
```
ltrung, line 51: Can we create  a script to get prev_version (in node) and if that script returns no then quit.

ltrung (approved):

.github/workflows/release-backwards-compatiblity.yml
devalevenkatesh: Yes I have not added any checks just checking with what was existing, was refreshing some GitHub actions stuff doing it. But yes, 3.0.0 or any first major version needs to be skipped.

.github/workflows/release-backwards-compatiblity.yml
devalevenkatesh: Yes, saw the slack message was just testing. will do :)

.github/workflows/release-backwards-compatiblity.yml
devalevenkatesh: Fixed now.

.github/workflows/release-backwards-compatiblity.yml
devalevenkatesh: Created a node script and just call that in GA.

.github/script/backward-compatibility-run-check.js
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * This script simply checks whether the version we are releasing is either the first NPM version for any new major version or
+ * a beta version. For both, we do not want to run backward compatibility checks since both will have backward in-compatible breaking changes.
+ * 
+ * For e.g: 3.0.0, 3.0.0-beta.0, 4.0.0 and so on.
+ */
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const versionToRelease = require('../../package.json').version;
+if (versionToRelease.includes('beta') || (/^[0-9]+\.0\.0$/g).test(versionToRelease)) {
```
ltrung, line 12: Actually can we call `get-pre-release-name` script to get the pre-release name ? Try to avoid hard code beta in case we change it later.

script/backward-compatibility-demo-setup.js
```diff
@@ -0,0 +1,36 @@
...
+
+logger.log('Building package...');
+spawnOrFail('npm', ['run build']);
+logger.log('Building done!');
+
+logger.log('Packaging...');
+spawnOrFail('npm', ['pack']);
+logger.log('Packaging done!');
+
+const previousRelease = (spawnOrFail('npm', ['view amazon-chime-sdk-js@latest version'], { skipOutput: true })).trim();
```
ltrung, line 23: Can we just use `get-prev-version` script here?

.github/script/backward-compatibility-run-check.js
```diff
@@ -0,0 +1,16 @@
...
+  console.log('no');
```
ltrung, line 13: Can we print out something more meaningful and return a value instead of console.log?

nainkunal933: +1

.github/script/backward-compatibility-run-check.js
nainkunal933, line 12: Can you please add a comment to explain the regex part of the if block? Just add a small comment on what the check is for. Regex is always hard to read on its face.
```(/^[0-9]+\.0\.0$/g)```

Comment (devalevenkatesh) [reply .github/script/backward-compatibility-run-check.js]: Fixed it, that makes sense!

File: .github/script/backward-compatibility-run-check.js (hunk 5 above)
Comment (devalevenkatesh) [reply .github/script/backward-compatibility-run-check.js]: Do you guys have an example to return a value and catch it in GitHub action run? I see all scripts to be logging to console and accessing through a bash call in GA hence checking. I am re-searching more around this.

Comment (devalevenkatesh) [reply .github/script/backward-compatibility-run-check.js]: I think I have to use `process.exit` on the node script and then based on the job success failure run the actual checks.

File: .github/script/backward-compatibility-run-check.js (hunk 3 above)
Comment (devalevenkatesh) [reply .github/script/backward-compatibility-run-check.js]: Yes, I had added an explanation at the top of the script but now I also added a comment for the regex 👍

File: .github/workflows/release-backwards-compatiblity.yml [hunk 6]
```diff
@@ -21,10 +21,22 @@ env:
...
+          fetch-depth: 0
+      - name: Version check
+        run: node .github/script/backward-compatibility-run-check.js
+
   release-integ-test:
     name: Run Backwards Compatible Integration Tests with Release Tarball
     runs-on: ubuntu-latest
-
+    needs: check-backward-compatibility-run
+    if: ${{ success() }}
```
ltrung, line 39: Sorry not sure we want this as it will show the workflow as fail right (as the previous job fail? )

ltrung (approved):

.github/script/backward-compatibility-run-check.js
devalevenkatesh: So I discussed with Trung and could not find a way to return a boolean and catch it in GitHub actions, hence, reverted to `console.log` as we have in other scripts. I extended the log though from "yes" or "no" to a detailed log and check it using `contains` in GitHub action `if` conditional.

.github/workflows/release-backwards-compatiblity.yml
devalevenkatesh: Discussed offline, but did this: https://github.com/aws/amazon-chime-sdk-js/pull/2077#discussion_r817968614

PR #2102, opened 2022-03-11
Update publish github workflow to remove useless if check

nainkunal933:
**Issue #:**
Publish workflow failed due to buggy if check: https://github.com/aws/amazon-chime-sdk-js/actions/runs/1966093922
**Description of changes:**
Removing the if check as it is not needed.
**Testing:**

NA

**Checklist:**

NA

NA

NA

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -96,13 +96,5 @@ jobs:
         run: |
           demo_current_version=$(.github/script/get-demo-browser-current-version)
           echo "Current demo version: " $demo_current_version
-          echo "Current released npm version:" ${{ steps.npm_version.outputs.npm_version }}
-          if [[ *$demo_current_version* != "^#${{ steps.npm_version.outputs.npm_version }}" ]];
-          then
-            echo "Demo npm version is not the same as the current released version"
-            exit 1
-          else
-            npm run build --app=meetingReadinessChecker
-            script/github-action-awscli-installation
-            node ./.github/script/call-canary-deploy-demo-prod $demo_current_version
-          fi
+          script/github-action-awscli-installation
+          node ./.github/script/call-canary-deploy-demo-prod $demo_current_version


.github/workflows/publish.yml
```diff
@@ -96,13 +96,6 @@ jobs:
...
-          then
-            echo "Demo npm version is not the same as the current released version"
-            exit 1
-          else
-            npm run build --app=meetingReadinessChecker
-            script/github-action-awscli-installation
-            node ./.github/script/call-canary-deploy-demo-prod $demo_current_version
-          fi
\ No newline at end of file
+          npm run build --app=meetingReadinessChecker
```
ltrung, line 99: This is not needed as deploy script always rebuild the app now.

.github/workflows/publish.yml
```diff
@@ -96,13 +96,6 @@ jobs:
...
+          script/github-action-awscli-installation
```
ltrung, line 100: Where do we set this https://github.com/aws/amazon-chime-sdk-js/blob/main/.github/workflows/deploy.yml#L17? deploy script needs $NAME to be set and I don't see one in publish.yml or here.

.github/workflows/publish.yml
nainkunal933: Correct. Removing this.

.github/workflows/publish.yml
nainkunal933: It is being set right here: https://github.com/aws/amazon-chime-sdk-js/blob/main/.github/workflows/publish.yml#L45-L46

ltrung: Did you test it to make sure the node child process get access to that NAME variable?

nainkunal933: There is a node process getting started here but ultimately the node script is calling `deploy-canary-demo` bash script which is executing in the same environment. So, the NAME variable should be there. 

I ran the script as part of my testing but did not verify the NAME variable availability explicitly. I will rerun a test now.

nainkunal933: Test: https://github.com/aws/amazon-chime-sdk-js/actions/runs/1971286943

ltrung (approved):

PR #2104, opened 2022-03-11
[Release-2.x] Update publish github workflow to remove useless if check (#2102)

nainkunal933:
**Issue #:**

**Description of changes:**
Duplicate of https://github.com/aws/amazon-chime-sdk-js/pull/2102
**Testing:**

**Checklist:**

NA

NA

NA

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -96,13 +96,5 @@ jobs:
         run: |
           demo_current_version=$(.github/script/get-demo-browser-current-version)
           echo "Current demo version: " $demo_current_version
-          echo "Current released npm version:" ${{ steps.npm_version.outputs.npm_version }}
-          if [[ *$demo_current_version* != "^#${{ steps.npm_version.outputs.npm_version }}" ]];
-          then
-            echo "Demo npm version is not the same as the current released version"
-            exit 1
-          else
-            npm run build --app=meetingReadinessChecker
-            script/github-action-awscli-installation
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -96,13 +96,5 @@ jobs:
         run: |
           demo_current_version=$(.github/script/get-demo-browser-current-version)
           echo "Current demo version: " $demo_current_version
-          echo "Current released npm version:" ${{ steps.npm_version.outputs.npm_version }}
-          if [[ *$demo_current_version* != "^#${{ steps.npm_version.outputs.npm_version }}" ]];
-          then
-            echo "Demo npm version is not the same as the current released version"
-            exit 1
-          else
-            npm run build --app=meetingReadinessChecker
```
michhyun1, line 105: we dont need this?

xuesichao (approved):

.github/workflows/publish.yml
nainkunal933: No. After looking at the workflow it turns out that this if statement will never be true.

nainkunal933: This if statement's string comparison was the reason why the script was failing.

xuesichao: Why we need to remove this:
>npm run build --app=meetingReadinessChecker

ltrung: The deploy script already build it

PR #2110, opened 2022-03-15
Added IOS audio tests

LeviSklaroff:
**Issue #:**
IOS audio tests fail from a selenium script timeout error. This is because currently for IOS Safari the script timeout is by default undefined which results in our asynchronous script being timeout instantly.
**Description of changes:**
Override the default script timeout for selenium which fixes IOS audio tests
**Testing:**

**Checklist:**

### DIFF
--- a/integration/js/utils/WebdriverSauceLabs.js
+++ b/integration/js/utils/WebdriverSauceLabs.js
@@ -203,6 +203,15 @@ class SaucelabsSession {
       .withCapabilities(cap)
       .forBrowser(capabilities.browserName)
       .build();
+
+    // Selenium is supposed to have a default timeout of 30 seconds, but on IOS Safari
+    // it is undefined. This will override the timeout in this case or any similar cases.
+    const defaultTimeout = 30000;
+    let timeouts = await driver.manage().getTimeouts()
+    if (timeouts.script == undefined || timeouts.script < defaultTimeout) {
+      await driver.manage().setTimeouts({script: defaultTimeout})
+    }
+    
     return new SaucelabsSession(driver, domain, appName);
   }
   constructor(inDriver, domain, appName) {


ltrung:
Can we provide more context of the changes (is there a bug report link somewhere, or what the issue is) ?

integration/js/utils/WebdriverSauceLabs.js
```diff
@@ -203,6 +203,9 @@ class SaucelabsSession {
       .withCapabilities(cap)
       .forBrowser(capabilities.browserName)
       .build();
+    const timeout = 600000;
```
ltrung, line 206: Should we only do it for iOS?

LeviSklaroff: We could, but I think it would be better to set it for everything. Selenium has a default timeout for all browsers, which could in theory change at some point, so it would be better to just make it uniform across the board

LeviSklaroff: Ok looked a bit more into it, and the default script timeout is supposed to be 30 seconds across the board, previously this was not the case. For IOS Safari it's undefined for some reason. I couldn't find a bug report for it, but I changed the code so now it will only override the timeout if it's undefined or less then 30 seconds.

integration/js/utils/WebdriverSauceLabs.js
```diff
@@ -203,6 +203,11 @@ class SaucelabsSession {
       .withCapabilities(cap)
       .forBrowser(capabilities.browserName)
       .build();
+    const timeout = 30000;
+    console.log();
+    let timeouts = await driver.manage().getTimeouts()
+    if(timeouts.script == undefined || timeouts.script < timeout)
```
ltrung, line 209: nit: You probably don't need `    const timeout = 30000;` and can just use the value in the if statement directly.

Also there should be a space between `if` and `(` and remove the empty `console.log`

LeviSklaroff: Updated, not sure how I ended up with that empty `console.log ` lol

ltrung (approved):

PR #2135, opened 2022-03-29
Remove MeetingSessionConfiguration dependency from MeetingSessionPOSTLogger

devalevenkatesh:
**Issue #:**

**Description of changes:**
- Renames `MeetingSessionPOSTLogger` to `POSTLogger`.
- Removed `MeetingSessionConfiguration` dependency from `MeetingSessionPOSTLogger`.
- Add `options` as optional last param in constructor to provide optional
`headers` and `metadata` for HTTP POST requests.
- Renamed `startLogPublishScheduler` to just `start` method.
- Updated the demo according to changes in JS SDK.
- Update migration doc for v3.
- Add / update existing unit testing for this change.
- Make `start` and `stop` private on `POSTLogger`.
- Update unit tests to use `destroy` and remove any private member accessing test earlier ignored using `ts-ignore`.

**Testing:**

Unit Tests
```
1982 passing (3m)
Statements   : 100% ( 9836/9836 )
Branches     : 100% ( 4290/4290 )
Functions    : 100% ( 1875/1875 )
Lines        : 100% ( 9724/9724 )
==============================================================================
```

Meeting browser demo

- Deployed a serverless browser meeting demo and checked:
  - Meeting events reporting.
  - Meeting event ingestion metrics reporting.
  - JS SDK browser logs posted successfully to AWS CloudWatch setup in demo like earlier.

Did same for the meeting readiness checker. 

Testing Steps:
1. Load up the serverless demo.
2. Open browser developers and click on the "Network" tab.
3. Join a deployed serverless meeting browser demo.
4. Notice network calls to `logs`, `log_event_ingestion` and `log_meeting_event` resulting in 200 response status code.

**Checklist:**

Yes

Yes

NA


guides/17_Migration_to_3_0.md
```diff
@@ -60,22 +71,98 @@ change it to
...
+With Amazon Chime SDK for JavaScript v3, we have removed `MeetingSessionConfiguration` dependency from the `MeetingSessionPOSTLogger`. The `MeetingSessionPOSTLogger` is now renamed to `POSTLogger`.
+
+Builders can add optional fields to `POSTLogger` which will be included when `POSTLogger` sends the HTTP POST logs to builder provided URL:
+
+1. HTTP POST request `headers`.
+2. Informational `metadata`. For example, `meetingId`, `attendeeId`, `appName`. This `metadata` depends on builders and `POSTLogger` will include this `metadata` as part of the request body when making the HTTP POST requests to builder provided URL.
+
+#### Constructor
+
+Before in 2.x:
```
ltrung, line 85: Kyu helped me update mine in the device controller PR but he grouped the before and after in one code block and only use // Before and // After so you can share the same common code such as meetingSessionConfiguration

guides/17_Migration_to_3_0.md
```diff
@@ -60,22 +71,98 @@ change it to
...
+  2000, // LOGGER_INTERVAL_MS
+  URL,
+  LogLevel.INFO
+);
+
+// You need responses from server-side Chime API. See below for details.
+const meetingResponse = // The response from the CreateMeeting API action.
+const attendeeResponse = // The response from the CreateAttendee API action.
+
+const metadata = {
```
ltrung, line 140: Seem like this could be in the same example as above.

src/logger/POSTLogger.ts
```diff
@@ -53,6 +62,14 @@ export default class MeetingSessionPOSTLogger implements Destroyable {
     window.removeEventListener('unload', this.eventListener);
   }
 
+  get metadata(): Record<string, string> | undefined {
```
ltrung, line 65: Do we need to do this or can we just set metadata to be a public prop?

src/logger/POSTLogger.ts
```diff
@@ -91,7 +108,7 @@ export default class MeetingSessionPOSTLogger implements Destroyable {
     return this.logCapture.length;
   }
 
-  startLogPublishScheduler(batchSize: number): void {
+  start(): void {
```
ltrung, line 111: Should this be private?

test/logger/POSTLogger.test.ts
```diff
@@ -102,31 +89,21 @@ describe('MeetingSessionPOSTLogger', () => {
...
-        BASE_URL,
-        LogLevel.INFO
-      );
+      const logger = new POSTLogger(batchSize, intervalMs, BASE_URL, LogLevel.INFO);
       logger.info('info');
       logger.error('error');
+      logger.debug('error');
       expect(logger.getLogCaptureSize()).is.equal(2);
       logger.stop();
       done();
```
ltrung, line 98: Do we need done for this test?

src/logger/POSTLogger.ts
```diff
@@ -100,7 +117,7 @@ export default class MeetingSessionPOSTLogger implements Destroyable {
         return;
       }
       this.lock = true;
-      const batch = this.logCapture.slice(0, batchSize);
+      const batch = this.logCapture.slice(0, this.batchSize);
       const body = this.makeRequestBody(batch);
       try {
         const response = await fetch(this.url, {
```
ltrung, line 123: Where do we use meta_data here?

src/logger/POSTLogger.ts
devalevenkatesh: It seems this class was implemented to help builders start stop the logger. I have kept the internal working same as is so dint touch these existing APIs. I removed the `batchSize` as it is already a class member variable which `start` can implicitly use.

Do you still want to make them private?

src/logger/POSTLogger.ts
devalevenkatesh: I kept it as getter setter so that if in future we want to validate on size or properties, a set method would be a good way to control it. In case of public, we would have no such control. Hence, I think it would be good to use setter and getter here.

src/logger/POSTLogger.ts
ltrung: Just nit comment so up to you but does not seem like builders want to call start since it is already called in constructor.

src/logger/POSTLogger.ts
ltrung: Don't think change it to get/set from public is a breaking change right? I prefer to do it whenever we need to do it :)

src/logger/POSTLogger.ts
devalevenkatesh: But I agree that for the POSTLogger both `start` and `stop` can be private and if builders want to stop logging, then they have to just destroy the logger can create a new one.

guides/17_Migration_to_3_0.md
devalevenkatesh: It just shows adding metadata after the object creation. First example, shows it as part of the constructor.

test/logger/POSTLogger.test.ts
```diff
@@ -0,0 +1,279 @@
...
+  });
+
+  describe('logging level', () => {
+    it('should log info with LogLevel.INFO and ignore the debug', async () => {
+      const logger = new POSTLogger(batchSize, intervalMs, BASE_URL, LogLevel.INFO);
+      logger.info('info');
+      logger.error('error');
+      logger.debug('error');
+      expect(logger.getLogCaptureSize()).is.equal(2);
+      await wait(1000);
```
ltrung, line 91: Any reason it has to be that big? We should try to do min wait time if possible.

guides/17_Migration_to_3_0.md
ltrung: Then could we add one sentence explain this use case?

guides/17_Migration_to_3_0.md
```diff
@@ -159,13 +159,86 @@ handle them please remove.
...
+#### Constructor
+
+```js
+
+// You need responses from server-side Chime API. See below for details.
+const meetingResponse = // The response from the CreateMeeting API action.
+const attendeeResponse = // The response from the CreateAttendee API action.
+
+// Before in 2.x
+const meetingSessionConfiguration = new MeetingSessionConfiguration(meetingResponse, attendeeResponse);
```
simmkyu, line 180: nit: Syntax highlighting is broken... Too 

image

guides/17_Migration_to_3_0.md
devalevenkatesh: Yup makes sense, so updated my docs with this pattern.

guides/17_Migration_to_3_0.md
devalevenkatesh: I added a brief explanation when that can be a use case, good point!

src/logger/POSTLogger.ts
devalevenkatesh: Changed metadata to a public property now.

src/logger/POSTLogger.ts
```diff
@@ -2,35 +2,44 @@
...
 
   constructor(
-    private name: string,
-    private configuration: MeetingSessionConfiguration,
     private batchSize: number,
     private intervalMs: number,
     private url: string,
     private level = LogLevel.WARN,
-    private headers?: Record<string, string>
+    options?: {
```
simmkyu, line 31: For consistency, we have two options.

1. `batchSize`, `intervalMs`, `url`, and `level` can also be under `options`. (Especially, `level` is also an optional field.)

    For example, we had to specify `LOGGER_BATCH_SIZE` and `LOGGER_INTERVAL_MS` to indicate 1st and 2nd parameters. With options, we won't need these comments.

    ```js
    const logger = new POSTLogger(
      20, // LOGGER_BATCH_SIZE
      2000, // LOGGER_INTERVAL_MS
      URL,
      LogLevel.INFO,
      postLoggerOptions
    );
    ```
2. Or you can make `metadata` and `headers` as individual parameters. JS SDK uses this pattern more.

test/logger/POSTLogger.test.ts
devalevenkatesh: so I removed the `intervalScheduler` and used `wait` and not using `done` pattern anymore.

test/logger/POSTLogger.test.ts
devalevenkatesh: I have removed the wait from most places where sending is not actually checked, and updated other places based on the logs and time. Thanks and this is good point to change!

guides/17_Migration_to_3_0.md
```diff
@@ -159,13 +159,102 @@ handle them please remove.
...
+  'SDK',
+  configuration,
+  20, // LOGGER_BATCH_SIZE
+  2000, // LOGGER_INTERVAL_MS
+  URL,
+  LogLevel.INFO
+);
+
+// After in 3.x
+const options: POSTLoggerOptions = {
```
simmkyu, line 198: Remove `: POSTLoggerOptions` since snippets are for JavaScript

guides/17_Migration_to_3_0.md
```diff
@@ -159,13 +159,102 @@ handle them please remove.
...
+// After this POSTLogger will include the metadata in HTTP POST request body when sending logs.
+logger.metadata = metadata;
+```
+
+### Set `headers` when creating `POSTLogger`
+
+You may want to add custom authentication `headers` with each HTTP POST request. You can provide such `headers` when constructing the `POSTLogger`.
+
+```typescript
+const options: POSTLoggerOptions = {
```
simmkyu, line 241: Delete `: POSTLoggerOptions`

guides/17_Migration_to_3_0.md
```diff
@@ -159,13 +159,102 @@ handle them please remove.
...
+  url: 'URL TO POST LOGS',
+  headers: {
+    Chime-Bearer: 'authentication-token',
```
simmkyu, line 244: `Chime-Bearer` to `'Chime-Bearer'` (can't use `-` in an object key)

guides/17_Migration_to_3_0.md
```diff
@@ -159,13 +159,102 @@ handle them please remove.
...
+```
+
+### Set `metadata` after `POSTLogger` creation
+
+This helps with Web application development using React frameworks or other UI frameworks where the component getting the server responses comes later based on an HTML form submission and not on initial load of the Web application.
+
+You may want to add logging in Providers, Hooks (if using React) in your application. Since, we have removed the `MeetingSessionConfiguration` dependency now, you can create `POSTLogger` earlier in the application flow and you can add `metadata` once you receive API responses from server upon a HTML form submission. This way, you can update `POSTLogger`'s `metadata` as per your application needs.
+
+```typescript
+const options: POSTLoggerOptions = {
```
simmkyu, line 217: Delete `: POSTLoggerOptions` and `: Record<string, string>`

guides/17_Migration_to_3_0.md
```diff
@@ -159,13 +159,102 @@ handle them please remove.
...
+  metadata: {
+    appName: 'SDK',
+    meetingId: meetingResponse.Meeting.MeetingId,
+    attendeeId: attendeeResponse.Attendee.AttendeeId,
+  },
+};
+const logger = new POSTLogger(options);
+```
+
+### Set `metadata` after `POSTLogger` creation
```
simmkyu, line 210: How about just adding this code to the example above?
- Setting `metadata` will also help client applications with no UI frameworks.
- I think the provider and hook instruction can be into a Chime SDK React Component Library doc.

```js
// You can also set new metadata after initializing POSTLogger.
// For example, you can set metadata after receiving API responses from your server application.
logger.metadata = {
  appName: 'Your app name',
  meetingId: meetingResponse.Meeting.MeetingId,
  attendeeId: attendeeResponse.Attendee.AttendeeId,
};
```

src/logger/POSTLoggerOptions.ts
```diff
@@ -0,0 +1,38 @@
...
+   */
+  headers?: Record<string, string>;
+
+  /**
+   * Log sending interval. Default value is 2000ms.
+   */
+  intervalMs?: number;
+
+  /**
+   * Level of logging. Check [[LogLevel]] for more information.
```
simmkyu, line 29: Specify the default value. `LogLevel.WARN` as in other comments

guides/17_Migration_to_3_0.md
```diff
@@ -159,13 +159,102 @@ handle them please remove.
 - StateMachineTransitionFailed
 - AudioDeviceSwitched
 
+## `MeetingSessionPOSTLogger`
+
+We have renamed `MeetingSessionPOSTLogger` to `POSTLogger` and removed the `MeetingSessionConfiguration` dependency. You don't need to pass the `MeetingSessionConfiguration` object to the `POSTLogger` constructor anymore.
+
+You can pass below `POSTLoggerOptions` to the `POSTLogger` constructor:
+```typescript
+interface POSTLoggerOptions {
```
simmkyu, line 168: In README.md and this migration guide, we provide JavaScript snippets. JS users won't know `Record` or `interface`.

simmkyu (commented):
The code looks good to me!

Instead of three sub-sections, you can put them into a single section to keep the migration guide shorter. Also, `POSTLoggerOptions` doc has complete information. Resolved conversations had detailed feedback.

---

## `MeetingSessionPOSTLogger` to `POSTLogger`

We have renamed `MeetingSessionPOSTLogger` to `POSTLogger` and removed the `MeetingSessionConfiguration` dependency. You don't need to pass the `MeetingSessionConfiguration` object to the `POSTLogger` constructor anymore.

```js
// You need responses from server-side Chime API. See below for details.
const meetingResponse = // The response from the CreateMeeting API action.
const attendeeResponse = // The response from the CreateAttendee API action.

// Before
const meetingSessionConfiguration = new MeetingSessionConfiguration(meetingResponse, attendeeResponse);
const meetingSessionPOSTLogger = new MeetingSessionPOSTLogger(
  'SDK',
  configuration,
  20, // LOGGER_BATCH_SIZE
  2000, // LOGGER_INTERVAL_MS
  URL,
  LogLevel.INFO
);

// After
const logger = new POSTLogger({
  url: 'URL TO POST LOGS',
});
```

You can create a `POSTLogger` object with `headers`, `logLevel`, `metadata`, and other options. See the `POSTLoggerOptions` documentation for more information.

```js
const logger = new POSTLogger({
  url: 'URL TO POST LOGS',
  // Add "headers" to each HTTP POST request
  headers: {
    'Chime-Bearer': 'authentication-token'
  },
  logLevel: LogLevel.INFO,
  metadata: {
    appName: 'Your app name',
    meetingId: meetingResponse.Meeting.MeetingId,
    attendeeId: attendeeResponse.Attendee.AttendeeId,
  },
});

// You can also set new metadata after initializing POSTLogger.
// For example, you can set metadata after receiving API responses from your server application.
logger.metadata = {
  appName: 'Your app name',
  meetingId: meetingResponse.Meeting.MeetingId,
  attendeeId: attendeeResponse.Attendee.AttendeeId,
};
```

src/logger/POSTLoggerOptions.ts
```diff
@@ -0,0 +1,38 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import LogLevel from './LogLevel';
+
+export default interface POSTLoggerOptions {
+  /**
+   * URL to send logs.
+   */
+  url: string;
```
simmkyu, line 10: no action required:
We can also take a `Request` object or `URL`, but we can start with `string` only and add more later without breaking changes.

guides/17_Migration_to_3_0.md
devalevenkatesh: Thanks for the catch! Removed it.

src/logger/POSTLoggerOptions.ts
devalevenkatesh: Sure thanks for the suggestion will track internally.

guides/17_Migration_to_3_0.md
devalevenkatesh: Ok I got it, thought we can use typescript as well. Thanks for the pointer!

src/logger/POSTLoggerOptions.ts
devalevenkatesh: Updated it thanks for the catch!

guides/17_Migration_to_3_0.md
devalevenkatesh: Thanks for the concise write-up have replaced existing with your provided write-up.

guides/17_Migration_to_3_0.md
devalevenkatesh: Great catch! I tested this with meetingReadinessChecker and then realized but forgot to update here. Have updated now.

guides/17_Migration_to_3_0.md
```diff
@@ -159,13 +159,86 @@ handle them please remove.
 - StateMachineTransitionFailed
 - AudioDeviceSwitched
 
+### `MeetingSessionPOSTLogger` to `POSTLogger`
```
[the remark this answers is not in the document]
devalevenkatesh: I think this was missed due to rebase, was following earlier format in the doc. I have replaced your provided documentation now and fixed the markdown.

src/logger/POSTLogger.ts
devalevenkatesh: Used the first approach since we have many properties some optional and some with default value and only URL as required. Thanks for the suggestion!

simmkyu (approved):

PR #2144, opened 2022-04-01
Fix join with no audio when WebAudio is enabled

ltrung:
**Description of changes:**
- Fix a bug where joining without selecting any audio device failed when Web Audio is on by making sure there is an inner device exists before returning.
- Fix a typo in `DefaultSignalingClient` that checks for the wrong method.
- Fix a minor log info for video input ended event, we say `resetting to null device` when we just stop the video input.
- Fix a minor demo bug where selecting a video input from the dropdown will auto start the video input stream even when the video button is not enabled yet. To enable it you will then have to click the video button which will then call `startVideoInput` again which will throw an error if the previous `startVideoInput` from selecting a video input is not completed yet.

**Testing:**

- In Chrome, enable Web Audio and `Join with view-only mode`.
- Verify that you can join successfully and show up in roster and view other videos.

**Checklist:**

### DIFF
--- a/guides/17_Migration_to_3_0.md
+++ b/guides/17_Migration_to_3_0.md
@@ -38,10 +38,10 @@ In v3, you should call `stopAudioInput` to stop sending an audio stream when you
 const observer = {
   audioVideoDidStop: async sessionStatus => {
     // v3
-    await meetingSession.audioVideo.stopVideoInput();
+    await meetingSession.audioVideo.stopAudioInput();
 
     // Or use the destroy API to call stopAudioInput and stopVideoInput.
-    meetingSession.audioVideo.deviceController.destroy()
+    meetingSession.deviceController.destroy();
   },
 };
 meetingSession.audioVideo.addObserver(observer);
@@ -64,7 +64,10 @@ await meetingSession.audioVideo.startVideoInput(videoInputDeviceInfo.deviceId);
 In v3, you should call `stopVideoInput` to stop the video input stream.
 
 ```js
-// v3
+// Before
+await meetingSession.audioVideo.chooseVideoInputDevice(null);
+
+// After
 await meetingSession.audioVideo.stopVideoInput();
 ```
 
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -565,12 +565,27 @@ export default class DefaultDeviceController
     return this.videoInputQualitySettings;
   }
 
-  acquireAudioInputStream(): Promise<MediaStream> {
-    return this.acquireInputStream('audio');
+  async acquireAudioInputStream(): Promise<MediaStream> {
+    if (!this.activeDevices['audio']) {
+      this.logger.info(`No audio device chosen, creating empty audio device`);
+      await this.startAudioInput(null);
+    }
+
+    if (this.useWebAudio) {
+      const dest = this.getMediaStreamDestinationNode();
+      return dest.stream;
+    }
+    return this.activeDevices['audio'].stream;
   }
 
-  acquireVideoInputStream(): Promise<MediaStream> {
-    return this.acquireInputStream('video');
+  async acquireVideoInputStream(): Promise<MediaStream> {
+    if (!this.activeDevices['video']) {
+      throw new Error(`No video device chosen`);
+    }
+    if (this.chosenVideoInputIsTransformDevice()) {
+      return this.chosenVideoTransformDevice.outputMediaStream;
+    }
+    return this.activeDevices['video'].stream;
   }
 
   async acquireDisplayInputStream(
@@ -876,8 +891,14 @@ export default class DefaultDeviceController
   private async handleDeviceStreamEnded(kind: 'audio' | 'video', deviceId: string): Promise<void> {
     try {
       if (kind === 'audio') {
+        this.logger.warn(
+          `Audio input device which was active is no longer available, resetting to null device`
+        );
         await this.startAudioInput(null); //Need to switch to empty audio device
       } else {
+        this.logger.warn(
+          `Video input device which was active is no longer available, stopping video`
+        );
         await this.stopVideoInput();
       }
     } catch (e) {
@@ -1191,9 +1212,6 @@ export default class DefaultDeviceController
         // Hard to test, but the safety check is worthwhile.
         /* istanbul ignore else */
         if (this.activeDevices[kind] && this.activeDevices[kind].stream === newDevice.stream) {
-          this.logger.warn(
-            `${kind} input device which was active is no longer available, resetting to null device`
-          );
           this.handleDeviceStreamEnded(kind, newDeviceId);
           delete newDevice.endedCallback;
         }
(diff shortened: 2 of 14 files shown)


guides/17_Migration_to_3_0.md
```diff
@@ -64,7 +64,10 @@ await meetingSession.audioVideo.startVideoInput(videoInputDeviceInfo.deviceId);
 In v3, you should call `stopVideoInput` to stop the video input stream.
 
 ```js
-// v3
+// Before
```
xuesichao, line 67: Do we need to mention explicitly that `startVideoInput` API does not accept `null` as argument since it's a change of parameter type.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -565,12 +565,27 @@ export default class DefaultDeviceController
...
+      return dest.stream;
+    }
+    return this.activeDevices['audio'].stream;
   }
 
-  acquireVideoInputStream(): Promise<MediaStream> {
-    return this.acquireInputStream('video');
+  async acquireVideoInputStream(): Promise<MediaStream> {
+    if (!this.activeDevices['video']) {
+      throw new Error(`no video device chosen`);
```
simmkyu, line 583: Nit: `No video` to match the `acquireAudioInputStream`'s error message

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -565,12 +565,27 @@ export default class DefaultDeviceController
     return this.videoInputQualitySettings;
   }
 
-  acquireAudioInputStream(): Promise<MediaStream> {
-    return this.acquireInputStream('audio');
+  async acquireAudioInputStream(): Promise<MediaStream> {
```
xuesichao, line 568: Nice to split them 👍

simmkyu:
Destroy change is missing?

src/devicecontroller/DeviceController.ts
```diff
@@ -86,15 +86,15 @@ export default interface DeviceController {
...
   /**
    * Selects a video input device to use. The constraint may be a device id,
-   * `MediaTrackConstraint`, `MediaStream` (containing video track), or `null` to
-   * indicate no device. The promise will resolve indicating success or it will
-   * throw an appropriate error indicating the failure.
+   * `MediaTrackConstraint`, `MediaStream` (containing video track). It may also be an {@link VideoTransformDevice}
+   * to apply video transform.
+   * The promise will resolve indicating success or it will throw an appropriate error indicating the failure.
    */
   startVideoInput(device: VideoInputDevice): Promise<MediaStream | undefined>;
```
xuesichao, line 93: Suggest removing `null` from `AudioInputDevice` type and add it in the `startAudioInput` API:
```ts
// Now
type AudioInputDevice = Device | AudioTransformDevice | null;
startAudioInput(device: AudioInputDevice): Promise<MediaStream | undefined>;

// Suggest
type AudioInputDevice = Device | AudioTransformDevice;
startAudioInput(device: AudioInputDevice | null): Promise<MediaStream | undefined>;
```

* Keep `AudioInputDevice` type and `VideoInputDevice` type consistent.
* Easier to tell the type difference between parameter of `startAudioInput` and `startVideoInput`.

```ts
type VideoInputDevice = Device | VideoTransformDevice;
startVideoInput(device: VideoInputDevice): Promise<MediaStream | undefined>;
```

ltrung: Feel like null is a valid input for audio so I would rather keep it that way. Audio and video work different so there is no need for exact consistency. In the future, where we can truly join with no audio then we can just drop it from AudioInputDevice.

guides/17_Migration_to_3_0.md
ltrung: Don't think its needed since we already add the before example.

src/devicecontroller/DefaultDeviceController.ts
simmkyu: Nice suggestion, since the device controller now has two separate "start" APIs for audio and video!

ltrung: > Destroy change is missing?

I dropped it for now as it causes the update in all docs file (not sure why yet).

ltrung:
> Destroy change is missing?

I dropped it for now as it causes the update in all docs file (not sure why yet).

src/devicecontroller/DeviceController.ts
xuesichao: Although `null` is a valid input but in fact there is no `null` device, it's just a special case. The essence of `Device` and `TransformDevice` is `MediaStream` obtained from browser. And `null` isn't a valid way to request it. 

If the type is called `startAudioInputParamType`, then I think it's ok to add `null`. But now it stands for a type of device, so it makes more sense to me to remove it from `AudioInputDevice` type.

ltrung: Seem like niche change to me. Just like MediaTrackConstraints is not a real device. The end result is that null generates an audio input change so it is an input type. Device is kinda a bad name since it can be stream and other stuff that is not device. Maybe the name should be `AudioInputType` instead of `AudioInputDevice`  but I am not going to change it now.

xuesichao: Agree, `device` is a bad naming. Yeah we can change this later.

xuesichao:
>I dropped it for now as it causes the update in all docs file (not sure why yet).

Feel like it's kinda putting the cart before the horse. Don't think drop the reasonable improvement just because of too many doc change is a right thing.

If we really cares about it, maybe we could consider adding `/doc/*` to `.gitignore`.

ltrung:
> > I dropped it for now as it causes the update in all docs file (not sure why yet).
> 
> Feel like it's kinda putting the cart before the horse. Don't think drop the reasonable improvement just because of too many doc change is a right thing.
> 
> If we really cares about it, maybe we could consider adding `/doc/*` to `.gitignore`.

Can't ignore docs as it is our public documentation. By drop it I mean not including it in the same PR as the other one which is a more higher priority. I will come back to it tomorrow unless you want me to stay in office to finish that today :)

xuesichao (approved):
Wish we could still have the `Destroyable` change, but anyway it's not the scope of this PR so approved.

xuesichao (dismissed):

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -565,12 +565,27 @@ export default class DefaultDeviceController
...
+    if (!this.activeDevices['audio']) {
```
devalevenkatesh, line 569: Do we assign `null` to activeDevices['audio'], if yes does this code path execute say when we have selected `null` audio input device already?

Since, this is only related to Web Audio, should we call `startAudioInput(null)` explicitly under if `this.useWebAudio` is `true`? Looks like this function will execute and create empty device if current selection for audio input is `null`.

src/signalingclient/DefaultSignalingClient.ts
```diff
@@ -431,7 +431,7 @@ export default class DefaultSignalingClient implements SignalingClient {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const GlobalAny = global as any;
     GlobalAny['window'] &&
-      GlobalAny['window']['addEventListener'] &&
+      GlobalAny['window']['removeEventListener'] &&
```
devalevenkatesh, line 434: Thanks for fixing this!

src/devicecontroller/DefaultDeviceController.ts
ltrung: no we only assign activeDevices['audio'].stream = <stream> from null. This is case is relevant to both webAudio is enabled/disabled (you need a dummy audio to join a call). In both case, we need empty device created first then attach it to the webAudio node if webAudio is enabled.

devalevenkatesh (approved):

src/devicecontroller/DefaultDeviceController.ts
devalevenkatesh: Got it thanks for explanation!

simmkyu (approved):

PR #2151, opened 2022-04-04
[V3] Handle multiple start/stop audio/video input calls

ltrung:
**Description of changes:**
Handle multiple start/stop audio/video input calls by calling them in order instead of throwing errors.

**Testing:**

In meeting demo, regression testing switch audio/video input.

**Checklist:**


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -216,14 +217,14 @@ export default class DefaultDeviceController
   }
 
   async startAudioInput(device: AudioInputDevice): Promise<MediaStream | undefined> {
+    return (await this.audioInputQueue.add(() => this.startAudioInputInner(device))) as MediaStream;
```
simmkyu, line 220: A better way would be to use a TypeScript generic instead of type casting.

```
add<T>(func: () => Promise<T>): Promise<T> {
```

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -71,8 +72,8 @@ export default class DefaultDeviceController
   private readonly useWebAudio: boolean = false;
   private readonly useMediaConstraintsFallback: boolean = true;
 
-  private isStartAudioInputInProgress: boolean = false;
-  private isStartVideoInputInProgress: boolean = false;
+  private audioInputQueue: PromiseQueue = new PromiseQueue();
```
xuesichao, line 75: nit: This naming might be confusing. How about something like below?:
`audioInputQueue` - >`audioInputTaskQueue`
`startAudioInputInner` -> `startAudioInputTask`
`stopAudioInputInner` -> `stopAudioInputTask`

xuesichao (approved):
LGTM. Thanks!

src/devicecontroller/DefaultDeviceController.ts
ltrung: Thank you very much. I tried <T> this morning but probably did it wrong as it did not work :p.

src/devicecontroller/DefaultDeviceController.ts
ltrung: Nice suggestion. Updated.

xuesichao (approved):

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -374,16 +374,15 @@ export default class DefaultDeviceController
   }
 
   async startVideoInput(device: VideoInputDevice): Promise<MediaStream | undefined> {
+    return await this.videoInputTaskQueue.add(() => this.startVideoInputTask(device));
```
simmkyu, line 377: - Do I understand correctly that builders cannot handle an error anymore? i.e., The start input API does not throw an error.
- `await` only in the start API?

simmkyu: NVM. Using `await` will let `add` throw an error so builders can handle it accordingly.

simmkyu: I discussed this offline with Trung. The APIs behave like the following:

```js
try {
  await startVideoInput(deviceId) // Suppose that this call throws an error.

  // Not reached
  await startVideoInput(deviceId2)
} catch (error) {
  // Catch and handle an error.
}
```

```js
try {
  startVideoInput(deviceId) // Suppose that this call throws an error.

  // Reached
  startVideoInput(deviceId2)
} catch (error) {}
```

---

Only one concern we discussed is that if the browser's permission dialog appears, the queue will hold subsequent promise objects for a long time, confusing builders.

simmkyu (approved):

PR #2156, opened 2022-04-06
Reuse media capture s3 buckets for release demo

ltrung:
**Description of changes:**
- Specified a fix s3 prefix for media capture s3 buckets so we reuse them for each release instead of creating a new one for every release.

**Testing:**

Manually deploy demo and verify that s3 buckets are not re-created.

**Checklist:**


script/release.js
```diff
@@ -8,12 +8,12 @@ const currentBranch = (spawnOrFail('git', [' branch --show-current'], { skipOutp
 const deployDemo = (version) => {
   const formattedVersion = `${version.replace(/\./g, "-")}`;
   process.chdir(path.join(__dirname, '../demos/serverless'));
-  const meetingDemoName = `chime-sdk-demo-${formattedVersion}`;
+  const meetingDemoName = `chime-sdk-demo-global-${formattedVersion}`;
   logger.log(`Deploying ${meetingDemoName} ...`);
-  spawnOrFail('npm', [`run deploy -- -b ${meetingDemoName} -s ${meetingDemoName} -o ${meetingDemoName} -u false`], { printErr: true });
+  spawnOrFail('npm', [`run deploy -- -b ${meetingDemoName} -s ${meetingDemoName} -o chime-sdk-demo-global -u false`], { printErr: true });
```
LeviSklaroff, line 13: Why are we changing the -o param?

ltrung: So we can test media capture in both global and regional demo.  Also we change it to a fix name so we do not create new S3 buckets every time for each release.

LeviSklaroff: Gotcha, I misunderstood what the `-o` option was doing

LeviSklaroff (approved):

PR #2160, opened 2022-04-07
Update getMediaType() to check the property kind instead of mediaType

xuesichao:
**Issue #:**
The `rawMetricReport` of type `remote-inbound-rtp` is not stored when using Chrome.

**Description of changes:**
Update `getMediaType()` to check the property `kind` instead of `mediaType`.

**Testing:**

Yes, meeting demo. Run an e2e test using Firefox and Chrome, and use Network Link Conditioner to simulate 30% `packetsLoss`. Confirm the metrics works well in both browser.

**Checklist:**

Yes.

No.

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Update `getMediaType` method to check the property `kind` instead of `mediaType` of a `RawMetricReport`.
+
 ### Fixed
 
 ## [3.0.0] - 2022-03-30
--- a/src/statscollector/StatsCollector.ts
+++ b/src/statscollector/StatsCollector.ts
@@ -393,7 +393,7 @@ export default class StatsCollector {
    * Returns the MediaType for a RawMetricReport.
    */
   private getMediaType(rawMetricReport: RawMetricReport): MediaType {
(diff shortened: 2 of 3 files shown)


LeviSklaroff (approved):

devalevenkatesh (commented):
Couple of questions:
1. Have we tested mobile browsers and Safari, just to cover the base on the fix?
2. Is this also affecting the downstream stats? This is upstream based hence did the `mediaType` worked for downstream metric map or we need `kind` there as well?

xuesichao:
@devalevenkatesh
> Have we tested mobile browsers and Safari, just to cover the base on the fix?

Tested that it works for Safari. Didn't test for mobile as the WebRTC metrics is not well supported in mobile browsers.

> Is this also affecting the downstream stats? This is upstream based hence did the mediaType worked for downstream metric map or we need kind there as well?

This change will not affect downstream or any other stream. So the `mediaType` is obsolete. And `kind` is the replacement for it. Check here: https://w3c.github.io/webrtc-stats/#obsolete-rtcrtpstreamstats-members. 

And also

I connfirmed that:
* Firefox supports `inbound-rtp`, `outbound-rtp`, `remote-inbound-rtp` and `remote-outbound-rtp`. All of them contain property `kind`.
* Chrome supports `inbound-rtp`, `outbound-rtp`, `remote-inbound-rtp` and `remote-outbound-rtp`. All of them contain property `kind`.
* Safairi supports `inbound-rtp` and `outbound-rtp`. Both of them contain property `kind`.

PR #2162, opened 2022-04-08
No longer stop video stream when calling stopLocalVideoTile

ltrung:
**Issue #:**
- In V3, we decouple device controller from other workflows and only manage streams in device controller. However, `disconnectVideoStreamFromVideoElement` (which can be called by `stopLocalVideoTile`) still tries to stop media stream as a workaround to prevent crashes in older Safari versions. 

**Description of changes:**
- Remove the workaround so `disconnectVideoStreamFromVideoElement` no longer stops media stream. Test in Safari 13 and 14 and do not observe any crash.

**Testing:**

- Join a meeting demo,  turn on camera, then turn it off (test in older version of Safari if possible).

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Removed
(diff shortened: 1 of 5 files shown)


simmkyu (approved):
Are unit tests working without any change?

xuesichao (approved):

xuesichao (approved):

PR #2163, opened 2022-04-12
Move subscribe to MediaStreamBrokerObserver to end of connection

ltrung:
**Issue #:**
In Safari, joining really quick in the preview window in the demo before the audio input selection finish initializing will cause an error `failed to replace local audio input` to be thrown. This is because we try to replace local audio when there is no connection yet. This is not a new issue but in v2 we try and catch the error from `restartLocalAudio` in device controller (https://github.com/aws/amazon-chime-sdk-js/blob/v2.30.0/src/devicecontroller/DefaultDeviceController.ts#L1625).

**Description of changes:**
Only listen to audio/video media stream input changed when audio video session is connected. In other words, move `addMediaStreamBrokerObserver` at the end of connection when all connection tasks have finished.

Also change to subscribe/unsubscribe for every connection (including reconnection) and disconnection since we create new transceiver every time so this theoretically happen during reconnection too.

**Testing:**

- Click fast in preview window on Safari to join meeting
- Verify that there is no error thrown in console. 

**Checklist:**

### DIFF
--- a/src/audiovideocontroller/DefaultAudioVideoController.ts
+++ b/src/audiovideocontroller/DefaultAudioVideoController.ts
@@ -614,7 +614,6 @@ export default class DefaultAudioVideoController
       this.forEachObserver(observer => {
         Maybe.of(observer.audioVideoDidStartConnecting).map(f => f.bind(observer)(false));
       });
-      this._mediaStreamBroker.addMediaStreamBrokerObserver(this);
       this.eventController?.publishEvent('meetingStartRequested');
     }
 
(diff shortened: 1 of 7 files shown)


src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -785,7 +785,7 @@ export default class DefaultAudioVideoController
           this.configuration.connectionTimeoutMs
         ),
       ];
-
+      this.cleanUpMediaStreamsAfterStop();
```
devalevenkatesh, line 788: So we just remove the observers and does that help cleaning up the streams as well? Or cleaning up means setting active audio-video input devices to undefined.

PR #2179, opened 2022-04-21

dpwspoon:

**Issue #:**

**Description of changes:**

Main goal is to have credentials and endpoint URL refresh on reconnect.  Moving this logic internal of the Messaging Session did create async behavior in syncronous connect method.  Rev1 exposes this with an new connectAsync method.  On the fence of whether this is useful outside of testing or not so can have it limited to unit test changes (which need it to maintain assumptions)

**Testing:**

This can be tested by demo app by running a client and killing network and reconnecting.  I did this test on earlier internal version and will do it again on final version of this cr 

**Checklist:**

No, it complains about Changelog needing update.  Is this an automated process or do you manually add entries.  Can you point me to directions on how to do it properly (I didn't see in README though might have missed)

Yes I deprecate a field by adding doc and then add a new method.  This has not been reviewed.  Looking for feedback

It does not change wire protocol


src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -62,7 +62,15 @@ export default class DefaultMessagingSession implements MessagingSession {
 
   start(): void {
     if (this.isClosed()) {
-      this.startConnecting(false);
+        this.startConnecting(false);
```
ltrung, line 65: nit: extra space?

salsorrentino (dismissed):

src/messagingsession/MessagingSessionConfiguration.ts
```diff
@@ -30,13 +30,17 @@ export default class MessagingSessionConfiguration {
...
+   * the chimeClient on reconnect
+   *
    * The messaging session id is to uniquely identify this messaging session for the userArn.
    * If messaging session id is passed in as null, it will be automatically generated.
    */
   constructor(
     public userArn: string,
     public messagingSessionId: string | null,
-    public endpointUrl: string,
+    public endpointUrl: string | null,
```
ltrung, line 43: nit: Should be `undefined` instead of `null`.

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -103,8 +111,15 @@ export default class DefaultMessagingSession implements MessagingSession {
     });
   }
 
-  private startConnecting(reconnecting: boolean): void {
-    const signedUrl = this.prepareWebSocketUrl();
+  private async startConnecting(reconnecting: boolean) {
+    // reconnect needs to re-resolve endpoint url, which will also refresh credentials on client if they are expired
+    let endpointUrl = !reconnecting ? this.configuration.endpointUrl : null;
```
ltrung, line 116: Since we plan to deprecate the `configuration.endpointURL` I think we should change this logic always try to get the endpoint unless users still pass in the endpoint.
```
if (reconnecting || !this.configuration.endpointURL) {
  // Get endpoint
}
```

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -103,8 +111,15 @@ export default class DefaultMessagingSession implements MessagingSession {
...
+    if (endpointUrl == null) {
+      const endpoint = await this.configuration.chimeClient.getMessagingSessionEndpoint().promise();
```
ltrung, line 118: Does this syntax work with `aws-sdk-js` v3? We now support both aws-sdk v2 and v3 in `amazon-chime-sdk-js` v3 and the syntax seems different in v3:
https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-chime-sdk-messaging/classes/getmessagingsessionendpointcommand.html

ltrung (commented):
Could we update the demo to remove this line since we deprecate the endpointURL in configuration? 
https://github.com/aws/amazon-chime-sdk-js/blob/main/demos/browser/app/messagingSession/messagingSession.ts#L57

src/messagingsession/DefaultMessagingSession.ts
dpwspoon: I believe that is what this is doing ;-P, just doing it in a way so let endpointUrl is defined above.  If it is null in configuration then it will fetch it.

src/messagingsession/DefaultMessagingSession.ts
dpwspoon: Will update v3 but leave like it is here for v2

src/messagingsession/MessagingSession.ts
```diff
@@ -9,6 +9,11 @@ export default interface MessagingSession {
    */
   start(): void;
 
+  /**
+   * Start a messaging session.
+   */
+  startAsync(): Promise<void>;
```
devalevenkatesh, line 15: - We have to make this optional method as we do not know if some builders are implementing the MessagingSession. If they are, then they have to implement this method and hence that would be a breaking change for them. 
- We would also need to update the 2.x README under the MessagingSession on using the `startAsync` method over `start` so that builders know why this is updated.

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -103,8 +111,15 @@ export default class DefaultMessagingSession implements MessagingSession {
     });
   }
 
-  private startConnecting(reconnecting: boolean): void {
-    const signedUrl = this.prepareWebSocketUrl();
+  private async startConnecting(reconnecting: boolean) {
+    // reconnect needs to re-resolve endpoint url, which will also refresh credentials on client if they are expired
+    let endpointUrl = !reconnecting ? this.configuration.endpointUrl : null;
+    if (endpointUrl == null) {
```
xuesichao, line 117: Need to update `==` to `===`.

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -123,15 +138,15 @@ export default class DefaultMessagingSession implements MessagingSession {
...
+  private prepareWebSocketUrl(endpointUrl: string | null): string {
     const queryParams = new Map<string, string[]>();
     queryParams.set('userArn', [this.configuration.userArn]);
     queryParams.set('sessionId', [this.configuration.messagingSessionId]);
     return this.sigV4.signURL(
       'GET',
       'wss',
       'chime',
-      this.configuration.endpointUrl,
+      endpointUrl == null ? this.configuration.endpointUrl : endpointUrl,
```
xuesichao, line 149: Need to update `==` to `===`.

src/messagingsession/MessagingSession.ts
```diff
@@ -9,6 +9,11 @@ export default interface MessagingSession {
    */
```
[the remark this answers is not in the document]
simmkyu: I don't think we need to introduce a new API `startAsync` because the messaging session invokes `messagingSessionDidStart` after receiving the `SESSION_ESTABLISHED` message. i.e., Even if the `startAsync` execution ends, the messaging session has not yet started. What do you think?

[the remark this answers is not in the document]
devalevenkatesh: Good point! @dpwspoon could you please check this? We do not have to await the new endpointUrl resolving which should help with existing `start` API.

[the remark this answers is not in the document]
dpwspoon: managed to work around them with some timeouts...

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -29,6 +30,7 @@ export default class DefaultMessagingSession implements MessagingSession {
     private readonly reconnectController?: ReconnectController,
     private readonly sigV4?: SigV4
   ) {
+    this.isConnecting;
```
devalevenkatesh, line 33: Is there some assignment missing?

devalevenkatesh, line 33: Is there some assignment missing?

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -61,7 +63,7 @@ export default class DefaultMessagingSession implements MessagingSession {
   }
 
   start(): void {
-    if (this.isClosed()) {
+    if (this.isClosed() && !this.isConnecting) {
```
devalevenkatesh, line 66: Could you please explain why do we need the `isConnecting` here, is `start` getting called multiple times unintendedly?

devalevenkatesh, line 66: Could you please explain why do we need the `isConnecting` here, is `start` getting called multiple times unintendedly?

dpwspoon: It's needed for an existing unit test that detects that start isn't called twice.  With the async behavior in connect now the isClosed existing method is using the underlying WebSocket connection state which may or may not reflect open by the time the second call is made so we need an isConnecting state.

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -191,7 +194,7 @@ export default class DefaultMessagingSession implements MessagingSession {
       !this.isClosing &&
       this.canReconnect(event.code) &&
       this.reconnectController.retryWithBackoff(async () => {
-        this.startConnecting(true);
+        await this.startConnecting(true);
```
devalevenkatesh, line 197: `await`ing this should also not affect as for the successful connection we will be depending on the `messagingSesssionDidStart` right? Tagging Kyu (@simmkyu ) for checking once.

devalevenkatesh, line 197: `await`ing this should also not affect as for the successful connection we will be depending on the `messagingSesssionDidStart` right? Tagging Kyu (@simmkyu ) for checking once.

src/messagingsession/DefaultMessagingSession.ts
dpwspoon: ooops, this is me just learning typescript as working.  Defaulted in param above

src/messagingsession/DefaultMessagingSession.ts
dpwspoon: this is in the reconnect and want to await so don't have duplicate reconnects going at same time if takes a while to start

src/messagingsession/MessagingSession.ts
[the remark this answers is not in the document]
dpwspoon: messagingSessionDidStart is not entirely true cause there are some assumptions executed by unit tests of calling start and then stop with out listening to the observer... hence the timeouts added to those tests

src/messagingsession/DefaultMessagingSession.ts
devalevenkatesh: Yup I thought so, thanks for explaining.

devalevenkatesh: Yup I thought so, thanks for explaining.

README.md
```diff
@@ -242,11 +242,10 @@ const logger = new ConsoleLogger('SDK', LogLevel.INFO);
 
 // You will need AWS credentials configured before calling AWS or Amazon Chime APIs.
 const chime = new Chime({ region: 'us-east-1' });
-const endpoint = await chime.getMessagingSessionEndpoint().promise();
 
 const userArn = /* The userArn */;
 const sessionId = /* The sessionId */;
-const configuration = new MessagingSessionConfiguration(userArn, sessionId, endpoint.Endpoint.Url, chime, AWS);
+const configuration = new MessagingSessionConfiguration(userArn, sessionId, undefined, chime, AWS);
```
devalevenkatesh, line 248: Suggest adding a explanation here as well for why builders need to pass `undefined` for the `endpointUrl`.

devalevenkatesh, line 248: Suggest adding a explanation here as well for why builders need to pass `undefined` for the `endpointUrl`.

devalevenkatesh (commented):
Few comments but looks good to me. You may have to run the `npm run build:release` again or commit the doc changes by running `npm run doc`.

From a failed check:
```
Error: there are uncommitted changes:
 M docs/classes/defaultmessagingsession.html
```

src/messagingsession/DefaultMessagingSession.ts
simmkyu: Right, `await` is unnecessary because `this.isConnecting` will be `true` anyway.

devalevenkatesh: But the `startConnecting` actually does not check for `isConnecting` may be instead of doing the check for `isConnecting` in `start` would it be better to do it in `startConnecting`?

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -121,21 +130,14 @@ export default class DefaultMessagingSession implements MessagingSession {
       }
     });
     this.setUpEventListeners();
+    this.isConnecting = false;
```
simmkyu, line 133: If any error occurs in this method, `this.isConnecting` will always be `true`. For example, `getMessagingSessionEndpoint` throws an error if the customer passes the Chime client with expired credentials.

How about using try/catch/finally block?

```
try {
  await ...
} catch {
} finally {
  this.isConnecting = false;
}
```

dpwspoon: thanks, updated

src/messagingsession/DefaultMessagingSession.ts
devalevenkatesh: Yes, but as per this discussion: https://github.com/aws/amazon-chime-sdk-js/pull/2179/files#r860311122, the `await` does not gurantee that the first start is successful, the start being successful is only guranteed when the messagingSessionDidStart is triggered.

README.md
dpwspoon: Not sure the cognitive overload on unimportant detail is needed for readme when already documented in doc.

devalevenkatesh: So the first place builders are going to look are at the README and seeing an `undefined` field is weird to start with, but agree they can go to `MessagingSessionConfiguration` documentation which as an extra hop. Can skip.

src/messagingsession/DefaultMessagingSession.ts
simmkyu: oh I understand what you meant. For example, if the WebSocket receives the `close` event twice in a short period, the messaging session can run `startConnecting` twice.

If we check for `this.isConnecting` in the `startConnecting`, we can guarantee that the messaging session executes `startConnecting` once at a time.

devalevenkatesh: @dpwspoon seems like to avoid running into duplicate reconnects, moving the `isConnecting` check inside the `startConnecting` will help over using `await`. Also, if waiting for reconnection means an end to end success, should we set the `isConnecting` to `false` in the `messagingSessionDidStart`?

dpwspoon: Plus didStart might never trigger say in auth failure

dpwspoon: `messagingSessionDidStart` only triggers if successful.   The await here is to make sure any errors bubble up.  Not to await for a success.  The await might not be strictly necessary but it doesn't hurt I don't think and it might be necessary

devalevenkatesh (approved):

PR #2180, opened 2022-04-22

dpwspoon:

**Issue #:**

**Description of changes:**

**Testing:**

Tested on 2.x line with chime demo and this is cherry pick of that but will try testing using https://github.com/aws/amazon-chime-sdk-js/blob/main/demos/browser/app/messagingSession/messagingSession.ts

Start the messaging demo. Force the network to shutdown and then reopen.  Should reconnect.

**Checklist:**

yes

~ sort of.  I made a parameter optional

No


CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
...
+
+### Added
+
+### Removed
+
+### Changed
+
+### Fixed
+
+- MessagingSession reconnects with refreshed endpoint and credentials if needed
```
ltrung, line 18: Could we mention that endpoint URL is deprecated in configuration and we will call `getMessagingSessionEndpoint` automatically ?

ltrung: Also need to update README in the example:
https://github.com/aws/amazon-chime-sdk-js/blob/main/README.md#messaging-session

CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
...
+
```
xuesichao, line 19: We haven't release 3.1.0 version. I think you should add your CHANGELOG under 3.1.0 section.
And remove 3.1.1 section.

CHANGELOG.md
dpwspoon: updated thx

CHANGELOG.md
xuesichao: @dpwspoon Hi, when do we schedule this release? if it's before the release of 3.1.0. Maybe we should update the version to 3.0.1, if it's a patch version.

Sorry I didn't catch this in the previous comment.

src/messagingsession/MessagingSessionConfiguration.ts
```diff
@@ -30,13 +30,17 @@ export default class MessagingSessionConfiguration {
   /**
    * Constructs a MessagingSessionConfiguration optionally with userArn, messaging session id, a messaging session
    * endpoint URL, and the chimeClient.
+   *
+   * endpointUrl is deprecated as internally it is set on connect via chimeClient if null, and re-resolved (whether null or not) from
```
devalevenkatesh, line 34: Since you are using `undefined` over `null` could you update this docstring accordingly?

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- MessagingSession reconnects with refreshed endpoint and credentials if needed.  EndpointUrl on `MessagingSessionConfiguration` is deprecated as it is resolved by calling getMessagingSessionEndpoint internally
```
devalevenkatesh, line 24: nit: Could you please backtick all coding constructs here? Also end the second sentence with a ".".

`MessagingSession` reconnects with refreshed endpoint and credentials if needed.  `EndpointUrl` on `MessagingSessionConfiguration` is deprecated as it is resolved by calling `getMessagingSessionEndpoint` internally.

README.md
```diff
@@ -243,11 +243,10 @@ const logger = new ConsoleLogger('SDK', LogLevel.INFO);
 
 // You will need AWS credentials configured before calling AWS or Amazon Chime APIs.
 const chime = new ChimeSDKMessagingClient({ region: 'us-east-1'});
-const endpoint = await chime.send(new GetMessagingSessionEndpointCommand());
 
 const userArn = /* The userArn */;
 const sessionId = /* The sessionId */;
-const configuration = new MessagingSessionConfiguration(userArn, sessionId, endpoint.Endpoint.Url, chime);
+const configuration = new MessagingSessionConfiguration(userArn, sessionId, null, chime);
```
devalevenkatesh, line 249: `null` -> `undefined` since we are using `undefined` in the method signature.

demos/browser/app/messagingSession/messagingSession.ts
```diff
@@ -54,10 +54,9 @@ export class DemoMessagingSessionApp implements MessagingSessionObserver {
       try {
         const response = await this.fetchCredentials();
         const chime = new ChimeSDKMessagingClient({ region: 'us-east-1', credentials: response });
-        const endpoint = await chime.send(new GetMessagingSessionEndpointCommand());
         this.userArn = (document.getElementById('userArn') as HTMLInputElement).value;
         this.sessionId = (document.getElementById('sessionId') as HTMLInputElement).value;
-        this.configuration = new MessagingSessionConfiguration(this.userArn, this.sessionId, endpoint.Endpoint.Url, chime);
+        this.configuration = new MessagingSessionConfiguration(this.userArn, this.sessionId, null, chime);
```
devalevenkatesh, line 59: Same as https://github.com/aws/amazon-chime-sdk-js/pull/2180/files?file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.ts&show-viewed-files=true#r856577470

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -104,7 +106,16 @@ export default class DefaultMessagingSession implements MessagingSession {
   }
 
   private async startConnecting(reconnecting: boolean): Promise<void> {
-    const signedUrl = await this.prepareWebSocketUrl();
+    // reconnect needs to re-resolve endpoint url, which will also refresh credentials on client if they are expired
+    let endpointUrl = !reconnecting ? this.configuration.endpointUrl : null;
```
devalevenkatesh, line 110: Should we use `undefined` here as well? `this.configuration.endpointUrl` can either be string | undefined, so if that value is undefined, here endpointUrl will be undefined and not strict equal null. I think you just updated the type to use undefined on the constructor but not elsewhere.

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -104,7 +106,16 @@ export default class DefaultMessagingSession implements MessagingSession {
...
+    if (endpointUrl === null) {
+      const endpoint = await this.configuration.chimeClient.send(
```
devalevenkatesh, line 112: So if `this.configuration.endpointUrl` is `undefined` while connecting at the start, then here, do you want to set the `this.configuration.endpointUrl = endpoint.Endpoint.Url` ? The docstring says so:
"endpointUrl is deprecated as internally it is set on connect via chimeClient if null.."

CHANGELOG.md
xuesichao: @ltrung @devalevenkatesh how do you think?

test/messagingsession/DefaultMessagingSession.test.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { GetMessagingSessionEndpointCommand } from '@aws-sdk/client-chime-sdk-messaging';
```
devalevenkatesh, line 4: I think we can add unit tests to test:
1. Checking the MessagingSessionConfiguration has a valid endpointUrl if passed as `undefined`. As the startConnecting should set one.
2. In reconnect unit tests, add a unit test to see whether we can mock a chimeClient to send a new endpointUrl and assert that code path works.

src/messagingsession/DefaultMessagingSession.ts
dpwspoon: I don't think we want to expose it in the configuration as that should not exist and just be an internal thing

test/messagingsession/DefaultMessagingSession.test.ts
dpwspoon: Will do 2, as stated above we don't want to set value on MessagingSessionConfiguration so won't do 1

src/messagingsession/DefaultMessagingSession.ts
dpwspoon: updated thx

README.md
dpwspoon: update thx

CHANGELOG.md
dpwspoon: updated thx

src/messagingsession/MessagingSessionConfiguration.ts
dpwspoon: updated thx

src/messagingsession/MessagingSessionConfiguration.ts
```diff
@@ -30,13 +30,17 @@ export default class MessagingSessionConfiguration {
   /**
    * Constructs a MessagingSessionConfiguration optionally with userArn, messaging session id, a messaging session
    * endpoint URL, and the chimeClient.
+   *
+   * endpointUrl is deprecated as internally it resolved on connect via chimeClient if undefined, and always re-resolved
```
devalevenkatesh, line 34: nit: "..it is resolved on connect..". Also could you please end this with a ".".

src/messagingsession/DefaultMessagingSession.ts
devalevenkatesh: Got it, so should the docstring contain why builders or when rather they can provide their own URL? Also, what is the impact when re-connecting if the builder provided URL at the start is different than the internally resolved one?

CHANGELOG.md
devalevenkatesh: So, this chanelog can be under 3.1.0 we will later fix the changelog once the patch is released. I think I had checked the same with Trung. After release we will need to add 3.0.1 as a patch with this change.

devalevenkatesh: The changelog update is a bit of improvement we have to make once we go through this hotfix release once and fix any hiccups if seen. But yes, we would like to add as part of the PR after we release this time.

src/messagingsession/DefaultMessagingSession.ts
dpwspoon: The builder should never set this URL directly.  If they do, and in all examples or demos they do it by calling getMessagingSessionEndpoint and then passing it in.  If they don't get the URL by calling getMessagingSessionEndpoint then there is no guarantee it will connect to anything and thus is broken if they have a hardcoded value.

This calling of `getMessagingSessionEndpoint` is what will be done internally via the PR, so this field is really strictly deprecated and should not be used as it has no purpose now.  There is no real use case to get the URL from the configuration object, which is why it doesn't really make sense to do extra work to keep it up to date.

In retrospect, the original implementation should have never exposed this URL and done the resolving on connect/reconnect like this PR implements.

devalevenkatesh: Got it thanks for the explanation!

devalevenkatesh (dismissed):

devalevenkatesh (approved):

PR #2192, opened 2022-04-28
Re-add layers allocation header extension and avoid changing and duplicate extension ids

shi-su:
**Issue #:**

**Description of changes:**
Re-add layers allocation header extension.

According to https://github.com/aws/amazon-chime-sdk-js/pull/1889, a changing header extension ID would result in a hard failure in chrome. Therefore, this PR added extra logic to check if there is changing extension IDs and if the extension is already in the SDP.

**Testing:**
Added unit tests for the changes and smoke tested with two to three users

**Checklist:**

n/a

n

n


src/sdp/SDP.ts
```diff
@@ -375,15 +377,22 @@ export default class SDP {
...
-  withVideoLayersAllocationRtpHeaderExtension(): SDP {
-    const sections = SDP.splitSections(this.sdp);
+  withVideoLayersAllocationRtpHeaderExtension(previousSdp: SDP): SDP {
+    const url = `http://www.webrtc.org/experiments/rtp-hdrext/video-layers-allocation00`
 
+    // According to https://webrtc.googlesource.com/src/+/b62ee8ce94e5f10e0a94d6f112e715cc4d0cd9dc,
+    // RTP header extension ID change would result in a hard failure. Therefore if the extension exists
+    // in the previous SDP, use the same extension ID to avoid the failure. Otherwise use a new ID
+    let previousId = previousSdp ? previousSdp.getRtpHeaderExtensionId(url) : -1;
+    const id = previousId === -1 ? this.getUniqueRtpHeaderExtensionId(SDP.splitLines(this.sdp)) : previousId;
```
hensmi-amazon, line 387: To make this super safe, can we check that this ID is still unique and just not add it if so?

src/sdp/SDP.ts
```diff
@@ -375,15 +377,22 @@ export default class SDP {
...
+
+    const sections = SDP.splitSections(this.sdp);
     const newSections = [];
     for (let section of sections) {
-      if (/^m=video/.test(section)) {
+      if (/^m=video/.test(section) && SDP.getRtpHeaderExtensionIdInSection(section, url) === -1) {
+        // Add RTP header extension only when it does not already exist
```
hensmi-amazon, line 393: To be safe, should we check to see if its different and override it if so?

src/sdp/SDP.ts
shi-su: Not sure if I fully understood this comment. Do you mean checking if `previousId` exists, checking for if it is unique in the current SDP?

src/sdp/SDP.ts
shi-su: In case that there is a different extension ID than previous SDP, the problem is a lot bigger since there could be multiple changes in the initial SDP. In that case, simply overriding the extension ID may not solve the problem since the id in previous SDP might be used by some other extensions. My understanding is that we should probably treat the initial SDP's extension ID as ground truth here since if the initial SDP is incorrect, it does not seem fixable in this function.

hensmi-amazon: So this is actually the issue that we previously ran into that caused all the issues. When the header extension showed up again, it had a different ID (yes this is a browser bug, but since SDP munging is sorta in the 'technically works but not official' region of support I just want to be careful :) ). What about just sanity checking and removing the line if it exists?

src/sdp/SDP.ts
hensmi-amazon: like if the first SDP uses 1-14 so we use 15. But for some reason browser adds a different extension for 15. We just bail out and don't add it.

Not sure if you saw in the previous PR, but we actually confirm whether the extension is negotiated before enabled any behavior based off of it. So bailing out and not adding is totally fine.

shi-su: I opened another PR https://github.com/aws/amazon-chime-sdk-js/pull/2195 for this. The change https://github.com/aws/amazon-chime-sdk-js/pull/2195/files#diff-d6a73856184ac9b6920494e5e8fcc8be715f515d30c8871cc9a8891ebc7c87fdR397 should avoid this behavior

src/sdp/SDP.ts
shi-su: In the new PR, I added https://github.com/aws/amazon-chime-sdk-js/pull/2195/files#diff-d6a73856184ac9b6920494e5e8fcc8be715f515d30c8871cc9a8891ebc7c87fdR411 to override the extension ID if changed, it also removes the extension if the ID changed and the old ID is used by some other extensions

PR #2193, opened 2022-04-29
Bug/msg session reconnect

dpwspoon:
**Issue #:**

**Description of changes:**

Fix issue with recent change to getMessagingSessionEndpoint resolve on connect where it would not reconnect if getMessagingSessionEndpoint failed.

**Testing:**

Start the messaging demo app in browser.  Create app instance user.  Connect.  Disconnect wifi, reconnect, and see websocket get establish

**Checklist:**

yes

yes

yes

### DIFF
--- a/src/messagingsession/DefaultMessagingSession.ts
+++ b/src/messagingsession/DefaultMessagingSession.ts
@@ -107,6 +107,15 @@ export default class DefaultMessagingSession implements MessagingSession {
   private async startConnecting(reconnecting: boolean): Promise<void> {
     this.isConnecting = true;
     try {
+      if (!reconnecting) {
+        this.reconnectController.reset();
+      }
+      if (this.reconnectController.hasStartedConnectionAttempt()) {
+        this.reconnectController.startedConnectionAttempt(false);
+      } else {
+        this.reconnectController.startedConnectionAttempt(true);
+      }
+
       // reconnect needs to re-resolve endpoint url, which will also refresh credentials on client if they are expired.
       let endpointUrl = !reconnecting ? this.configuration.endpointUrl : undefined;
       if (endpointUrl === undefined) {
@@ -118,14 +127,7 @@ export default class DefaultMessagingSession implements MessagingSession {
 
       const signedUrl = this.prepareWebSocketUrl(endpointUrl);
       this.logger.info(`opening connection to ${signedUrl}`);
-      if (!reconnecting) {
-        this.reconnectController.reset();
-      }
-      if (this.reconnectController.hasStartedConnectionAttempt()) {
-        this.reconnectController.startedConnectionAttempt(false);
-      } else {
-        this.reconnectController.startedConnectionAttempt(true);
-      }
+
       this.webSocket.create(signedUrl, [], true);
       this.forEachObserver(observer => {
         if (observer.messagingSessionDidStartConnecting) {
@@ -191,16 +193,21 @@ export default class DefaultMessagingSession implements MessagingSession {
     }
   }
 
+  private retryConnection(): boolean {
+    return this.reconnectController.retryWithBackoff(async () => {
+      try {
+        await this.startConnecting(true);
+      } catch (e) {
+        // getMessagingSessionEndpoint can fail, so need to retry here
+        this.retryConnection();
+      }
+    }, null);
+  }
+
   private closeEventHandler(event: CloseEvent): void {
     this.logger.info(`WebSocket close: ${event.code} ${event.reason}`);
     this.webSocket.destroy();
-    if (
-      !this.isClosing &&
-      this.canReconnect(event.code) &&
-      this.reconnectController.retryWithBackoff(async () => {
-        await this.startConnecting(true);
-      }, null)
-    ) {
+    if (!this.isClosing && this.canReconnect(event.code) && this.retryConnection()) {
       return;
     }
     this.isClosing = false;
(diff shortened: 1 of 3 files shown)


src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -191,16 +193,21 @@ export default class DefaultMessagingSession implements MessagingSession {
     }
   }
 
+  private retryConnection(): boolean {
+    return this.reconnectController.retryWithBackoff(async () => {
+      try {
+        await this.startConnecting(true);
+      } catch (e) {
+        // getMessagingSessionEndpoint can fail, so need to retry here
+        this.retryConnection();
```
dpwspoon, line 202: should this be schedule next tick or something so call stack doesn't get large?

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -191,16 +193,21 @@ export default class DefaultMessagingSession implements MessagingSession {
     }
   }
 
+  private retryConnection(): boolean {
+    return this.reconnectController.retryWithBackoff(async () => {
+      try {
+        await this.startConnecting(true);
+      } catch (e) {
```
devalevenkatesh, line 200: So this behavior is a new thing? Should not the existing code run into this as well?

src/messagingsession/DefaultMessagingSession.ts
devalevenkatesh, line 202: Should there be a threshold on how long one can retry connecting? I think the threshold is handled by `retryWithBackoff` already so it should execute only when the `retryConnection` returns `true` right?

I think this should be return

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -107,6 +107,15 @@ export default class DefaultMessagingSession implements MessagingSession {
...
+      }
+      if (this.reconnectController.hasStartedConnectionAttempt()) {
+        this.reconnectController.startedConnectionAttempt(false);
+      } else {
+        this.reconnectController.startedConnectionAttempt(true);
+      }
+
       // reconnect needs to re-resolve endpoint url, which will also refresh credentials on client if they are expired.
       let endpointUrl = !reconnecting ? this.configuration.endpointUrl : undefined;
       if (endpointUrl === undefined) {
```
devalevenkatesh, line 121: I think you also have to handle the error here, sorry I missed that there is no `catch` here:
https://github.com/aws/amazon-chime-sdk-js/pull/2179#discussion_r860332917

devalevenkatesh: The `getMessagingSessionEndpoint()` can error out here as well.

src/messagingsession/DefaultMessagingSession.ts
devalevenkatesh, line 202: I think `this.retryConnection()` needs to return a boolean value to check whether to trigger a reconnection or not right?

Something like:
```
private retryConnection(): boolean {
  try {
   if (this.reconnectController.retryWithBackoff(async () => await this.startConnecting(true))) {
     return true;
   } else {
     return false;
   }
  } catch (error) {
    this.logger.error(`${error}`);
    return retryConnection();
  }
}
```

src/messagingsession/DefaultMessagingSession.ts
dpwspoon: There is a finally block around this whole thing to reset the boolean.  Not sure what if anything we should do with the caught exception.  Thoughts?

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -126,6 +115,30 @@ export default class DefaultMessagingSession implements MessagingSession {
...
+      // reconnect needs to re-resolve endpoint url, which will also refresh credentials on client if they are expired.
+      let endpointUrl = !reconnecting ? this.configuration.endpointUrl : undefined;
+      if (endpointUrl === undefined) {
+        try {
+          const endpoint = await this.configuration.chimeClient
+            .getMessagingSessionEndpoint()
+            .promise();
+          endpointUrl = endpoint.Endpoint.Url;
+        } catch (e) {
+          const closeEvent = new CloseEvent('close', {
```
devalevenkatesh, line 128: Could you print the error here before creating the CloseEvent?
```
this.logger.error(${error});
```

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -191,16 +204,18 @@ export default class DefaultMessagingSession implements MessagingSession {
...
-      !this.isClosing &&
-      this.canReconnect(event.code) &&
-      this.reconnectController.retryWithBackoff(async () => {
-        await this.startConnecting(true);
-      }, null)
-    ) {
+    if (event.code !== 4999) {
+      this.webSocket.destroy();
+    }
+    if (!this.isClosing && this.canReconnect(event.code) && this.retryConnection()) {
```
devalevenkatesh, line 218: Here, `canReconnect` does not handle `4999` and will return `false`, so, should it now handle this as we are throwing a custom `CloseEvent` with `4999` code and expect a retry and a reconnect?

dpwspoon: `canReconnect` does handle 4999 cause `closeCode > 4000 ` other than the ones special case returns true.

src/messagingsession/DefaultMessagingSession.ts
dpwspoon: This is already logged when handled in the closeEventHandler so don't think duplicate logging is needed here: `  this.logger.info(`WebSocket close: ${event.code} ${event.reason}`);`

devalevenkatesh: Yeah, but the reason is logged, not sure if `error` can be something different hence suggested.

src/messagingsession/DefaultMessagingSession.ts
devalevenkatesh: Ah, my bad, missed the `>`, thanks for correcting.

src/messagingsession/DefaultMessagingSession.ts
devalevenkatesh: Also, good solution I think.

devalevenkatesh (approved):

PR #2195, opened 2022-05-02
Re-add layers allocation header extension and avoid changing and duplicate extension ids

shi-su:
**Issue #:**

**Description of changes:**
Re-add layers allocation header extension.

According to https://github.com/aws/amazon-chime-sdk-js/pull/1889, a changing header extension ID would result in a hard failure in chrome. Therefore, this PR added extra logic to check if there is changing extension IDs and if the extension is already in the SDP.

**Testing:**
Added unit tests for the changes and smoke tested with two to three users

1. Turn on simulcast and throttle the network to force browser to drop one layer. Verified switching to the other stream without freeze;
2. Turn on and off video on sender side and verify the sender does not drop (Verify the issue in https://github.com/aws/amazon-chime-sdk-js/pull/1889 does not happen)

**Checklist:**

y

n

n


CHANGELOG.md
```diff
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
ltrung, line 8: Don't do unreleased. Please put it under 3.2.0 which is the next release version

shi-su: Updated change log and put in 3.2.0

src/task/SetLocalDescriptionTask.ts
```diff
@@ -33,6 +33,12 @@ export default class SetLocalDescriptionTask extends BaseTask {
     const sdpOfferInit = this.context.sdpOfferInit;
     let sdp = sdpOfferInit.sdp;
 
+    if (this.context.browserBehavior.hasChromiumWebRTC()) {
```
ltrung, line 36: nit: We generally try to add new API that is specific to the browser behavior (in this case something like supportsVideo...HeaderExtension).

hensmi-amazon: I think technically (maybe a bit overkill) the more future prepped (i.e. in case FF/others decide to add support/support munging) check would be [this.meetingSessionContext.transceiverController.localVideoTransceiver().sender.getParameters](https://github.com/aws/amazon-chime-sdk-js/blob/8e7c0402c89837fe2dd623075003bbf679d28247/src/audiovideocontroller/DefaultAudioVideoController.ts#L964) . Since we are checking with that to see whether it ends up being negotiated. Limiting to just chrome is fine though (the helper function as trung mentioned would be nice as well).

hensmi-amazon (dismissed):

src/task/SetLocalDescriptionTask.ts
shi-su: Not sure what is the desired check here. Are we sticking to limiting to chrome? @ltrung what do you think?

hensmi-amazon: Let's just limit to chrome for now, but we can abstract that choice (and technically allow builder override) by adding a function like `supportsVideoLayersHeaderExtensionMunging` to https://github.com/aws/amazon-chime-sdk-js/blob/main/src/browserbehavior/DefaultBrowserBehavior.ts .

I was just mentioning that we could theoretically make the check broader, as long as the browser supports `getParameters` we can check whether the extension was negotiated or not. Not important though.

shi-su: Okay, added a function `supportsVideoLayersAllocationRtpHeaderExtension`.

hensmi-amazon (approved):

PR #2203, opened 2022-05-04
Show correct media region in demo meeting window

pracheth:
**Issue #:**

**Description of changes:**
* Show correct media region in meeting window
* Also adds a tooltip for region dropdown which shows `Used only for meeting creation. Ignored if meeting already exists` when hoverd

**Testing:**
Tested by deploying demo to account and check region displayed in meeting window

Yes
* Deploy demo
* Join a meeting X in region `us-east-1`. This will lead to meeting creation.
* Check the region displayed in meeting window on top right is `us-east-1`
* Leave call
* Join same meeting X but choose a different region in dropdown. 
* Verify that the region displayed in meeting window on top right is still `us-east-1` which is the region the meeting is hosted in. (Earlier this used to display the region selected in dropdown which is incorrect)

**Checklist:**

Y

N

N


ltrung (approved):
Thanks for fixing this :)

PR #2204, opened 2022-05-04
Reset state in case audio video controller is reused

hensmi-amazon:
**Issue #:** None. Chime Application was having issues on reuse due to stale state. 

**Description of changes:** See comment.

**Testing:** Chime application tested reuse of the audio video controller, and no longer had stale state (leading to missing tiles).

No

**Checklist:**

y

n

n

### DIFF
--- a/src/audiovideocontroller/DefaultAudioVideoController.ts
+++ b/src/audiovideocontroller/DefaultAudioVideoController.ts
@@ -793,6 +793,9 @@ export default class DefaultAudioVideoController
     }
     this.sessionStateController.perform(SessionStateControllerAction.FinishDisconnecting, () => {
       if (!reconnecting) {
+        // Do a hard reset of the signaling client in case this controller is reused;
+        // this will also can `this.meetingSessionContext` to be reset if reused.
+        this.meetingSessionContext.signalingClient = null;
         this.notifyStop(status, error);
       }
     });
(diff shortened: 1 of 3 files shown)


src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -793,6 +793,9 @@ export default class DefaultAudioVideoController
     }
     this.sessionStateController.perform(SessionStateControllerAction.FinishDisconnecting, () => {
       if (!reconnecting) {
+        // Do a hard reset of the signaling client in case this controller is reused;
+        // this will also can `this.meetingSessionContext` to be reset if reused.
+        this.meetingSessionContext.signalingClient = null;
```
ltrung, line 798: Do you need to reset it here as well? https://github.com/aws/amazon-chime-sdk-js/blob/main/src/audiovideocontroller/DefaultAudioVideoController.ts#L736

ltrung:
Can we put more details on the description in a general case (without mentioning Chime App specifically) since I assume other builders can have the same issue?

src/audiovideocontroller/DefaultAudioVideoController.ts
hensmi-amazon: Sure, I'm not super familiar with that path but I think it makes sense to reset it there too (or at least it doesn't hurt).

ltrung (approved):

PR #2213, opened 2022-05-09
Pre run script automation

jhaarpan:
**Issue #:**

**Description of changes:**
Created a workflow to update the ID for the pre run script for integration tests
We need to update the ID because the 'script.sh' file expires every two months on Saucelabs which causes our canaries to fail
I am not using a personal access token as it is not required. We can see [here](https://github.com/marketplace/actions/create-pull-request#action-inputs) that if the workflow is not running `on: push` or `on: pull_request`, we can use the default `GITHUB_TOKEN`.

**Testing:**

**Checklist:**


.github/workflows/update-prerun-script.yml
```diff
@@ -0,0 +1,45 @@
...
+    name: Download pre run script from S3 bucket
+    runs-on: ubuntu-latest
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+      - run: aws s3 cp s3://saucelabs-prebuild-script/script.sh ./
```
michhyun1, line 22: what is the point of downloading this in a separate job? Each job runs in a different context, right?

michhyun1, line 22: Does this  job need to finish first before the update-pre-run-script job?

jhaarpan: Yes it does

.github/workflows/update-prerun-script.yml
```diff
@@ -0,0 +1,93 @@
...
+          delete-branch: true
+  
+  update-pre-run-script-previous-version:
+    name: Update pre run script on Saucelabs for previous version
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Package
+        uses: actions/checkout@v3
+        with:
+          ref: "release-${{ needs.get-previous-version.outputs.previous_version_number }}.x"
```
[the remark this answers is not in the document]
jhaarpan: Yes it is the correct syntax. Took this code from the existing workflow

.github/workflows/update-prerun-script.yml
```diff
@@ -0,0 +1,93 @@
...
+    steps:
+      - name: Checkout Package
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Get Version
+        id: get-version
+        run: |
+          current_version=$(.github/script/get-current-version)
+          echo "::set-output name=version-number::$((${current_version%%.*} - 1))"
```
michhyun1, line 31: why cant you just add this to the job below? Why do you need a second job?

.github/workflows/update-prerun-script.yml
```diff
@@ -0,0 +1,93 @@
...
+          aws-region: us-east-1
+      - run: aws s3 cp s3://saucelabs-prebuild-script/script.sh ./
+      - name: Update pre run script and create PRs
+        run: |
+          RESPONSE="$(curl -u "$SAUCE_USERNAME:$SAUCE_ACCESS_KEY" —location --request POST 'https://api.us-west-1.saucelabs.com/v1/storage/upload' --form 'payload=@"script.sh"' --form 'name="blur_replacement_script"')"
+          SCRIPT_ID=$(jq -r '.item.id' <<< "$RESPONSE")
+          echo "Updating integration tests pre-run script ID for origin branch"
+          sed -i -E "s/storage.*'/storage:$SCRIPT_ID'/g" ${GITHUB_WORKSPACE}/integration/js/utils/WebdriverSauceLabs.js
+          rm -rf script.sh
+      - name: Create Pull Request
```
ltrung, line 87: Do we really need a github action for this? Can you just use regular GitHub CLI commands?

.github/workflows/update-prerun-script.yml
```diff
@@ -0,0 +1,93 @@
...
+        with:
+          ref: "release-${{ needs.get-previous-version.outputs.previous_version_number }}.x"
+          fetch-depth: 0
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+      - run: aws s3 cp s3://saucelabs-prebuild-script/script.sh ./
```
ltrung, line 79: Should there be a name for this step?

.github/workflows/update-prerun-script.yml
ltrung: Agree with Michael. This can be organized better. Seem like you need 3 jobs. First one is to download the scripts and upload to SL. Then you can have another 2 jobs to update it in current and prev_version branch.

.github/workflows/update-prerun-script.yml
jhaarpan: The action gets the credentials directly. I will have to add credentials to GitHub secrets to use the CLI commands

michhyun1: which credentials is this using? I dont see how this has the correct permisisons

jhaarpan: This is using the default `GITHUB_TOKEN` according to [this](https://github.com/marketplace/actions/create-pull-request#action-inputs)

michhyun1: Oh right, I rmember now - Github token should be sufficient. We tend to stay away from 3rd party actions as much as possible though. Is it possible to just have a bash script do this with github cli?

jhaarpan: I did try using GitHub CLI in one of my previous commits. It did not work as it said it did not have enough permissions to create a PR which is why I am using this action now. I will have to create a PAT to add permissions to create a PR

michhyun1: If its using the GITHUB_TOKEN, that env variable is available in this context, meaning you should be able to write a script as long as you make sure it has the github token env variable attached to the job that runs the script

.github/workflows/update-prerun-script.yml
michhyun1: I would combine "get-previous-version" and "update-pre-run-script"

jhaarpan: Done

.github/workflows/update-prerun-script.yml
```diff
@@ -0,0 +1,86 @@
+name: Update pre build script on Saucelabs
+
+on:
+  schedule:
+    - cron: '*/10 * * * *'
+#  schedule:
+#    - cron: '0 0 1 */2 *'
```
michhyun1, line 7: delete?

.github/workflows/update-prerun-script.yml
```diff
@@ -0,0 +1,86 @@
...
+        id: upload-pre-run-script
+        run: |
+          RESPONSE="$(curl -u "$SAUCE_USERNAME:$SAUCE_ACCESS_KEY" —location --request POST 'https://api.us-west-1.saucelabs.com/v1/storage/upload' --form 'payload=@"script.sh"' --form 'name="blur_replacement_script"')"
+          SCRIPT_ID=$(jq -r '.item.id' <<< "$RESPONSE")
+          rm -rf script.sh
+          echo "::set-output name=script-id::$SCRIPT_ID"
+      
+  create-pr:
+    name: Create PR for origin branch
+    runs-on: ubuntu-latest
```
michhyun1, line 41: If this job depends on the previous job, doesnt it have to have the `needs: update-pre-run-script` property set?

.github/workflows/update-prerun-script.yml
```diff
@@ -0,0 +1,61 @@
...
+          echo "Updating integration tests pre-run script ID for origin branch"
+          git branch update-pre-run-script-id
+          git checkout update-pre-run-script-id
+          sed -i -E "s/storage.*'/storage:$SCRIPT_ID'/g" ${GITHUB_WORKSPACE}/integration/js/utils/WebdriverSauceLabs.js
+          rm -rf script.sh
+      - uses: EndBug/add-and-commit@v9
+        with:
+          add: '${GITHUB_WORKSPACE}/integration/js/utils/WebdriverSauceLabs.js'
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
```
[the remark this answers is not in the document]
michhyun1: please use a commit hash instead of `v4`

.github/workflows/update-prerun-script.yml
```diff
@@ -0,0 +1,86 @@
...
+          current_version=$(.github/script/get-current-version)
+          PREVIOUS_VERSION=${current_version%%.*} - 1
+      - name: Update pre run script ID
+        run: |
+          echo "Updating integration tests pre-run script ID for previous major version branch"
+          git checkout release-$PREVIOUS_VERSION.x
+          SCRIPT_ID=${{ needs.update-pre-run-script.outputs.script-id }}
+          sed -i -E "s/storage.*'/storage:$SCRIPT_ID'/g" ${GITHUB_WORKSPACE}/integration/js/utils/WebdriverSauceLabs.js
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
```
michhyun1, line 81: please use commit hash

.github/workflows/update-prerun-script.yml
jhaarpan: I need to change it to this cron once I test my changes after the code is merged in
Currently I am going to run it every 10 mins just to check if my code works. After that it will run every two months

.github/workflows/update-prerun-script.yml
jhaarpan: Added

.github/workflows/update-prerun-script.yml
jhaarpan: Done

michhyun1 (approved):

Issue #2210, opened 2022-05-09
Frame size changes when camera is stopped & restarted with video processor applied.


hajime-goshima:
### What happened and what did you expect to happen?

Frame size changes when camera is stopped & restarted with video processor applied.
I was hoping that it would continue to transmit at the frame size set on the device selection screen.
Reproduced in Chime SDK Browser Demo v3.1.0.

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

- Select 720p on the device selection screen and join the meeting.
- Turn on WebRTC Stats Display and confirm that it is transmitting in 720p.

1. Confirm that the camera is transmitting in 720p when it is stopped and started.
2. Verify that the camera is transmitting at 720p when background blur is applied.
3. After that, when the camera is stopped and started, it changes to 480p.

### Amazon Chime SDK for JavaScript version

3.1.0

### What browsers are you seeing the problem on?

Chrome on Ubuntu, Chrome on Windows

### Browser version

101.0.4951.41（Official Build） （64 ビット）

### Meeting and Attendee ID Information.

Meeting ID: 32f1ebaa-71ef-4436-a66a-cafa04662713
Attendee ID: 28b90267-43c9-e682-4946-a9c1203cec6f

### Browser console logs

[console.log](https://github.com/aws/amazon-chime-sdk-js/files/8648271/console.log)

ltrung:
@hajime-goshima Thanks for reporting. I have been able to reproduce and confirmed this was a bug in v3. We will working on a fix.

ltrung:
@hajime-goshima The fix in #2216 has been merged to `main`. Can you try that out and let us know if it resolves the issue on your end?

hajime-goshima:
@ltrung 
I confirmed that it is working perfectly in Chime SDK Browser Demo App.
Thank you for the quick fix.

xuesichao:
Close the issue as the fix has been released.

PR #2216, opened 2022-05-10
Do not lose resolution when toggle transform video

ltrung:
**Issue #2210 :**
In V3, when toggle video with some video transformation (such as background blur), the video resolution will be default from browser (e.g., in Chrome its 480p).
This is because in V3, we would not try to modify if the input device pass to `chooseInputIntrinsicDevice` is a `MediaTrackConstraints` (whereas previously in v2, we will try to add in video resolution, framerate, or audio augmented properpties like `googEchoCancellation`).
However, in `intrinsicDevice` in `DefaultVideoTransformDevice` always returns `MediaTrackConstraints` even when the device is device Id. The code is unnecessary because it is the same logic as in `DefaultDeviceController`.

This is not an issue with `VoiceFocusTransformDevice`, because there we try to override some default constraints such as turn off `echoCancellation` from browser etc.

**Description of changes:**
- Change `intrinsicDevice` in `DefaultVideoTransformDevice` to just return the original device information.
- Delete the code path in `DefaultDeviceController` where we try to re-apply the video resolution and framerate when reusing the current video stream. That should not be needed. This is needed before in case the original input is a media track constraints and we always try to augment video resolution and frame rate.
- Unrelated but update the meeting demo to disable changing video input quality in simulcast case as we only support 720p for simulcast right now. Using other video input resolution might cause issues.

**Testing:**

Test case 1:
- Join a meeting and turn on video and background blur.
- Turn on video stats and confirm your video resolution is correct (default is 540p in the meeting demo).
- Toggle video off then on again.
- Confirm your video resolution is still the same as before.

Test case 2:
- Enable simulcast
- In the preview screen make sure the video input quality is disable and locked to 720p.
- Join and turn on video and verify that your video is 720p.

**Checklist:**

Yes

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -693,6 +693,7 @@ export class DemoMeetingApp
             ) as HTMLSelectElement;
             videoInputQuality.value = '720p';
             this.audioVideo.chooseVideoInputQuality(1280, 720, 15);
+            videoInputQuality.disabled = true;
           }
 
           // `this.primaryExternalMeetingId` may by the join request
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -356,17 +356,7 @@ export default class DefaultDeviceController
 
     // Update device and stream
     this.chosenVideoTransformDevice = device;
-    const newMediaStream = this.activeDevices['video'].stream;
     this.logger.info('video transform device uses previous stream');
-
-    // Input is not a MediaStream. Update constraints
-    if (!(inner as MediaStream).id) {
-      const constraint = inner as MediaTrackConstraints;
-      constraint.width = constraint.width || this.videoInputQualitySettings.videoWidth;
-      constraint.height = constraint.height || this.videoInputQualitySettings.videoHeight;
-      constraint.frameRate = constraint.frameRate || this.videoInputQualitySettings.videoFrameRate;
-      await newMediaStream.getVideoTracks()[0].applyConstraints(constraint);
-    }
 
     // `transformStream` will start processing.
     this.logger.info('apply processors to transform');
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
+- Fix issue where video resolution and framerate changes when toggle video transform.
 
 ## [3.1.0] - 2022-04-07
 
--- a/demos/browser/package-lock.json
+++ b/demos/browser/package-lock.json
@@ -41,7 +41,7 @@
       }
     },
     "../..": {
-      "version": "3.1.0",
+      "version": "3.2.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-crypto/sha256-js": "^2.0.1",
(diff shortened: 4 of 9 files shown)


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -356,18 +356,8 @@ export default class DefaultDeviceController
...
 
-    // Input is not a MediaStream. Update constraints
-    if (!(inner as MediaStream).id) {
-      const constraint = inner as MediaTrackConstraints;
-      constraint.width = constraint.width || this.videoInputQualitySettings.videoWidth;
-      constraint.height = constraint.height || this.videoInputQualitySettings.videoHeight;
-      constraint.frameRate = constraint.frameRate || this.videoInputQualitySettings.videoFrameRate;
-      await newMediaStream.getVideoTracks()[0].applyConstraints(constraint);
-    }
-
```
xuesichao, line 370: Could you remind me what this change is for? I can't remember. The rest changes looks good to me.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -693,6 +693,7 @@ export class DemoMeetingApp
             ) as HTMLSelectElement;
             videoInputQuality.value = '720p';
             this.audioVideo.chooseVideoInputQuality(1280, 720, 15);
+            videoInputQuality.disabled = true;
```
xuesichao, line 696: For understanding: if simulcast is enabled, we want to fix the resolution to 720p for upstreaming and don't want this to be changed. Is this change for this?

ltrung: It is just demo change but its confusing if you allow them to change in the demo to 540p but the video quality in meeting is still 720p

src/devicecontroller/DefaultDeviceController.ts
ltrung: It's in the PR description:
>Delete the code path in DefaultDeviceController where we try to re-apply the video resolution and framerate when reusing the current video stream. That should not be needed. This is needed before in case the original input is a media track constraints and we always try to augment video resolution and frame rate.

demos/browser/app/meetingV2/meetingV2.ts
xuesichao: Cool, same as I thought, thanks for explaining!

src/devicecontroller/DefaultDeviceController.ts
xuesichao: Oh, I forgot to check description. Thanks!

xuesichao (approved):
LGTM!

hensmi-amazon (approved):

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -356,18 +356,8 @@ export default class DefaultDeviceController
 
     // Update device and stream
     this.chosenVideoTransformDevice = device;
-    const newMediaStream = this.activeDevices['video'].stream;
```
simmkyu, line 359: I don't see any change in the device controller unit tests. Do we have any to cover this?

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -356,18 +356,8 @@ export default class DefaultDeviceController
...
     this.logger.info('video transform device uses previous stream');
 
-    // Input is not a MediaStream. Update constraints
-    if (!(inner as MediaStream).id) {
```
simmkyu, line 363: I don't see any change in the device controller unit tests. Do we have any tests for this?

ltrung: The if statement is covered by the standard video transform test and the else is covered by this test https://github.com/aws/amazon-chime-sdk-js/blob/main/test/devicecontroller/DefaultDeviceController.test.ts#L1294. It is really direct related to this code block so I decided not to remove it.

simmkyu (approved):

Issue #1076, opened 2021-02-19
unbindVideoElement API does not remove MediaStream


p-foucht:
**Describe the bug**
The `unbindVideoElement` API does not remove the MediaStream from the video element. Since the `bindVideoElement` is responsible for attaching the stream, the `unbindVideoElement` call should remove it.

This is needed when moving video elements around the DOM. Currently, unbinding and rebinding the same `VideoTile` will create two videos playing the same stream.

**Workaround**
You can workaround this by manually remove the `srcObject` of the video element, or disposing of the video element altogether.

```
const videoEl = document.querySelector(...);
audioVideo.unbindVideoElement(3);
videoEl.srcObject = null;
```

**Fix**
To not break backwards compatability, allow passing some optional config when calling `unbindVideoElement` that will detach the `MediaStream`.

**Platform**
- All platforms

michhyun1:
We want to invlude this change in the next major version update - it would be a breaking change to the current api definition.

michhyun1:
unbindVideoElement now sets the srcObject for the video element to null.

PR #2217, opened 2022-05-11
Update unbindVideoElement

devalevenkatesh:
**Issue #:**
- https://github.com/aws/amazon-chime-sdk-component-library-react/issues/771
- https://github.com/aws/amazon-chime-sdk-js/issues/1076

Libraries or apps depending on JS SDK can leak memory when the video element bound to a video tile is not cleaned correctly. Based on the above issues, the `unbindVideoElement` API is used for cleanup. For example, in [Amazon Chime SDK react component library](https://github.com/aws/amazon-chime-sdk-component-library-react), as part of the cleanup function in React's `useEffect` (or similarly in various apps) we call `unbindVideoElement`  for cleanup. But this does not really dissociate the media stream from the now un-bounded video element. Builders have to explicitly address this on their side. Builders expect that the `unbindVideoElement` should do the opposite of `bindVideoElement` as per #1076 .

In Safari specifically, memory leak is observed if the un-bounded video element's stream is not cleaned up.
Check below graph, the memory keeps increasing steadily in a pagination app where the video elements are binded and un-binded often on page changes.
srcObject-not-null

I have checked multiple approaches and found that with JS SDK v3, we can now disconnect the `srcObject` from the bounded video element as part of un-bind. Below graph shows the memory usage when `srcObject` is set to `null`:
srcObject-null

Note: Check the "page" memory category.

It is also recommended in WebRTC to set this and we could not do so in v2 as we were also cleaning the media stream tracks to avoid crashing in Safari 12.
https://webrtchacks.com/srcobject-intervention/

With V3, we no more clear tracks and hence can fix this. We are still looking for the impact and will gain some more info by sharing this PR with builders and see if there are any issues.

**Since, this is a behavior change, added a `cleanUpVideoElement` optional parameter in `unbindVideoElement` API defaulting to `true`. If builders face any issue and this `srcObject` cleanup is un-intended, builders can call `unbindVideoElement(tileId, false)` to fallback to previous behavior.**

**Description of changes:**
- With V3, we do not cleanup stream tracks bound to a video element and hence leveraged that to cleanup the `srcObject` as part of `unbindVideoElement` API. Note that the `DefaultVideoTile` currently does that internally in [`disconnectVideoStreamFromVideoElement`](https://github.com/aws/amazon-chime-sdk-js/blob/main/src/videotile/DefaultVideoTile.ts#L73-L99) function hence used it.
- This should now address https://github.com/aws/amazon-chime-sdk-js/issues/627#issuecomment-736107449.
- Fixes unit test in `POSTLogger` which was not reaching the 100% threshold.

**Testing:**

Tested with custom pagination sample using React component library. [Sample link](https://github.com/devalevenkatesh/test-amazon-chime-sdk-react-components/tree/main/pagination/without-VideoPriorityBasedPolicy)

1. Started the demo app.
2. Join 6 attendees. (Join first from Chrome and FF)
3. Load the demo in Safari, start recording the memory by going to "Timelines" in network tab.
4. Join the meeting in Safari.
5. Demo automatically changes through the pages so that multiple video elements get binded and un-binded to get the graph over 90 seconds.
6. Compare with current graph and check the significant memory usage is reduced.

Tested the JS SDK video test app to specifically test the `unbindVideoElement` API.
1. Run the `video_test` app under `integration/js/app` folder.
2. Join a meeting.
3. Click `startLocalVideoTile`.
4. Click `bindVideoElement` setting `tileId` to 1 and video element id as `video-0`.
5. Verify that the video shows up.
6. Click `stopLocalVideoTile`.
7. Verify the stream is now `null`.
8. Click `unbindVideoElement` and see that the tile state shows `boundVideoElement` as `null`.
9. Call the `unbindVideoElement` with `cleanUpVideoElement` true and false and verify that when `false` the un-bounded HTML video element still has the stream attached through `srcObject`. When true, the video element does not have the stream attached anymore.

**Checklist:**

Yes

11. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved?
Yes, modified `unbindVideoElement` API. The PR is for review.

12. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved?
No


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 ### Changed
+- Clear srcObject in unbindVideoElement API.
```
devalevenkatesh, line 18: I will clean this up and few docs need update. Please ignore for now.

guides/03_API_Overview.md
```diff
@@ -300,7 +300,7 @@ To stop sharing video with others, call meetingSession.audioVideo.[stopLocalVide
...
 
-To unbind a tile, call meetingSession.audioVideo.[unbindVideoElement(tileId)](https://aws.github.io/amazon-chime-sdk-js/interfaces/audiovideofacade.html#unbindvideoelement).
+To unbind a tile, call meetingSession.audioVideo.[unbindVideoElement(tileId)](https://aws.github.io/amazon-chime-sdk-js/interfaces/audiovideofacade.html#unbindvideoelement). Note that, this will also cleanup the HTML video element's `srcObject` by default. Call `unbindVideoElement(tileId, false)` to avoid the video element cleanup. Check [this PR](https://github.com/aws/amazon-chime-sdk-js/pull/2217) description for more details.
```
simmkyu, line 303: - `Note that, this` => `Note that this` (Remove `,`)
- `cleanup` => `clean up`

src/videotilecontroller/VideoTileController.ts
```diff
@@ -23,10 +23,12 @@ export default interface VideoTileController {
   /**
    * Unbinds the video element from the tile if it exists for the provided tileId.
    * The video tile's bounded video element and that element's width and height are set to null.
+   * @param cleanUpVideoElement By default, the bounded video element's srcObject is also set to null using [[disconnectVideoStreamFromVideoElement]].
```
simmkyu, line 26: nit: I don't think you need to mention a non-public method: `disconnectVideoStreamFromVideoElement`.

src/videotile/DefaultVideoTile.ts
```diff
@@ -62,11 +62,10 @@ export default class DefaultVideoTile implements DevicePixelRatioObserver, Video
   }
 
   /**
-   * Disconnect a video stream to a video element by clearing the srcObject of the video element.
-   * This will also stop all the tracks of the current stream in the srcObject.
+   * Disconnect a video stream from a video element by clearing the srcObject of the video element.
```
simmkyu, line 65: nit: We have used four phrases: `clear`, clear out`, `clean`, and `clean up`. Optionally, we can explicitly state that set `HTMLMediaElement.srcObject` to `null` in some comments.

devalevenkatesh: Fixed thanks for the suggestion. Updated elsewhere to use `HTMLVideoElement.srcObject` to `null` over others.

src/videotilecontroller/VideoTileController.ts
devalevenkatesh: Makes sense, removed and rephrased 🙏

guides/03_API_Overview.md
devalevenkatesh: Updated, thanks!

src/videotilecontroller/DefaultVideoTileController.ts
```diff
@@ -56,8 +57,21 @@ export default class DefaultVideoTileController implements VideoTileController {
...
 
-  unbindVideoElement(tileId: number): void {
-    this.bindVideoElement(tileId, null);
+  unbindVideoElement(tileId: number, cleanUpVideoElement: boolean = true): void {
+    const tile = this.getVideoTile(tileId);
+    if (tile === null) {
+      this.logger.warn(`Ignoring video element unbinding for unknown tile id ${tileId}`);
+      return;
+    }
+    if (cleanUpVideoElement) {
```
ltrung, line 66: nit: Can we remove the else layer?
```
this.logger.info(`Unbinding the video element`);
tile.bindVideoElement(null);
if (cleanUpVideoElement) {
 this.logger.info(`Unbinding and cleaning up the video element`);
 DefaultVideoTile.disconnectVideoStreamFromVideoElement(tile.stateRef().boundVideoElement, false);
}
```

simmkyu: `tile.stateRef().boundVideoElement` will return a wrong object after `tile.bindVideoElement(null)`?

simmkyu:
```js
this.logger.info(`Unbinding the video element`);

const videoElement = tile.stateRef().boundVideoElement;
tile.bindVideoElement(null);
if (cleanUpVideoElement) {
  this.logger.info(`Cleaning up the video element`);
  DefaultVideoTile.disconnectVideoStreamFromVideoElement(videoElement, false);
}
```

devalevenkatesh: Yes, that is correct Kyu. After `tile.bindVideoElement(null)`, the `boundVideoElement` is `null` hence, have to maintain the ref before `bindVideoElement(null)` call.

devalevenkatesh: Also, I just wanted to avoid stale `videoElement` reference cause, if `cleanupVideoElement` is `false` then no use for the maintained ref. Since, that is not the default behavior is it fine to do that way then?

```
const videoElement = tile.stateRef().boundVideoElement;
tile.bindVideoElement(null);
if (cleanUpVideoElement) {
  DefaultVideoTile.disconnectVideoStreamFromVideoElement(videoElement, false);
}
```

devalevenkatesh: Looks like @simmkyu provided the code snippet already, so `videoElement` reference wont be used if `videoElement` is `false`, is that fine 😅 ?

simmkyu: Not using `videoElement` in the `else` case should be fine. I don't have a strong preference between `else` vs without `else`.

devalevenkatesh: I will update as suggested I also do not have a preference, but the suggested looks more cleaner. Thanks!

devalevenkatesh: Updated the code, thanks!

ltrung (approved):

PR #2229, opened 2022-05-13
[Demo] Make sure to only deploy fetch_credentials lambda for messaging session demo

ltrung:
**Description of changes:**
Modify the CloudFormation template to only deploy fetch_credentials lambda for messaging session demo. 

**Testing:**

- Verify that calling fetch_credentials endpoint will result in internal server in meeting demo but works in messaging session demo
- Check CF template to make sure the lambda is not created for meeting demo.

**Checklist:**

### DIFF
--- a/demos/serverless/template.yaml
+++ b/demos/serverless/template.yaml
@@ -32,8 +32,14 @@ Parameters:
     Description: Prefix of S3 bucket to write capture artifacts.  The bucket name will be suffixed with the region.
     Default: ""
     Type: String
+  UseFetchCredentialLambda:
+    Description: Set to true to deploy a lambda to return AWS credentials with limited privileges. This should be disabled in most demo.
+    Default: false
+    Type: String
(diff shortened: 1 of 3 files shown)


demos/serverless/template.yaml
```diff
@@ -32,8 +32,14 @@ Parameters:
     Description: Prefix of S3 bucket to write capture artifacts.  The bucket name will be suffixed with the region.
     Default: ""
     Type: String
+  UseFetchCredentialLambda:
+    Description: Set to true to deploy a lambda to return AWS credentials. This should be disabled in most demo.
```
nainkunal933, line 36: Can you please update this description to specify which AWS credentials? I will prefer it to be something like:
`Set to true to deploy a lambda to return AWS credentials with limited privileges`.

ltrung: Updated.

nainkunal933 (approved):

Issue #2235, opened 2022-05-17
realtimeUnsubscribeFromVolumeIndicator causes a fatal error when attendeeId match not found


schontz:
### What happened and what did you expect to happen?

Calling `realtimeUnsubscribeFromVolumeIndicator` with an `attendeeId` for which there are no subscriptions causes a fatal error that is not catch-able by the application. For example:

```javascript
try {
  // even if 'some-attendee' is a valid ID but if we did not first call realtimeSubscribeToVolumeIndicator,
  // it will fail and leave the meeting
  rt.realtimeUnsubscribeFromVolumeIndicator('some-attendee', () => {})
} catch (error) {
  // never fails, but internally creates an error which is considered fatal
}
```

I would expect that this kind of invalid call would not fail, or only fail in a way that the local code would catch. I would not expect it to cause a fatal error.

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

See above

### Amazon Chime SDK for JavaScript version

3.1.0

### What browsers are you seeing the problem on?

Chrome

### Browser version

101

### Browser console logs

```
[ERROR] [pfid: ***] MeetingsSDK::realtime error: TypeError: Cannot read properties of undefined (reading 'indexOf'): TypeError: Cannot read properties of undefined (reading 'indexOf')
    at t.default.realtimeUnsubscribeFromVolumeIndicator
    at t.default.realtimeUnsubscribeFromVolumeIndicator
    at tl.realtimeUnsubscribeFromVolumeIndicator
```

hensmi-amazon:
PR merged.

PR #2237, opened 2022-05-17
Prevent realtimeUnsubscribeFromVolumeIndicator error when no matching attendeeId is found

schontz:
**Issue #:** #2235

**Description of changes:** 
`realtimeUnsubscribeFromVolumeIndicator` aggressively errors when a matching `attendeeId` cannot be found. This results in being disconnected.

I updated the code to check for a valid array for `attendeeId`. There is no longer anything that would cause the try/catch to occur, so I removed that code to fix coverage.

**Testing:**

The demo application does not exercise this bug. The easiest way to exercise the bug is to call `realtimeUnsubscribeFromVolumeIndicator` with an invalid `attendeeId`.

**Checklist:**

Yes

No

No


ltrung:
Close in favor of #2239

PR #2239, opened 2022-05-17
Prevent realtimeUnsubscribeFromVolumeIndicator error when no matching attendeeId is found

ltrung:
**Issue #2235 :**
`realtimeUnsubscribeFromVolumeIndicator` throw a fatal error when there are no subscriptions for the  given`attendeeId`. This results in being disconnected.

**Description of changes:**
Check that the subscription for the given `attendeeId` exists to prevent an error thrown when calling `indexOf`. There is no longer anything that would cause the try/catch to occur, so I removed that code to fix coverage.

**Testing:**

No

**Checklist:**

### DIFF
--- a/test/realtimecontroller/DefaultRealtimeController.test.ts
+++ b/test/realtimecontroller/DefaultRealtimeController.test.ts
@@ -548,6 +548,16 @@ describe('DefaultRealtimeController', () => {
       expect(state.volumeIndicatorCallbacks).to.equal(expectedResult);
     });
 
+    it('will not error if unsubscribed from a non-exist attendeeId', () => {
+      const callback = (
+        _attendeeId: string,
+        _volume: number | null,
+        _muted: boolean | null,
+        _signalStrength: number | null
+      ): void => {};
+      rt.realtimeUnsubscribeFromVolumeIndicator('bar-attendee', callback);
+    });
+
     it('will tolerate an exception thrown in a volume indicator callback', () => {
       const sentAttendeeId = 'foo-attendee';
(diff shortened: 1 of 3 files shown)


test/realtimecontroller/DefaultRealtimeController.test.ts
```diff
@@ -548,6 +548,16 @@ describe('DefaultRealtimeController', () => {
       expect(state.volumeIndicatorCallbacks).to.equal(expectedResult);
     });
 
+    it('will not error if unsubscribed from a non-exist attendeeId', () => {
+      const callback = (
```
schontz, line 552: nit: You should be able to have a simple callback, i.e., `const callback = () => {}`.

schontz (approved):

test/realtimecontroller/DefaultRealtimeController.test.ts
```diff
@@ -1195,29 +1205,6 @@ describe('DefaultRealtimeController', () => {
       expect(fatal.calledWith(matchUn)).to.be.true;
     });
 
-    it('handles broken volume callbacks', () => {
```
LeviSklaroff, line 1198: Why was this removed?

ltrung: As commented in the PR description, there is nothing in the code that can throw error anymore so this test is no longer valid.
> There is no longer anything that would cause the try/catch to occur, so I removed that code to fix coverage.

LeviSklaroff (approved):

Issue #2242, opened 2022-05-18
chooseAudioOutput doesn't work unless mute is toggled


vsimeikoKG:
### What happened and what did you expect to happen?

Chime SDK doesn't respect selected audio output device unless mute/unmute is toggled.

It worked correctly in the latest Chime v2.x.x version, but after update it seems to be broken.

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

**Steps to Reproduce:**
1) Make sure you have at least 2 audio output devices available.
2) Select non-default audio output device.
3) Join meeting with any number of users.
4) When any other users speaks the audio is coming from the default audio output device.
5) User unmutes and then mutes himself.
6) Audio is coming from the selected audio output now.

**Reference code:**
```js
if (!state.audioElement) {
  document.body.append(audioElement);
}

await meetingSession.audioVideo.listAudioOutputDevices();
await meetingSession.audioVideo.bindAudioElement(audioElement);

if (canSelectAudioOutput()) {
  await meetingSession.audioVideo.chooseAudioOutput(state.audioOutputDeviceID);
}

console.log('SET OUTPUT DEVICE', state.audioOutputDeviceID);
```

### Amazon Chime SDK for JavaScript version

3.2.0

### What browsers are you seeing the problem on?

Chrome

### Browser version

Version 101.0.4951.64 (Official Build) (x86_64)

### Meeting and Attendee ID Information.

Meeting ID - 09bba375-8db6-430f-b542-f01755ca0706
Attendee 1 ID - 01b746da-085c-2cf9-340a-821599d5c795
Attendee 2 ID - 300061a3-5a79-34ff-8c7b-1222492f3b88

### Browser console logs

Logs for the attendee 2.

[logs.txt](https://github.com/aws/amazon-chime-sdk-js/files/8719032/logs.txt)

ltrung:
[...]

PR #2244, opened 2022-05-18
Fix issue where audio output is not updated if chosen before meeting starts

ltrung:
**Issue #2242:**
Calling `chooseAudioOutput` before calling `audioVideo.start` does not work because audio video controller only subscribes to `audioOutputChange` event during connection. This is changed in PR #2163.

**Description of changes:**
Subscribe to `audioOutputChange` in audio mix controller instead of audio video controller (note that audio video controller still handles audio and video input change events as before). This works as audio video controller does not actually handle anything with audio output and just call audio mix controller anyway.

**Testing:**

- Join a meeting in Chrome and select a non-default audio output in the preview screen
- When in a meeting, verify that you can hear audio from the one you selected before

**Checklist:**


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Replace `startVideoInput(null)` and `startAudioInput(null)` with`stopVideoInput` and `stopAudioInput` for video, audio test in meeting readiness checker to stop video, audio input.
 - Replace the deprecated API `getRTCPeerConnectionStats` with `metricsDidReceive` in meeting readiness checker. 
 - Prevent `realtimeUnsubscribeFromVolumeIndicator` from causing a fatal error when there are no subscriptions for the `attendeeId`.
+- Subscribe to `audioOutputDidChange` in audio mix controller to fix the issue where the audio output is not updated 
+  before meeting start.
```
xuesichao, line 21: nit: better not to split into two lines

xuesichao (approved):
Looks good to me. It's good to have test steps in the description section. But I guess it's hard to test and maybe not necessary.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -170,7 +170,9 @@ export default class DefaultAudioVideoController
       this,
       this._logger
     );
-    this._audioMixController = new DefaultAudioMixController(this._logger);
+    const audioMixController = new DefaultAudioMixController(this._logger);
+    this._audioMixController = audioMixController;
+    this._mediaStreamBroker.addMediaStreamBrokerObserver(audioMixController);
```
simmkyu, line 175: Should we also remove the observer in the `destroy` method in the case the customer calls `meetingSession.destory`?

https://github.com/aws/amazon-chime-sdk-js/blob/ae910a680c8d0de920c8bdc639e43f37eb621446/src/meetingsession/DefaultMeetingSession.ts#L135-L137

simmkyu (changes requested):

src/audiovideocontroller/DefaultAudioVideoController.ts
ltrung: Oops. Forget about it. Updated to remove during destroy.

simmkyu (dismissed):

simmkyu (approved):

PR #2246, opened 2022-05-19
Fix Saucelabs screen resolution to display meeting ID

akashuc:
**Issue #:**

**Description of changes:**
Currently Meeting ID is hidden/cropped in saucelabs session. This change adds the screen resolution option to saucelabs config to enable sessions to display meeting ID correctly. 

**Testing:**
No. This change is a part of integration test running on Saucelabs. It can be tested in Saucelabs which now will display the meeting ID without getting cropped in the Saucelabs sessions.

**Checklist:**

Yes

NA

NA

### DIFF
--- a/integration/js/utils/WebdriverSauceLabs.js
+++ b/integration/js/utils/WebdriverSauceLabs.js
@@ -128,6 +128,7 @@ const getSauceLabsConfig = (capabilities) => {
     name: capabilities.name,
     tags: [capabilities.name],
     seleniumVersion: '3.141.59',
+    screenResolution: '1280x960',
     tunnelIdentifier: process.env.JOB_ID,
     ...((capabilities.platform.toUpperCase() !== 'LINUX' &&
       !((capabilities.name).includes('ContentShare'))) && {


xuesichao (approved):

integration/js/utils/WebdriverSauceLabs.js
```diff
@@ -128,6 +128,7 @@ const getSauceLabsConfig = (capabilities) => {
     name: capabilities.name,
     tags: [capabilities.name],
     seleniumVersion: '3.141.59',
+    screenResolution: '1280x960',
```
ltrung, line 131: Does this work in mobile? Have you verified?

ltrung (changes requested):
Make sure to verify that it works in mobile (iOS and Android).

integration/js/utils/WebdriverSauceLabs.js
devalevenkatesh: I think the better fix is to fix the CSS in the meeting demo. Could you please first try that?

akashuc: This method is not used in mobile. Please let me know if I missed something.

ltrung: Ok confirmed that this method is not used for mobile so I am good. @devalevenkatesh Think we can use this for now and fix CSS later (we probably want to own it) :)

devalevenkatesh: Cool, yup agree.

devalevenkatesh (approved):

ltrung (approved):

Issue #2059, opened 2022-02-24
NScaleVideoUplinkBandwidthPolicy spamming the logs potentially causing audio quality issues


justindarc:
### What happened and what did you expect to happen?

The `NScaleVideoUplinkBandwidthPolicy` was seen spamming the logs 7,300 times without taking any action to actually change the scale factor. This seems to indicate some degree of recurring work that is happening under certain conditions far more frequently than it should. This user also experienced sender audio quality loss during this time.

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

Unsure what the steps are to reproduce. Hardware details specific to this user are as follows:

```
{
  "hardware": {
    "manufacturer": "LENOVO",
    "model": "20XHCTO1WW",
    "virtual": false,
    "cpu": {
      "manufacturer": "AMD",
      "brand": "Ryzen 7 PRO 5850U with Radeon Graphics",
      "speed": 1.9,
      "numCores": 16,
      "numPhysicalCores": 8
    },
    "audioDevices": [
      {
        "name": "USB Audio Device",
        "manufacturer": "(Generic USB Audio)",
        "driver": null,
        "default": null,
        "status": "3"
      },
      {
        "name": "USB Audio Device",
        "manufacturer": "(Generic USB Audio)",
        "driver": null,
        "default": null,
        "status": "3"
      },
      {
        "name": "Realtek High Definition Audio",
        "manufacturer": "Realtek",
        "driver": null,
        "default": null,
        "status": "3"
      },
      {
        "name": "DisplayLink USB Audio Adapter",
        "manufacturer": "DisplayLink",
        "driver": null,
        "default": null,
        "status": "3"
      },
      {
        "name": "AMD High Definition Audio Device",
        "manufacturer": "Advanced Micro Devices",
        "driver": null,
        "default": null,
        "status": "3"
      },
      {
        "name": "AMD Audio Device",
        "manufacturer": "AMD",
        "driver": null,
        "default": null,
        "status": "3"
      }
    ],
    "bluetoothDevices": [
      {
[...]

PR #2250, opened 2022-05-24
Ignore NScale uplink policy indexes that haven't changed the number of participants

hensmi-amazon:
**Issue #:** https://github.com/aws/amazon-chime-sdk-js/issues/2059

**Description of changes:** I haven't been able to reproduce the log spam/perf issues given the original report corresponded to a particular remote client. However this change should be low risk as the policy is a small amount of code and its easy to verify the behavior will be equivalent through inspection.

The reset interface addition is so that this doesn't skip recalculation on reconnection if it happens to have the same number of sending videos as before.

**Testing:**
Smoke tested FF and Chrome in three video call. Tested toggle. Tested reconnection

Not beyond sanity checks.

**Checklist:**

y

n

n


test/videouplinkbandwidthpolicy/NScaleVideoUplinkBandwidthPolicy.test.ts
```diff
@@ -149,6 +149,15 @@ describe('NScaleVideoUplinkBandwidthPolicy', () => {
       [26, new DefaultVideoCaptureAndEncodeParameter(320, 192, 15, 600, false, 1)],
     ]);
 
+    it('returns the expected values when no index is provided yet', () => {
+      const expectedParams = new DefaultVideoCaptureAndEncodeParameter(0, 0, 0, 0, false, 1);
```
hensmi-amazon, line 153: This seems strange but isn't actually any change from what this would have been returned. 

Also related: https://github.com/aws/amazon-chime-sdk-js/blob/main/src/task/JoinAndReceiveIndexTask.ts#L121

ltrung:
Could you document in the PR description on why resetting the nscale policy would stop the spamming from happening (code path would be great)?

ltrung (approved):

PR #2276, opened 2022-06-03
Add a workaround for a Safari black screen issue

simmkyu:
**Issue #:**
- N/A

**Description of changes:**
- Add a workaround for a black screen issue (https://bugs.webkit.org/show_bug.cgi?id=241152). A video element with the following two properties shows a black screen in iOS if the element is outside the viewport.
  - muted = true
  - srcObject = null

**Testing:**

Yes.

1 Open iOS Safari in the portrait orientation.
2 Open the [Chime SDK serverless demo](https://github.com/aws/amazon-chime-sdk-js/tree/main/demos/serverless) in the iOS Safari tab.
3 Join a meeting.
4 Choose the camera button to enable a video and scroll down to ensure that a video tile works okay.
5 Scroll top and zoom in to the camera button. Ensure that you do not see video tiles.
  Screen-Shot-2022-05-27-at-10 17 09-AM
6 Choose the camera button to disable the video.
7 Wait for 5 seconds.
8 Choose the camera button to enable the video.
9 Scroll down and confirm that the video tile becomes black.

**Checklist:**

Yes

No

No


src/videotile/DefaultVideoTile.ts
```diff
@@ -58,6 +59,14 @@ export default class DefaultVideoTile implements DevicePixelRatioObserver, Video
 
     if (videoElement.srcObject !== videoStream) {
       videoElement.srcObject = videoStream;
+
+      // In Safari, a hidden video element can show a black screen.
+      // See https://bugs.webkit.org/show_bug.cgi?id=241152 for more information.
+      if (new DefaultBrowserBehavior().requiresVideoPlayWorkaround() && videoElement.paused) {
+        try {
+          videoElement.play();
+        } catch (error) {}
```
ltrung, line 68: Should we log the error in case it happens?

ltrung:
Should we update our known issues?

src/videotile/DefaultVideoTile.ts
simmkyu: Yes. But we have to use `console.error()` because this static method in `DefaultVideoTile` does not have access to a logger. What do you think about using `console.error`?

simmkyu:
> Should we update our known issues?

Yes, I'll update the known issues after merging this PR.

src/videotile/DefaultVideoTile.ts
ltrung: I think console.error for now is fine. It is better to log error to screen. We can move console to logger when there are more places we need to log.

src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -224,6 +230,10 @@ export default class DefaultBrowserBehavior implements BrowserBehavior, Extended
     );
   }
 
+  private isSafari(): boolean {
+    return this.browser.name === 'safari' || this.isIOSSafari();
```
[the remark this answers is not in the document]
simmkyu: Interesting. We have been skipping iOS Web View that also uses the same WebKit engine.

src/videotile/DefaultVideoTile.ts
simmkyu: Actually, `try/catch` block is not needed because we call `HTMLMediaElement.play()` without the `await` keyword.

```js
try {
  videoElement.play();
} catch (error) {
  // Never reached even if play() throws an error
}
```

```js
try {
  await videoElement.play();
} catch (error) {
  // play() throws an error!
}
```

I pushed a revision removing the `try/catch` block.

ltrung (approved):

PR #2285, opened 2022-06-09
Update bootstrap to 5 and move additional options to modal

hensmi-amazon:
**Issue #:** None

**Description of changes:** See title. Being on latest bootstrap just allows us to use latest functionality if we ever need it, I was looking to do some refactoring and figured it'd be nice to do (I'm also due for some spring cleaning).

Switching additional options to modal gives me less guilt when adding to it. I may eventually compartmentalize it in the JS code as well so that it is easy for copy pasting builders to remove.

Screen Shot 2022-06-10 at 9 36 32 AM

Screen Shot 2022-06-09 at 12 51 08 PM

**Testing:**
Tested each app to make sure they look about the same. Pressed a lot of buttons.

Sanity checking all buttons still work.

**Checklist:**

y

y

n

### DIFF
--- a/demos/browser/app/meetingReadinessChecker/meetingReadinessChecker.html
+++ b/demos/browser/app/meetingReadinessChecker/meetingReadinessChecker.html
@@ -19,19 +19,19 @@
       <h1 class="h3 mb-3 font-weight-normal">Amazon Chime SDK Readiness Checker</h1>
       <div class="row mt-3">
         <label for="audio-output block" class="text-center p-2">Speaker</label>
-        <select id="audio-output" class="custom-select" style="width:100%"></select>
+        <select id="audio-output" class="form-select" style="width:100%"></select>
       </div>
       <div class="row mt-3">
         <label for="audio-input block">Microphone</label>
-        <select id="audio-input" class="custom-select" style="width:100%"></select>
+        <select id="audio-input" class="form-select" style="width:100%"></select>
       </div>
       <div class="row mt-3">
         <label for="video-input block">Camera</label>
-        <select id="video-input" class="custom-select" style="width:100%"></select>
+        <select id="video-input" class="form-select" style="width:100%"></select>
       </div>
       <div class="row mt-3">
         <label for="inputRegion block">Media Region</label>
-        <select id="inputRegion" class="custom-select" style="width:100%">
+        <select id="inputRegion" class="form-select" style="width:100%">
           <option value="us-east-1" selected>United States (N. Virginia)</option>
           <option value="ap-northeast-1">Japan (Tokyo)</option>
           <option value="ap-southeast-1">Singapore</option>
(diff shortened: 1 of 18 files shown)


demos/browser/app/meetingReadinessChecker/meetingReadinessChecker.html
```diff
@@ -19,19 +19,19 @@
...
       </div>
       <div class="row mt-3">
         <label for="video-input block">Camera</label>
-        <select id="video-input" class="custom-select" style="width:100%"></select>
+        <select id="video-input" class="form-select" style="width:100%"></select>
       </div>
       <div class="row mt-3">
         <label for="inputRegion block">Media Region</label>
-        <select id="inputRegion" class="custom-select" style="width:100%">
+        <select id="inputRegion" class="form-select" style="width:100%">
```
devalevenkatesh, line 34: Are these class changes coming from bootstrap 5 update?

demos/browser/app/meetingV2/meetingV2.html
```diff
@@ -55,120 +57,147 @@ <h1 class="h3 mb-3 font-weight-normal">Join a meeting</h1>
         </select>
       </div>
       <div class="row mt-3">
-        <div class="col-12">
```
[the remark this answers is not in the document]
hensmi-amazon: Seems like we should just change the defaults to what people are usually clicking :). Also as a tradeoff i was going to add a button that skips the device preview screen in join.

Do we have integration tests which run against main demo? Do they actually click or just marked as checked (i.e. they should still be able to reach the options even if not shown).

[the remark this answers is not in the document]
devalevenkatesh: Yes so our integration tests run against the kitchen sink demo 😓 . As long as they working we should be fine.

integration/js/pages/AppPage.js
```diff
@@ -124,6 +126,18 @@ class AppPage {
     await clickElement(this.driver, authenticateButton);
   }
 
+  async openAdditionalOptions() {
```
devalevenkatesh, line 129: This is an extra step for the integration tests 😅 @ltrung any preference?
 This PR adds a modal for the additional options, for choosing an option do I have to open the additional options and then choose something? If so, these I think are currently choosing options based on tests and for example, EventReporting test does not run as part of PR checks but as part of canaries. Can you once confirm whether all those current option choosing are handled?

Ref: https://github.com/aws/amazon-chime-sdk-js/blob/4a403f6a37ea84d16e06c84dbfa54a8fdcf0fc26/integration/js/steps/AuthenticateUserStep.js#L62-L77

devalevenkatesh: This makes may be UX better for end users but extra work for the integration test. I am fine to choose UX over small extra integration test work.

hensmi-amazon: Yes I checked the options selection buttons function on `AppPage` and confirmed all tests which called them (well, they all just call `AuthenticateUserStep` are wrapped in the open function). Though now I do see that the `selectRegion` is not in the right place as that should be outside additional options.

PR #2290, opened 2022-06-10
Bumps typedoc to 0.22.17 and remove chai-fetch-mock

xuesichao:
**Issue #:**

Solve the dependency resolving issue after upgrading to latest npm version. See details in the description of PR #2283.
I will raise another PR to revert changes in PR #2283.

**Description of changes:**
* Bumps `typedoc` from 0.19.2 to 0.22.17.
* Install new dependency `typedoc-plugin-merge-modules` to merge multiple modules into one when generating docs.
  * We need to generate docs for two parts, one is `./src` (JS SDK source code0 and the other is `./guides` (our guides documentation). Thus we need to specify two entry points and `typedoc` will generate two modules and add a sub-path to documentation URL. 
  * For example, from "/amazon-chime-sdk-js/docs/modules/FAQs.html" to "/amazon-chime-sdk-js/docs/modules/**guides**/FAQs.html". 
  * This plugin helps us to merge multiple modules into one package in the same way as typedoc generated the documentation before.
* Remove `chai-fetch-mock` because it's never used.
* Remove `config/typedoc.json` and move its content to `guides/tsconfig.json`
* Regenerate the docs

**Testing:**

N/A

**Checklist:**

Yes

N/A

N/A


guides/tsconfig.json
```diff
@@ -6,5 +6,21 @@
...
+    "exclude": [
+      "src/signalingprotocol/SignalingProtocol.js",
+      "src/signalingprotocol/SignalingProtocol.d.ts",
+      "src/signalingprotocol/ScreenSignalingProtocol.js",
+      "src/signalingprotocol/ScreenSignalingProtocol.d.ts"
+    ],
+    "excludePrivate": true,
+    "media": "./media",
+  }
+}
```
devalevenkatesh, line 26: nit: Add a new line

script/generate-docs
```diff
@@ -19,4 +19,4 @@ STDERR.puts "Generating docs against #{remote}/#{branch}."
 # Explicitly use the guides/tsconfig.json file, because it includes
 # both the source and docs. The build itself just uses the root tsconfig,
 # which just hits src.
-`node node_modules/typedoc/bin/typedoc --tsconfig guides/tsconfig.json --options config/typedoc.json --gitRemote #{remote} --gitRevision #{branch}`
+`node node_modules/typedoc/bin/typedoc --plugin typedoc-plugin-merge-modules --tsconfig guides/tsconfig.json --gitRemote #{remote} --gitRevision #{branch}`
```
devalevenkatesh, line 22: Just confirming using this plugin the URLs do not change right?

xuesichao: Yes, I mentioned in the PR description

guides/tsconfig.json
xuesichao: Updated

docs/modules/videoprocessor.html
```diff
@@ -319,8 +319,62 @@ <h3>Custom processor usage during meeting preview</h3>
 		<div class="col-4 col-menu menu-sticky-wrap menu-highlight">
 			<nav class="tsd-navigation primary">
 				<ul>
-					<li class="globals  ">
-						<a href="../globals.html"><em>Globals</em></a>
+					<li class=" ">
+						<a href="../modules.html">Exports</a>
```
devalevenkatesh, line 323: seems like global is now a module, does this affect the URL?

xuesichao: It only affects this single page: https://aws.github.io/amazon-chime-sdk-js/globals.html, it will become `https://aws.github.io/amazon-chime-sdk-js/modules.html`. For rest of the guide, class, interface, the URL remains the same.

devalevenkatesh: Can you add a custom globals.html to generated docs folder and mention to check modules.html instead? Cause this means its breaking for documentation, and would be good to let builders know to go to right place.

ltrung: Is it easier to just rename modules.html to globals.html in our post generation script?

xuesichao: I think we can do that. The git change is actually to rename the `globals.html` to `modules.html`. We can revert this for sure.

xuesichao: oh wait, basically all `.html` files have a link to `modules.html`, we also need to replace these links.

xuesichao: Alright, I think the best solution is to duplicate the `modules.html` and rename it to `global.html`. So we keep `globals.html` and `modules.html` at the same time. So we don't need to update those links, and they can still link to `modules.html`.

I will go ahead with this solution. Let me know if you have any concern.

devalevenkatesh (approved):

PR #2292, opened 2022-06-13
Add support for configuring video send codec

hensmi-amazon:
**Issue #:** None, though issues with iOS H.264  indicate an API like this would be useful.

**Description of changes:**  Add APIs `setVideoCodecSendPreferences` and `setContentShareVideoCodecPreferences` to allow configuration of codec being used to send. See the [JS SDK guide](https://aws.github.io/amazon-chime-sdk-js/modules/videocodecs.html) for more details.

**Testing:**

Smoke tested 3 video calls that send video codec was expected. Through some backend modifications, tested fallback by using a codec which was not supported in Firefox. When firefox joined, the sender switched back to vp8.

* Select an option other then 'Meeting Default' for video send codec.
* Confirm in chrome://webrtc-internals/ that the send codec is as expected, e.g.
Screen Shot 2022-06-13 at 9 45 08 AM

* The fallback logic is a bit trick to test unless you have an android device without H.264 decode support. However you can also have a load test client join, since they are using an older version of Chrome that doesn't use the FFMpeg decoder by default. In that case, if you select H.264 (not Meeting default) you should see it switch to VP8 when the other client joins.

**Checklist:**

Y

N

N


guides/19_Video_Codecs.md
```diff
@@ -0,0 +1,71 @@
...
+* Some older Android devices do not have hardware H.264 codecs and therefore will not be able to encode or decode H.264.
+* Some custom or old Chrome builds may not have H.264 enabled.
+
+If your application encounters significant end-user issues with not being able to receive H.264 (which will currently be replaced by a static VP8 video stream with an error message), you may want to consider calling [`setVideoCodecSendPreferences`](https://aws.github.io/amazon-chime-sdk-js/interfaces/audiovideocontrollerfacade.html#setvideocodecsendpreferences) with either VP8 only or VP8 as a fallback.
+
+#### Avoiding Browser/Device Specific Bugs
+Occasionally, Browser or OS updates (particularily on mobile) may bring along bugs and issues related to specific codecs (particularily when hardware is in use). In this case, you may not want to wait on JS SDK versions to manually change the default send codec. For example, if you detect that iOS clients are sending significantly lower bitrates then expected, it may be due to a recent release impacting hardware encoders. You can for clients to fallback to VP8, e.g.:
```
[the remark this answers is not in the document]
hensmi-amazon: Oops typo. I can change 'fallback' to 'choose' though technically you are just falling back manually at application level (different then the 'fallback' built into the api) :).

src/task/ReceiveVideoStreamIndexTask.ts
```diff
@@ -206,6 +213,43 @@ export default class ReceiveVideoStreamIndexTask
...
+          newMeetingSupportedVideoSendCodecPreferences.push(capability);
+          break;
+        }
+      }
+      // We don't need to renegotiate if we aren't sending a codec that is no longer supported in the call.
+      if (
+        this.context.currentVideoSendCodec &&
+        capability.equals(this.context.currentVideoSendCodec)
+      ) {
+        willNeedUpdate = true;
```
ltrung, line 237: Not sure if this is correct but I don't think the code here matches the comment.

src/sdp/SDP.ts
```diff
@@ -658,4 +659,169 @@ export default class SDP {
...
+        const priority1 = preferences.findIndex(capability => {
+          return codecNamesToPayloadTypes.get(capability.codecName) === name1;
+        });
+        const priority2 = preferences.findIndex(capability => {
+          return codecNamesToPayloadTypes.get(capability.codecName) === name2;
+        });
+        return priority1 - priority2;
+      }
+    );
+    mline.splice(3, 0, ...orderedPreferedPayloadTypes.values());
```
ltrung, line 756: Could we put a comment on why it starts from 3 :D You have something similar in the function below.

src/sdp/SDP.ts
```diff
@@ -658,4 +659,169 @@ export default class SDP {
...
+      // E.g. 'a=rtpmap:125 H264/90000'
+      const payloadMatch = /^a=rtpmap:([0-9]+)\s/.exec(line); // Get the payload type
+      if (
+        payloadMatch === null ||
+        payloadMatch.length < 2 ||
+        payloadMatch[1] !== highestPriorityPayloadType
+      ) {
+        continue;
+      }
+      const lineTokens = line.split(' '); // Previous check gurantees this to be valid
```
ltrung, line 799: nit: typo: `gurantees`

src/sdp/SDP.ts
```diff
@@ -658,4 +659,169 @@ export default class SDP {
...
+    );
+    mline.splice(3, 0, ...orderedPreferedPayloadTypes.values());
+    lines[0] = mline.join(' ');
+
+    // Note that nothing in the RFCs require `a=rtpmap` lines to be reordered
+
+    return lines.join(SDP.CRLF) + SDP.CRLF;
+  }
+
+  // Returns undefined if there is no video send section or no codecs in the send section
```
ltrung, line 764: Could we document what it actually returns?  There is no interface for this anymore so I think we should document these two public functions in more details.

src/sdp/SDP.ts
```diff
@@ -658,4 +659,169 @@ export default class SDP {
...
+    if (mlineTokens.length < 4) {
+      return undefined;
+    }
+    const highestPriorityPayloadType = mlineTokens[3];
+    let highestPriorityCodecName: string = undefined;
+    let highestPriorityClockRate: string = undefined;
+    let highestPriorityFmtpLine: string = undefined;
+    for (const line of lines) {
+      if (highestPriorityCodecName !== undefined) {
+        continue; // Already found it
```
ltrung, line 788: Shouldn;t we just break after we find highest priority codec then?

ltrung (commented):
Could you update the PR description (you can just copy the CHANGELOG) ?

guides/19_Video_Codecs.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+This implies the following:
+* Ignoring other restrictions, the sending client will transmit video into the meeting using the H.264 video codec.
+* The sending client *will not* fallback to any other codec if another attendee joins without an H.264 decoder (which may occur on some browsers, see notes below). That receiving attendee will not receive this video.
+* If the sending client does not support encoding H.264, then it will fall back to any video codec it supports, that is also supported by the meeting. Currently that would be VP8 only.
+
+For more complex fallback behavior, consider the following example:
+```
+this.meetingSession.audioVideo.setVideoCodecSendPreferences([VideoCodecCapability.h264(), VideoCodecCapability.vp8()]);
```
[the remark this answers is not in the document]
hensmi-amazon: Yes, I mention this in the next section.

guides/19_Video_Codecs.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+This implies the following:
+* Ignoring other restrictions, the sending client will transmit video into the meeting using the H.264 video codec.
+* The sending client *will* fallback to VP8 other codec if another attendee joins with H.264 decode support, but does have VP8 decode support.
+* If the sending client does not support encoding H.264, then it will fall back to VP8 (not technically a change in behavior, but will be more relevant when accounting for additional codecs).
+
+A similar API [`ContentShareControllerFacade.setContentShareVideoCodecPreferences`](https://aws.github.io/amazon-chime-sdk-js/interfaces/contentsharecontrollerfacade.html#setcontentsharevideocodecpreferences) exists to control content share video send codecs and it is used in identical fashion, e.g.:
```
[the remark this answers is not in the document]
hensmi-amazon: It is extremely relevant even with just VP8 and H.264. For example safari screen share has issue with H.264 encoder at high resolution, and additionally technically the only 'safe' way to do >720p is with VP8 (H.264 technically only requires decoders to support 720p max). This is more critical with future codecs, some which are better at screen share then others :).

src/sdp/SDP.ts
hensmi-amazon: sure!

src/sdp/SDP.ts
hensmi-amazon: sure, originally had this in a `forEach` function :)

src/sdp/VideoCodecCapability.ts
```diff
@@ -0,0 +1,71 @@
...
+      mimeType: 'video/H264',
+      sdpFmtpLine: 'level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42e01f',
+    });
+  }
+
+  /**
+   * Returns the configuration of H.264 recommended by the SDK
+   */
+  static h264(): VideoCodecCapability {
+    return this.h264ConstrainedBaselineProfile();
```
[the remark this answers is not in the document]
hensmi-amazon: it refers to the `42e01f` . https://en.wikipedia.org/wiki/Advanced_Video_Coding#Profiles . I don't really go into detail here because there is only one option (and they can also just call `h264`).

src/task/ReceiveVideoStreamIndexTask.ts
hensmi-amazon: Will remove the double negative. Comment was referring to the non-existent `else` branch.

src/task/ReceiveVideoStreamIndexTask.ts
```diff
@@ -206,6 +213,43 @@ export default class ReceiveVideoStreamIndexTask
...
+      }
+    }
+
+    if (newMeetingSupportedVideoSendCodecPreferences.length > 0) {
```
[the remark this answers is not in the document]
hensmi-amazon: imo the variable name is long to the point where is seems a bit clunky to read but sure.

guides/19_Video_Codecs.md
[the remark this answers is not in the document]
ltrung: Ok, but is the 90% time ppl will just expect the same codec for both? If so, we can make content share to copy the setting from the sharer and they can override it if they want :) If you think most of the time ppl will set them up differently between video and content share then its fine :D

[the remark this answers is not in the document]
hensmi-amazon: I do not think it would be 90%. Given the possible large impact on CPU/video quality I would like builders to make each choice explicitly. Again the performance/quality impact of different codecs varies **significantly** between screen content and camera content. This is more of an advanced API anyways; I just don't want to make builders shoot themselves in the foot just so I can avoid having them call 2 functions.

ltrung (approved):

ltrung (approved):

PR #2294, opened 2022-06-13
Add client side APIs and implementation for server side network adaption

hensmi-amazon:
**Issue #:** None.

**Description of changes:** See docstrings and guide.

**Testing:**

Various adhoc and experimental testing. A snippet is below:

In both of these tests, the network is unconstrained for 60 seconds, then constrained to 1Mbps for 60 seconds, and then unconstrained.

First we can look at the current policy:

Recovery from the initial constraint occurs with prolonged loss, mostly due to a lack of limit on retransmissions and pacing, which were also added with this change. Eventually the policy does flatten out at 3 videos, but then tries to subscribe to a 4th while under constraint, leading to heavy loss.and a revert to 2 videos. 

Screen Shot 2022-06-07 at 4 05 12 PM

With the changes in place the same scenario looks like the following.

Screen Shot 2022-06-07 at 3 58 37 PM

In this it can still be seen that there is loss on constraint, and future iterations will improve this response time, but it can also be seen that there is no spike of retransmissions as before, the backend is actually using the REMB to limit retransmissions, even if there is a delay in downgrade/pausing from the client. Under constraint and recovery was can see the bitrate closely follows the bandwidth estimation, follows a reasonably tight AIMD pattern without loss under the constraint, and exponentially grows when the constraint is lifted.

* Select `Use Priority-Based Downlink Policy` in additional options in supporting browser (e.g. Chrome).
* Select `Server Side Network Adaption` -> `Enable Bandwidth Probing`.
* Smoke test 1-1 and multi-video calls. Use network link conditioner or similar to reduce downlink network, and allow recovery.

**Checklist:**

y

n

n


src/task/JoinAndReceiveIndexTask.ts
```diff
@@ -111,9 +112,19 @@ export default class JoinAndReceiveIndexTask extends BaseTask {
       this.context.previousSdpOffer = null;
       this.context.serverSupportsCompression = false;
 
-      this.context.signalingClient.join(
-        new SignalingClientJoin(this.context.meetingSessionConfiguration.applicationMetadata)
+      const join = new SignalingClientJoin(
+        this.context.meetingSessionConfiguration.applicationMetadata
       );
+      if (this.context.videoDownlinkBandwidthPolicy.wantsServerSideNetworkAdaption !== undefined) {
+        switch (this.context.videoDownlinkBandwidthPolicy.wantsServerSideNetworkAdaption()) {
```
ltrung, line 119: Will the other case here is serverSideNetworkAdaptation? If that is the case, should we just make one property in SignalingJoin constructor for `wantsServerSideNetworkAdaption` and pass through the properties there.
A bit confuse why we need to break out these enum values into 2 different boolean props.

src/videodownlinkbandwidthpolicy/VideoDownlinkBandwidthPolicy.ts
```diff
@@ -54,6 +56,11 @@ export default interface VideoDownlinkBandwidthPolicy {
    */
   removeObserver?(observer: VideoDownlinkObserver): void;
 
+  /**
+   * (Optional) Call and observer function on all added `VideoDownlinkObserver`
+   */
+  forEachObserver?(observerFunc: (observer: VideoDownlinkObserver) => void): void;
```
ltrung, line 62: Why do we need to add this to the interface? Do you call it somewhere outside of the video priority class?

src/task/SubscribeAndReceiveSubscribeAckTask.ts
```diff
@@ -176,22 +192,73 @@ export default class SubscribeAndReceiveSubscribeAckTask extends BaseTask {
...
+      const configuration = new SignalingClientVideoSubscriptionConfiguration();
+      const mid = attendeeIdToMid.get(preference.attendeeId);
+      if (mid === undefined) {
+        this.context.logger.warn(`Could not find MID for attendee ID: ${preference.attendeeId}`);
+        continue;
+      }
+      configuration.mid = mid;
+      configuration.attendeeId = preference.attendeeId;
+      configuration.groupId = attendeeIdToGroupId.get(preference.attendeeId);
+      configuration.priority = Number.MAX_SAFE_INTEGER - preference.priority;
```
ltrung, line 255: Maybe we need a comment in the proto or here to explain the logic of priority in the backend?

src/task/JoinAndReceiveIndexTask.ts
hensmi-amazon: Ya...I've been meaning to do that (originally was using bools on the config as well), just a bit awkward because I need to deprecate the previous flags (should be fine to do before release :) ).

src/task/SubscribeAndReceiveSubscribeAckTask.ts
hensmi-amazon: Sure. My intention was to avoid the classic "This video was set to high priority...by setting it to a low value" statement which always confuses people lol.

src/videodownlinkbandwidthpolicy/VideoDownlinkBandwidthPolicy.ts
hensmi-amazon: Oops that was for future change, will remove.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -1029,7 +1033,21 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
...
     return this.videoPreferences || this.defaultVideoPreferences;
   }
+
+  wantsServerSideNetworkAdaption(): ServerSideNetworkAdaption {
+    return this.videoPriorityBasedPolicyConfig.serverSideNetworkAdaption;
+  }
+
+  getVideoPreferences(): VideoPreferences {
+    let preferences = this.getCurrentVideoPreferences();
+    if (!preferences) {
```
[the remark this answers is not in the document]
hensmi-amazon: The first subscribe (audio only) is oddly timed and seems to start before this class receives an index. Will add comment.

src/task/JoinAndReceiveIndexTask.ts
```diff
@@ -75,7 +80,21 @@ export default class JoinAndReceiveIndexTask extends BaseTask {
             }
 
             context.serverSupportsCompression = joinAckFrame?.wantsCompressedSdp;
-
+            if (
+              joinAckFrame?.serverSideNetworkAdaptionOverride !== undefined &&
+              context.videoDownlinkBandwidthPolicy.overrideServerSideNetworkAdaption !== undefined
+            ) {
+              const serverSideNetworkAdaptionOverride: SdkServerSideNetworkAdaption =
+                joinAckFrame?.serverSideNetworkAdaptionOverride;
```
ltrung, line 88: nit: No need for ? right since it should be covered in the if statement.

src/task/SubscribeAndReceiveSubscribeAckTask.ts
ltrung: But if it is new in the backend, wouldn't it better to be consistent?

src/signalingclient/DefaultSignalingClient.ts
```diff
@@ -107,13 +109,31 @@ export default class DefaultSignalingClient implements SignalingClient {
...
+    joinFrame.serverSideNetworkAdaptionOverrideSupport = settings.serverSideNetworkAdaptionOverrideSupport.map(
+      this.convertServerSideNetworkAdaptionEnumToSignaled
+    );
     const message = SdkSignalFrame.create();
     message.type = SdkSignalFrame.Type.JOIN;
     message.join = joinFrame;
     this.sendMessage(message);
   }
 
+  private convertServerSideNetworkAdaptionEnumToSignaled(
```
ltrung, line 124: nit: Seem like this should be a helper function somewhere

src/videodownlinkbandwidthpolicy/VideoDownlinkBandwidthPolicy.ts
```diff
@@ -62,4 +64,29 @@ export default interface VideoDownlinkBandwidthPolicy {
...
+  /**
+   * Values which this policy supports being overriden to. The server may use this to transition
+   * clients to values of `ServerSideNetworkAdaption`. Override this function
+   * to return an empty list to disable any server override attempts.
+   */
+  serverSideNetworkAdaptionOverrideSupport?(): ServerSideNetworkAdaption[];
+
+  /**
+   * Dynamically switch any client behavior to the adaption type which this policy indicated support
+   * to from `supportsServerSideNetworkAdaptionOverride`.
```
ltrung, line 83: typo: `serverSideNetworkAdaptionOverrideSupport` ?

guides/11_Priority_Based_Downlink_Policy.md
```diff
@@ -113,6 +114,29 @@ The following presets are provided:
...
 * *Stable Network Preset*: `VideoPriorityBasedPolicyConfig.StableNetworkPreset` configures slow network issue response delay and quick recovery delay. Select stable network preset when the network is expected to stay stable, for example in unified communications client or Ethernet connections. It will try to 'wait-out' network downturns in case they are a temporary spike.
 
+### Enabling Server Side Network Adaption Features <span id="server-side-network-adaption"><span>
+
+`VideoPriorityBasedPolicy` has support for configuring additional features on the backend. These features are currently opt-in but may become defaults in the future. To enable pacing and probing on the existing receive side estimator, and disable the policy based stream probing which can often oversubscribe over network constraints and lead to video freezes, set the following on `VideoPriorityBasedPolicyConfig` before injecting:
```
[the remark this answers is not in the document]
hensmi-amazon: I added a note about the fact the value may be override if set to default, but I didn't not bother mentioning the downlink policy functions since those are largely internal. They work fine if not set and if a builder really wants to fiddle around there are docstrings :).

src/signalingclient/DefaultSignalingClient.ts
hensmi-amazon: moved!

src/task/JoinAndReceiveIndexTask.ts
hensmi-amazon: Fixed! would have been impossible to get to 100% coverage if left as is :P

src/task/SubscribeAndReceiveSubscribeAckTask.ts
hensmi-amazon: Discussed offline.

XHatan (approved):
Nice work ! Should probably resolve integration test failure

ltrung (approved):

PR #2323, opened 2022-06-23
Fix VF bug when toggling mute and then quickly changing device

LeviSklaroff:
**Issue #:**
- When device was muted as a normal device and is then change to null and unmute then swapped to a valid device Amazon Voice Focus stops working
- When device is muted/unmuted and then swapped shortly after Amazon Voice Focus stops working

**Description of changes:**
- Keep track of mute state in device controller and apply to newly selected transform devices, to prevent the mute state from being made out of sync. 
- Added new behavior to meeting demo that will prevent resources from being cleaned up after `audioVideo.stop()` this allows for reuse of meeting components
**Testing:**
- Tested that VF will work under the following condtions:
     - When device is muted when it's null and then swapped to a valid device Amazon Voice Focus stops working
     - When device is muted/unmuted and then swapped shortly after Amazon Voice Focus stops working

No

**Checklist:**

### DIFF
--- a/src/devicecontroller/DefaultDeviceController.ts
+++ b/src/devicecontroller/DefaultDeviceController.ts
@@ -75,6 +75,8 @@ export default class DefaultDeviceController
   private audioInputTaskQueue: PromiseQueue = new PromiseQueue();
   private videoInputTaskQueue: PromiseQueue = new PromiseQueue();
 
+  private muted: boolean = false;
+
   // This handles the dispatch of `mute` and `unmute` events from audio tracks.
   // There's a bit of a semantic mismatch here if input streams allow individual component tracks to be muted,
   // but addressing that gap is not feasible in our stream-oriented world.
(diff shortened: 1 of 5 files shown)


src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -692,6 +695,7 @@ export default class DefaultDeviceController
     if (!audioDevice) {
       return;
     }
+    this.muted = !enabled;
```
ltrung, line 698: is a better fix to make this one async?

ltrung (commented):
Could you add more details about the race condition part?

src/devicecontroller/DefaultDeviceController.ts
LeviSklaroff: So it's actually not entirely a race condition, it's also just the mute state not being kept track of. There’s two cases where it's triggered, one is when the device is changed while it's in the middle of muting the voice focus transform. I believe this is because it creates a new transform device with the old transform device's mute state. This could be fixed through making this async or even keeping the promise and awaiting it before starting a device. The other case is when the device is set to null or an intrinsic device in this case if the device is swapped back and there was a change in the mute state before swapping, the transform will be out of sync with the mute state. This case can not be fixed by making it async.

PR #2344, opened 2022-06-30
Add src folder to include the source code in node module

michhyun1:
**Issue #:**

**Description of changes:**

Add the source folder so that the src folder is included in node package that is published to npm.

This will increase the initial bundle size, but technically will only affect builders applications should they choose to enable the source maps for this module.
 
Builders can specify whether or not to include the source map when they build their apps. Meaning, we can add the src folder to our npm module, but it doesnt mean that the builders applications will need to load the src code for our amazon-chime-sdk-js module as long as they specify to 'GENERATE_SOURCEMAP=false' (for Create react app applications).

I think it'd be good, since we are open source. It will allow builders to debug much more easily and perhaps even make more contributions to this repository.

Before (not including src folder):
```
npm notice package size:  534.9 kB
npm notice unpacked size: 3.0 MB
```

After (including src folder):
```
npm notice package size:  769.5 kB
npm notice unpacked size: 4.8 MB
```

**Testing:**

npm run build, npm pack, compare the package.

Install that tarball to a new Create React App application and run the app, see that you can see the source code for amazon-chime-sdk-js for debugging purposes. Set a breakpoint on the JS SDK source code and see that it works.

Disable sourcemaps for amazon-chime-sdk-js by setting the GERNATE_SOURCEMAPS=false property when building my create react application. Verify in the source tab of the developer tools that source map and src code for amazon-chime-sdk-js is not bundled in my application build.

**Checklist:**

yes

no

no


xuesichao (approved):

michhyun1:
> That is quite an increase in term of size. Are we sure it is fine? We already have some GitHub issues about our package size.

Yeah, it is a large increase, and I'm not sure if this is still acceptable. @simmkyu can you provide any guidance on this package size?

michhyun1:
Builders can specify whether or not to include the source map when they build their apps. Meaning, we can add the src folder to our npm module, but it doesnt mean that the builders applications will need to load the src code for our amazon-chime-sdk-js module as long as they specify to 'GENERATE_SOURCEMAP=false'.

I think it'd be good, since we are open source. It will allow builders to debug much more easily and perhaps even make more contributions to this repository. @ltrung @simmkyu What do you think?

michhyun1:
We should include this in V4 - but perhaps not now.

PR #2354, opened 2022-07-07
Fix meeting demo join in older Safari

ltrung:
**Issue #:**
In Safari 15.3 or older, joining the meeting demo will result in a crash due to this recent change here https://github.com/aws/amazon-chime-sdk-js/blob/main/demos/browser/app/meetingV2/meetingV2.ts. 
Submitter event is not supported before Safari 15.4 

**Description of changes:**

Move the quick join event from submit to click (does not make sense to have 2 button trigger submit in a form anyway).

**Testing:**

Join meeting in previous Safari 15.3 or older version and verify there is no crash. 
Also test quick join and verify that it works (default audio input is selected when joining meeting).

**Checklist:**

2. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? N/A

3. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved? N/A

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -628,149 +628,14 @@ export class DemoMeetingApp
       (document.getElementById('primary-meeting-external-id') as HTMLInputElement).value = "";
     });
 
+    document.getElementById('quick-join').addEventListener('click', e => {
+      e.preventDefault();
+      this.redirectFromAuthentication(true);
+    });
+
     document.getElementById('form-authenticate').addEventListener('submit', e => {
       e.preventDefault();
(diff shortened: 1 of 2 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -3674,6 +3539,140 @@ export class DemoMeetingApp
     }
   }
 
+  private redirectFromAuthentication(quickjoin: boolean = false): void {
```
nainkunal933, line 3542: I am bit confused as to why this method is named `redirectFromAuthentication`, can you please explain? It seems a bit vague. 

This method basically changes the application flow from `flow-authenticate` to `flow-devices` or `flow-meeting`.

PR #2380, opened 2022-07-28
Fix security vulnerabilities

devalevenkatesh:
**Issue #:**

**Description of changes:**
- Tried updating the security deps but somehow ran into `typescript` update issues.
- `npm audit fix` fixed all fine.
- Test running the demo locally with various video APIs.
- Also updated the referenced local `amazon-chime-sdk-js` dependency.
- Remove `markdown-it` from root and update `integration/js/app` directory.
- Move `@types/dom-mediacapture-record` from JS SDK root to browser demo
   Browser demo uses `MediaRecorder` hence needs `@types/dom-mediacapture-record`
   but the dependency was installed in JS SDK root, hence, move it to browser demo.
- Removed `test/dommediarecordermock/DOMMediaRecorderMock.ts` file as it was not referenced anywhere in the repo.

**Testing:**

Tested running the `integration/js/app` video test app locally with various APIs and they all work as expected.

**Checklist:**

NA

NA

NA


integration/js/app/package-lock.json
```diff
@@ -55,6 +55,7 @@
         "fs-extra": "^8.1.0",
         "git-rev-sync": "^3.0.2",
         "longjohn": "^0.2.12",
+        "markdown-it": "^13.0.1",
```
simmkyu, line 58: This PR seems to add a new dependency `markdown-it` to the integration test. Can you confirm?

devalevenkatesh: I think this came when I moved the `markedown-it` from `dependencies` to `devDependencies` in last PR. Checked and removing this as we do not need in JS SDK but use it in `demos/browser` only. Thanks for the catch!

simmkyu (dismissed):

integration/js/app/package-lock.json
devalevenkatesh: Updated.

simmkyu (approved):

ltrung (approved):

PR #2388, opened 2022-08-03
Fix video play abort in Safari 15.5 / 15.6 (latest)

devalevenkatesh:
**Issue #:**
From browser compatibility test failures we realized that video test is failing in Safari macOS latest. Turning the video ON results into a fatal error.
```
[Error] Fatal error: this was going to be caught, but should not have been thrown. – AbortError: The operation was aborted.
AbortError: The operation was aborted.
    fatal (Prod:2:986193)
    fatal
    (anonymous function) (Prod:2:985999)
[Log] [DEMO] AbortError: The operation was aborted. (Prod, line 2)
[Error] Unhandled Promise Rejection: AbortError: The operation was aborted.
    (anonymous function)
    rejectPromise
```

Browser compatibility test failure:
```
Generating failure report
=========================================
Run configuration:
 Client 1 : MAC  latest
Failure details:
 [Safari] Check the local video to be: video
=========================================
```

Failed test runs:
- https://github.com/aws/amazon-chime-sdk-js/actions/runs/2769578869
- https://github.com/aws/amazon-chime-sdk-js/runs/7591695192?check_suite_focus=true
- https://github.com/aws/amazon-chime-sdk-js/actions/runs/2760845837

**Root cause**
For a WebKit bug we had added a workaround to play a video element which is out of viewport.
PR: https://github.com/aws/amazon-chime-sdk-js/pull/2276
WebKit Bug: https://bugs.webkit.org/show_bug.cgi?id=241152
```
// In Safari, a hidden video element can show a black screen.
// See https://bugs.webkit.org/show_bug.cgi?id=241152 for more information.
if (new DefaultBrowserBehavior().requiresVideoPlayWorkaround() && videoElement.paused) {
  videoElement.play();
}
```

If I comment this code from [here](https://github.com/aws/amazon-chime-sdk-js/blob/main/src/videotile/DefaultVideoTile.ts#L63-L67), there is no more `AbortError`.

I tested on macOS Safari (15.5) and iOS with Safari (15.6) and both work fine when tested with the test steps in #2276, hence, reverting the added workaround.

**Description of changes:**
- Revert the workaround commit. It was added for another black video tile issue reported by a customer through support. That issue was not fixed with #2276  workaround and was a CSS issue reported in Webkit bug: https://bugs.webkit.org/show_bug.cgi?id=241303
- We have not had other issues from builder where the workaround #2276 is needed.

**Testing:**

Testing using the JS SDK browser demo app.
- Join a meeting with fatal error checked under additional options.
- Turn ON the video.
- There should be no fatal error.

Also tested the test steps in #2276 and see that the video is coming up fine after zooming in so that the video element when turned ON, it is out of the viewport. When scrolled down to get the video in viewport, I can see the video element fine.

**Checklist:**

Yes

NA

NA


xuesichao:
How does the Safari video tile workaround cause the `The operation was aborted` error?

devalevenkatesh:
> How does the Safari video tile workaround cause the `The operation was aborted` error?

It is due the `videoElement.play` API that was added in the #2276 . There is no root cause from logs why that happens but after lot of trial and error I found out AbortError around HTML video element play APIs reported by lot of people generally. Plus, the abort error happened when the video is toggle to ON and thus found that reverting that workaround mitigates the issue. We can add the workaround later when we get those reports on black video tile if needed.

devalevenkatesh:
Closing this PR looks like the black video tile now comes back and is not fixed if we do not use videoElement.play API when multiple attendees are in a meeting. With single attendee, the video element works fine without black video tile if out of viewport, but if remote attendees are in a meeting then that results into a black video tile if videoElement.play workaround is removed so have to look more.

PR #2390, opened 2022-08-05
Add sending audio failure and recovery events

avinashmidathada:
**Issue #:** None

**Description of changes:**

*What?*

Added two new meeting events: `sendingAudioFailed` to indicate a failure in sending audio packets out and `sendingAudioRecovered` to indicate recovery after a `sendingAudioFailed` event. We rely on the `packetsSent` metric from RTCPeerConnection's getStats() method to trigger these events, by tracking the no. of times packetsSent is 0 consecutively and storing it in `connectionHealthData` in a new field called `consecutiveStatsWithNoAudioPacketsSent`.

The events are triggered using a new `SendingAudioFailureConnectionHealthPolicy`.  It introduces two new properties `sendingAudioFailureSamplesToConsider` and `sendingAudioFailureInitialWaitTimeMs` inside `ConnectionHealthPolicyConfiguration` for configuring it's sensitivity for triggering the events. The policy also closely resembles the existing `UnusableAudioWarningConnectionHealthPolicy` having a similar cool down time and maximum warnings logic. I have also refactored the code in MonitorTask to create a new `applyHealthPolicy()` method to reduce redundancy in the health policy evaluation.

Also, added a new device to microphone drop down called `No Audio` which is used for integration testing.

*Why?*

The `sendingAudioFailed` event will serve both as a diagnosis tool as it's ingested into our event ingestion flow but more importantly as a callback for builders to notify the user or to prompt them to take an action such as a device change or application restart. The `sendingAudioRecovered` event can be used to notify user that the sending audio is now back to normal.

**Testing:**

- Added integration tests to verify that the events are being published as expected.

1. Run Browser Demo App in Chrome or Safari (doesn't work in Firefox)
2. Connect to a meeting with `Enable event reporting` checked and open the Developer Tools Console
3. After joining, change microphone input to No Audio by clicking the mic drop down
4. Wait for about 3 secs and verify that there's a log starting with `Received an event: ` and the name of the event is `sendingAudioFailed`
5. Change the microphone back to the previous one or any functioning audio input
6. Wait 1 sec and verify that another similar log is printed, this time with `sendingAudioRecovered`

**Checklist:**

Yes

N/A

7. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved?
N/A


CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
hensmi-amazon, line 8: We don't used `[Unreleased]` section anymore. This can just go under `[3.7.0]` (which hasn't released yet, the date is just when the previous was released afaik).

CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
...
+
+### Added
+
+- Added two new meeting events: `sendingAudioFailed` to indicate a failure in sending audio packets out and `sendingAudioRecovered` to indicate recovery after a `sendingAudioFailed` event
```
hensmi-amazon, line 12: Not sure if this was discussed, did we consider `audioSendXXX` or `audioSendingXXX`? Feels more natural to namespace the events that way.

...and now i see receivingAudioDropped...but also audioInputFailed....argh. I'm still inclined to ignore `receivingAudioDropped` and prefix with `audio` first, what do you think?

CHANGELOG.md
```diff
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
...
+- Added a new ConnectionHealthPolicy `SendingAudioFailureConnectionHealthPolicy` to evaluate and trigger the above events. Also added `sendingAudioFailureSamplesToConsider` and `sendingAudioFailureInitialWaitTimeMs` to `ConnectionHealthPolicyConfiguration` fields for configuring the new health policy.
+- Added a new field `consecutiveStatsWithNoAudioPacketsSent` in `ConnectionHealthData` for evaluating the `SendingAudioFailureConnectionHealthPolicy`.  
+
+### Changed
+- Refactored `connectionHealthDidChange()` in `MonitorTask` by creating a new `enactHealthPolicy()` method to reduce redundancy in the health policy evaluation.
```
hensmi-amazon, line 17: nit: `enact` is uncommon vocabulary. `apply`? I guess i see now its a private fn so not a big deal.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1640,6 +1640,10 @@ export class DemoMeetingApp
       case 'meetingReconnected':
       case 'receivingAudioDropped':
       case 'signalingDropped':
+      case 'sendingAudioFailed':
+      // We can move sendingAudioRecovered below to include meetingHistory attribute if we re-use it for
```
hensmi-amazon, line 1644: Lol @ leaving a TODO but not marking it a TODO since we aren't supposed to do that i guess :P.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2920,6 +2924,11 @@ export class DemoMeetingApp
       }
     }
 
+    if (value === 'No Audio') {
+      // An empty media stream destination without any source connected to it
```
hensmi-amazon, line 2928: Whats the purpose of this, e.g. if a builder were to ask? That should probably go into the comment.

guides/06_Meeting_Events.md
```diff
@@ -248,6 +252,22 @@ JavaScript uses the browser's [`getUserMedia` API](https://developer.mozilla.org
 
 <br>
 
+### Audio transmission related events
+
+The Javascript SDK publishes the following events to report issues with audio transmission:
```
hensmi-amazon, line 257: I believe it should always be `Chime SDK for JavaScript` .

hensmi-amazon: 'audio transmission and receipt'?

guides/06_Meeting_Events.md
```diff
@@ -248,6 +252,22 @@ JavaScript uses the browser's [`getUserMedia` API](https://developer.mozilla.org
...
+
+- The `sendingAudioFailed` event is triggered when audio packets are not being sent out consistently. This event can be used to alert the user that their audio is not reaching other attendees or to ask them to take a manual action such as changing the input audio device or restarting the application. When there's a recovery in the audio packets sent, the `sendingAudioRecovered` event is published which can be used to acknowledge that the user action did succeed.  Unlike the `receivingAudioDropped` event there are no additional callbacks which are invoked for this case. There are few caveats attached to these events listed below:
+  - The `sendingAudioFailed` event may not trigger accurately in Firefox when the user switches from a working audio input device to a non-working one, this is due to a potential bug in the WebRTC's getStats() implementation within Firefox.
```
hensmi-amazon, line 262: How should a builder compensate for this?

guides/06_Meeting_Events.md
```diff
@@ -248,6 +252,22 @@ JavaScript uses the browser's [`getUserMedia` API](https://developer.mozilla.org
...
+  - The `sendingAudioFailed` event may not trigger accurately in Firefox when the user switches from a working audio input device to a non-working one, this is due to a potential bug in the WebRTC's getStats() implementation within Firefox. 
+  - In certain genuine usecases where applications want to allow users to join as an observer without any audio input, we may trigger the `sendingAudioFailed` event incorrectly in Firefox and Safari. This is related to an [existing bug](https://github.com/aws/amazon-chime-sdk-js/issues/474) which actually sends such attendees into a reconnect loop and eventually results in meeting failure.
```
hensmi-amazon, line 263: Same for this

guides/06_Meeting_Events.md
```diff
@@ -248,6 +252,22 @@ JavaScript uses the browser's [`getUserMedia` API](https://developer.mozilla.org
...
+
+
+Once published these events undergo a cooldown period to avoid inundating the event observer
```
hensmi-amazon, line 266: inundate (fun word choice though) -> overwhelm. Reducing vocabulary set/complexity improves understand-ability, especially given we aren't translating these guides :)

integration/js/SendingAudioFailureAndRecoveryTest.js
```diff
@@ -0,0 +1,38 @@
...
+    constructor(name, kiteConfig) {
+        super(name, kiteConfig, "Sending Audio Failure and Recovery Test");
+    }
+
+    async runIntegrationTest() {
+        this.numberOfParticipant = 1;
+        const session = this.seleniumSessions[0];
+        const attendeeID = uuidv4();
+
+        await OpenAppStep.executeStep(this, session);
```
hensmi-amazon, line 18: Hmmm this uses browser demo instead of custom integration test app right?...this is fine but ideally we wouldn't be using the demo app for this.

src/connectionhealthpolicy/BaseConnectionHealthPolicy.ts
```diff
@@ -10,12 +10,22 @@ export default class BaseConnectionHealthPolicy implements ConnectionHealthPolic
...
+    canonicalNameForLogging?: string
+  ) {
     this.minHealth = configuration.minHealth;
     this.maxHealth = configuration.maxHealth;
     this.currentHealth = configuration.initialHealth;
     this.currentData = data.clone();
+    this.canonicalNameForLogging = canonicalNameForLogging;
+  }
+
+  getCanonicalNameForLogging(): string {
```
hensmi-amazon, line 27: can't we just make the member public instead of manually writing a getter? we can always do `get XXX` later.

src/connectionhealthpolicy/BaseConnectionHealthPolicy.ts
```diff
@@ -10,12 +10,22 @@ export default class BaseConnectionHealthPolicy implements ConnectionHealthPolic
   protected minHealth: number;
   protected maxHealth: number;
   protected currentHealth: number;
+  readonly canonicalNameForLogging?: string;
```
hensmi-amazon, line 13: Anything wrong with just calling this `name`? Can't imagine we would have another name member in the future.

src/connectionhealthpolicy/ConnectionHealthData.ts
```diff
@@ -84,6 +87,9 @@ export default class ConnectionHealthData {
   setConsecutiveStatsWithNoPackets(stats: number): void {
     this.consecutiveStatsWithNoPackets = stats;
   }
+  setConsecutiveStatsWithNoAudioPacketsSent(stats: number): void {
```
hensmi-amazon, line 90: all these unnecessary setters for public members lol.

src/connectionhealthpolicy/ConnectionHealthPolicyConfiguration.ts
```diff
@@ -24,4 +24,6 @@ export default class ConnectionHealthPolicyConfiguration {
   missedPongsUpperThreshold = 4;
   maximumAudioDelayMs = 60000;
   maximumAudioDelayDataPoints = 10;
+  sendingAudioFailureSamplesToConsider = 2;
```
hensmi-amazon, line 27: Nit: I know *none* of the other members do this, but can we comment why we chose these values here or in the guide?

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
```diff
@@ -0,0 +1,63 @@
...
+    super(configuration, data, 'Sending Audio Health');
+    this.secondsToConsider = configuration.sendingAudioFailureSamplesToConsider;
+    this.initialWaitTimeMs = configuration.sendingAudioFailureInitialWaitTimeMs;
+    this.maximumTimesToWarn = configuration.maximumTimesToWarn;
+    this.coolDownTimeMs = configuration.cooldownTimeMs;
+    this.lastWarnTimestampMs = 0;
+    this.warnCount = 0;
+  }
+
+  private isSendingAudioUnHealthy(): boolean {
```
hensmi-amazon, line 34: nit: 'unhealthy' is one word, so don't need the caps.

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  health(): number {
+    const didWarnRecently = Date.now() - this.lastWarnTimestampMs < this.coolDownTimeMs;
+    if (this.isSendingAudioUnHealthy()) {
+      if (this.currentHealth > this.minimumHealth() && !didWarnRecently) {
+        this.logger.debug(`Sending Audio is unhealthy for ${this.secondsToConsider} seconds`);
+        this.warnCount++;
+        if (this.warnCount > this.maximumTimesToWarn) {
+          this.logger.debug(
```
hensmi-amazon, line 50: should this be debug level?

src/connectionmonitor/SignalingAndMetricsConnectionMonitor.ts
```diff
@@ -116,9 +116,23 @@ export default class SignalingAndMetricsConnectionMonitor
     if (packetsReceived === 0 || fractionPacketsLostInbound > 0) {
       this.connectionHealthData.setLastPacketLossInboundTimestampMs(Date.now());
     }
+    const potentialAudioPacketsSent = metricReport.audioPacketsSent;
```
hensmi-amazon, line 119: nit: why the 'potential' prefix?

src/task/MonitorTask.ts
```diff
@@ -296,11 +298,41 @@ export default class MonitorTask
...
+      () => {
+        this.context.eventController?.publishEvent('sendingAudioRecovered');
+      }
+    );
+  }
+
+  private enactHealthPolicy(
+    healthPolicy: BaseConnectionHealthPolicy,
+    connectionHealthData: ConnectionHealthData,
+    unHealthyCallback: () => void,
```
hensmi-amazon, line 324: nit: unhealthy is one word

CHANGELOG.md
avinashmidathada: Yes this was discussed and the names were suggested by @simmkyu. I personally feel `sendingAudioFailed` should tie up closely with `receivingAudioDropped` because it's related to transmission/reception as opposed to something like `audioInputXXX`. That said, this is not a strong opinion of mine so if any other reviewer agrees I will change it to `audioSendingXXX`

demos/browser/app/meetingV2/meetingV2.ts
avinashmidathada: :sweat_smile: Well it's supposed to serve like a note to the person implementing the `sendingAudioSilenced` event but I guess i could track it offline

src/connectionhealthpolicy/ConnectionHealthPolicyConfiguration.ts
avinashmidathada: Thanks for reminding, I wanted to add that info. I'm inclined to adding it to the description of the PR so it's easily visible

guides/06_Meeting_Events.md
avinashmidathada: Do you mean, how should builders handle the events or are you specifically asking about the Firefox issue?

demos/browser/app/meetingV2/meetingV2.ts
avinashmidathada: It's primarily for testing the `sendingAudioFailed` event, I don't see it being useful otherwise. Will add that as a comment

integration/js/SendingAudioFailureAndRecoveryTest.js
avinashmidathada: I was unaware that we have a custom integration test app, let me look into that

src/connectionhealthpolicy/BaseConnectionHealthPolicy.ts
avinashmidathada: Curse of overthinking :sigh:. will change

src/connectionhealthpolicy/ConnectionHealthData.ts
avinashmidathada: Haha, but in my case this proved to be useful for unit testing though!

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2519,7 +2523,7 @@ export class DemoMeetingApp
 
   async populateAudioInputList(): Promise<void> {
     const genericName = 'Microphone';
-    let additionalDevices = ['None', '440 Hz', 'Prerecorded Speech', 'Echo'];
+    let additionalDevices = ['None', '440 Hz', 'Prerecorded Speech', 'Echo', 'No Audio'];
```
devalevenkatesh, line 2526: In description you mentioned this does not work in firefox, is it due to firefox not supporting the dependent WebRTC metric?

If thats the case, should we populate this based on browser now? We do this for simulcast, video priority options while joining the demo hence checking.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2920,6 +2924,11 @@ export class DemoMeetingApp
       }
     }
 
+    if (value === 'No Audio') {
+      // An empty media stream destination without any source connected to it
+      return DefaultDeviceController.getAudioContext().createMediaStreamDestination().stream;
```
devalevenkatesh, line 2929: This creates a new `AudioContext` if one not existing, not sure in testing whether you see the audio content been re-used or created. If created do we have to [`close`](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/close) it to avoid any memory leak? I know this is a demo change but still checking. Will using plain [`MediaStream`](https://developer.mozilla.org/en-US/docs/Web/API/MediaStream/MediaStream) API help here?

guides/06_Meeting_Events.md
```diff
@@ -102,6 +102,8 @@ The Chime SDK for JavaScript sends these meeting events.
 |`videoInputFailed`      |The camera selection failed.
 |`signalingDropped`      |The WebSocket failed or closed with an error.
 |`receivingAudioDropped` |A significant number of receive-audio packets dropped.
+|`sendingAudioFailed`    |Audio packets are not being sent out for a considerable timespan
```
devalevenkatesh, line 105: nit: End with a dot.

Also, would it be better for builders to understand if we link to a documentation or add explicit details on configuring the timespan here over "considerable timespan"?

guides/06_Meeting_Events.md
```diff
@@ -102,6 +102,8 @@ The Chime SDK for JavaScript sends these meeting events.
...
+|`sendingAudioRecovered` |Recovery from `sendingAudioFailed` as audio packets are being sent.
```
devalevenkatesh, line 106: Same as above, may be: "Audio packets send successful after failure."

guides/06_Meeting_Events.md
devalevenkatesh: Please correct me if we anywhere document what the builders are supposed to do if they receive this event? If not, we can add more here as well as add an use case under README for surfacing it more.

guides/06_Meeting_Events.md
devalevenkatesh: Just to add to Henry's first comment: The Amazon Chime SDK for JavaScript :)

guides/06_Meeting_Events.md
```diff
@@ -248,6 +252,22 @@ JavaScript uses the browser's [`getUserMedia` API](https://developer.mozilla.org
 
 <br>
 
+### Audio transmission related events
+
+The Javascript SDK publishes the following events to report issues with audio transmission:
+
+- The `receivingAudioDropped` event is triggered when the packet loss for receiving audio is high. This can be used to warn users about potential network connectivity issues but if you are already using the [connectionDidSuggestStopVideo](https://aws.github.io/amazon-chime-sdk-js/interfaces/audiovideoobserver.html#connectiondidsuggeststopvideo) or [connectionDidBecomePoor](https://aws.github.io/amazon-chime-sdk-js/interfaces/audiovideoobserver.html#connectiondidbecomepoor) callbacks you do not need to take any action as these callbacks are triggered on the same Predicate.
```
devalevenkatesh, line 259: nit: Predicate -> predicate?

guides/06_Meeting_Events.md
```diff
@@ -248,6 +252,22 @@ JavaScript uses the browser's [`getUserMedia` API](https://developer.mozilla.org
...
+
+- The `sendingAudioFailed` event is triggered when audio packets are not being sent out consistently. This event can be used to alert the user that their audio is not reaching other attendees or to ask them to take a manual action such as changing the input audio device or restarting the application. When there's a recovery in the audio packets sent, the `sendingAudioRecovered` event is published which can be used to acknowledge that the user action did succeed.  Unlike the `receivingAudioDropped` event there are no additional callbacks which are invoked for this case. There are few caveats attached to these events listed below:
```
devalevenkatesh, line 261: nit: Suggest dropping "being".
I believe, this is fine"...when audio packets are not sent out..".

guides/06_Meeting_Events.md
devalevenkatesh: I think the prior if I understand that correctly.

For the firefox issue, should we then file a bug with them? Changing audio device failure affects the user experience in a meeting hence if this occurs would be good to track it as a browser bug or provide a workaround in such a case.

guides/06_Meeting_Events.md
```diff
@@ -248,6 +252,22 @@ JavaScript uses the browser's [`getUserMedia` API](https://developer.mozilla.org
...
+  - The `sendingAudioFailed` event may not trigger accurately in Firefox when the user switches from a working audio input device to a non-working one, this is due to a potential bug in the WebRTC's getStats() implementation within Firefox. 
+  - In certain genuine usecases where applications want to allow users to join as an observer without any audio input, we may trigger the `sendingAudioFailed` event incorrectly in Firefox and Safari. This is related to an [existing bug](https://github.com/aws/amazon-chime-sdk-js/issues/474) which actually sends such attendees into a reconnect loop and eventually results in meeting failure.
+
+
+Once published these events undergo a cooldown period to avoid inundating the event observer
+in case of an erratic network/audio device. In addition to this, within a connection, the number of times the events are published is also capped to a maximum . `cooldownTimeMs` and `maximumTimesToWarn` are the respective thresholds for this behaviour whose values are defined in [ConnectionHealthPolicyConfiguration](https://aws.github.io/amazon-chime-sdk-js/classes/connectionhealthpolicyconfiguration.html).
```
devalevenkatesh, line 267: nit: Remove an extra space after the "maximum".

src/connectionhealthpolicy/ConnectionHealthPolicyConfiguration.ts
devalevenkatesh: I think adding a comment here would be helpful over PR description as once this is merged its extra steps to get to the PR :)

We do that in [`MeetingSessionConfiguration`](https://github.com/aws/amazon-chime-sdk-js/blob/main/src/meetingsession/MeetingSessionConfiguration.ts).

devalevenkatesh, line 27: [For understanding]: Failure samples here means audio packets metric or how?

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
```diff
@@ -0,0 +1,63 @@
...
+  private warnCount: number;
+  private lastWarnTimestampMs: number;
+
+  constructor(
+    private logger: Logger,
+    configuration: ConnectionHealthPolicyConfiguration,
+    data: ConnectionHealthData
+  ) {
+    super(configuration, data, 'Sending Audio Health');
+    this.secondsToConsider = configuration.sendingAudioFailureSamplesToConsider;
```
devalevenkatesh, line 26: [For understanding]: Failure samples here means audio packets metric or how? Though private, the `secondsToConsider` might need a better naming 😅 .

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
```diff
@@ -0,0 +1,63 @@
...
+    const hasEnoughTimeElapsedToEvaluateStatus = !this.currentData.isConnectionStartRecent(
+      this.initialWaitTimeMs
+    );
+    const areAudioPacketsNotBeingSent =
+      this.currentData.consecutiveStatsWithNoAudioPacketsSent >= this.secondsToConsider;
+    return hasEnoughTimeElapsedToEvaluateStatus && areAudioPacketsNotBeingSent;
+  }
+
+  health(): number {
+    const didWarnRecently = Date.now() - this.lastWarnTimestampMs < this.coolDownTimeMs;
```
devalevenkatesh, line 44: Probably we should return if `didWarnRecently` is `true`, cause, then no need to calculate `isSendingAudioUnhealthy` and we can save some computations.

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
```diff
@@ -0,0 +1,63 @@
...
+    this.lastWarnTimestampMs = 0;
+    this.warnCount = 0;
+  }
+
+  private isSendingAudioUnHealthy(): boolean {
+    const hasEnoughTimeElapsedToEvaluateStatus = !this.currentData.isConnectionStartRecent(
+      this.initialWaitTimeMs
+    );
+    const areAudioPacketsNotBeingSent =
+      this.currentData.consecutiveStatsWithNoAudioPacketsSent >= this.secondsToConsider;
```
devalevenkatesh, line 39: Can we use `configuration.sendingAudioFailureSamplesToConsider` over `this.secondsToConsider` and just drop that variable? I am a bit confused one maps to seconds and one maps to sample count.

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
```diff
@@ -0,0 +1,63 @@
...
+    const didWarnRecently = Date.now() - this.lastWarnTimestampMs < this.coolDownTimeMs;
+    if (this.isSendingAudioUnHealthy()) {
+      if (this.currentHealth > this.minimumHealth() && !didWarnRecently) {
+        this.logger.debug(`Sending Audio is unhealthy for ${this.secondsToConsider} seconds`);
+        this.warnCount++;
+        if (this.warnCount > this.maximumTimesToWarn) {
+          this.logger.debug(
+            'SendingAudioFailure maximum warnings were breached. Falling back to reporting healthy.'
+          );
+          return this.maximumHealth();
```
devalevenkatesh, line 53: So why not `currentHealth`, technically, the experience would be still bad right?

src/task/MonitorTask.ts
```diff
@@ -296,11 +298,41 @@ export default class MonitorTask
...
+        this.context.eventController?.publishEvent('sendingAudioFailed');
+      },
+      () => {
+        this.context.eventController?.publishEvent('sendingAudioRecovered');
+      }
+    );
+  }
+
+  private enactHealthPolicy(
+    healthPolicy: BaseConnectionHealthPolicy,
```
devalevenkatesh, line 322: Since all health policies implement `ConnectionHealthPolicy` suggest this to type to `ConnectionHealthPolicy` (interface) and not `BaseConnectionHealthPolicy` (class).

src/task/MonitorTask.ts
```diff
@@ -296,11 +298,41 @@ export default class MonitorTask
...
+    connectionHealthData: ConnectionHealthData,
+    unHealthyCallback: () => void,
+    healthyCallback?: () => void
```
devalevenkatesh, line 325: Why is `healthyCallback` optional and the `unhealthyCallback` required? Since all policies provides both and its private function should we just make both required OR both optional?

src/task/MonitorTask.ts
```diff
@@ -296,11 +298,41 @@ export default class MonitorTask
...
+  ): void {
+    healthPolicy.update(connectionHealthData);
+    const healthValue = healthPolicy.healthIfChanged();
+    if (healthValue !== null) {
+      this.logger.info(`${healthPolicy.getCanonicalNameForLogging()} value is now ${healthValue}`);
+      if (healthValue <= healthPolicy.minimumHealth()) {
+        unHealthyCallback.bind(this)();
+      } else {
+        Maybe.of(healthyCallback).map(f => f.bind(this)());
```
devalevenkatesh, line 334: `healthyCallback` also does not need to map on observers since it is not an array? Sorry correct me if wrong.

demos/browser/app/meetingV2/meetingV2.ts
pracheth: Can maybe update the comment to say it's currently only used in the integration test for the new event?

src/connectionhealthpolicy/ConnectionHealthPolicyConfiguration.ts
avinashmidathada: Makes sense but unlike [MeetingSessionConfiguration](https://github.com/aws/amazon-chime-sdk-js/blob/main/src/meetingsession/MeetingSessionConfiguration.ts) where we documented the meaning of each property, what Henry was asking for is the rationale for choosing the value (in this case 2) which in this case is a bit verbose.

The rationale here is "2000 ms is the upper bound for median (p50) of the `attendeePresenceDurationMs` values across all meetings using event ingestion and reporting the `attendeePresenceReceived` event."  I feel this would be better off inside the guide rather than the code. Let me know if you disagree

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
avinashmidathada: Was confused because out of the two existing health policies one has logging at warn level and the other has no logging at all. I'll change this to warn as it provides better visibility and there's not a lot of performance impact either (2 logs every 1 min in worst case)

src/connectionmonitor/SignalingAndMetricsConnectionMonitor.ts
avinashmidathada: Was just following the norm set by the code written above. But since it's a bit confusing, I'll just remove the declaration and just check whether the `audioPacketsSent` metric is present or not and use it

demos/browser/app/meetingV2/meetingV2.ts
avinashmidathada: The metric is supported by Firefox, but it's not working as expected when I change the device to `No Audio`  and I suspect the issue is with Firefox's implementation of getStats() and I'm planning to confirm that and report a bug to them.

But I don't fully get your point about populating "this" based on browser, do you mean adding `No Audio` to the additionalDevices based on browser?

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
avinashmidathada: Sure, that makes sense!

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
avinashmidathada: `didWarnRecently` is only used when we're in the unhealthy path (i.e. returning minimumHealth) because I wanted to make sure that we notify asap when we recover, so we need to calculate the `isSendingAudioUnhealthy` each time.

demos/browser/app/meetingV2/meetingV2.ts
devalevenkatesh: Yup. I thought that this is not supported in FF hence suggested to populate it based on browser but if it is supported, then, it would be good to check the issue and file a bug with FF directly.

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
avinashmidathada: Yeah, I admit it's a bit confusing, I think just using `sendingAudioFailureSamplesToConsider` everywhere is clear.

 PS: The reason I used `secondsToConsider` is because we only collect RTC stats once every second

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
avinashmidathada: Yeah, the experience would still be bad but this is how `UnusableAudioWarningConnectionHealthPolicy` also works. I think the rationale with returning maximum health is that we don't want to flood the observer with too many invocations which might directly lead to user level warnings on UI if builders don't have any extra handling in between.

I don't think an ignore flag makes sense here, because `maximumTimesToWarn` itself is a configurable value. Instead maybe I can create a separate `sendingAudioFailureMaximumTimesToWarn` which has a higher value (say 5 instead of 2)? Let me know your thoughts.

src/task/MonitorTask.ts
avinashmidathada: Sure I was hesitant because I needed to add a readonly property for capturing the name of the health policy for logging purpose and I was not sure if I can add that to the interface, but it doesn't seem to be an issue, will change

src/task/MonitorTask.ts
avinashmidathada: `ReconnectionHealthPolicy` doesn't have a healthyCallback, that's why I put it as optional. I can make both of them optional for more flexibility

src/task/MonitorTask.ts
avinashmidathada: Sorry, I don't understand your comment. Based on my understanding `Maybe` can be used with any function to safely invoke after checking for null/undefined. But I don't really need it here because I'm not using the return value after invocation anyways and not chaining any calls.  I can avoid using Maybe by doing the following:

```
if (healthyCallback) {
  healthyCallback.bind(this)();
}
```

Is that what you're suggesting?

src/connectionhealthpolicy/ConnectionHealthPolicyConfiguration.ts
devalevenkatesh: No strong opinions here either way is fine for me.

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
devalevenkatesh: So better to move `const didWarnRecently = Date.now() - this.lastWarnTimestampMs < this.coolDownTimeMs;` inside the `this.isSendingAudioUnHealthy()` check then?

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
devalevenkatesh: I think using `maximumTimesToWarn` is fine for now and we can add extra variable if needed later. 

If ignoring then why would we flood the observers? Instead of checking for `maximumHealth` value will have to also check whether `health()` returns ignore. But anyways if other policies are using the same I am fine with keeping it that way.

src/task/MonitorTask.ts
devalevenkatesh: Sorry I was not clear earlier. But you answered it and thanks for explanation. I was unaware on the `Maybe.of` returning a [`map`](https://github.com/aws/amazon-chime-sdk-js/blob/main/src/utils/Types.ts#L38). 

If you are making both callbacks optional then its needed for both now. Elsewhere, we follow the same so its fine to use here as well.

src/connectionhealthpolicy/SendingAudioFailureConnectionHealthPolicy.ts
avinashmidathada: We call the `healthIfChanged()` method for each health policy which returns the health value if the health has changed, null if the health hasn't changed so by ignoring the unhealthy state and returning `maximumHealth()` there would be no change and hence observers won't be notified.

integration/js/SendingAudioFailureAndRecoveryTest.js
devalevenkatesh: Yes we do have to specifically test the video APIs here:
https://github.com/aws/amazon-chime-sdk-js/tree/main/integration/js/app

For this test, may it is good to add an audio demo and test them under it? @ltrung any suggestions? Then may be we also do not have to block on the firefox issue with no audio at least for the main PR merge.

integration/js/SendingAudioFailureAndRecoveryTest.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+
+module.exports = SendingAudioFailureAndRecoveryTest;
+
+(async () => {
+    const kiteConfig = await TestUtils.getKiteConfig(__dirname);
+    let test = new SendingAudioFailureAndRecoveryTest('Sending Audio Failure and Recovery Test', kiteConfig);
+    await test.run();
+})();
```
devalevenkatesh, line 38: nit: Add newline for EOF

integration/js/checks/SendingAudioFailureAndRecoveryCheck.js
```diff
@@ -0,0 +1,29 @@
...
+    async run() {
+        const passed = await this.page.sendingAudioCheck(this, this.expectedEvent, this.waitTimeMs);
+        if (!passed) {
+            throw new KiteTestError(Status.FAILED, 'Sending audio failure/recovery check failed');
+        }
+        this.finished("sending_audio_failure_and_recovery_check_complete");
+    }
+}
+
+module.exports = SendingAudioFailureAndRecoveryCheck;
```
devalevenkatesh, line 29: nit: Add newline for EOF

integration/js/steps/SelectNoAudioInputStep.js
```diff
@@ -0,0 +1,28 @@
+const AppTestStep = require('../utils/AppTestStep');
+
+class SelectNoAudioInputStep extends AppTestStep {
+    constructor(kiteBaseTest, sessionInfo) {
+        super(kiteBaseTest, sessionInfo);
+    }
+
+    static async executeStep(KiteBaseTest, sessionInfo, useMedical) {
```
devalevenkatesh, line 8: Sorry dint understood this `useMedical` use here, what is this for?

demos/browser/app/meetingV2/meetingV2.ts
ltrung: What is the different between `No Audio` and `None` ?

integration/js/SendingAudioFailureAndRecoveryTest.js
ltrung: Ideally, we want separate demo for video and audio testing and not keep adding test stuffs in the kitchen sink demo. In this case, however, I don't think it matters that much as it is just an event check and there is not much changes we make to the demo.

guides/06_Meeting_Events.md
avinashmidathada: Got it, I'm figuring out how to file an issue with FF with reproducible steps in a standalone manner

avinashmidathada (commented):
Replying to some comments

guides/06_Meeting_Events.md
avinashmidathada: After some more deep dive realized that this is not actually a bug with Firefox's `getStats()` implementation. Rather, it's just the way we're testing  that's causing the disparity between browsers. 

Chromium's implementation for an empty MediaStreamDestination doesn't generate any audio packets whereas in Firefox it always defaults to sending frames of silence. (I verified this on the server side by logging the packets as they're received)

I've just about tried every possible way to induce a "no audio" scenario using Web Audio APIs but had no success. Here are my findings:
- Any kind of `AudioScheduledSourceNode` such as [AudioBufferSourceNode](https://developer.mozilla.org/docs/Web/API/AudioBufferSourceNode), [ConstantSourceNode](https://developer.mozilla.org/docs/Web/API/ConstantSourceNode) and [OscillatorNode](https://developer.mozilla.org/docs/Web/API/OscillatorNode) do not work as the documentation clearly states they output silence when not playing. 
- Tried adding a DelayNode with a large delay on top of these to try and induce "no audio" temporarily but even when the delay is in effect silent audio packets are generated.
-  Also looked at AudioWorkletNode with custom AudioWorkletProcessor but the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/AudioWorkletProcessor) states that the output of the node by default holds silence. I tried deleting the values in the array but that throws an error too.

So, I'm just going to remove this line from the guide and call out that the` No Audio` device doesn't work in Firefox in the demo app

devalevenkatesh: Checking if we have considered not using Web Audio APIs and use MediaStream APIs? Or that also does not help?

avinashmidathada: Yes, I have tried using MediaStream API first. But, I couldn't find a good way to create a dummy MediaStreamTrack that produces no audio. Even the hacky ways mentioned [here](https://github.com/web-platform-tests/wpt/issues/7424) like `new RTCPeerConnection().addTransceiver('audio').receiver.track` don't seem to do the trick because FF still sends silent packets. There are many ways to mock a MediaStreamTrack but so far none of them worked while running the application for me

Also, we can't get away with not passing a MediaStreamTrack by simple constructing an empty MediaStream, because we have an explicit [condition to throw an error](https://github.com/aws/amazon-chime-sdk-js/blob/main/src/audiovideocontroller/DefaultAudioVideoController.ts#L1056-L1058) when that happens

CHANGELOG.md
```diff
@@ -9,11 +9,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added two new meeting events: `sendingAudioFailed` to indicate a failure in sending audio packets out and `sendingAudioRecovered` to indicate recovery after a `sendingAudioFailed` event
```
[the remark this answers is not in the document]
devalevenkatesh: In these change logs some code related constructs are back-ticked and some are not, could you please make them consistent? For example, ConnectionHealthPolicy, SignalingAndMetricsConnectionMonitor, signalingClient.

demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -2920,7 +2922,16 @@ export class DemoMeetingApp
       }
     }
 
+    if (value === 'No Audio') {
+      // An empty media stream destination without any source connected to it, so it doesn't generate any audio
+      // This is currently only used for integration testing of 'sendingAudioFailed' and 'sendingAudioRecovered' events.
+      // Note: This only works in Chromium based browsers and doesn't work in Firefox because Firefox generates
```
devalevenkatesh, line 2928: So should we attach the 'No Audio' option only for browsers other than Firefox? If we run into any error/fatal error then probably we should do that. What do you think?

guides/06_Meeting_Events.md
```diff
@@ -204,6 +206,8 @@ The following table lists available states.
 |`videoInputFailed`|The camera selection failed.
 |`videoInputSelected`|The camera was selected.
 |`videoInputUnselected`|The camera was removed. You called `meetingSession.audioVideo.stopVideoInput` with `null`.
+|`sendingAudioFailed`|Audio packets are not being sent out for a considerable timespan
```
devalevenkatesh, line 209: The state description here for both the new events is different from the ones above under "Meeting events and attributes", can you please keep them same?

src/connectionhealthpolicy/BaseConnectionHealthPolicy.ts
```diff
@@ -10,12 +10,18 @@ export default class BaseConnectionHealthPolicy implements ConnectionHealthPolic
   protected minHealth: number;
   protected maxHealth: number;
   protected currentHealth: number;
+  readonly name: string;
 
-  constructor(configuration: ConnectionHealthPolicyConfiguration, data: ConnectionHealthData) {
+  constructor(
+    configuration: ConnectionHealthPolicyConfiguration,
+    data: ConnectionHealthData,
+    name: string
```
devalevenkatesh, line 18: Since customer may be overriding these classes (we cannot say for sure), this would be a breaking change as we are introducing `name` as new required parameter. Could you make this `name` optional to preserve backward compatibility?

src/connectionhealthpolicy/ConnectionHealthPolicy.ts
```diff
@@ -6,6 +6,11 @@ import ConnectionHealthData from './ConnectionHealthData';
 /* A ConnectionHealthPolicy reduces the current state of a ConnectionHealthData
    object into a single number between a min and max range.*/
 export default interface ConnectionHealthPolicy {
+  /**
+   * A unique readable name for the purpose of logging
+   */
+  readonly name: string;
```
devalevenkatesh, line 12: We have to make this optional as well to preserve backward compatibility.

src/task/MonitorTask.ts
```diff
@@ -267,23 +275,17 @@ export default class MonitorTask
...
-    if (reconnectionValue !== null) {
-      this.logger.info(`reconnection health is now: ${reconnectionValue}`);
-      if (reconnectionValue === 0) {
-        this.context.audioVideoController.handleMeetingSessionStatus(
-          new MeetingSessionStatus(MeetingSessionStatusCode.ConnectionHealthReconnect),
-          null
-        );
-      }
-    }
+    this.applyHealthPolicy(this.reconnectionHealthPolicy, connectionHealthData, () => {
```
devalevenkatesh, line 278: Since you addressed the comments which changed this, could you please check the change log and PR description once  and update it accordingly? For example: `enact` -> `apply

src/task/MonitorTask.ts
```diff
@@ -296,11 +298,41 @@ export default class MonitorTask
...
           Maybe.of(observer.connectionDidBecomeGood).map(f => f.bind(observer)());
         });
       }
+    );
+
+    this.applyHealthPolicy(
+      this.sendingAudioFailureHealthPolicy,
+      connectionHealthData,
+      () => {
+        this.context.eventController?.publishEvent('sendingAudioFailed');
```
devalevenkatesh, line 313: Sorry, dint got this point earlier, so for the `receivingAudioDropped` event we also publish `attributes` with it:
- https://github.com/aws/amazon-chime-sdk-js/blob/main/src/task/MonitorTask.ts#L289
- https://github.com/aws/amazon-chime-sdk-js/blob/main/src/task/MonitorTask.ts#L408-L428

I recommend we also include the relevant attributes with both of these new events. Also, we would like to update [this](https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/06_Meeting_Events.md#meeting-attributes) section in the Meeting events guide so that builders know which attributes is available under which meeting event.

src/connectionhealthpolicy/BaseConnectionHealthPolicy.ts
avinashmidathada: Ah that makes sense, I initially planned to make it optional but didn't see a real need for it, but it makes sense that customers may implement the interface or extend this class. Will change!

src/task/MonitorTask.ts
avinashmidathada: I intentionally left those additional attributes out because the [standard attributes](https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/06_Meeting_Events.md#standard-attributes) seemed like they're enough for the `sendingAudioXXX` events and I wanted to be frugal with the data we send. Let me know what you think.

If we decide to add additional attributes I think we can include the following:
- meetingDurationMs
- meetingStartDurationMs
- signalingOpenDurationMs
- iceGatheringDurationMs
- attendeePresenceDurationMs

And leave out maxVideoTileCount and poorConnectionCount. Thoughts?

demos/browser/app/meetingV2/meetingV2.ts
avinashmidathada: It doesn't throw an error/fatal per se, rather it just doesn't work as intended. I do think it's better to hide it in FF just for the sake of avoiding any confusion.

src/task/MonitorTask.ts
devalevenkatesh: So basically if in situation like these, there are existing meeting events that also trigger which include above attributes then we should be fine to drop the attributes altogether. If not then would be good to add them.

avinashmidathada: As discussed, we cannot rely on existing meeting events so I added the attributes I listed above

devalevenkatesh (approved):

devalevenkatesh:
You can have to re-sync the PR and resolve conflicts, then the checks should pass.

PR #2400, opened 2022-08-15

dpwspoon:
**Issue #:**

Fixs 4401 from https://github.com/aws/amazon-chime-sdk-js/issues/2372

**Description of changes:**

MessagingSession reconnects with refreshed endpoint and credentials if needed.  This is an updated version of https://github.com/aws/amazon-chime-sdk-js/pull/2180 that was reverted

**Testing:**

Tested with demo app (see below)

1. Run the messagingSession demo here: https://github.com/aws/amazon-chime-sdk-js/tree/main/demos/browser. `npm run start --app=messagingSession`
2. connect with an app instance user
3. Disconnect Wifi
4. wait ~20 secs
5. Reconnect Wifi and see reconnect

Then test again with the following diff to see working on v2 version of client:

```
diff --git a/demos/browser/app/messagingSession/messagingSession.ts b/demos/browser/app/messagingSession/messagingSession.ts
index 2757af4d..9f408593 100644
--- a/demos/browser/app/messagingSession/messagingSession.ts
+++ b/demos/browser/app/messagingSession/messagingSession.ts
@@ -20,7 +20,7 @@ import {
   Versioning,
 } from 'amazon-chime-sdk-js';
 
-import { ChimeSDKMessagingClient } from '@aws-sdk/client-chime-sdk-messaging';
+import { ChimeSDKMessaging } from '@aws-sdk/client-chime-sdk-messaging';
 
 export class DemoMessagingSessionApp implements MessagingSessionObserver {
   static readonly BASE_URL: string = [
@@ -56,7 +56,7 @@ export class DemoMessagingSessionApp implements MessagingSessionObserver {
     document.getElementById('connect').addEventListener('click', async () => {
       try {
         const response = await this.fetchCredentials();
-        const chime = new ChimeSDKMessagingClient({ region: 'us-east-1', credentials: response });
+        const chime = new ChimeSDKMessaging({ region: 'us-east-1', credentials: response });
         this.userArn = (document.getElementById('userArn') as HTMLInputElement).value;
         this.sessionId = (document.getElementById('sessionId') as HTMLInputElement).value;
         this.prefetchSortByUnread = (document.getElementById('prefetchSortByUnread') as HTMLInputElement).checked;

```

During the testing suggest looking at networking console to see disconnect, retries, and retry success when wifi comes back up

**Checklist:**

 yes

yes these have been reviewed

no


src/messagingsession/MessagingSessionConfiguration.ts
```diff
@@ -47,13 +47,17 @@ export default class MessagingSessionConfiguration {
   /**
    * Constructs a MessagingSessionConfiguration optionally with userArn, messaging session id, a messaging session
    * endpoint URL, and the chimeClient.
+   *
+   * endpointUrl is deprecated and should not be used. Internally it is resolved on connect via chimeClient if undefined, and
```
duh17, line 51: nit: would suggest mark it deprecated with `@deprecated`  https://code.visualstudio.com/updates/v1_49#_deprecated-tag-support-for-javascript-and-typescript

duh17 (approved):

package.json
```diff
@@ -98,6 +98,7 @@
   },
   "dependencies": {
     "@aws-crypto/sha256-js": "^2.0.1",
+    "@aws-sdk/client-chime-sdk-messaging": "^3.0.0",
```
devalevenkatesh, line 101: Should we make this a peer dependency? So if builder is using AWS SDK v2 they do not have to install `@aws-sdk/client-chime-sdk-messaging` if they are using v3, they can install this dependency on their own? @ltrung any discussion on this?

devalevenkatesh: Actually we use it nvm thats invalid comment ^

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -105,7 +107,32 @@ export default class DefaultMessagingSession implements MessagingSession {
   }
 
   private async startConnecting(reconnecting: boolean): Promise<void> {
-    const signedUrl = await this.prepareWebSocketUrl();
+    // reconnect needs to re-resolve endpoint url, which will also refresh credentials on client if they are expired
+    let endpointUrl = !reconnecting ? this.configuration.endpointUrl : undefined;
+    if (endpointUrl === undefined) {
+      try {
+        if (this.configuration.chimeClient.getMessagingSessionEndpoint instanceof Function) {
```
devalevenkatesh, line 114: If the `chimeClient` is v3 version then, `this.configuration.chimeClient.getMessagingSessionEndpoint` would error right since `getMessagingSessionEndpoint` may not be in `chimeClient`? So do we have to do a validity check here?

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -105,7 +107,32 @@ export default class DefaultMessagingSession implements MessagingSession {
...
+          endpointUrl = (
+            await this.configuration.chimeClient.getMessagingSessionEndpoint().promise()
+          ).Endpoint.Url;
+        } else {
+          endpointUrl = (
+            await this.configuration.chimeClient.send(new GetMessagingSessionEndpointCommand({}))
+          ).Endpoint.Url;
+        }
+      } catch (e) {
```
devalevenkatesh, line 123: Would be good to add a log here on successful `endpointUrl` retrieval and same for error as well.

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -105,7 +107,32 @@ export default class DefaultMessagingSession implements MessagingSession {
...
+        const closeEvent = new CloseEvent('close', {
+          wasClean: false,
+          code: 4999,
+          reason: 'Failed to getMessagingSessionEndpoint',
```
devalevenkatesh, line 127: Since `getMessagingSessionEndpoint` is an function on AWS SDK v2, this would be irrelevant to AWS SDK v3. So should we change it to just 'Failed to get messaging session endpoint URL'?

src/messagingsession/DefaultMessagingSession.ts
dpwspoon: thx will update.

src/messagingsession/DefaultMessagingSession.ts
dpwspoon: thx will update

src/messagingsession/DefaultMessagingSession.ts
dpwspoon: Not follow... on v3 client getMessagingSessionEndpoint will be null so fail the instanceof Function check and use the GetMessagingSessionCommand.

devalevenkatesh: Cool just wanted to check whether you have already checked `getMessagingSessionEndpoint` validity in here.

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -105,7 +107,32 @@ export default class DefaultMessagingSession implements MessagingSession {
   }
 
   private async startConnecting(reconnecting: boolean): Promise<void> {
-    const signedUrl = await this.prepareWebSocketUrl();
+    // reconnect needs to re-resolve endpoint url, which will also refresh credentials on client if they are expired
+    let endpointUrl = !reconnecting ? this.configuration.endpointUrl : undefined;
```
ltrung, line 111: nit: is this better to read?
```
let endpointUrl = this.configuration.endpointUrl;
// reconnect needs to re-resolve endpoint url, which will also refresh credentials on client if they are expired
if (reconnecting || endpointUrl === undefined) {
```

src/messagingsession/DefaultMessagingSession.ts
ltrung: Isn't that why we check whether `getMessagingSessionEndpoint` is a function?

package.json
devalevenkatesh, line 101: So this is adding 1.5mb size to JS SDK. Just a thought: What if we take in a callback in `MessagingSessionConfiguration` and that provides us with updated URL? The `startConnecting` function can internally call this callback for getting the refreshed credentials and the endpoint.

Not a great builder experience as they have to now implement a callback in their app, but better than adding 1.5Mb size to JS SDK package? Not sure how many people are using messaging + meeting session and for those using only meeting session its an un-needed dependency.

cc: @ltrung @simmkyu

ltrung: I prefer the increase in size than adding a cumbersome callback. Also seem like you force builders to do a migration then.

src/messagingsession/DefaultMessagingSession.ts
devalevenkatesh: Yup It should be fine I resolved this :)

ltrung:
Can we verify with both v2 and v3 (different importing) to make sure it works ?

package.json
dpwspoon: +1 with what ltrung.  You can't use the messaging sdk with out the aws sdk anyways.  In the future, modularization of the amazon-chime-sdk-js will make the concerns about package size moot

src/messagingsession/DefaultMessagingSession.ts
dpwspoon: yeah, think so.  Will update thx

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -191,16 +218,18 @@ export default class DefaultMessagingSession implements MessagingSession {
...
     this.logger.info(`WebSocket close: ${event.code} ${event.reason}`);
-    this.webSocket.destroy();
-    if (
-      !this.isClosing &&
-      this.canReconnect(event.code) &&
-      this.reconnectController.retryWithBackoff(async () => {
-        this.startConnecting(true);
-      }, null)
-    ) {
+    if (event.code !== 4999) {
```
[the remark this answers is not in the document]
dpwspoon: yeah will update, thx.  4999 is just an artificial internal code.  The number was chosen cause it isn't standardized/used externally and will likely never be picked by anyone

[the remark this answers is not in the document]
devalevenkatesh: So in that case any reason why we are adding it now? I see it may be due to the custom `CloseEvent` triggered with 4999 code?

dpwspoon:
> Can we verify with both v2 and v3 (different importing) to make sure it works ?

Done, and updated directions on pr on how to do

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -105,7 +107,37 @@ export default class DefaultMessagingSession implements MessagingSession {
...
+            await this.configuration.chimeClient.getMessagingSessionEndpoint()
+          ).Endpoint.Url;
+        } else {
+          endpointUrl = (
+            await this.configuration.chimeClient.send(new GetMessagingSessionEndpointCommand({}))
+          ).Endpoint.Url;
+        }
+        this.logger.debug(`Messaging endpoint resolved to: ${endpointUrl}`);
+      } catch (e) {
+        this.logger.error("DPW " + e)
```
ltrung: Is this intended?

dpwspoon: this diff is old I believe (was having some struggles commiting changes) please see latest

test/messagingsession/DefaultMessagingSession.test.ts
```diff
@@ -233,6 +298,48 @@ describe('DefaultMessagingSession', () => {
...
+          didStartCount++;
+        },
+        messagingSessionDidStop(_event: CloseEvent): void {
+          expect(didStartConnecting).to.be.eq(2);
+          expect(didStartCount).to.be.eq(1);
+          expect(getMessSessionCnt).to.be.eq(2);
+          done();
+        },
+      });
+      messagingSession.start();
```
devalevenkatesh, line 340: With v3, you have to use `await` as the `messagingSession.start` returns a promise now:
https://github.com/aws/amazon-chime-sdk-js/blob/main/guides/17_Migration_to_3_0.md#remove-aws-global-object-from-messagingsessionconfigurationts

Also could you please update the migration doc for the same and make the `endpointUrl` `undefined` there as well?

dpwspoon: > Also could you please update the migration doc for the same and make the endpointUrl undefined there as well?

Done

ltrung (dismissed):

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -105,7 +107,35 @@ export default class DefaultMessagingSession implements MessagingSession {
...
+        if (this.configuration.chimeClient.getMessagingSessionEndpoint instanceof Function) {
+          endpointUrl = (await this.configuration.chimeClient.getMessagingSessionEndpoint())
+            .Endpoint.Url;
+        } else {
+          endpointUrl = (
+            await this.configuration.chimeClient.send(new GetMessagingSessionEndpointCommand({}))
+          ).Endpoint.Url;
+        }
+        this.logger.debug(`Messaging endpoint resolved to: ${endpointUrl}`);
+      } catch (e) {
```
devalevenkatesh, line 123: nit: e -> error

Also, would be good to log the actual error as the `CloseEventHandler` will only print the custom event.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fix `AbortError` when turning video ON in Safari.
+- `MessagingSession` reconnects with refreshed endpoint and credentials if needed.  EndpointUrl on `MessagingSessionConfiguration` is deprecated as it is resolved by calling `getMessagingSessionEndpoint` internally.
```
devalevenkatesh, line 23: nit: Backtick `EndpointUrl`

devalevenkatesh (dismissed):

Issue #2420, opened 2022-09-02
Choppy audio with Web Audio on Windows


justindarc:
### What happened and what did you expect to happen?

When joining a call with Web Audio/Voice Focus enabled on Windows, audio may become choppy when running on battery power.

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

1. Be on a Windows laptop connected to AC power
2. Join a call with Chime SDK with Web Audio and Voice Focus enabled
3. After joining the call, disconnect AC power and run on battery power
4. Other meeting participants will begin receiving "choppy audio" from you

### Amazon Chime SDK for JavaScript version

3.2.0

### What browsers are you seeing the problem on?

Chrome, Edge, Electron

### Browser version

105

### Browser console logs

No evidence of issue in logs.

michhyun1:
Can we close this issue? @justindarc

justindarc:
Yes, this was fixed with #2421

PR #2421, opened 2022-09-02
Use 'playback' latency hint for AudioContext on Windows to reduce choppy audio

justindarc:
**Issue #:** #2420

**Description of changes:** Uses a default "playback" `latencyHint` when creating the `AudioContext` on Windows. Also adds a `setDefaultLatencyHint` API to `DefaultDeviceController` to allow for overriding.

**Testing:**

Yes.

1. Be on a Windows laptop connected to AC power
2. Join a call with Chime SDK with Web Audio and Voice Focus enabled
3. After joining the call, disconnect AC power and run on battery power
4. Other meeting participants will no longer receive "choppy audio" from you

**Checklist:**

2. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? Yes. No.

3. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved? No.


michhyun1:
Requires changelog entry. Thanks!

hensmi-amazon (approved):

michhyun1:
Some conflicts to resolve.

src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -83,6 +87,14 @@ export default class DefaultBrowserBehavior implements BrowserBehavior, Extended
     return this.isFirefox();
   }
 
+  isMacOS(): boolean {
```
ltrung, line 90: We usually avoid create public platform APIs like this and instead create a wrapper on what the intention of the check is such as `needToSetLatencyForWebAudio`.

justindarc: Ack

justindarc:
Addressed comments.

hensmi-amazon (approved):
🚀

ltrung (approved):

src/browserbehavior/BrowserBehavior.ts
```diff
@@ -27,6 +27,11 @@ export default interface BrowserBehavior {
    */
   hasFirefoxWebRTC(): boolean;
 
+  /**
+   * Returns whether the browser requires the "playback" latency hint for Web Audio
+   */
+  requiresPlaybackLatencyHintForAudioContext(): boolean;
```
devalevenkatesh, line 33: If wont make this optional this would be a breaking change right? Not sure if builders are using this interface but to consider if they are using it this would be a breaking change I believe.

Tagging @justindarc @ltrung

PR #2434, opened 2022-09-13
Report encode/decode time and implementation metric

shi-su:
add implementation name as a dimension.

**Issue #:**

**Description of changes:**
Report encode/decode time and if encoder/decoder is in hardware, and add encoder/decoder name as a metric dimension

**Testing:**
Added unit tests to cover the changes

**Checklist:**

y

n

n


src/clientmetricreport/GlobalMetricReport.ts
```diff
@@ -4,4 +4,5 @@
 export default class GlobalMetricReport {
   previousMetrics: { [id: string]: number } = {};
   currentMetrics: { [id: string]: number } = {};
+  stringValues: { [id: string]: string } = {};
```
hensmi-amazon, line 7: maybe `currentDimensions` or just `dimensions` instead of `stringValues`?

src/clientmetricreport/StreamMetricReport.ts
```diff
@@ -10,4 +10,5 @@ export default class StreamMetricReport {
   direction: Direction;
   previousMetrics: { [id: string]: number } = {};
   currentMetrics: { [id: string]: number } = {};
+  stringValues: { [id: string]: string } = {};
```
hensmi-amazon, line 13: same

src/clientmetricreport/ClientMetricReport.ts
```diff
@@ -297,6 +333,14 @@ export default class ClientMetricReport {
     },
     frameHeight: { transform: this.identityValue, type: SdkMetric.Type.VIDEO_DECODE_HEIGHT },
     frameWidth: { transform: this.identityValue, type: SdkMetric.Type.VIDEO_DECODE_WIDTH },
+    totalDecodeTime: {
```
hensmi-amazon, line 336: should this name be changed?

src/clientmetricreport/ClientMetricReport.ts
```diff
@@ -329,6 +373,22 @@ export default class ClientMetricReport {
     }
   }
 
+  /**
+   *  Dimensions derived from metric
+   */
+  readonly streamMetricDimensionMap: {
```
hensmi-amazon, line 379: Nit: Don't think we need 'Metric' in any of this or related naming?

hensmi-amazon (commented):
lgtm just some naming nits

src/clientmetricreport/GlobalMetricReport.ts
shi-su: I do not think they are just dimension, it can also be metric derived from string value. For example, whether decoder/encoder is in hardware is one of the derived metric.

src/clientmetricreport/ClientMetricReport.ts
shi-su: This key is just about the field name in webrtc report where the metric is derived from, we cannot change that naming as it comes from webrtc..

src/clientmetricreport/GlobalMetricReport.ts
shi-su: Updated it to `currentStringMetrics`. Also added comments that this is a workaround in case metric values are strings.

src/clientmetricreport/StreamMetricReport.ts
shi-su: Updated it to `currentStringMetrics`. Also added comments that this is a workaround in case metric values are strings.

src/clientmetricreport/ClientMetricReport.ts
shi-su: Updated naming as suggested.

hensmi-amazon (approved):
lgtm

src/clientmetricreport/ClientMetricReport.ts
```diff
@@ -126,6 +126,35 @@ export default class ClientMetricReport {
     return Number(metricReport.currentMetrics[metricName] * 1000);
   };
 
+  millisecondsPerSecond = (metricName?: string, ssrc?: number): number => {
```
ltrung, line 129: Can we name this to be clearer what it does? milliseconds per second is kinda confusing

shi-su: Thanks for the comment. Created https://github.com/aws/amazon-chime-sdk-js/pull/2443 to update function naming

PR #2446, opened 2022-09-20
Messaging session unauthorized bug fix

paparekh-amzn:
**Issue #:**
#2446

**Description of changes:**
- Fix behavior of websocket disconnects before a session is connected 
  - Session.start() method promise shall fail if there is a websocket disconnection before a session is established
  - Earlier, session.start() waited till session establish process began, now it blocks till session establish process completes
- Queue messages before SESSION_ESTABLISHED and publish them when a session is established to avoid dropping any messages
  - There was a possibility of dropping messages just in case some message arrived earlier than SESSION_ESTABLISHED message
  - We queue the message now and send it all together when a session is established

**Testing:**
Tested locally by using the messagingSession demo in demos/browser and simulating Unauthorized by providing an invalid UserArn. I saw the expected behavior of promise failing which was not the case before the change.

Tested Queueing of messages and sending it on Session establish by manually initializing a queue with some messages and observed SessionEstablished message along with Queue messages being published on Session establishment together.

Added Unit Tests.

- Setup AWS credentials for an AWS account which has Chime messaging SDK resources
  - vim ~/.aws/credentials and copy access id, secret key and session token
- `cd demos/browser`
- `npm run start --app=messagingSession`
- Go to http://127.0.0.1:8080/ in your Browser and Open Developer Console by Right click -> Inspect -> Console tab
- Provide any invalid UserArn, SessionId is optional and observe console logs using Inspector. There should be a log.error with reason code 4401.

**Checklist:**

Yes

No

No

Yes

### DIFF
--- a/demos/browser/package-lock.json
+++ b/demos/browser/package-lock.json
@@ -1,7204 +1,8 @@
 {
   "name": "amazon-chime-sdk-js-demo",
   "version": "2.0.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
-    "": {
-      "name": "amazon-chime-sdk-js-demo",
-      "version": "2.0.0",
-      "license": "Apache-2.0",
-      "dependencies": {
-        "@aws-sdk/client-chime-sdk-messaging": "^3.47.0",
-        "amazon-chime-sdk-js": "file:../..",
-        "aws-sdk": "^2.1216.0",
-        "bootstrap": "^5.1.3",
-        "compression": "^1.7.4",
-        "jquery": "^3.5.1",
-        "lodash": "^4.17.20",
-        "markdown-it": "^13.0.1",
-        "popper.js": "^1.15.0",
-        "util": "^0.12.4"
-      },
-      "devDependencies": {
-        "@types/bootstrap": "^5.1.6",
-        "@types/dom-mediacapture-record": "^1.0.11",
-        "autoprefixer": "^10.2.6",
-        "cross-spawn": "^7.0.3",
-        "csp-html-webpack-plugin": "^5.1.0",
-        "css-loader": "^6.2.0",
-        "html-webpack-plugin": "^5.3.2",
-        "open-iconic": "^1.1.1",
-        "postcss-loader": "^6.1.1",
-        "react-dev-utils": "^12.0.1",
-        "sass": "^1.26.11",
-        "sass-loader": "^12.1.0",
-        "style-loader": "^3.2.1",
-        "ts-loader": "^9.3.0",
-        "typescript": "4.2.3",
-        "webpack": "^5.51.1",
-        "webpack-cli": "^4.8.0",
-        "webpack-dev-server": "^4.9.0"
-      }
-    },
-    "../..": {
-      "version": "3.8.0",
-      "license": "Apache-2.0",
-      "dependencies": {
-        "@aws-crypto/sha256-js": "^2.0.1",
-        "@aws-sdk/client-chime-sdk-messaging": "^3.0.0",
-        "@aws-sdk/util-hex-encoding": "^3.47.0",
-        "@types/ua-parser-js": "^0.7.35",
-        "aws-sdk": "^2.1216.0",
-        "detect-browser": "^5.2.0",
-        "pako": "^2.0.4",
-        "protobufjs": "^6.11.3",
-        "resize-observer": "^1.0.0",
-        "ua-parser-js": "^1.0.1"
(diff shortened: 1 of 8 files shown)


src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -107,6 +111,14 @@ export default class DefaultMessagingSession implements MessagingSession {
   }
 
   private async startConnecting(reconnecting: boolean): Promise<void> {
+    await this.startConnectingInternal(reconnecting);
+    return await new Promise<void>((resolve, reject) => {
+      this.bootstrapResolved = resolve;
```
dpwspoon, line 116: Think you can just store the promise directly and then use .resolve or .reject on it: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject

paparekh-amzn: Tried with the approach you mentioned and faced some issues before I went with this approach recommended in SO.

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -224,24 +236,39 @@ export default class DefaultMessagingSession implements MessagingSession {
...
+          });
+        }
       } else if (!this.isSessionEstablished) {
         // SESSION_ESTABLISHED is not guaranteed to be the first message, and in rare conditions a message or event from
         // a channel the member is a member of might arrive prior to SESSION_ESTABLISHED.  Because SESSION_ESTABLISHED indicates
-        // it is safe to bootstrap the user application with out any race conditions in losing events we opt to drop messages prior
-        // to SESSION_ESTABLISHED being received
+        // it is safe to bootstrap the user application without any race conditions in losing events we opt to store messages prior
+        // to SESSION_ESTABLISHED being received and send when once SESSION_ESTABLISHED.
+        this.preBootstrapMessages.push(message);
```
ltrung, line 255: On average  how many messages would be in this queue and how often this happens? Do we need to limit the size of this queue?

demos/browser/package-lock.json
```diff
@@ -1,7204 +1,8 @@
 {
   "name": "amazon-chime-sdk-js-demo",
   "version": "2.0.0",
-  "lockfileVersion": 2,
```
simmkyu, line 4: Please use the later versions of NPM. The`lockfileVersion` should remain the same.

src/messagingsession/DefaultMessagingSession.ts
simmkyu, line 116: In a slow (or almost offline) internet connection, the `startConnecting` will take a while to resolve. I wonder if we need to add a timeout or retry logic.

Also, if customers rely on the existing behavior, they may experience unexpected user experiences. For example, if customers update the UI after `await messagingSession.start()`, they will see some delay now.

paparekh-amzn: This seems like an existing issue even without this change. I agree that a timeout or retry logic shall be helpful but I think that could be tackled separately considering this fix is a result of a customer ticket.

demos/browser/package-lock.json
paparekh-amzn: Sure, will do

src/messagingsession/DefaultMessagingSession.ts
paparekh-amzn: We believe this to be a rare race condition after introducing prefetch feature. Maximum number of messages we sent via Prefetch is 50.

src/messagingsession/DefaultMessagingSession.ts
simmkyu: Right. The `startConnecting` previously called other async APIs like `getMessagingSessionEndpoint`. `getMessagingSessionEndpoint` will be slow too.

ltrung (commented):
Can we update the testing instructions step-by-step so QA could follow?

ltrung (approved):

src/messagingsession/DefaultMessagingSession.ts
```diff
@@ -51,6 +54,7 @@ export default class DefaultMessagingSession implements MessagingSession {
 
     CSPMonitor.addLogger(this.logger);
     CSPMonitor.register();
+    this.preBootstrapMessages = [];
```
dpwspoon, line 57: nit bootstrap isn't a thing.  Let's use a different term like sessionEstablished

simmkyu (approved):

dpwspoon (approved):

Issue #2374, opened 2022-07-21
Simultcast - `remoteVideoSourcesDidChange` called with incorrect number of videoSources


kelvin2200:
### What happened and what did you expect to happen?

3rd attempt to enable simulcast in our app.

`remoteVideoSourcesDidChange` is called with an incorrect number of videoSources

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

We haven't tested to see if this can be reproduced in a simple setup, so it may apply only to our app.

Using:
- `DefaultSimulcastUplinkPolicy`.
- `config.serverSideNetworkAdaption = ServerSideNetworkAdaption.BandwidthProbing`

Layout (devices are of no relevance, except for the number of attendees):
- Device A (local)
- Device B (remote)
- Device C (remote)
- all devices have their cameras ON

When **Device C** hits refresh : 

- **A** will see **B** as having turned off its camera (also its stream disappears), though **B** has not taken any action.
- `remoteVideoSourcesDidChange` is triggered twice; once with the remaining remote video source (**B**) and the second time with an empty array.
- weirdly enough, we can't reproduce in less or more than 3 attendees
- cannot reproduce when not using simulcast

### Amazon Chime SDK for JavaScript version

3.6.0

### What browsers are you seeing the problem on?

any

### Browser version

any

### Browser console logs

no logs needed

hensmi-amazon:
Hi @kelvin2200 we will try to reproduce. Is this reproducible without `config.serverSideNetworkAdaption = ServerSideNetworkAdaption.BandwidthProbing` ?

xuesichao:
I was able to reproduce this issue using JS SDK browser demo.

Steps:
1. Three attendees (1, 2, 3) join the meeting enabling `simulcast`, `priorityBasedDownlinkPolicy` with `BandwidthProbing`.
2. All of three enable camera.
3. Refresh page in the window of attendee 3.
4. From attendee 1's window, attendee 3's video tile disappears, and attendee 2's video tile disappears shortly and appears.
[...]

PR #2452, opened 2022-09-22
 Fixed unnecessarily long freezes when switching simulcast streams.

hensmi-amazon:
**Issue #:** https://github.com/aws/amazon-chime-sdk-js/issues/2374 

**Description of changes:**

* Fix `active` check in demo which unnecessarily hides videos during rebind. Upon inspection...I found `active` to be quite arbitrarily calculated so I added some recommendations to docstring.
* Use `subscribeFrameSent` on remote video updates so we will cache the previous index and be able to skip subscribes on simulcast stream switches when senders have switched layers.

Note i have made some backend changes too, which should be out within a week or so, but those are not coupled to these changes, just solve similar problems.

**Testing:**
Ignoring backend fixes, these two changes can be verified with the following reduced steps.

* Start 2 attendees, one with simulcast. No other settings need to be enabled.
* Start sending on the simulcast attendee.
* Join the conference with a 3rd attendee.
* Leave the conference with the 3rd attendee.
* Verify no pauses (e.g. 1 second or longer, ofc depends on RTT).

See steps above.

**Checklist:**

y

n

n


ltrung (approved):

src/videotile/VideoTileState.ts
```diff
@@ -26,7 +26,9 @@ export default class VideoTileState {
   isContent: boolean = false;
 
   /**
-   * Indication of whether the tile has active video stream.
+   * Indication of whether the tile has active video stream. This value may be somewhat arbitrary in how it is calculated,
+   * it is recommended to used the more specific values (e.g. `paused`, `boundVideoElement`, and `boundVideoStream`) to
```
devalevenkatesh, line 30: nit: used the -> use

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -952,6 +952,7 @@ export default class DefaultAudioVideoController
     if (added.length !== 0 || removed.length !== 0) {
       return false;
     }
+    context.videoStreamIndex.subscribeFrameSent();
```
devalevenkatesh, line 955: [For understanding]: So the above `context.signalingClient.remoteVideoUpdate` means sending of a subscribe frame?

hensmi-amazon: Its not actually sending a subscribe, I will update the comment in docstring and above usage. Alternatively I could create an alias function `cacheCurrentIndex` but maybe will save for a new major rev :)

devalevenkatesh: Cool, thanks for explaining and elaborating the docstring.

devalevenkatesh (approved):

PR #2461, opened 2022-09-30
Add the maximum audio output gain for the readiness checker's audio o…

simmkyu:
…utput test

**Issue #:**
N/A

**Description of changes:**
- Add the `audioOutputGain` and `audioOutputFreqency` fields (optional) in the MeetingReadinessCheckerConfiguration. Customers can use this value to adjust the Play Tone sound in the readiness checker.

**Testing:**
- Tested it by passing a MeetingReadinessCheckerConfiguration object with updated fields.

- No, I haven't included it in a demo application.

**Checklist:**

Yes

Yes

No


xuesichao (dismissed):

devalevenkatesh (approved):

src/meetingreadinesschecker/MeetingReadinessCheckerConfiguration.ts
```diff
@@ -16,4 +16,14 @@ export default class MeetingReadinessCheckerConfiguration {
    * If null, it will use the default value.
    */
   waitDurationMs: number = 3000;
+
+  /**
+   * Specify the frequency of oscillation in hertz for the audio output test.
+   */
+  audioOutputFrequency?: number = 440;
```
ltrung, line 23: Do we need optional `?` if we already use default value?

simmkyu: Without `?`, TS throws
> Property 'audioOutputFrequency' is missing in type 'Child' but required in type 'Parent'.

```
class Parent {
  name: string = 'hey';
  age: number = 12
  audioOutputFrequency = 440;
}

class Child implements Parent {
  name = 'Name';
  age = 99;
}
```


I don't think SDK users implement this configuration, but using `?` ensures backward compatibility.

ltrung: Interesting, I thought implements only use for interface :)

ltrung (approved):

PR #2469, opened 2022-10-12
simplified media capture demo deployment and added automatic concatenation for media capture demo

tenge-amzn:
**Description of changes:**
1. Some of the default boolean parameter like `useChimeSdkMeetings` or `useChimeSdkMediaPipelines` didn't work because they were set as boolean and compared as string in the `ensureRegion()`. Fixed.
2. Rewrote the media live connector and media capture section in Readme.
3. Enabled media capture video compositing for the demo.
4. Added concatenation for media capture.
5. Removed the complicated media capture s3 bucket regional parameter, as media capture no longer requires a specific bucket region, and this simplifies the deployment workflow.
6. Moved the media capture bucket creation to cloudformation, so that we don't need to create a new stack every time. Just updating the previous stack. The resource creation depends on the condition that a bucket name is passed in.
7. Decoupled media pipeline service principal and meetings SP. 

**Testing:**
Deployed and worked fine.

**Checklist:**

Have you successfully run `npm run build:release` locally?
Yes

Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved?
No.

Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved?
No

Yes

### DIFF
--- a/demos/serverless/deploy.js
+++ b/demos/serverless/deploy.js
@@ -14,7 +14,8 @@ let enableTerminationProtection = false;
 let disablePrintingLogs = false;
 let chimeEndpoint = 'https://service.chime.aws.amazon.com';
 let chimeSDKMeetingsEndpoint = 'https://service.chime.aws.amazon.com';
-let chimeMediaPipelinesServicePrincipal = 'mediapipelines.chime.amazonaws.com'
+let chimeMediaPipelinesServicePrincipal = 'mediapipelines.chime.amazonaws.com';
+let chimeMeetingsServicePrincipal = 'chime.amazonaws.com';
 
 // Supported regions for the Amazon Chime SDK Meetings namespace
 // https://docs.aws.amazon.com/chime/latest/dg/sdk-available-regions.html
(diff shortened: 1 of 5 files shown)


demos/serverless/template.yaml
```diff
@@ -74,10 +75,60 @@ Globals:
...
+                - - 'arn:aws:s3:::'
+                  - !Ref ChimeMediaCaptureS3Bucket
+              - !Join
+                - ''
+                - - 'arn:aws:s3:::'
+                  - !Ref ChimeMediaCaptureS3Bucket
+                  - '/*'
+            Principal: 
+              Service: 
+                - 'mediapipelines.chime.amazonaws.com'
```
lognoel, line 122: SP should be variable

lognoel (changes requested):
please make SP variable

demos/serverless/deploy.js
```diff
@@ -70,8 +51,7 @@ function usage() {
   console.log(`  -p, --service-principal              Service principal for media pipelines related resources, default is '${chimeMediaPipelinesServicePrincipal}'`)
   console.log(`  -t, --enable-termination-protection  Enable termination protection for the Cloudformation stack, default is false`);
   console.log(`  -l, --disable-printing-logs          Disable printing logs`);
-  console.log(`  -o, --capture-output-prefix          Prefix for S3 bucket name`);
-  console.log(`  -i, --opt-in-regions                 Comma separated list of additional opt-in regions to enable for media capture`);
```
anuranduttaroy, line 74: This will break the [deploy-canary-demo](https://github.com/aws/amazon-chime-sdk-js/blob/main/script/deploy-canary-demo) script as discussed offline.

anuranduttaroy (changes requested):

yuqil-amzn:
Also I was told by sdk team them fork and merge doesn't trigger the CI/CD (could confirm with the team). Maybe you can just create a branch and have PR.

demos/serverless/deploy.js
```diff
@@ -221,60 +209,13 @@ function ensureTools() {
...
-  }
-}
-
 function copyAssets() {
   fs.copySync('../browser/dist/speech.mp3', 'src/speech.mp3');
   fs.copySync('../browser/dist/speech_stereo.mp3', 'src/speech_stereo.mp3');
 }
 
 function ensureRegion() {
-  if (useChimeSDKMeetings === 'true') {
```
devalevenkatesh, line 277: The `useChimeSDKMeetings` and similar variables comes from CloudFormation and CloudFormation parameters do not support boolean, thus String was used and compared here.

tenge-amzn: why line 7 initiated as boolean?

tenge-amzn:
need to park this, and resubmit after we move the canary out of here.

PR #2472, opened 2022-10-17
Add average bitrate as a second criterion for stream selection.

shi-su:
**Issue #:**

**Description of changes:**
This is useful for the SVC stream selection usecase where the maximum bitrate advertised is the same.

**Testing:**

This change is not testable at the time as VP9 SVC is not supported. Will add QA test cases prior to SVC rollout.

**Checklist:**

y

n

n

### DIFF
--- a/src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
+++ b/src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
@@ -286,7 +286,10 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
     // Sort streams by bitrate ascending.
     remoteInfos.sort((a, b) => {
       if (a.maxBitrateKbps === b.maxBitrateKbps) {
-        return a.streamId - b.streamId;
+        if (a.avgBitrateKbps === b.avgBitrateKbps) {
+          return a.streamId - b.streamId;
+        }
+        return a.avgBitrateKbps - b.avgBitrateKbps;
       }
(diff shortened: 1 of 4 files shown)


src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -286,7 +286,10 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
     // Sort streams by bitrate ascending.
     remoteInfos.sort((a, b) => {
       if (a.maxBitrateKbps === b.maxBitrateKbps) {
-        return a.streamId - b.streamId;
+        if (a.avgBitrateKbps === b.avgBitrateKbps) {
```
hensmi-amazon, line 289: This actually has a very significant downside in that variation in sender bitrate will cause the received video selection to flip flop. E.g. if you are picking between two 200 kbps streams (1, 2), and your downlink bitrate is 200 kbps you will always subscribe to 1. However after this change, if 2 starts sending 199 kbps it will pause 1 and pick 2. Then if 2 starts sending 201 kbps it will reverse again.

It's a bit of a contrived example but hopefully makes the issue clear, we already have enough complaints for the policy flip-flopping (especially before bandwidth probing). Is it possible to update the max bitrates from the backend?

shi-su: Yes, this might trigger more flip-flopping. Let me look into how we may calculate max bitrate for downsampled SVC streams. We can set max bitrate from the backend, but calculating it could be a little tricky.

PR #2474, opened 2022-10-18
Publish audio and video failed event if device label trigger fails

ltrung:
**Description of changes:**
Publish audio and video failed event if device label trigger fails. Previously if list audio and video devices fail we do not register any audio/video failed meeting events.

Also in Firefox, even if users deny permissions to audio and video input, device Id are still populated (in Chrome, they are blank) so I update the demo to not use the deviceId if the device label is empty (i.e., no permission) so the demo does not crash when it tries to start audio/video input with the fake deviceId.
**Testing:**

Use serverless demo and join a meeting but deny permission to audio/video input. You should see a log_event requests on the consoles for audio input failed and video input failed event. You may also see an audio input selected event as we select a dummy audio stream.

**Checklist:**


michhyun1 (dismissed):

hensmi-amazon (approved):

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -826,6 +826,12 @@ export default class DefaultDeviceController
         }
       } catch (err) {
         this.logger.info('unable to get media device labels');
+        this.eventController?.publishEvent('audioInputFailed', {
```
devalevenkatesh, line 829: Just checking if firefox sends two same `audioInputFailed` when the permissions are blocked or a single one.

ltrung: I mentioned on the ticket but the reason Firefox sent previously is due to we tried to select a device with empty label so i fixed that in the demo.

devalevenkatesh (approved):

PR #2522, opened 2022-12-09
Remove UserJoinedMeetingCheck and do it as part of the JoinMeeting Step instead

michhyun1:
**Issue #:**

**Description of changes:**
We don't need UserJoinedMeetingCheck, since we always do it in the Join Meeting step anyways. 

Also, I noticed that we used to look for the "meeting-flow" id, and wait until it is visible, which consistently took a minute or longer - even though the saucelabs video shows that the in-meeting view is rendered. We can just wait until the meeting's  <audio> element is able to be located (not visible) so that we can confirm that the in meeting view is rendered. 

**Testing:**

**Checklist:**


integration/js/ContentShareOnlyAllowTwoTest.js
```diff
@@ -15,7 +20,16 @@ const { v4: uuidv4 } = require('uuid');
  */
 class ContentShareOnlyAllowTwoTest extends SdkBaseTest {
   constructor(name, kiteConfig) {
-    super(name, kiteConfig, "ContentShareOnlyAllowTwoTest");
+    super(name, kiteConfig, 'ContentShareOnlyAllowTwoTest');
+  }
+
+  // Commented out the UserJoinedMeetingCheck because we already check it in the JoinMeetingStep
+  async addUserToMeeting(attendee_id, sessionInfo) {
```
devalevenkatesh, line 27: Checking if this method is added, because SDKTestUtils one has that last `UserJoinedMeetingCheck`?

devalevenkatesh (approved):

integration/js/ContentShareOnlyAllowTwoTest.js
michhyun1: that utils method is used in a lot of other places but I needed to change it, so instead of changing that, I created a new one here without one step

PR #2526, opened 2022-12-14
Add destroy method to VoiceFocus

richhx:
**Issue #:** N/A

**Description of changes:**
During the `VoiceFocus.init()`, a worker is spawned to help load the models and decide which execution model to take (e.g. inline, SAB, etc). Currently, the only way to destroy that worker is by creating an `AudioWorkletNode` and stopping the node in an `AudioContext`. As a result, it's possible that this worker leaks as we've seen when a client constantly calls `VoiceFocus.init()`.

This revision looks to address this problem by adding a simple `destroy()` method that will force terminate the worker. For example, this can be used in the case of the React component library where the worker needs to be terminated on component unmount.

This revision will also introduce a static `destroyVoiceFocus` method inside of the `VoiceFocusDeviceTransformer` class. The reason it must be static is because the `create` function is also static. 

This will allow us to stop the worker thread associated with a Transformer.

**Testing:**
Verified that internal unit tests pass.

Manually call `VoiceFocus.init()` in the demo application in a loop. Verify that workers are  constantly created. After you add the `vf.destroy()` call, the workers should no longer leak.

**Checklist:**

Yes.

No.

No.

Yes


libs/voicefocus/voicefocus.js
```diff
@@ -292,6 +299,13 @@ class VoiceFocus {
             return node;
         });
     }
+    destroy() {
```
devalevenkatesh, line 302: Checking how this can be used in our meetingV2 demo? Accompanying demo changes will help in testing and understanding the usage. Also, update to the documentation if needed would also be helpful.

michhyun1:
I'll checkout this pull request and test it and get back to you!

libs/voicefocus/voicefocus.js
richhx: It's technically not needed in the meetingV2 demo since it creates a voice focus device and the device is stopped when the participant leaves (thus cleaning up the worker): https://github.com/aws/amazon-chime-sdk-js/blob/main/demos/browser/app/meetingV2/meetingV2.ts#L1826.

We can potentially use the destroy by adding it to the transformer first and then calling it here https://github.com/aws/amazon-chime-sdk-js/blob/7428c620fa0233d5c8e88587cea5bbd2068a1409/demos/browser/app/meetingV2/meetingV2.ts#L480-L483 in the case that VF is not supported.

I'll wait for @michhyun1 to  give this a try and verify the fix before making those changes though.

michhyun1:
I think you need to do a `git pull origin main --rebase`, its very behind main branch. 

Other than that, I tested this change, and confirmed it works! I see in the Task Manager that the worker thread is terminated as expected.

CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
ltrung, line 8: Current unreleased version is 3.10.0.

guides/09_Amazon_Voice_Focus.md
```diff
@@ -369,6 +369,19 @@ await deviceController.startAudioInput(this.voiceFocusDevice);
 console.log('Amazon Voice Focus switched to new device', newDevice);
 ```
 
+## Destroying Amazon Voice Focus worker thread from your application
+
+In some cases, you might want to completely stop the worker thread that runs the `VoiceFocusDeviceTransformer`. In that case, you can use the `destroyVoiceFocus` method to completely remove the voice focus worker thread. After destroying the `VoiceFocusDeviceTransformer`, you'd now have to create a new one. This might be useful if you intend to completely remove Voice Focus from your application at some point.
```
ltrung, line 374: Could we give a specific example and when we should use this method here?

guides/09_Amazon_Voice_Focus.md
```diff
@@ -369,6 +369,19 @@ await deviceController.startAudioInput(this.voiceFocusDevice);
 console.log('Amazon Voice Focus switched to new device', newDevice);
 ```
 
+## Destroying Amazon Voice Focus worker thread from your application
+
+In some cases, you might want to completely stop the worker thread that runs the `VoiceFocusDeviceTransformer`. In that case, you can use the `destroyVoiceFocus` method to completely remove the voice focus worker thread. After destroying the `VoiceFocusDeviceTransformer`, you'd now have to create a new one. This might be useful if you intend to completely remove Voice Focus from your application at some point. If there was a `VoiceFocusTransformerDevice` associated with the destroyed transformer, that `VoiceFocusTransformerDevice` will stop working, as the transformer has been destroyed. You can also call `VoiceFocusTransformerDevice.stop()` to stop the worker thread, but that assumes that you've already created a `VoiceFocusTransformerDevice`. If you havent created a `VoiceFocusTransformerDevice` already, you should use `VoiceFocusDeviceTransformer.destroyVoiceFocus` to destroy the Voice Focus worker thread.
```
simmkyu, line 374: - `the voice focus worker thread` => `the Voice Focus worker thread
- `havent` => `haven't`

guides/09_Amazon_Voice_Focus.md
```diff
@@ -369,6 +369,19 @@ await deviceController.startAudioInput(this.voiceFocusDevice);
...
+
+```typescript
+const transformer = VoiceFocusDeviceTransformer.create(spec, { logger });
```
simmkyu, line 377: Use `await` since `create` returns a promise.

guides/09_Amazon_Voice_Focus.md
```diff
@@ -369,6 +369,19 @@ await deviceController.startAudioInput(this.voiceFocusDevice);
...
+
+VoiceFocusDeviceTransformer.destroyVoiceFocus(transformer);
```
simmkyu, line 379: Use `await` too. Otherwise, the `stopped` message is incorrect until a promise is resolved.

libs/voicefocus/voicefocus.js
```diff
@@ -255,6 +256,9 @@ class VoiceFocus {
     }
     createNode(context, options) {
         var _a;
+        if (this.internal.isDestroyed) {
+            throw new Error("Unable to create node because VoiceFocus worker has been destroyed.");
```
simmkyu, line 260: Use single quotes. Do you have a corresponding TypeScript file that outputs to this JavaScript?

libs/voicefocus/voicefocus.js
```diff
@@ -292,6 +299,13 @@ class VoiceFocus {
             return node;
         });
     }
+    destroy() {
+        const { worker } = this.internal;
+        if (worker) {
+            worker.terminate();
+        }
```
simmkyu, line 306: You can use optional chaining here too.
```
this.internal.worker?.terminate();
```

src/voicefocus/VoiceFocusDeviceTransformer.ts
```diff
@@ -228,6 +229,14 @@ export class VoiceFocusDeviceTransformer {
     }
   }
 
+  /**
+   * Destroy the Voice Focus instance and worker thread associated with the transformer.
+   */
+  static async destroyVoiceFocus(transformer: VoiceFocusDeviceTransformer): Promise<void> {
+    /* istanbul ignore next */
```
simmkyu, line 236: Avoid skipping coverage; writing tests, if possible, is recommended.

src/voicefocus/VoiceFocusDeviceTransformer.ts
```diff
@@ -301,8 +310,8 @@ export class VoiceFocusDeviceTransformer {
     preload: boolean
   ): Promise<[VoiceFocus, VoiceFocusTransformDeviceDelegate]> {
     const delegate = new VoiceFocusTransformDeviceDelegate();
-    const vf = await VoiceFocus.init(config, { delegate, preload, logger: this.vfLogger });
-    return [vf, delegate];
+    this.vf = await VoiceFocus.init(config, { delegate, preload, logger: this.vfLogger });
```
simmkyu, line 313: If someone calls this internal method consecutively, could there be any memory leaks since subsequent calls merely reassign a new VF?

test/voicefocus/VoiceFocusDeviceTransformer.test.ts
```diff
@@ -232,6 +232,26 @@ describe('VoiceFocusDeviceTransformer', () => {
...
+            spec,
+            {
+              preload: false,
+              logger,
+            },
+            configWithURL
+          );          
+          await VoiceFocusDeviceTransformer.destroyVoiceFocus(transformer);
+          expect(transformer.isSupported()).to.be.true;
+          expect(transformer).to.not.be.undefined;
```
simmkyu, line 251: How do these two `expect` calls verify that the transformer has been destroyed?

guides/09_Amazon_Voice_Focus.md
michhyun1: oh yes, thanks.

libs/voicefocus/voicefocus.js
michhyun1: cool!

src/voicefocus/VoiceFocusDeviceTransformer.ts
michhyun1: I just need to skip the transformer.vf? case. transformer.vf should always be non-null, but added a `?` after vf just in case.

src/voicefocus/VoiceFocusDeviceTransformer.ts
michhyun1: Yes, but technically the create() funciton inside VoiceFocusDeviceTransformer is a static function, so someone should be able to create multiple transformers already - that is not a bug, that is how it was originally implemented. The piece we were missing is a static destroy() function, which I added in this PR.

Now, at least developers can tie up any loose memory leaks that they might've opened if they created multiple VoiceFocusDeviceTransformers.

test/voicefocus/VoiceFocusDeviceTransformer.test.ts
michhyun1: fixed

ltrung (approved):

PR #2529, opened 2022-12-20
Added new video metrics

paulhinks:
**Issue #:**

**Description of changes:**

**Testing:**

**Checklist:**

Yes.

No.

No.


src/clientmetricreport/ClientMetricReport.ts
```diff
@@ -341,6 +369,11 @@ export default class ClientMetricReport {
       transform: this.isHardwareImplementation,
       type: SdkMetric.Type.VIDEO_DECODER_IS_HARDWARE,
     },
+    freezeCount:  { transform: this.deltaValue, type: SdkMetric.Type.VIDEO_FREEZE_COUNT },
```
hensmi-amazon, line 372: `countPerSecond` seems like what should be used here and below. I assume this is so, if we wanted we could get an accurate accumulation on the backend? If thats the case we need to include duration on `SdkStreamMetricFrame` so that we can get per second values. Though i recommend just  using `countPerSecond` for now for consistency and saving the latter for a followup PR where we change all of the metrics.

src/clientmetricreport/ClientMetricReport.ts
```diff
@@ -341,6 +369,11 @@ export default class ClientMetricReport {
...
+    totalFreezesDuration:  { transform: this.deltaValueMS, type: SdkMetric.Type.VIDEO_FREEZE_DURATION },
```
hensmi-amazon, line 373: Same for this and `averageTimeSpentPerSecondInMilliseconds`

src/clientmetricreport/ClientMetricReport.ts
```diff
@@ -440,6 +473,36 @@ export default class ClientMetricReport {
       media: MediaType.VIDEO,
       dir: Direction.UPSTREAM,
     },
+    videoQualityReason: {
```
hensmi-amazon, line 476: As in other PRs, these should have unit in name

src/clientmetricreport/ClientMetricReport.ts
paulhinks: The frame does have a duration in it, so this value can be used directly. I was hoping not to have to take an average of an average. If consistency is more important I can do that for now.

paulhinks: Changed transforms to countPerSecond and averageTimeSpentPerSecondInMilliseconds. Added Ms to Duration metrics name. Added an enum for the video quality reasons.

hensmi-amazon (approved):

PR #2558, opened 2023-01-25
Add flag to MeetingSessionConfiguration to disable periodic keyframe requests to content sender

hensmi-amazon:
**Issue #:** None

**Description of changes:** Title explains this, and docstring provides more context:
```
  /**
   * The Amazon Chime SDK media backends currently send a keyframe request to content share senders every 10 seconds to help mitigate
   * decoder issues on receivers. This flag requests the backed to disable that feature.
   *
   * Setting this flag to true may or may not lead to issues with content received for your application, as it depends on browsers used
   * and whether they have fixed previous issues leading to the introduction of this periodic keyframe request. It will however
   * reduce CPU consumption on content senders which no longer have to generate as many expensive keyframes.
   */
  disablePeriodicKeyframeRequestOnContentSender: boolean = false;
```

**Testing:** Tested with backend changes to confirm keyframe request only exists when flag is not existent or false.

1. Enable checkbox `Disable periodic keyframe request to content senders` in additional options.
2. Join meeting and enable content share.
3. In WebRTC internals, the keyframes sent metric should stay flat:
Screen Shot 2023-01-24 at 4 04 53 PM

**Checklist:**

y

y

y


ltrung (commented):
What is the reason behind this flag?  Not sure what the problem we are trying to solve by turning off this behavior.
Nvm I missed the documentation apparently

PR #2561, opened 2023-01-25
Update meeting readiness checker logic for fast join

aman-aguptaaa:
**Issue #:** None

**Description of changes:**
*What?*
- Updated `checkAudioConnectivity` logic in meeting readiness checker to check audio packets sent and audio packet received metrics in addition to attendee presence flag, as part of fast join changes. In case both audio packets sent and audio packet received metrics are greater than 0, and attendee presence flag is set to true, then only audio connection will be categorized as successful.
- Added no Audio microphone device option inside meetingReadinessChecker demo app to simulate the failure of `checkAudioConnectivity`

*Why?*
- As part of changes being made to support listen only attendees use cases, NoDevice use case, and allow clients with microphone issues, where mic packet does not go out quickly during setup, to atleast join the call asap and receive conference audio, we want to update meeting readiness checker logic accordingly.

**Testing:**

- Running MeetingReadinessChecker Demo App in the browser with a genuine microphone option selected, should pass the Audio Connectivity check in the app.
- Running MeetingReadinessChecker Demo App in the chrome browser with no Audio microphone option selected, should fail the Audio Connectivity check in the app with reason `AudioNotReceived`.

**Checklist:**

Yes

N/A

N/A


CHANGELOG.md
```diff
@@ -9,10 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added observer for audio packets sent and audio packet received metrics inside `checkAudioConnectivity` in meeting readiness checker
```
devalevenkatesh, line 12: nit:
- Added -> Add.
- End all these CHANGELOGs with a "."

CHANGELOG.md
```diff
@@ -9,10 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
 ### Removed
 
 ### Changed
 
+- Update `checkAudioConnectivity` logic in meeting readiness checker to check audio packets sent and audio packet received metrics in addition to attendee presence flag, as part of fast join changes
```
devalevenkatesh, line 18: You can drop ", as part of fast join changes" or you can re-word this to:

"To support fast join, update `DefaultMeetingReadinessChecker`'s `checkAudioConnectivity` logic to check for audio packets sent and received metrics using `metricsDidReceive` observer in addition to attendee presence condition"

With this you can drop the above "Add observer..." changelog I think.

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
```diff
@@ -270,7 +270,16 @@ export default class DefaultMeetingReadinessChecker implements MeetingReadinessC
 
   async checkAudioConnectivity(audioInputDevice: Device): Promise<CheckAudioConnectivityFeedback> {
     let audioPresence = false;
+    const checkAudioConnectivityMetrics = {
```
devalevenkatesh, line 273: Is just `audioConnectivityMetrics` fine? The underlying attributes can also be just `packetsSent` and `packetsReceived` as the name of this JS object already has audio naming in it:
```
const audioConnectivityMetrics = {
  packetsSent: 0,
  packetsReceived: 0,
}
```

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
```diff
@@ -270,7 +270,16 @@ export default class DefaultMeetingReadinessChecker implements MeetingReadinessC
...
+      audioPacketsSent: 0,
+      audioPacketsReceived: 0,
+    };
     const audioVideo = this.meetingSession.audioVideo;
+
+    const checkAudioConnectivityMetricsObserver: AudioVideoObserver = this.getCheckAudioConnectivityMetricsObserver(
```
devalevenkatesh, line 279: Probably inline the observer here itself like the `checkVideoConnectivity` approach over a class method? 

If we are making it a class level method then would be better to make it re-usable so that `metricsDidReceive` is attached once and populates both audio and video metrics and each check relies on that. That would be out of scope for this though.

test/dommock/DOMMockBehavior.ts
```diff
@@ -68,7 +68,7 @@ export default class DOMMockBehavior {
   mediaDeviceOnDeviceChangeSupported: boolean = true;
   rtcPeerConnectionGetStatsSucceeds: boolean = true;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  rtcPeerConnectionGetStatsReport: { [name: string]: any } = {};
+  rtcPeerConnectionGetStatsReports: [{ [name: string]: any }] = [{}];
```
devalevenkatesh, line 71: For array we can follow the existing pattern: 
```
rtcPeerConnectionGetStatsReports: { [name: string]: any }[] = [{}];
```

test/dommock/DOMMockBuilder.ts
```diff
@@ -858,19 +858,24 @@ export default class DOMMockBuilder {
...
-            id: 'RTCInboundRTPAudioStream',
-            ssrc: 1,
-            timestamp: 1,
-            type: 'inbound-rtp',
-            kind: 'video',
-            ...mockBehavior.rtcPeerConnectionGetStatsReport,
-          },
-        ];
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const reports: [{ [name: string]: any }] = [{}];
```
devalevenkatesh, line 862: Same comment for the array type above.

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
aman-aguptaaa: Will create inline observer for now.

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- To support fast join, update DefaultMeetingReadinessChecker's checkAudioConnectivity logic to check for audio packets sent and received metrics using metricsDidReceive observer in addition to attendee presence condition.
+- Modify `rtcPeerConnectionGetStatsReport` variable inside `DOMMockBehavior` class to `rtcPeerConnectionGetStatsReports` to contain multiple getStatsReports.
```
ltrung, line 17: We don't need to include changes in unit tests

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- To support fast join, update DefaultMeetingReadinessChecker's checkAudioConnectivity logic to check for audio packets sent and received metrics using metricsDidReceive observer in addition to attendee presence condition.
```
ltrung, line 16: What does fast join mean? Don't think we have any documentation mentions that term so you would want to explain in more details.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- To support use case that allows attendees with mic issue to at least join the call asap and receive conference audio, update `checkAudioConnectivity` logic inside `DefaultMeetingReadinessChecker` accordingly, to check for audio packets sent and received metrics using `metricsDidReceive` observer in addition to attendee presence condition.
```
pracheth, line 16: Maybe say `To support future use case`?

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
```diff
@@ -270,7 +270,25 @@ export default class DefaultMeetingReadinessChecker implements MeetingReadinessC
...
+      packetsSent: 0,
+      packetsReceived: 0,
+    };
     const audioVideo = this.meetingSession.audioVideo;
+
+    const checkAudioConnectivityMetricsObserver: AudioVideoObserver = {
+      metricsDidReceive(clientMetricReport: ClientMetricReport) {
+        clientMetricReport.getRTCStatsReport().forEach(report => {
+          if (report.type === 'outbound-rtp' && report.mediaType === 'audio') {
+            audioConnectivityMetrics.packetsSent = report.packetsSent;
```
pracheth, line 283: Hmm why are we checking packets sent here as part of connectivity check? The idea is to allow an attendee to join the call even if they are not sending any audio packets. So a check for packets being received, in addition to the presence, should be enough.

aman-aguptaaa: Please correct me if I am wrong here. My understanding is that - Prior to the fast join, attendee presence used to indicate that the audio packet is being sent and received by the attendee. So, as part of *AudioConnectivity* check in meetingReadinessChecker, we want to check whether the attendee is able to send and receive the audio packets. 
Now, after fast join, audio packets sent and received in addition to attendee presence should account for audio connectivity check. 
And if we are to show the connectivity where attendee can join the call even if they are not sending any audio packets, we should show a separare connectivity check - Say *AudioReceiveConnectivity* check inside meetingReadinessChecker. What do you think ?

pracheth: Currently, with fast join not enabled, attendee presence message is sent out by the server on receiving an audio packet from the client. So the connectivity check today, checks one way audio. 

When we deploy fast join, presence message will be received by client on the signaling channel without it needing to send out an audio packet. But to ensure the clients media channel is connected, we are adding the check for received packets. 

Sending of packets is very dependent on the audio device being used. I think sending packet health should be a separate check and audio connectivity check should include the presence and received audio check.

aman-aguptaaa: Thanks for the clarification.

avinashmidathada: I agree with Pracheth's thought process here,but given that [`checkVideoConnectivity`](https://github.com/aws/amazon-chime-sdk-js/blob/553c2d2f8f18abd2c3daef487ac5e4173afa93c6/src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts#L312-L353)  uses packetsSent to determine success and returns a VideoNotSent if it fails, I'm inclined to just keep a single connectivity check and add AudioNotSent to the CheckAudioConnectivityFeedback enum and return that if packetsSent is not greater than zero. Thoughts?

aman-aguptaaa: Agree, packetsSent metric in addition to attendee Presence can indicate audio connectivity, plus the packetsSent metric is consistent with the current checkVideoConnectivity as well.

avinashmidathada: I agree "Connectivity" is not clearly defined here. But, I think by checking all three things, we'll cover all grounds and your rationale makes sense to me.

pracheth: What if a listen only attendee is using the readiness checker before joining the meeting? Given that we won't require builders to have a "fake" audio stream sending muted packets soon, this attendees readiness check will fail as they will never send out packets.

pracheth: Previously, rather currently, the `checkAudioConnectivity` function only checked for presence and given that the server waits for a mic packet to send that out today, it was technically on testing the sending of audio. We weren't checking for packets being received anywhere. Note that the presence message is received by the client on the signaling channel.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- To support future use case that allows attendees with mic issue to at least join the call asap and receive conference audio, update `checkAudioConnectivity` logic inside `DefaultMeetingReadinessChecker` accordingly, to check for audio packets received metrics using `metricsDidReceive` observer in addition to attendee presence condition. Previously the checkAudioConnectivity used to implicitly test that audio was both being sent out and being received but now it's only checking that the audio is being received successfully.
```
pracheth, line 16: Couple of edits: 
`To support future use case of allowing users to join a meeting without the need for a microphone or empty audio stream`
And
`Previously, the audio connectivity check used to implicitly test that audio was being sent out from the client and being received by the server. Now this check will include a test for attendee presence, showing that the signal connection has been established, and audio packets being received, showing that the data connection has been established.`

aman-aguptaaa (commented):
Addressed all comments

CHANGELOG.md
aman-aguptaaa: Updated. Thanks

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
aman-aguptaaa: Updated the checkAudioConnectivity logic as per the suggestions. Thanks

pracheth (approved):

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
```diff
@@ -270,7 +270,21 @@ export default class DefaultMeetingReadinessChecker implements MeetingReadinessC
...
+    const audioConnectivityMetrics = {
+      packetsReceived: 0,
+    };
     const audioVideo = this.meetingSession.audioVideo;
+
+    const checkAudioConnectivityMetricsObserver: AudioVideoObserver = {
+      metricsDidReceive(clientMetricReport: ClientMetricReport) {
+        clientMetricReport.getRTCStatsReport().forEach(report => {
+          if (report.type === 'inbound-rtp' && report.mediaType === 'audio') {
+            audioConnectivityMetrics.packetsReceived = report.packetsReceived;
```
avinashmidathada, line 282: Nit: We can use += instead so that we don't fail the test based on a single report where there might be a disconnect due to network.

avinashmidathada (approved):

src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts
aman-aguptaaa: this metric - report.packetsReceived - will give the latest packet received metric total count upto now. so, we will be safe.

avinashmidathada: Yeah my bad, was confusing this with observable metrics. I can see you're accessing the raw report so we're good!

demos/browser/app/meetingReadinessChecker/meetingReadinessChecker.ts
```diff
@@ -181,8 +181,18 @@ export class DemoMeetingApp {
     }
   }
 
-  getAudioInputDeviceID = (): string => {
+  getAudioInputDevice = (): string | MediaStream => {
     const dropdownList = document.getElementById('audio-input') as HTMLSelectElement;
+    if(dropdownList.value == 'No Audio') {
```
ltrung, line 186: Where is this drop down value from? 
Also

aman-aguptaaa: This value is selected from the mic drop down inside MeetingReadinessChecker demo browser app.
MeetingReadinessChecker mic dropdown

ltrung: Can you point me to the code where its added? Its not there in my local repo.

aman-aguptaaa: L500-L505 of [demos/browser/app/meetingReadinessChecker/meetingReadinessChecker.ts]([demos/browser/app/meetingReadinessChecker/meetingReadinessChecker.ts](https://github.com/aws/amazon-chime-sdk-js/pull/2561/commits/6b747db1a72684c799b52afb1e4fa65d25ffe9e3#diff-d37b035014cecb05ff1a11a4ea1543eb5be17eb6e249101b484744ecd08c4909)) file

ltrung (approved):

PR #2613, opened 2023-03-25
Remove entry from meetingsTable when meeting ends

anilkumarkodali:
**Issue #:**
Unable to create a meeting with the same name with in a browser session. I either have to use a new name every time I need to run a test or restart the app to use the same name.

**Description of changes:**
Deleting the entry from the temporary cache(meetingsTable) when the meeting is ended.

**Testing:**
Tested by creating the meeting with same name with in a browser session.

- Create a meeting with title "Test"
- End meeting and recreate with the same title. The meeting would be created without complaining about the meeting not being found.

**Checklist:**

Yes

No

No

### DIFF
--- a/demos/browser/server.js
+++ b/demos/browser/server.js
@@ -199,6 +199,8 @@ function serve(host = '127.0.0.1:8080') {
         await client.deleteMeeting({
           MeetingId: meetingTable[requestUrl.query.title].Meeting.MeetingId,
         }).promise();
+        //Remove the meeting from table(so that a new meeting can be created with the same name with in a session)
+        delete meetingTable[requestUrl.query.title];
         respond(response, 200, 'application/json', JSON.stringify({}));
       } else if (request.method === 'POST' && requestUrl.pathname === '/startCapture') {
         if (captureS3Destination) {
--- a/demos/serverless/src/handlers.js
+++ b/demos/serverless/src/handlers.js
@@ -202,6 +202,7 @@ exports.end = async (event, context) => {
 
   // End the meeting. All attendee connections will hang up.
   await client.deleteMeeting({ MeetingId: meeting.Meeting.MeetingId }).promise();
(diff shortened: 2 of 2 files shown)


devalevenkatesh (approved):

demos/browser/server.js
```diff
@@ -199,6 +199,8 @@ function serve(host = '127.0.0.1:8080') {
         await client.deleteMeeting({
           MeetingId: meetingTable[requestUrl.query.title].Meeting.MeetingId,
         }).promise();
+        //Remove the meeting from table(so that a new meeting can be created with the same name with in a session)
+        delete meetingTable[requestUrl.query.title];
```
ltrung, line 203: This change only apply for local dev. How about our severless demo? For that one, we currently set ttl to 24h https://github.com/aws/amazon-chime-sdk-js/blob/main/demos/serverless/src/handlers.js#L544

anilkumarkodali: I thought of updating the serverless demo too, but wasn't sure if there was a reason for setting that 24 ttl. I have the code for that as well, but did not push it. I can push that as well if you are okay with that. That way we delete the entry from dynamo if meeting ends and if all participants leave without ending the meeting, the entry would be deleted after 24 hours. Let me know.

anilkumarkodali: @ltrung - committed the handler.js code for your reference. One thing to note is there is a slight delay in UI from when the end meeting button is clicked to taking back the new meeting landing page that is introduced by the dynamo delete call. I don't think it's a big deal, but wanted to make a note of it.

ltrung (approved):

ltrung:
Sorry. Forget to point out but in order to run the approval checks, you have to clone and submit your pull request. Currently, fork does not work.

ltrung:
This actually fail the MeetingEnd test where we verified that you could not createMeeting once a meeting is deleted so closing for now.

PR #2618, opened 2023-03-30
Pass Client UTC Offset inside join message over signalling channel

aman-aguptaaa:
**Issue #:** None

**Description of changes:**
*What?*
Add Client UTC Offset inside join message (which is sent over signalling channel)

*Why?*
With the aim of improving audio quality for customers, we want to map the timezone of attendees to server's AWS region  being connected to by them. 

**Testing:**
- Used demo application to verify that the client UTC offset parameter is being passed inside join message.

Yes, Running MeetingV2 Demo App in the browser should pass the client UTC offset parameter inside join message. You can filter the logs in console of browser by searching for the term `client_utc_offset`

**Checklist:**

Yes

 N/A

 N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [3.13.0] - 2023-03-28
 
 ### Added
-- Add Client UTC Offset inside join message which is sent over signalling channel.
+- Send client UTC offset with attendee JOIN frame over signalling channel.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [3.13.0] - 2023-03-28
 
 ### Added
+- Add Client UTC Offset inside join message which is sent over signalling channel.
 
 ### Removed
 
@@ -17,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [3.12.0] - 2023-02-14
 
-### Added
+### Added 
 
 ### Removed
 
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
     "@aws-sdk/util-hex-encoding": "^3.47.0",
     "@types/ua-parser-js": "^0.7.35",
     "detect-browser": "^5.2.0",
+    "moment-timezone": "^0.5.42",
     "pako": "^2.0.4",
     "protobufjs": "^6.11.3",
     "resize-observer": "^1.0.0",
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [3.13.0] - 2023-03-28
 
 ### Added
+- Add Client UTC Offset inside join message which is sent over signalling channel.
```
devalevenkatesh, line 11: Can make this builder friendly: "Send client UTC offset with attendee `JOIN` frame over signalling channel."

package.json
```diff
@@ -102,6 +102,7 @@
     "@aws-sdk/util-hex-encoding": "^3.47.0",
     "@types/ua-parser-js": "^0.7.35",
     "detect-browser": "^5.2.0",
+    "moment-timezone": "^0.5.42",
```
devalevenkatesh, line 105: Is there a browser native API over adding a dependency? If not hard enough, we would like to limit the dependencies.

src/utils/Utils.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import moment = require('moment-timezone');
```
devalevenkatesh, line 4: Its 2.99 MB when un-packed dependency per NPM, could we do this on backend?

package.json
aman-aguptaaa: Removed the dependency.

src/utils/Utils.ts
aman-aguptaaa: Removed the dependency.

CHANGELOG.md
aman-aguptaaa: Changed. Thanks.

src/utils/Utils.ts
```diff
@@ -36,3 +36,20 @@ export function toLowerCasePropertyNames(input: any): any {
     return result;
   }, {});
 }
+
+/**
+ * Get UTC offset in (+|-)HH:mm format
+ * E.g. For Asia/Calcutta timezone, +05:30 UTC offset value is returned
+ */
+export function getUTCOffsetString(utcOffsetInteger: number): string {
```
devalevenkatesh, line 44: - Can function be named to `getFormattedOffset(utcOffset: number): string`?
- Do we have to consider DST difference?

aman-aguptaaa: getTimezoneOffset considers daylight savings as well. Ref [this](https://stackoverflow.com/questions/11887934/how-to-check-if-dst-daylight-saving-time-is-in-effect-and-if-so-the-offset).

aman-aguptaaa: Will rename to getFormattedOffset

devalevenkatesh (approved):

ltrung (approved):

PR #2621, opened 2023-04-05
Fix false positive sendingAudioFailed events

avinashmidathada:
**Issue #:**

We have received customer reports where `sendingAudioFailed` event has caused some unnecessary confusion as it was triggered during the start of a meeting but was immediately being followed by a `sendingAudioRecovered` event, implying that there was some sort of a delay in sending out audio packets during the connection initiation time. 

We were aware of this and originally attempted to circumvent the problem by using a initial wait time i.e. `sendingAudioFailureInitialWaitTimeMs` whose default value is 3 secs. But, even with this there are some false-positives so we're also adding an additional condition on top, by making sure we wait till the meeting session is connected before we check for sending audio failure.

**Description of changes:**

The fix involves evaluating the SendingAudioFailureConnectionHealthPolicy only after a session is "connected" i.e. the policy is evaluated only if the `audioVideoDidStart()` observer is invoked, so we only evaluate the policy after ICE negotiation is complete or attendeePresence is received (if `attendeePresenceTimeoutMs` is configured), whichever happens later. Additionally, the unused code block for `connectWithTasks` has been removed with this commit.

PS: Note that, when `attendeePresenceTimeoutMs` is configured, and there's an issue causing audio packets to not be sent, we will **not** trigger `sendingAudioFailed`.  Usually triggering the event would've been useful as sending an audio packet was required for completing authentication and receiving attendee presence, but with the fast join rollout on server side, we no longer require the mic packet from client in order to send attendee presence. So, attendee presence will be received as long as our signalling channel is working properly.

**Testing:**
- Modified unit tests to cover the new logic.

Pre-requisite: A way to induce delay in Downlink Delay. In MacOS, use Network Link Conditioner and create a custom profile which has Downlink Delay set to 1000 ms with all other parameters being good.

1. Run Browser Demo App in Chrome or Safari (doesn't work in Firefox as `No Audio` microphone is required to repro)
2. Connect to a meeting normally i.e. land on the Preview page. Open the Developer Tools Console
3. Change microphone input to No Audio by clicking the Microphone drop down. Use Network Link Conditioner or equivalent to induce the 1000 ms downlink delay. 
4. Join the meeting and wait until you see `sendingAudioFailed` event. Check the logs above to make sure `meetingStartSucceeded` event was published prior to that event. 

Note that due to the nature of the issue it's not 100% guaranteed that the same test would not pass with an SDK version prior to this commit but it's highly likely it does fail.
 
**Checklist:**

Yes

 N/A

 N/A


devalevenkatesh (dismissed):

src/task/MonitorTask.ts
```diff
@@ -306,18 +307,41 @@ export default class MonitorTask
...
+    // We want to wait until we're connected to avoid false-positives.
+    this.shouldEvaluateSendingAudioFailurePolicy = true;
+  }
+
+  audioVideoDidStartConnecting(reconnecting: boolean): void {
+    // Similar to the start of a meeting, we don't want to evaluate sendingAudioFailureHealthyPolicy
+    // until we're connected. The expectation here is that the flag will be set to true again when
+    // audioVideoDidStart() is eventually called.
+    /* istanbul ignore else */
+    if (reconnecting) {
```
ltrung, line 336: nit: Couldn't we always set this to false ? No need for the if check.

src/task/MonitorTask.ts
```diff
@@ -306,18 +307,41 @@ export default class MonitorTask
...
+          const attributes = this.generateBaseAudioVideoEventAttributes();
+          this.context.eventController?.publishEvent('sendingAudioRecovered', attributes);
+        }
+      );
+    }
+  }
+
+  audioVideoDidStart(): void {
+    // We want to wait until we're connected to avoid false-positives.
+    this.shouldEvaluateSendingAudioFailurePolicy = true;
```
ltrung, line 328: nit: Could generalize this flag to something like `isMeetingConnected` as it can be used in other places.

src/task/MonitorTask.ts
avinashmidathada: Agreed, I just wanted to make it more explicit in the context of the flag's purpose, but as you suggested below I'm changing the variable name to `isMeetingConnected` so I can remove the if condition here and just set it to false.

devalevenkatesh (approved):

pracheth (approved):

PR #2646, opened 2023-04-20
Fix protobuf script and add join flag for layers alloc header extension

bjason:
**Issue #:**

**Description of changes:**

Protobuf generate script fails because of the jsdoc issue. Also add a join flag for layers allocation header extension with default value of true. The header extension was added in https://github.com/aws/amazon-chime-sdk-js/pull/2195.

**Testing:**

Unit test and smoke test with the demo app.

no

**Checklist:**

y

n

n

### DIFF
--- a/script/generate-signaling-protocol
+++ b/script/generate-signaling-protocol
@@ -10,21 +10,23 @@ end
 ['SignalingProtocol'].each do |protocol|
   dir = protocol.downcase
 
+  # please change the protobuf dependency in package.json to the following:
+  # "protobufjs": "git://github.com/VitalyName/protobuf.js.git#236a7c757a5eada7926fdff149d0b49ada838275"
+  # This is a walkaround for https://github.com/protobufjs/protobuf.js/issues/1414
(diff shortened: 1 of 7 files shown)


ltrung (approved):

mklingb:
LGTM

script/generate-signaling-protocol
```diff
@@ -10,21 +10,23 @@ end
...
+
   verbose "rm -rf src/#{dir}"
   verbose "mkdir -p src/#{dir}"
 
   # workaround for https://github.com/hegemonic/requizzle/issues/5
-  verbose "brew install node@10"
-  node = "#{`brew --prefix node@10`.strip}/bin/node"
+  # node@14 will work with libicu4c 72
+  # verbose "brew install node@14"
+  node = "#{`brew --prefix node@14`.strip}/bin/node"
```
ltrung, line 23: Can we update to Node 18 or 20? Node 14 is no longer supported.

PR #2647, opened 2023-04-26
Update VideoFxProcessor CSP Guide

hunnorth:
**Issue #:NA**

**Description of changes:**
Update VideoFxProcessor CSP Guide to be as Restrictive as Possible

**Testing:**

**Checklist:**

yes

no

no

### DIFF
--- a/guides/21_Background_Filter_VideoFX_Processor.md
+++ b/guides/21_Background_Filter_VideoFX_Processor.md
@@ -97,6 +97,17 @@ One can read further on the Content Security Policy’s documentation for WebAss
 ## Optional CSP Directives ##
 ### Background Image Policy ###
 To use the background replacement filter with a dynamically loaded background image, `VideoFxProcessor` must be able to access the image\. Include a `connect-src` directive with the domain that hosts the image\.
+## Example CSP Declaration ##
+Here is an example csp declaration to permit the use of the `VideoFxProcessor`\. Please note that the definitions in the `connect-src` tag are not `VideoFxProcessor` specfic but are related to video/audio in Chime meetings.
+```
+<meta http-equiv="Content-Security-Policy" 
+  content= "base-uri         'self';
+  connect-src      'self' https://*.chime.aws wss://*.chime.aws https://*.amazonaws.com wss://*.chime.aws https://*.ingest.chime.aws;
+  script-src       'self' blob: 'wasm-unsafe-eval' https://*.sdkassets.chime.aws;
+  script-src-elem  'self' blob: https://*.sdkassets.chime.aws;
+  worker-src       'self' blob: https://*.sdkassets.chime.aws;
+">
+```
 ## CSP Errors ##
 If any any of the required directives are omitted then the VideoFxProcessor will not be able to properly instantiate and the processor will be unsupported\. Additionally, the following (or similar) errors will likely appear in the browser console:
 ```
--- a/guides/21_Background_Filter_VideoFX_Processor.md
+++ b/guides/21_Background_Filter_VideoFX_Processor.md
@@ -13,7 +13,7 @@ The new background blur and replacement filters are integrated into the [browser
   - [WebGL2 support](#webgl2-support)
   - [Content delivery and bandwidth](#delivery-caching-bandwidth)
   - [Browser compatibility](#filters-browser-compat)
-- [Using a content-security policy](#content-security)
+- [Using a content security policy](#content-security)
 - [Adding background filters to your application](#add-filters)
   - [Checking for support before offering a filter](#support-check)
   - [Creating a VideoFxConfig object](#create-videofxconfig)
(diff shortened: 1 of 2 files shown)


guides/21_Background_Filter_VideoFX_Processor.md
```diff
@@ -75,38 +75,34 @@ The following table lists the browsers and version that support background filte
...
-Refused to connect to
-'https://static.sdkassets.chime.aws/bgblur/workers/worker.js…'
-because it violates the document's content security policy.
+script-src       'self' blob: https://*.sdkassets.chime.aws
+script-src-elem  'self' blob: https://*.sdkassets.chime.aws
 ```
-
-The following example shows a typical content security policy\. It includes the tags needed to access the resources needed to run the filters\. You can adapt this example as needed\. 
-
+Please note that `script-src` and `script-src-elem` directives are both required for full support on Safari and Firefox browsers\.
```
ltrung, line 87: Nice. Thank you for writing this up.

guides/21_Background_Filter_VideoFX_Processor.md
```diff
@@ -75,38 +75,34 @@ The following table lists the browsers and version that support background filte
...
+### Worker Policy ###
+The `VideoFxProcessor` loads JavaScript classes as a blob to run a web worker thread, which processes video using Machine Learning (ML) models\. To grant an application access to fetching and using this worker, include:
+```
+worker-src       'self' blob: https://*.sdkassets.chime.aws
+```
+### WebAssembly Policy ###
+The `VideoFxProcessor` loads a WebAssembly (Wasm) module from the same Amazon-owned CDN\. In Chrome 95 and later, compiled Wasm modules cannot be passed across multiple module boundaries\. To allow fetching and instantiating these modules, please also include `'wasm-unsafe-eval'` to the `script-src` directive\. 
+
+One can read further on the Content Security Policy’s documentation for WebAssembly [here](https://github.com/WebAssembly/content-security-policy/blob/main/proposals/CSP.md)\.  Ideally the `wasm-eval` tag would be better used, however it yet to gain official standardization. Correspondence on the directive can be found [here](https://bugs.chromium.org/p/v8/issues/detail?id=7041)\.
```
ltrung, line 96: So we do not need both `wasm-eval ` and `wasm-unsafe-eval` as before?

guides/21_Background_Filter_VideoFX_Processor.md
```diff
@@ -75,38 +75,34 @@ The following table lists the browsers and version that support background filte
...
+## Optional CSP Directives ##
+### Background Image Policy ###
+To use the background replacement filter with a dynamically loaded background image, `VideoFxProcessor` must be able to access the image\. Include a `connect-src` directive with the domain that hosts the image\.
+## CSP Errors ##
+If any any of the required directives are omitted then the VideoFxProcessor will not be able to properly instantiate and the processor will be unsupported\. Additionally, the following (or similar) errors will likely appear in the browser console:
 ```
-<head>
```
ltrung, line 98: Can we keep this section ? It is useful to have a sample CSP setup

guides/21_Background_Filter_VideoFX_Processor.md
hunnorth: Correct. `wasm-eval` is not yet standardized and will often times produce an error along the lines of:
 `The source list for Content Security Policy directive <directive-x> contains an invalid source: ''wasm-eval''. It will be ignored.`

guides/21_Background_Filter_VideoFX_Processor.md
hunnorth: Sure, I can add it back. But the difficulty will be drawing the line for exactly what the example is going to support. 

If you prefer the csp demo be specifically for the videofxprocessor it would look like this:
```
<meta http-equiv="Content-Security-Policy" 
content= "base-uri         'self';
          script-src       'self' blob: 'wasm-unsafe-eval' https://*.sdkassets.chime.aws;
          script-src-elem  'self' blob: https://*.sdkassets.chime.aws;
          worker-src       'self' blob: https://*.sdkassets.chime.aws;
">
```

But in order to support the actual video/audio that is essential to running a meeting (but not relevant to the videofxprocessor), we would also need to add this line:
```
connect-src      'self' https://*.chime.aws wss://*.chime.aws https://*.amazonaws.com wss://*.chime.aws https://*.ingest.chime.aws;
```

I am fine with either option here. Including the secondary aspect might cause some confusion. But not including it could also lead developers to think the example is an out of box csp configuration for their application. This confusion is what lead me to omit the example all together and alternatively just specify individual directives and their corresponding justification.

ltrung: I think its better to have a minimum set that ppl can just copy and it works even though it may include more thing than just videofxprocessor. You can also make a note that we also include the essential needed for audio video in the sample.

hunnorth: Sounds good. Added.

ltrung (approved):

Issue #2636, opened 2023-04-11
checkBrowserSupportAndFeatureConfiguration returns null after browser detection


mgautam98:
### What happened and what did you expect to happen?

Using chime in Android webview chime client throws following error. From what I checked it is failing wile reading `browser.name` in `checkBrowserSupportAndFeatureConfiguration`. 

Device details
- Pixel 6 API 33
- Android 13.0 Google APIs | arm64
- Chrome 101.0.4951.74

```
Error in meetingHandler try block TypeError: Cannot read properties of null (reading 'name')
    at s.name (asset?key=gautamx%2Fgautamx_LiveVideoShoppingBuyerFrontendServiceCR%2FLiveVideoShoppingBuyerFrontendServiceAssets%2F9cb1ee89cc42a0fc8b18d5b49ba7f4f82fd27c16.js.lz4&ext=js&AUIClients&AUIClients/LiveVideoShoppingBuyerFrontendServiceAssets:31:325553)
    at t.default.checkBrowserSupportAndFeatureConfiguration (asset?key=gautamx%2Fgautamx_LiveVideoShoppingBuyerFrontendServiceCR%2FLiveVideoShoppingBuyerFrontendServiceAssets%2F9cb1ee89cc42a0fc8b18d5b49ba7f4f82fd27c16.js.lz4&ext=js&AUIClients&AUIClients/LiveVideoShoppingBuyerFrontendServiceAssets:31:456764)
    at new t.default (asset?key=gautamx%2Fgautamx_LiveVideoShoppingBuyerFrontendServiceCR%2FLiveVideoShoppingBuyerFrontendServiceAssets%2F9cb1ee89cc42a0fc8b18d5b49ba7f4f82fd27c16.js.lz4&ext=js&AUIClients&AUIClients/LiveVideoShoppingBuyerFrontendServiceAssets:31:454596)
    at S (asset?key=gautamx%2Fgautamx_LiveVideoShoppingBuyerFrontendServiceCR%2FLiveVideoShoppingBuyerFrontendServiceAssets%2F9cb1ee89cc42a0fc8b18d5b49ba7f4f82fd27c16.js.lz4&ext=js&AUIClients&AUIClients/LiveVideoShoppingBuyerFrontendServiceAssets:31:932544)
[...]

PR #2648, opened 2023-04-26
Add fallback value for detect() when it can not parse UserAgent

xuesichao:
**Issue #2636 :**

**Description of changes:**
We're using `detect-browser` library's `detect()` method to parse the UserAgent.

https://github.com/aws/amazon-chime-sdk-js/blob/26327a523539cd0b0b0538d12d9d9df42263ce6e/src/browserbehavior/DefaultBrowserBehavior.ts#L11
And later we're accessing the `name`, `os`, `version` property, and maybe even `type` property in the future of this `browser` in `DefaultBrowserBehavior`.

https://github.com/aws/amazon-chime-sdk-js/blob/26327a523539cd0b0b0538d12d9d9df42263ce6e/src/browserbehavior/DefaultBrowserBehavior.ts#L57-L59
https://github.com/aws/amazon-chime-sdk-js/blob/26327a523539cd0b0b0538d12d9d9df42263ce6e/src/browserbehavior/DefaultBrowserBehavior.ts#L69-L71
https://github.com/aws/amazon-chime-sdk-js/blob/26327a523539cd0b0b0538d12d9d9df42263ce6e/src/browserbehavior/DefaultBrowserBehavior.ts#L226-L228

However if the UserAgent is not supported, `detect()` returns `null`. When `browser` is `null` it will throw error and stop page from rendering.
>TypeError: Cannot read properties of null (reading 'name')

This change adds a fallback value to avoid such errors.
https://github.com/aws/amazon-chime-sdk-js/blob/c1660fe8671c0df7adfd0339b728c03ed619acc8/src/browserbehavior/DefaultBrowserBehavior.ts#L11-L17

And we already log a warning when browser is not supported. So no need to add an extra warning message.
https://github.com/aws/amazon-chime-sdk-js/blob/26327a523539cd0b0b0538d12d9d9df42263ce6e/src/meetingsession/DefaultMeetingSession.ts#L157-L163

**Testing:**

Yes, in meetingV2 demo. 
1. In Chrome browser, follow this [guide](https://www.searchenginejournal.com/change-user-agent/368448/#close) to custom UserAgent. Use `FakeBrowser/26.7.2.100 (Android/13/sdk_gphone64_arm64)` as the value.
2. Refresh the page and load the meeting demo.
3. Verify the meeting demo is rendered correctly.
4. In debugger console, enter `this.navigator.userAgent` and verify the printed value is `FakeBrowser/26.7.2.100 (Android/13/sdk_gphone64_arm64)`

**Checklist:**

Yes

No

No

### DIFF
--- a/src/browserbehavior/DefaultBrowserBehavior.ts
+++ b/src/browserbehavior/DefaultBrowserBehavior.ts
@@ -8,7 +8,13 @@ import BrowserBehavior from './BrowserBehavior';
 import ExtendedBrowserBehavior from './ExtendedBrowserBehavior';
 
 export default class DefaultBrowserBehavior implements BrowserBehavior, ExtendedBrowserBehavior {
-  private readonly browser = detect();
+  private readonly FALLBACK_BROWSER = {
+    type: 'browser',
+    name: 'unknown',
+    version: 'unknown',
+    os: 'unknown',
+  };
+  private readonly browser = detect() || this.FALLBACK_BROWSER;
   private readonly uaParserResult =
     navigator && navigator.userAgent ? new UAParser(navigator.userAgent).getResult() : null;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,22 +5,25 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-
 ## [3.14.0] - 2023-04-11
 
 ### Added
+
 * Add new guide for `VideoFXProcessor`.
 
 ### Removed
 
 ### Changed
 
+* Add fallback value for `DefaultBrowserBehavior.browser` when `detect()` can not parse browser UserAgent.
+
 ### Fixed
 
 ## [3.13.0] - 2023-03-28
 
 ### Added
-- Send client utc offset with attendee JOIN frame over signalling channel. 
+
+- Send client utc offset with attendee JOIN frame over signalling channel.
 - Add an unified interface (videofx processor) with new implementation for ML features, namely background blur 2.0 and background replacement 2.0. This includes improvement to the underlying model for better segmentation, better image processing algorithms to blur the background, and other miscellanies bug-fixes/features.
 
 ### Removed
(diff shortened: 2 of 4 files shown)


src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -8,7 +8,13 @@ import BrowserBehavior from './BrowserBehavior';
 import ExtendedBrowserBehavior from './ExtendedBrowserBehavior';
 
 export default class DefaultBrowserBehavior implements BrowserBehavior, ExtendedBrowserBehavior {
-  private readonly browser = detect();
+  private readonly FALLBACK_BROWSER = {
+    type: 'unknown',
```
ltrung, line 12: nit: Should `type` be `browser` ? From BrowserInfo that seems like the default value for browser info.
```
readonly name: Browser;
  readonly version: string;
  readonly os: OperatingSystem | null;
  readonly type = "browser";
```

ltrung:
Can you test this by set the userAgent to something else in browser and verify that it is crashed before the fix and not after? So QA could test it.

src/browserbehavior/DefaultBrowserBehavior.ts
xuesichao: The return type of `detect()` is `BrowserInfo | SearchBotDeviceInfo | BotInfo | NodeInfo | ReactNativeInfo | null`. 
* `BrowserInfo` has `type` of `"browser"`.
* `BotInfo` has `type` of `"bot"`.
* `NodeInfo` has `type` of `"node"`.
* `ReactNativeInfo` has `type` of `"react-native"`.

Setting `type` to `"browser"` works, but I think `"unknown"` should be more appropriate. How do you think?

ltrung: Yeah I dont really have a strong opinion on it. It is just that you already call the variable `Fallback_browser` which to indicate the type to be `browser`. Also the class name is also browserbehavior.

xuesichao: Yeah, let me change it to "browser". We don't use this `type` anywhere so it doesn't matter that much. And "browser" seems to be more consistent.

xuesichao: Updated

ltrung (approved):

Issue #2655, opened 2023-05-02
How to encourage sdk to prefer lower quality video instead of pausing or ending streams. 


CodySchaaf:
### What are you trying to do?

Currently our app mostly focuses on meetings with 2-4 people (some on mobile), as well as a lot of screen sharing. My main goal right now is for as little interruption to video as possible. If quality needs to be downgraded to ensure consistent video availability (especially screen sharing), that is fine. 

We have primarily been utilizing the code provided here https://github.com/aws-samples/amazon-chime-sdk/tree/main/apps/meeting/src/providers/VideoTileGridProvider

With the following configuration:

```ts
const config = VideoPriorityBasedPolicyConfig.Default;
config.serverSideNetworkAdaption = ServerSideNetworkAdaption.BandwidthProbing;

const priorityBasedPolicy = new VideoPriorityBasedPolicy(logger, config)

//....      

meetingSessionConfiguration.enableSimulcastForUnifiedPlanChromiumBasedBrowsers = true;

meetingSessionConfiguration.videoDownlinkBandwidthPolicy = priorityBasedPolicy;

try {
  await meetingManager.join(meetingSessionConfiguration, options);
  setJoiningMeeting(false);
} catch {
  setJoiningMeeting(false);
}
```

With this configuration we are getting a lot of pausing of video and screen sharing as bandwidth becomes constrained. Even on networks that seem relatively stable/strong. There has been no interruption on other video platforms (google meet/zoom) with these same networks/computers. 

We have also tried to switch to the `VideoAdaptiveProbePolicy` but that comes with the issue of when it decides to kill the screen share stream there is no way for our UI to react. 

We have also tried toying with 

```
videoPreferences.add(
          new VideoPreference(attendeeId, 1, TargetDisplaySize.High),
        );
```

And setting it to prefer Low target size, but that seems like it is more of a suggestion and still doesn't set a lower band for the quality of "do not pause". 

### Question

Is there a way to configure these policies further to encourage the sdk to prefer decreasing frame rate/quality to allow videos/screen sharing to stay active. Basically do anything possible to keep these enabled while not degrading audio quality. Our fallback right now is to revert to `AllHighestVideoBandwidthPolicy`, but we would prefer to allow the sdk to do its thing and adjust quality when it determines it needs to. 

Thanks so much for any suggestions.

hensmi-amazon:
[...]

Issue #2559, opened 2023-01-25
Simulcast issues with `ServerSideNetworkAdaptation.BandwidthProbing`


kelvin2200:
### What happened and what did you expect to happen?

Yet another attempt to implement simulcast has proven challenging.
I reported #2374 a few months ago, and it was reported as solved some time later, but I just saw #2550 which seems to be the same issue. Will that be released soon?

Apart from that, it seems that combining simulcast with BandwidthProbing leads to multiple issues.
all issues below assume that simulcast is enabled with the following config:
`const config = VideoPriorityBasedPolicyConfig.Default;
config.serverSideNetworkAdaption = ServerSideNetworkAdaption.BandwidthProbing;`

1. Incoming videos are jerky, sometimes freezing, other times laggish, and there seems to be a considerable drop in quality (compared to when BandwidthProbing is not used, for the same simulcast layer)
2.  At random intervals, random remote streams behave as if they were turned off and an on very fast (you just stop receiving the stream for about 1 sec, then it turns back on). In this particular situation `useAttendeeStatus().videoEnabled` becomes false, then true, just as if the user manually turned his cam off/on.
3. If some attendee suffers a temporary internet loss, when he reconnects, other streams (that never went offline) are lost on other attendees' side and they never recover (not the exact behavior as #2374 or #2550, but similar). It seems that one re-connecting to a meeting due to network issues, causes issues for other attendees. The only thing that works in this case, is either disconnecting and re-connecting to the meeting, or refreshing the page.

Providing a reproduction scenario for all these is very hard, as it took our QA team several days to find them. The one thing we know for sure is that they keep repeating, and there seems to be no pattern other than the combination between simulcast and BandwidthProbing (dunno... maybe it has something to do with the `VideoPriorityBasedPolicy` as well). Without BandwithProbing, the issues don't exist, but simulcast isn't really viable without it.

I'm really starting to feel that there's something wrong with this lib at a fundamental level. For 2 years we've been trying to implement simulcast in our app... still without success.

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

[...]

PR #2680, opened 2023-06-14
Add client side APIs and implementation for server side remote video quality  adaption

hensmi-amazon:
**Issue #:** No specific issue, but https://github.com/aws/amazon-chime-sdk-js/issues/2655 and https://github.com/aws/amazon-chime-sdk-js/issues/2559 are related.

**Description of changes:**

From the guide:
`VideoPriorityBasedPolicy` has support for enabling additional features on the backend related to network adaptation. These features lead to the following improvements relative to the default policy:

* Faster network constraint response time, e.g. compared to default it will downgrade and pause videos faster to prioritize audio.
* Avoidance of video upgrades that cause network overuse. The server side adaptation uses pacing and probing to determine headroom before upgrading videos.
* Stable video quality selections under constraint due to a backend calculated bandwidth estimation utilizing transport wide congestion control.
* Faster recovery time due to use of pacing and probing on backend.

**Testing:**

In isolated network testing we can (among others) run the following scenario to test the feature:
* 8 senders using the [default simulcast policy](https://github.com/aws/amazon-chime-sdk-js/blob/main/src/videouplinkbandwidthpolicy/DefaultSimulcastUplinkPolicy.ts) (150/350 kbps).
* Downlink not sending, using one of the three 4 policies mentioned above.
    * Begins unconstrained
    * At 30 seconds, constrained to 2 MB (for ref, Amazon only technically allocations 700/1300 kbps per non-dev/dev at 50% capacity).
    * At 60 seconds, toss a burst of TCP cross traffic at the downlink
    * At 90 seconds, toss a burst of UDP cross traffic at the downlink
    * Remove constraints at 120 seconds
 
With server side quality adaptation, we can see that we take advantage of simulcast and switch to the lower layers (not marked here but they are lower resolution), which brings us under the constraint without need to pause any videos.

Screen Shot 2023-06-13 at 1 45 39 PM

With the cross traffic it does lead us to paused videos (understandable) but recovery to the low simulcast streams occurs before the next event even occurs.

In the loss chart, we see just a second of loss events, which likely indicates that audio interruptions were negligible. 

Screen Shot 2023-06-13 at 1 45 39 PM(1)

Yes you can select the feature from the drop down menu. Internal documents cover more test cases.

**Checklist:**

y

y

n


src/task/MonitorTask.ts
```diff
@@ -191,11 +183,6 @@ export default class MonitorTask
   }
 
   metricsDidReceive(clientMetricReport: ClientMetricReport): void {
-    const defaultClientMetricReport = clientMetricReport as ClientMetricReport;
```
hensmi-amazon, line 194: Lol. Artifact of when we had a `Default` implementation of this.

src/task/MonitorTask.ts
```diff
@@ -228,45 +206,6 @@ export default class MonitorTask
         downlinkVideoStream.set(streamMetricReport[ssrc].streamId, streamMetricReport[ssrc]);
       }
     }
-
-    let fireCallback = false;
```
hensmi-amazon, line 232: These snippets no longer callback, and only log. They are not accurate when using the priority policy (before or after SSNA), and not worth fixing IMO, as since GA i have never seen these logs useful in my (or others) investigations, and they constantly become red herring when seen by builders. Let me know if anyone feels otherwise :)

rixiang-dev (approved):

src/videostreamindex/VideoStreamIndex.ts
```diff
@@ -89,6 +89,11 @@ export default interface VideoStreamIndex {
    */
   attendeeIdForStreamId(streamId: number): string;
 
+  /**
+   * Returns attendee id for a group id
+   */
+  attendeeIdForGroupId(groupId: number): string;
```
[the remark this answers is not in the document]
hensmi-amazon: Yup you are correct, good catch. Let me fix that up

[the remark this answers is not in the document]
hensmi-amazon: Well technically you have to reimplement the entire `AudioVideoController` to use your own index :D, but I fixed it up nonetheless :P

simmkyu (approved):

src/task/MonitorTask.ts
ltrung: Yeah, they usually just cause confusions for devs.

ltrung (approved):

Issue #2674, opened 2023-06-08
Uncaught TypeError: Cannot read property 'split' of null when using with react native while initializing default session


harryy2510:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `amazon-chime-sdk-js@3.14.1` for the project I'm working on.

The project is based on react native and utilises `amazon-chime-sdk-js` and `@aws-sdk/client-chime-sdk-messaging` for chat. While initializing default session on react native,`amazon-chime-sdk-js` throws error `Cannot read property 'split' of null` when it tried to find major version of browser.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/amazon-chime-sdk-js/build/browserbehavior/DefaultBrowserBehavior.js b/node_modules/amazon-chime-sdk-js/build/browserbehavior/DefaultBrowserBehavior.js
index 43cc745..6691c21 100644
--- a/node_modules/amazon-chime-sdk-js/build/browserbehavior/DefaultBrowserBehavior.js
+++ b/node_modules/amazon-chime-sdk-js/build/browserbehavior/DefaultBrowserBehavior.js
@@ -61,7 +61,7 @@ class DefaultBrowserBehavior {
         this.webkitBrowsers = ['crios', 'fxios', 'safari', 'ios', 'ios-webview', 'edge-ios'];
     }
     version() {
-        return this.browser.version;
+        return this.browser.version || "";
     }
     majorVersion() {
         return parseInt(this.version().split('.')[0]);
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

xuesichao:
Hi @harryy2510 , we are using `detect()` API of `detect-browser` library to retrieve the browser info in `DefaultBrwoserBehavior.js`. It seems for react native, the browser info from `detect()` does not contain a valid value for `version`, and in this cause your fix serves as a safe guard. 

I just checked the implementation of `detect-browser` and confirm it returns `null` for `version`. https://github.com/DamonOehlman/detect-browser/pull/137/files

I'll add your fix to our library. Thanks for sharing this with us!

xuesichao:
Duplicate of #1856 .

Issue #1856, opened 2021-12-07
AWS Chime SDK Doesn't work with React Native


robertjcolley:
### What happened and what did you expect to happen?

React Native supports websockets and can handle it. I expected messaging to work, but I got an error

> Error: null is not an object (evaluating 'this.version().split()')

I ended up tracing it to the detect-browser React Native implementation here:

https://github.com/DamonOehlman/detect-browser/blob/546e6f1348375d8a486f21da07b20717267f6c49/src/index.ts#L59-L65

which is referenced in 

https://github.com/aws/amazon-chime-sdk-js/blob/8af993e501fbf0c18b0b567e5e5a20114f08abd2/src/browserbehavior/DefaultBrowserBehavior.ts

The solution is:

1. detect-browser needs to update their implementation of React Native. I've made a PR to their project ([HERE](https://github.com/DamonOehlman/detect-browser/pull/172))
2. aws-chime-sdk-js needs to update detect-browser when this PR is merged in and detect-browser is updated

I've been able to successfully create a patch which I can use [patch-package](https://github.com/ds300/patch-package) to run in a post install script for the time being (see attached) and confirmed that this patch did correctly fix the issue and I was successfully able to use the SDK. I'm creating this issue here to track this issue until it is resolved, and will update this as I fix issues in detect-browser.

I had to rename the ".patch" file to ".patch.txt" to upload to GitHub:

[detect-browser+5.3.0.patch.txt](https://github.com/aws/amazon-chime-sdk-js/files/7671567/detect-browser%2B5.3.0.patch.txt)

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

Try creating a DefaultMessagingSession in React Native

### Amazon Chime SDK for JavaScript version

latest

### What browsers are you seeing the problem on?

React Native

### Browser version

0.65.1

### Meeting and Attendee ID Information.

N/A

### Browser console logs

N/A - see top section

simmkyu:
[...]

PR #2686, opened 2023-06-21
Add null check for DefaultBrowserBehavior.majorVersion()

xuesichao:
**Issue #1856, #2674 :**

**Description of changes:**
According to `detect-library` the `detect()` could return browser info with the value of `version` being `null`. 

For example: https://github.com/DamonOehlman/detect-browser/blob/546e6f1348375d8a486f21da07b20717267f6c49/src/index.ts#L59-L65.
```
export class ReactNativeInfo
  implements DetectedInfo<'react-native', 'react-native', null, null> {
  public readonly type = 'react-native';
  public readonly name: 'react-native' = 'react-native';
  public readonly version: null = null;
  public readonly os: null = null;
}
```

In this case our `majorVersion()` will throw an error because it try execute `null.split('.')`. This change add a null check to avoid such case.

**Testing:**

No. The change has been covered by unit test.

**Checklist:**

Yes

No

No

### DIFF
--- a/src/browserbehavior/DefaultBrowserBehavior.ts
+++ b/src/browserbehavior/DefaultBrowserBehavior.ts
@@ -64,7 +64,10 @@ export default class DefaultBrowserBehavior implements BrowserBehavior, Extended
     return this.browser.version;
   }
 
-  majorVersion(): number {
+  majorVersion(): number | null {
+    if (!this.browser.version) {
+      return null;
+    }
     return parseInt(this.version().split('.')[0]);
   }
 
--- a/test/browserbehavior/DefaultBrowserBehavior.test.ts
+++ b/test/browserbehavior/DefaultBrowserBehavior.test.ts
@@ -90,6 +90,18 @@ describe('DefaultBrowserBehavior', () => {
   });
 
   describe('platforms', () => {
+    it('can detect ReactNative', () => {
+      // Set ReactNative Environment. https://github.com/DamonOehlman/detect-browser/blob/546e6f1348375d8a486f21da07b20717267f6c49/src/index.ts#L216-L222
+      // @ts-ignore
+      navigator.product = 'ReactNative';
+      document = undefined;
+
+      expect(new DefaultBrowserBehavior().name()).to.eq('react-native');
+      expect(new DefaultBrowserBehavior().version()).to.eq(null);
+      expect(new DefaultBrowserBehavior().majorVersion()).to.eq(null);
+      expect(new DefaultBrowserBehavior().isSupported()).to.be.false;
+    });
+
     it('can detect Firefox', () => {
       setUserAgent(FIREFOX_MAC_USER_AGENT);
       expect(new DefaultBrowserBehavior().name()).to.eq('firefox');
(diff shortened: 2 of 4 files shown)


src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -64,7 +64,10 @@ export default class DefaultBrowserBehavior implements BrowserBehavior, Extended
     return this.browser.version;
   }
 
-  majorVersion(): number {
+  majorVersion(): number | null {
```
xuesichao, line 67: Not sure if it's fine to extend this return type and only in default implementation. The interface is still having `number` as return type.
https://github.com/aws/amazon-chime-sdk-js/blob/0aabd5aa821051e02569a0fb77736224416cdf3b/src/browserbehavior/BrowserBehavior.ts#L4-L13

test/browserbehavior/DefaultBrowserBehavior.test.ts
```diff
@@ -90,6 +90,18 @@ describe('DefaultBrowserBehavior', () => {
...
+    it('can detect ReactNative', () => {
+      // Set ReactNative Environment. https://github.com/DamonOehlman/detect-browser/blob/546e6f1348375d8a486f21da07b20717267f6c49/src/index.ts#L216-L222
+      // @ts-ignore
+      navigator.product = 'ReactNative';
+      document = undefined;
+
+      expect(new DefaultBrowserBehavior().name()).to.eq('react-native');
+      expect(new DefaultBrowserBehavior().version()).to.eq(null);
+      expect(new DefaultBrowserBehavior().majorVersion()).to.eq(null);
+      expect(new DefaultBrowserBehavior().isSupported()).to.be.false;
```
xuesichao, line 102: We still do not officially support React Native, so `isSupported()` returns `false`.

src/browserbehavior/DefaultBrowserBehavior.ts
michaelbachmann: I think you beat me to this by a few minutes :), I really appreciate you looking into this!

I tried out overriding the return type in a typescript playground

```
interface BrowserBehavior {
    majorVersion(): number
}

class DefaultBrowserBehavior implements BrowserBehavior {
    majorVersion(): number | null {
        return 1
    }
}
```

Typescript complains about it, but it might still work once it's transpiled to javascript.

```
Property 'majorVersion' in type 'DefaultBrowserBehavior' is not assignable to the same property in base type 'BrowserBehavior'.
  Type '() => number | null' is not assignable to type '() => number'.
    Type 'number | null' is not assignable to type 'number'.
      Type 'null' is not assignable to type 'number'.(2416)
```

[Could also consider a default value of like -1 for major version](https://github.com/aws/amazon-chime-sdk-js/pull/2686)?

xuesichao: @michaelbachmann I think -1 makes more sense. Besides this we think it's better to add fall back value for `browser.version` and `browser.os` if their values is `null` from `detect()`.

PR #2688, opened 2023-06-26
Fallback to -1 for browser majorVersion when version is null

xuesichao:
**Issue #1856, #2674 :**

**Description of changes:**
According to `detect-library` the `detect()` could return browser info with the value of `version` being `null`. 

For example: https://github.com/DamonOehlman/detect-browser/blob/546e6f1348375d8a486f21da07b20717267f6c49/src/index.ts#L59-L65.
```
export class ReactNativeInfo
  implements DetectedInfo<'react-native', 'react-native', null, null> {
  public readonly type = 'react-native';
  public readonly name: 'react-native' = 'react-native';
  public readonly version: null = null;
  public readonly os: null = null;
}
```

In this case our `majorVersion()` will throw an error because it try execute `null.split('.')`. This change add a null check to avoid such case. And give it a fallback value `-1`.

**Testing:**

No. The change has been covered by unit test.

**Checklist:**

Yes

No

No

### DIFF
--- a/src/browserbehavior/DefaultBrowserBehavior.ts
+++ b/src/browserbehavior/DefaultBrowserBehavior.ts
@@ -65,7 +65,7 @@ export default class DefaultBrowserBehavior implements BrowserBehavior, Extended
   }
 
   majorVersion(): number {
-    return parseInt(this.version().split('.')[0]);
+    return this.version() !== null ? parseInt(this.version().split('.')[0]) : -1;
   }
 
   osMajorVersion(): number {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 ### Changed
+
 - Update chime sdk messaging client version.
-- Clarify quotas for content-sharing publishing and subscriptions in API Overview. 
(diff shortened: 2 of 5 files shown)


src/browserbehavior/DefaultBrowserBehavior.ts
```diff
@@ -65,7 +65,7 @@ export default class DefaultBrowserBehavior implements BrowserBehavior, Extended
   }
 
   majorVersion(): number {
-    return parseInt(this.version().split('.')[0]);
+    return this.version() !== null ? parseInt(this.version().split('.')[0]) : -1;
```
simmkyu, line 68: Is there any possibility that `this.version` might return `undefined`?

simmkyu: How about returning `-1` for the `undefined` version too?

```js
typeof this.version() === 'string' ? parseInt(this.version().split('.')[0]) : -1;
```

xuesichao: According to `detect-browser` library, no. It could only be `string` or `null`. But falsy check is better than null check here. Could we update this in later if needed?

simmkyu: Thanks for checking!

simmkyu (approved):

PR #2699, opened 2023-07-03
Added recommendations to use server side network adaptation, and clarified support on all browsers

hensmi-amazon:
**Issue #:** None

**Description of changes:**

* Sprinkled some recommendations and deprecation around. Though there aren't any blockers to changing the default, I'm going to wait another month or so just to gather data on opt-in usage to better justify that. However I figured it would be useful to start recommending the new features sooner rather then later.
* Removed preset usage from demo since it doesn't have any impact on SSNA and just causes confusion.
* This also changes the default policy behavior to use 'max' instead of high, which will unnecessarily downgrade video sometimes.

**Testing:**
Smoke tested visual changes in demo. Otherwise this is just a documentation change.

No

**Checklist:**

y

n

n

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.html
+++ b/demos/browser/app/meetingV2/meetingV2.html
@@ -149,18 +149,12 @@
             <label for="priority-downlink-policy" class="form-check-label">Use Priority-Based Downlink
               Policy</label>
           </div>
-          <lable id="priority-downlink-policy-preset-title" for="priority-downlink-policy-preset" style="display: none; padding-top:5px">Policy Preset</lable>
-          <select id="priority-downlink-policy-preset" class="form-select" style="width:100%; display: none;">
-            <option value="default" selected>Default</option>
-            <option value="stable">Stable Network</option>
-            <option value="unstable">Unstable Network</option>
-          </select>
           <lable id="server-side-network-adaption-title" for="server-side-network-adaption" style="display: none; padding-top:5px">Server Side Network Adaption</lable>
           <select id="server-side-network-adaption" class="form-select" style="width:100%; display: none;">
-            <option value="default" selected>Default</option>
-            <option value="none">None</option>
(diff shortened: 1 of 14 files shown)


src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicy.ts
```diff
@@ -231,7 +231,7 @@ export default class VideoPriorityBasedPolicy implements VideoDownlinkBandwidthP
     const prefs = VideoPreferences.prepare();
 
     const numAttendees = attendeeIds.size;
-    let targetDisplaySize = TargetDisplaySize.High;
+    let targetDisplaySize = TargetDisplaySize.Maximum;
```
ltrung, line 234: Is this a bug fix?

demos/browser/app/meetingV2/meetingV2.html
```diff
@@ -149,18 +149,12 @@ <h5 class="modal-title" id="additional-options-modal-label">Additional Options</
             <label for="priority-downlink-policy" class="form-check-label">Use Priority-Based Downlink
               Policy</label>
           </div>
-          <lable id="priority-downlink-policy-preset-title" for="priority-downlink-policy-preset" style="display: none; padding-top:5px">Policy Preset</lable>
-          <select id="priority-downlink-policy-preset" class="form-select" style="width:100%; display: none;">
-            <option value="default" selected>Default</option>
```
ltrung, line 154: Do we want to remove it now as we still have not really removed it from our code yet we may want to keep it in order to test?

hensmi-amazon: This isn't currently in any QA regression tests AFAIK, so I think the benefits of not accidentally confusing a builder (or leading them to not use the latest SSNA) outweigh the benefits of being able to smoke test an infrequently used feature that we wouldn't fix issues for anyways (since we would just tell builders to use SSNA :P).

ltrung (approved):

PR #2739, opened 2023-08-23
Add a new meeting event "deviceLabelTriggerFailed"

simmkyu:
**Issue #:**
N/A

**Description of changes:**
- The Chime SDK for JavaScript invokes the device-label-trigger function when the device label is blank, such as in a private window. By default, the SDK requests access to both the microphone and camera to retrieve device labels. If either or both of these requests fail, the SDK will send a new meeting event `deviceLabelTriggerFailed`. (Before this PR, the SDK would emit `audioInputFailed` and `videoInputFailed` events simultaneously, which could lead to confusion.)

    If a custom function is provided via `meetingSession.audioVideo.setDeviceLabelTrigger`, the SDK will send this event if the custom function is unsuccessful.

**Testing:**

Yes.

1. Open the Chime SDK Serverless Demo in a private window, such as the Chrome Incognito window.
2. Open the browser console.
3. Enter your meeting title and name, then select the **Continue** button.
4. When the browser requests for mic and camera permission, deny it.
5. Confirm that the SDK sends the `deviceLabelTriggerFailed` event to the browser console.

**Checklist:**

Yes.

No.

No.


ltrung (dismissed):
So enumrateDevices will not throw any error?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Allow receiving notification messages from server and print messages received to console
+- Add a new meeting event, `deviceLabelTriggerFailed`, for device label trigger failures. By default, the Chime SDK for JavaScript requests access to the microphone and camera in order to retrieve device labels. The SDK will send the `deviceLabelTriggerFailed` event when either the microphone, camera or both requests fail. If a custom function is supplied with `meetingSession.audioVideo.setDeviceLabelTrigger`, the SDK will send this event when the custom function is not successful.
```
devalevenkatesh, line 13: Should we also mention that prior to this change, we would trigger two events: `audioInputFailed` and `videoInputFailed`? If anyone is relying on those events they can see why it was changed.

src/eventcontroller/EventName.ts
```diff
@@ -19,6 +19,7 @@ type EventName =
   | 'receivingAudioDropped'
   | 'sendingAudioFailed'
   | 'sendingAudioRecovered'
-  | 'backgroundFilterConfigSelected';
+  | 'backgroundFilterConfigSelected'
+  | 'deviceLabelTriggerFailed';
```
xuesichao, line 23: `deviceLabelTriggerFailed` may not be very accurate. For example there're two cases here to cause the failure:
1. camera, microphone device permission has been denied before join the meeting, in this case the device trigger is not triggered or (prompt does not show up).
2. end user denies the device permission in the pop up prompt.

Do we need to relate this event with deviceLabelTrigger? How about device permission acquisition failure?

For React SDK we have these states, just for reference.
https://github.com/aws/amazon-chime-sdk-component-library-react/blob/main/src/types/index.ts#L69-L74

```ts
export enum DeviceLabelTriggerStatus {
  UNTRIGGERED = 'UNTRIGGERED',
  IN_PROGRESS = 'IN_PROGRESS',
  GRANTED = 'GRANTED',
  DENIED = 'DENIED',
}
```

CHANGELOG.md
simmkyu: I will add this line in the next revision:
> Before this PR, the SDK would emit `audioInputFailed` and `videoInputFailed` events simultaneously, which could lead to confusion.

src/eventcontroller/EventName.ts
simmkyu: For both (1) and (2), the SDK invokes the `deviceLabelTrigger` function.

https://github.com/aws/amazon-chime-sdk-js/blob/20f2e1aa4c813ec1a885f581da0cb79c556d62c8/src/devicecontroller/DefaultDeviceController.ts#L824-L839



Naming this event presents some challenges for the following reasons:
- Subsequent functions like `enumerateDevices` may also generate an error.
- A developer could supply a custom function, such as performing an operation unrelated to device permissions, as explained here.
   https://github.com/aws/amazon-chime-sdk-js/blob/20f2e1aa4c813ec1a885f581da0cb79c556d62c8/src/devicecontroller/DeviceControllerFacade.ts#L25-L42


Given that our public API name `deviceLabelTrigger` is prone to causing an error, I have simply named it according to this API name.

src/devicecontroller/DefaultDeviceController.ts
```diff
@@ -830,12 +830,7 @@ export default class DefaultDeviceController
         }
       } catch (err) {
         this.logger.info('unable to get media device labels');
-        this.eventController?.publishEvent('audioInputFailed', {
-          audioInputErrorMessage: this.getErrorMessage(err),
-        });
-        this.eventController?.publishEvent('videoInputFailed', {
-          videoInputErrorMessage: this.getErrorMessage(err),
-        });
+        this.eventController?.publishEvent('deviceLabelTriggerFailed');
```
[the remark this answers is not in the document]
simmkyu: A good point! I'll add the `deviceLabelTriggerErrorMessage` to the https://github.com/aws/amazon-chime-sdk-js/blob/main/src/eventcontroller/DeviceEventAttributes.ts

ltrung (approved):

PR #2747, opened 2023-09-05
Block background throttling of video processing

hunnorth:
**Description of changes:**
Frames per second of video stream was being processed on the main thread using setTimeout. However, when a meeting window gets moved into the background, the main thread timer used in setTimeout gets throttled and the fps drops dramatically. To overcome this issue, this PR will move the setTimeout call into a worker thread that does not get throttled.

**Testing:**

Yes, start a meeting with at least one other attendee. Enable a video filtered and then hide the window (switch tabs). The other attendee should then observe no fps degradation in filter-enabled attendee.

**Checklist:**

yes

no

no


src/videoframeprocessor/DefaultVideoFrameProcessorPipeline.ts
```diff
@@ -45,8 +45,52 @@ export default class DefaultVideoFrameProcessorPipeline implements VideoFramePro
...
-  constructor(private logger: Logger, private stages: VideoFrameProcessor[]) {}
+  constructor(private logger: Logger, private stages: VideoFrameProcessor[]) {
+    this.workerTimer = this.createWorkerTimer();
+  }
+
+  /**
+   * Create a timer that exists within a web worker. This timer will be used to
+   * retrigger the process call whenever time expires.
+   */
+  private createWorkerTimer(): Worker {
```
hensmi-amazon, line 60: nit: This function could just fail and return undefined, which the calling code can use as an alternative to `isWorkerTimerSupported`. Then it could be moved outside of the class definition, in case we need to use this elsewhere.

hensmi-amazon: If you really want to have some fun and make this testable, you could make some `VideoFrameProcessorPipelineTimer` interface to make this easier to test :)

devalevenkatesh: Agree with Henry, this could be turned into a `util` and unit tested in itself.

kenpp:
Thank you for the development work. We have also encountered a similar issue in our service, as described by hunnorth.

Furthermore, we have verified that the problem is resolved when launching the demo application on the "video-ml/release" branch.

Could you please provide an estimate for when this release is expected? We would greatly appreciate it if you could release it as soon as possible.

src/videoframeprocessor/DefaultVideoFrameProcessorPipeline.ts
hensmi-amazon: Is it legal to resolve your own comments by pushing commits to someone else's PR? 🤔

src/videoframeprocessor/DefaultVideoTransformDevice.ts
```diff
@@ -31,7 +32,11 @@ export default class DefaultVideoTransformDevice
     private processors: VideoFrameProcessor[],
     private browserBehavior: BrowserBehavior = new DefaultBrowserBehavior()
   ) {
-    this.pipe = new DefaultVideoFrameProcessorPipeline(this.logger, this.processors);
+    this.pipe = new DefaultVideoFrameProcessorPipeline(
+      this.logger,
+      this.processors,
+      new DefaultVideoFrameProcessorTimer()
```
devalevenkatesh, line 38: This explicit passing is not needed as you create one by default when instantiating.

src/videoframeprocessor/DefaultVideoFrameProcessorTimer.ts
```diff
@@ -0,0 +1,70 @@
...
+      };
+      this.workerTimer.postMessage(delay);
+    } else {
+      this.lastTimeout = setTimeout(() => {
+        this.callback();
+      }, delay);
+    }
+  }
+
+  destroy(): void {
```
devalevenkatesh, line 61: Not necessary but the assigned `callback` could also be set back to `undefined`.

src/videoframeprocessor/VideoFrameProcessorTimer.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * [[VideoFrameProcessorTimer]] defines a timer used in the context of video frame processing.
+ */
+export default interface VideoFrameProcessorTimer {
```
devalevenkatesh, line 7: may not be needed now, but should we make `start` and `destroy` promise based? We can make this interface destroyable by extending existing `Destroyable` interface:
https://github.com/aws/amazon-chime-sdk-js/blob/faa246235609d63486a1d20ed54026ff78f82e50/src/destroyable/Destroyable.ts#L11

ltrung (approved):

src/videoframeprocessor/VideoFrameProcessorTimer.ts
hensmi-amazon: good point, will change! I wish js had a better way of transitioning from sync -> async APIs in a backwards compatible way.

hensmi-amazon:
@kenpp unfortunately even after this is merged it may be until December until we do another release. On a related note, I and others are no longer able to reproduce the original issue, but not sure why. Is it still reproducible on your machine?

devalevenkatesh (approved):

PR #2763, opened 2023-09-26
Limit max Redundant Audio Encoding packet size to 1000 bytes.

dinmin-amzn:
**Issue #:**

**Description of changes:**
In Chromium-based browsers, writing audio frames larger than 1000 bytes will cause an error to be thrown, so we limit the max Redundant Audio Encoding packet size to 1000 bytes. See https://crbug.com/1248479.

**Testing:**
Previously, starting a call with WebAudio, Redundant Audio Encoding, and fullband music stereo quality all enabled and introducing high uplink packet loss (>20%) would cause the Redundant Audio Encoding payloads to exceed the Chromium-imposed limit and audio sending would stop working for the rest of the meeting. This issue is fixed by this PR.

*Can these be tested using a demo application? Please provide reproducible step-by-step instructions.*
1. Have two attendees join a meeting using Chrome with WebAudio, Redundant Audio Encoding, and fullband music stereo quality all enabled
2. Enable audio sending and receiving for both clients
3. Introduce >20% uplink packet loss
4. Both attendees should still be able to receive and hear audio from the other attendee

**Checklist:**

y

n

n

### DIFF
--- a/src/redundantaudioencoder/RedundantAudioEncoder.ts
+++ b/src/redundantaudioencoder/RedundantAudioEncoder.ts
@@ -5,8 +5,9 @@ export default class RedundantAudioEncoder {
   // Each payload must be less than 1024 bytes to fit the 10 bit block length
   private readonly maxRedPacketSizeBytes = 1 << 10;
 
-  // Limit payload to 1100 bytes to handle small MTU
-  private readonly maxAudioRtpPacketSizeBytes = 1100;
+  // Limit payload to 1000 bytes to handle small MTU. 1000 is chosen because in Chromium-based browsers, writing audio
+  // frames larger than 1000 bytes will cause an error to be thrown. See https://crbug.com/1248479.
+  private readonly maxAudioRtpPacketSizeBytes = 1000;
 
   // Each payload can encode a timestamp delta of 14 bits
   private readonly maxRedTimestampOffset = 1 << 14;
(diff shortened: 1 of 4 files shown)


src/redundantaudioencoder/RedundantAudioEncoder.ts
```diff
@@ -236,12 +237,12 @@ export default class RedundantAudioEncoder {
     frame: RTCEncodedAudioFrame,
     controller: TransformStreamDefaultController
   ): void {
-    // const frameMetadata = frame.getMetadata();
```
ltrung, line 239: Is this a miss from previous PR? Does not seem related with the issue this PR tries to fix.

pracheth: Yeah this was missed from the previous CR

pracheth (dismissed):

src/redundantaudioencoder/RedundantAudioEncoder.ts
pracheth: @dinmin-amzn Curious why a unit test did not catch this. Maybe we can add one?

dinmin-amzn: Yes, let me add a unit test for this

mklingb (approved):

src/redundantaudioencoder/RedundantAudioEncoder.ts
dinmin-amzn: Updated the existing unit tests to better capture this case.

test/redundantaudioencoder/RedundantAudioEncoder.test.ts
```diff
@@ -574,20 +574,51 @@ describe('RedundantAudioEncoder', () => {
...
       encoder.setRedPayloadType(redPayloadType + 1);
+      frame = createRTCEncodedAudioFrame(packetizationTime * i, redPayloadType, 0);
+      frame.data = createRedPayload([header], [importantPayload]);
       encoder['senderTransform'](frame, controller);
-      expect(frame.data.byteLength).to.equal(0);
+      expect(frame.data.byteLength).to.equal(header.byteLength + importantPayload.byteLength);
     });
 
     it('forwards frames when the primary payload cannot be determined', () => {
+      const header = createRedHeader(
```
dinmin-amzn, line 600: I went ahead and made this test case more effective as well since it is similar in nature to the previous test case.

ltrung (approved):

PR #2766, opened 2023-09-29
Fix bug to allow use of Video Simulcast and Redundant Audio Encoding at the same time

dinmin-amzn:
**Issue #:**

**Description of changes:**
Fix bug to allow use of Video Simulcast and Redundant Audio Encoding at the same time. This was detected by Transcription canary failures and looking deeper, the canaries were failing because they were using Video Simulcast and Redundant Audio Encoding at the same time, which prevent sending and receiving both video and audio and thus caused the Transcription canary to fail since it wasn't receiving audio.

**Testing:**

1. Have two attendees join a meeting with Video Simulcast and Redundant Audio Encoding both enabled
2. Verify that both attendees can send and receive audio, video, and content share

**Checklist:**

y

n

n


src/transceivercontroller/DefaultTransceiverController.ts
```diff
@@ -586,13 +586,10 @@ export default class DefaultTransceiverController
...
-   * Adds a transceiver to the peer connection and performs additional necessary setup.
+   * Set up and add a video transceiver to the peer connection.
    */
-  private addTransceiver(
-    trackOrKind: string | MediaStreamTrack,
-    init?: RTCRtpTransceiverInit
-  ): RTCRtpTransceiver {
-    const transceiver = this.peer.addTransceiver(trackOrKind, init);
+  protected setupVideoTransceiver(init?: RTCRtpTransceiverInit): RTCRtpTransceiver {
+    const transceiver = this.peer.addTransceiver('video', init);
```
dinmin-amzn, line 592: I just did a small refactor with this function to make the intention more clear.

src/transceivercontroller/SimulcastTransceiverController.ts
```diff
@@ -98,11 +98,17 @@ export default class SimulcastTransceiverController extends DefaultTransceiverCo
         direction: 'inactive',
         streams: [this.defaultMediaStream],
       });
+
+      /* istanbul ignore next */
+      if (this.meetingSessionContext?.audioProfile.hasRedundancyEnabled()) {
+        // This will perform additional necessary setup for the audio transceiver.
+        this.setupAudioRedWorker();
+      }
```
dinmin-amzn, line 106: This is the actual fix for the issue.

src/transceivercontroller/DefaultTransceiverController.ts
```diff
@@ -586,13 +586,10 @@ export default class DefaultTransceiverController
...
   /**
-   * Adds a transceiver to the peer connection and performs additional necessary setup.
+   * Set up and add a video transceiver to the peer connection.
    */
-  private addTransceiver(
-    trackOrKind: string | MediaStreamTrack,
-    init?: RTCRtpTransceiverInit
-  ): RTCRtpTransceiver {
-    const transceiver = this.peer.addTransceiver(trackOrKind, init);
+  protected setupVideoTransceiver(init?: RTCRtpTransceiverInit): RTCRtpTransceiver {
```
devalevenkatesh, line 591: - Only optional arguments are not recommended, from earlier method do we need to keep `trackOrKind` around?
- We currently only explicilty want to pass 'video' so instead of creating a new method, should we use `addTransceiver` and pass 'video' to it? Not sure if this is being re-used elsewhere in this class and whether some string or media stream track is being passed.

src/transceivercontroller/SimulcastTransceiverController.ts
```diff
@@ -98,11 +98,17 @@ export default class SimulcastTransceiverController extends DefaultTransceiverCo
         direction: 'inactive',
         streams: [this.defaultMediaStream],
       });
+
+      /* istanbul ignore next */
+      if (this.meetingSessionContext?.audioProfile.hasRedundancyEnabled()) {
```
devalevenkatesh, line 103: Can `audioProfile` be `undefined` as well may not have `hasRedundancyEnabled` implemented?

src/transceivercontroller/SimulcastTransceiverController.ts
```diff
@@ -98,11 +98,17 @@ export default class SimulcastTransceiverController extends DefaultTransceiverCo
...
+        // This will perform additional necessary setup for the audio transceiver.
+        this.setupAudioRedWorker();
+      }
     }
 
     if (!this._localCameraTransceiver) {
       const encodingParams = Array.from(this.videoQualityControlParameterMap.values());
-      this._localCameraTransceiver = this.peer.addTransceiver('video', {
+      this._localCameraTransceiver = this.setupVideoTransceiver({
```
devalevenkatesh, line 111: Hmm, I dont understand the need for this. Id `this.peer.addTransceiver` not working?

src/transceivercontroller/SimulcastTransceiverController.ts
```diff
@@ -98,11 +98,17 @@ export default class SimulcastTransceiverController extends DefaultTransceiverCo
         direction: 'inactive',
         streams: [this.defaultMediaStream],
       });
+
+      /* istanbul ignore next */
```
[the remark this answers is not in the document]
dinmin-amzn: It's because of the safe navigation that we use here `this.meetingSessionContext?.audioProfile.hasRedundancyEnabled()`. I guess the "else" part of the safe navigation is untested. I couldn't find a way to ignore just the safe navigation so I ignored the whole thing instead.

src/transceivercontroller/DefaultTransceiverController.ts
mklingb: I thought that the capability to add an existing MediaStreamTrack (rather than having the PeerConnection create a new one) is important. But will need to check if that is being used.

dinmin-amzn: This function was actually added by the RED changes to setup passthrough transforms for video transceivers. Otherwise, RED would cause video send/receive to not work. `trackOrKind` is not needed if this is only intended to be used for video transceivers and I've updated all other uses of this function to not pass a string/media stream track in this PR.

src/transceivercontroller/SimulcastTransceiverController.ts
dinmin-amzn: Yes, we should add safe navigation to `audioProfile` as well since it can be `null`.

src/transceivercontroller/SimulcastTransceiverController.ts
dinmin-amzn: The `setupVideoTransceiver` does some additional setup after adding the video transceiver to the peer connection. I also sets up pass through transforms if RED is enabled, which allows video to be sent/received. This is actually an issue with [WebRTC Encoded Transform](https://www.w3.org/TR/webrtc-encoded-transform/) where enabling this feature makes it so that every single transceiver requires a transform or no media send/receive will work on that transceiver.

src/transceivercontroller/DefaultTransceiverController.ts
dinmin-amzn: Thinking more about this we would probably want to keep the ability to add a `MediaStreamTrack`, I will undo the refactors for this function as it is not necessary for the fix.

src/transceivercontroller/SimulcastTransceiverController.ts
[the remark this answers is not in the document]
dinmin-amzn: I added a test case and removed the `istanbul ignore`

src/transceivercontroller/DefaultTransceiverController.ts
dinmin-amzn: I reverted these refactors in latest revision.

src/transceivercontroller/SimulcastTransceiverController.ts
dinmin-amzn: Added safe navigation operators.

test/transceivercontroller/SimulcastTransceiverController.test.ts
```diff
@@ -71,6 +71,27 @@ describe('SimulcastTransceiverController', () => {
...
+      const audioTransceiver = transceivers[0];
+      expect(audioTransceiver.direction).to.equal('inactive');
+      expect(audioTransceiver.receiver.track.kind).to.equal('audio');
+      expect(audioTransceiver.sender.track.kind).to.equal('audio');
+
+      const videoTransceiver = transceivers[1];
+      expect(videoTransceiver.direction).to.equal('inactive');
+      expect(videoTransceiver.receiver.track.kind).to.equal('video');
+      expect(videoTransceiver.sender.track.kind).to.equal('video');
+    });
```
pracheth, line 93: nit: maybe also check that `tc['audioRedWorker']` is null/undefined?

pracheth (approved):

test/transceivercontroller/SimulcastTransceiverController.test.ts
dinmin-amzn: It looks like I can't do that since `audioRedWorker` would need to be `protected` instead of `private` in `DefaultTransceiverController` for it to be accessible (even with that syntax) in `SimulcastTransceiverController`

pracheth (approved):

devalevenkatesh (approved):

PR #2772, opened 2023-10-05
Save the redundant audio worker code during build time so that the worker code stays intact and is able to be loaded

dinmin-amzn:
**Issue #:** https://github.com/aws/amazon-chime-sdk-js/issues/2771
**Description of changes:**
The redundant audio worker code string may vary based on the build system of different projects. Because of this, certain build systems will cause invalid worker code to be generated and the redundant audio worker will not be able to run. Now, the redundant audio worker code is saved at build time so that all clients can run the redundant audio worker with the same exact code, regardless of the client's build system.

This issue was originally found in https://github.com/aws/amazon-chime-sdk-js/issues/2771.

**Testing:**

Reproduction steps for the original issue are mentioned in https://github.com/aws/amazon-chime-sdk-js/issues/2771. Following the same steps with this change should fix the issue.

**Checklist:**

y

n

n


hensmi-amazon (dismissed):

src/redundantaudioencoder/RedundantAudioEncoder.ts
```diff
@@ -1480,3 +1480,15 @@ namespace RedundantAudioEncoder {
     windowSize: number;
   }
 }
+
+/* istanbul ignore next */
```
ltrung, line 1484: Can't we just write unit test for this?

src/redundantaudioencoder/RedundantAudioEncoder.ts
```diff
@@ -1480,3 +1480,15 @@ namespace RedundantAudioEncoder {
...
+/**
+ * This function is only meant to be called within the RED worker to initialize and start RED functionality.
+ */
+function startRedWorker(): void {
+  RedundantAudioEncoder.shouldLogDebug = true;
```
pracheth, line 1489: This should be false, yes?

src/redundantaudioencoder/RedundantAudioEncoder.ts
dinmin-amzn: Yes, I can add unit tests for this

src/redundantaudioencoder/RedundantAudioEncoder.ts
dinmin-amzn: We want the RED worker to be able to emit logs, so this should be true

dinmin-amzn: The encoder file only has low frequency logs at the moment, so this shouldn't overload the main thread with too many messages

ltrung: Can we just based on the Log level defined ?

dinmin-amzn: Updated to configure this based on log level

config/tsconfig.red.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "../build",
+    "rootDir": "../src"
+  },
+  "include": [
```
ltrung, line 7: Why do we need a separate config for just RED?

script/generate-red-worker-code.js
```diff
@@ -0,0 +1,39 @@
...
+ * This script takes the transpiled JavaScript code for `RedundantAudioEncoder` and places all of that code into a
+ * string that can be used to run a worker. 
+ */
+
+const fs = require('fs');
+const RedundantAudioEncoder = require('../build/redundantaudioencoder/RedundantAudioEncoder.js').default;
+
+const debugLine = 'RedundantAudioEncoder.shouldLogDebug = true;';
+const withDebug = `${RedundantAudioEncoder.toString()}
+${debugLine}
```
ltrung, line 13: can we just move the debug like to here like other line to be consistent?

pracheth (commented):
Wondering if we should make these lines `INFO` rather than `DEBUG` while we're at it?

```
[DEBUG] ChimeLogger - [AudioRed] Initializing RedundantAudioEncoder
[DEBUG] ChimeLogger - [AudioRed] Setting up sender RED transform
[DEBUG] ChimeLogger - [AudioRed] Setting up receiver RED transform
[DEBUG] ChimeLogger - [AudioRed] red payload type set to 63
[DEBUG] ChimeLogger - [AudioRed] opus payload type set to 111
```

config/tsconfig.red.json
pracheth: Same question. Using the existing tsconfig file I see `RedundantAudioEncoder.js` being generated in the build folder? Doesn't that mean we don't need this extra tsconfig file?

src/redundantaudioencoder/RedundantAudioEncoderWorkerCode.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * This file was generated with the `generate-red-worker-code.js` script.
+ */
+
+export default class RedundantAudioEncoderWorkerCode {
```
ltrung, line 8: nit: Can we rename the folder to match with the file name (..workercode)?

script/generate-red-worker-code.js
```diff
@@ -0,0 +1,39 @@
...
+RedundantAudioEncoder.shouldReportStats = true;
+RedundantAudioEncoder.initializeWorker();
+`.replace(/"/g, '\'').replace(/\n/g, "\\n")
```
ltrung, line 16: Where does the `"` come from?

config/tsconfig.red.json
ltrung: I assume this is because we need to convert ts -> js for this file first and put it in the source folder so when we transpile for the whole folder they can refer to this file

dinmin-amzn: The idea for generating and building the worker code was:
1. Build the `RedundantAudioEncoder.js` file
2. Use the `RedundantAudioEncoder.js` file to generate `RedundantAudioEncoderWorkerCode.ts`
3. Now build everything using `npm run tsc:src` so that the generated `RedundantAudioEncoderWorkerCode.ts` file also gets built

So yes, we don't need this extra tsconfig file, but using the normal `tsconfig.json` would generate all of the built files for Step 1, then the `RedundantAudioEncoderWorkerCode.ts` file would be generated in Step 2, and then all of the files would be re-built again in Step 3. Having the separate config for RED was just a minor build time optimization so that we wouldn't be building all of the code twice during each `npm run build`.

dinmin-amzn:
> Wondering if we should make these lines `INFO` rather than `DEBUG` while we're at it?
> 
> ```
> [DEBUG] ChimeLogger - [AudioRed] Initializing RedundantAudioEncoder
> [DEBUG] ChimeLogger - [AudioRed] Setting up sender RED transform
> [DEBUG] ChimeLogger - [AudioRed] Setting up receiver RED transform
> [DEBUG] ChimeLogger - [AudioRed] red payload type set to 63
> [DEBUG] ChimeLogger - [AudioRed] opus payload type set to 111
> ```

Yes, we should make these `INFO` level

src/redundantaudioencoder/RedundantAudioEncoderWorkerCode.ts
dinmin-amzn: I will make this change

script/generate-red-worker-code.js
dinmin-amzn: Some of the comments in `RedundantAudioEncoder` class use double quotes. I can change the class to only use single quotes, which will remove the need for that replacement.

ltrung: Yeah I think for styling we always use ' in this repo so probably want to update the original file also.

dinmin-amzn: Sure, I'll leave the regex there for safety, but I'll change all of the double quotes in the original file to single quotes.

pracheth (approved):

pracheth (dismissed):

script/generate-red-worker-code.js
```diff
@@ -0,0 +1,68 @@
...
+  "compilerOptions": {
+    "outDir": "../build",
+    "rootDir": "../src"
+  },
+  "include": [
+    "../src/redundantaudioencoder/RedundantAudioEncoder.ts",
+  ]
+}
+`;
+fs.writeFileSync(`${configDir}/${redTsconfig}`, redTsconfigContent);
```
ltrung, line 27: could we add this temp file to .gitignore?

dinmin-amzn: sure, let me add it

script/generate-red-worker-code.js
```diff
@@ -0,0 +1,68 @@
...
+
+// Transpile `RedundantAudioEncoder.ts`.
+const command = 'npx tsc --build config/tsconfig.red.json';
```
ltrung, line 30: nit: Can we not depend on npx and just use the local node_module path to tsc? otherwise you would need to update READM to install it.

dinmin-amzn: Updated to use local `tsc` from `node_modules` in latest push

ltrung (approved):

pracheth (approved):

PR #2791, opened 2023-10-27
Do no enqueue any audio payloads larger than 1000 bytes

dinmin-amzn:
**Issue #:** n/a

**Description of changes:**
In Chromium-based browsers, writing audio payloads larger than 1000 bytes using the WebRTC Insertable Streams API (which is used to enable dynamic audio redundancy) will cause an error to be thrown and cause audio flow to permanently stop. See https://crbug.com/1248479.

**Testing:**

1. Have attendee A join with the `Set fullband music (stereo) quality` selected (this will cause the audio frames to be especially large due to the higher quality)
2. Have attendee B join normally
3. Start content share from attendee A, select a tab that actually has audio, and enable the `Also share tab audio` option
4. Use Network Link Conditioner to simulate >20% downlink packet loss (this will cause the downlink packets to contain up to 2 extra RED packets and have up to 3x the audio payload size)
5. Attendee B should *NOT* see the `Uncaught (in promise) DOMException: Failed to execute 'write' on 'UnderlyingSinkBase': Frame too large` error and Attendee B should receive audio normally

**Checklist:**

y

n

n


src/redundantaudioencoder/RedundantAudioEncoder.ts
```diff
@@ -227,6 +229,21 @@ export default class RedundantAudioEncoder {
     RedundantAudioEncoder.log(`redundancy ${this.redundancyEnabled ? 'enabled' : 'disabled'}`);
   }
 
+  /**
+   * Helper function to only enqueue audio frames if they do not exceed the audio payload byte limit imposed by
+   * Chromium-based browsers. Chromium will throw an error (https://crbug.com/1248479) if an audio payload larger than
+   * 1000 bytes is enqueued. Any controller that attempts to enqueue an audio payload larger than 1000 bytes will
+   * encounter this error and will permanently stop sending or receiving audio.
+   */
+  private maybeEnqueueAudioFrame(
```
hensmi-amazon, line 238: Since its such a short function we can probably make the name explicit, maybe `enqueueAudioFrameIfPayloadSizeIsValid`? Borderline too verbose, up to you. I myself have felt a bit mid towards the prefix 'maybe' after using it a bit over the years.

test/redundantaudioencoder/RedundantAudioEncoder.test.ts
```diff
@@ -541,6 +542,57 @@ describe('RedundantAudioEncoder', () => {
...
+        error: () => {},
+        terminate: () => {},
+      };
+    });
+
+    it('enqueues frames that are <= the audio payload size limit', () => {
+      const enqueueSpy = sinon.spy(controller, 'enqueue');
+
+      expect(enqueueSpy.calledOnce).to.be.false;
+      frame.data = new ArrayBuffer(encoder['maxAudioPayloadSizeBytes'] - 1);
```
hensmi-amazon, line 573: No way to test this without these private function accesses?

hensmi-amazon (dismissed):

src/redundantaudioencoder/RedundantAudioEncoder.ts
dinmin-amzn: We can go the verbose route as this is a private function anyway. I only used `maybe` since it looked like an existing pattern but I'm not strongly for that pattern either.

test/redundantaudioencoder/RedundantAudioEncoder.test.ts
dinmin-amzn: I can try to change this to test these cases without the need for private function accesses.

dinmin-amzn: I didn't avoid the private function accesses completely since it would have required a lot more mocking that seemed overkill for this small change, but I did update the tests to more closely resemble that actual edge cases that this change would protect against.

hensmi-amazon (approved):

pracheth (approved):

PR #2816, opened 2023-12-13
Fix content send with Send attendee capability

hensmi-amazon:
**Issue #:** None

**Description of changes:** With previous changes we would hit `this.context.currentVideoSendCodec === undefined` if attendee capability is Send (yes, JS Strict mode would have caught this :) ).

**Testing:**

Can send using Send AC again.

1. Pick Send for Content AC in additional options
2. Send content share and verify on remote attendee

**Checklist:**

y

n

n


pracheth (approved):

src/task/SetRemoteDescriptionTask.ts
```diff
@@ -46,9 +47,8 @@ export default class SetRemoteDescriptionTask extends BaseTask {
       this.context.videoSendCodecPreferences.length > 0
     ) {
       sdp = new SDP(sdp).withVideoSendCodecPreferences(
-        this.context.meetingSupportedVideoSendCodecPreferences !== undefined
-          ? this.context.meetingSupportedVideoSendCodecPreferences
-          : this.context.videoSendCodecPreferences
+        this.context.meetingSupportedVideoSendCodecPreferences ??
```
devalevenkatesh, line 50: I checked the support for `??` operator and it seems mostly its fine but some Chrome and Edge version will fail to parse it:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing
- https://docs.aws.amazon.com/chime-sdk/latest/dg/meetings-sdk.html#mtg-browsers

Until we bump that documentation probably we should still fallback to older version :)

devalevenkatesh (approved):

PR #2823, opened 2024-01-06
Fix endpoint override for demo deployment

ltrung:
**Description of changes:**
Fix endpoint override for demo deployment. After migrating to aws sdk js v3, the previous syntax for override endpoint did not work.

**Testing:**

Testing serverless demo with overriden endpoint

**Checklist:**

2. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved? N/A

3. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved? N/A


demos/serverless/deploy.js
```diff
@@ -11,7 +11,7 @@ let app = `meetingV2`;
 let useEventBridge = false;
 let enableTerminationProtection = false;
 let disablePrintingLogs = false;
-let chimeSDKMeetingsEndpoint = 'https://service.chime.aws.amazon.com';
+let chimeSDKMeetingsEndpoint = 'https://meetings-chime.us-east-1.amazonaws.com';
```
Chitz, line 14: Should we generate the endpoint based on region?

ltrung: Ideally yes but these values are default value and the default region value is us-east-1 so it does not make any difference.

demos/browser/server.js
```diff
@@ -24,13 +24,13 @@ const indexPage = fs.readFileSync(indexPagePath);
 
 const currentRegion = process.env.REGION || 'us-east-1';
 
-const chimeSDKMediaPipelines = new ChimeSDKMediaPipelines({ region: 'us-east-1' });
-chimeSDKMediaPipelines.endpoint = process.env.CHIME_SDK_MEDIA_PIPELINES_ENDPOINT || "https://media-pipelines-chime.us-east-1.amazonaws.com"
+const chimeSDKMediaPipelines = new ChimeSDKMediaPipelines({
+  region: 'us-east-1',
```
devalevenkatesh, line 28: Should we use `currentRegion` her as well?

ltrung: I think SDK media pipeline is currently hard-coded to us-east-1 by default. Thought about fixing it but decided not to as it is not related to this PR which is just to fix endpoint syntax for the SDK client.

devalevenkatesh: Got it, thanks!

PR #2825, opened 2024-01-08
Node v20 support

j0d0nn:
**Issue #2654**

**Description of changes:**

Adds v20 to the list of supported node versions and v10 to the supported npm versions. The only non-test code that was changed was the way to get at the webcrypto API, which now requires going through the `node:crypto` package.

**Testing:**

Tested using `npm run test` against a local node 20.10.0. I also installed the built package into my personal app that's using Node 20 and verified the functionality (audio only and data messages, but it works!).

**Checklist:**

1. Have you successfully run `npm run build:release` locally?
yes - the docs end up modified because the URLs point to my fork; not committing those.

2. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved?
no

3. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved?
no

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2827, opened 2024-01-18
Add support for node 20 and drop support for node < 18

xuesichao:
**Issue #2825:**

**Description of changes:**
Add support for node 20 and drop support for node < 18
* node < 18 has reached end of life, https://endoflife.date/nodejs

**Testing:**

Verified `npm install` works for demos and sdks. Tested meeting demo e2e.
#### Node 18
```
Statements   : 100% ( 11908/11908 )
Branches     : 100% ( 5399/5399 )
Functions    : 100% ( 2177/2177 )
Lines        : 100% ( 11730/11730 )

~/Projects/amazon-chime-sdk-js node20* 4m 13s                                                   10:50:49 PM
❯ node -v          
v18.17.1
```

#### Node 20
```
Statements   : 100% ( 11908/11908 )
Branches     : 100% ( 5399/5399 )
Functions    : 100% ( 2177/2177 )
Lines        : 100% ( 11730/11730 )

~/Projects/amazon-chime-sdk-js node20* 4m 16s                                                   10:55:43 PM
❯ node -v          
v20.11.0
```
N/A

**Checklist:**

Yes, succeeded.

No

No


src/utils/Utils.ts
```diff
@@ -69,3 +69,11 @@ export function getFormattedOffset(utcOffset: number): string {
 
   return `${offsetOperator}${offsetHours}:${offsetMinutes}`;
 }
+
+/**
+ * Returns a Crypto object, either from the window or from the node:crypto package.
+ * @returns Get the crypto object from the global scope.
+ */
+export function getCrypto(): Crypto {
```
xuesichao, line 77: When running unit tests, it's node env not browser env, we need this to be able to pass the unit test.
Check https://nodejs.org/api/crypto.html#determining-if-crypto-support-is-unavailable

ltrung: We do not support Node enviroment though. In our unit test can we catch the error and just return math.random instead?

xuesichao: We did mock `crypto.getRandomValues()` here:
https://github.com/aws/amazon-chime-sdk-js/blob/629b521b94354165200cf03b1353cefcb247ebb4/test/dommock/DOMMockBuilder.ts#L1526

But the problem I'm having now is, when running unit test using node 20, it throws this error:
```
     TypeError: Cannot set property crypto of #<Object> which has only a getter
      at new DOMMockBuilder (test/dommock/DOMMockBuilder.ts:1525:21)
      at Context.<anonymous> (test/messagingsession/MessagingSessionConfiguration.test.ts:16:22)
      at processImmediate (node:internal/timers:478:21)
```

I'm trying to resolve this and then we should be good.

ltrung: Hmm so maybe we can just check on whether window.crypto exists, log a warning and fall back to Math.random instead? We do not support Node so kinda prefer just fallback to something else.

integration/js/app/package.json
```diff
@@ -13,12 +13,12 @@
     "html-webpack-plugin": "^5.3.2",
     "ts-loader": "^9.1.1",
     "typescript": "^4.2.4",
-    "webpack": "^5.36.2",
-    "webpack-cli": "^4.6.0"
+    "webpack": "^5.89.0",
+    "webpack-cli": "^4.8.0"
```
xuesichao, line 17: need to update the webpack version to avoid this issue:
https://github.com/webpack/webpack/issues/14532

The webpack hashing algorithm uses `node:crypto`

src/utils/Utils.ts
xuesichao: Updated!

ltrung (approved):

PR #2830, opened 2024-01-23
Add support for H.264 profiles besides Constrained Baseline Profile

hensmi-amazon:
**Issue #:** None

**Description of changes:** Additional profiles are desired as they are often more likely to use hardware encoder then H.264 CBP. This change also includes changes that allows 'falling-forwards' to more preferred send codecs if limited receivers leave the call.

This also has a fix for the protobuf generation script since node@10 is not available on `brew` anymore.

**Testing:**

1. Select H.264 Baseline, Main, or High from Additional options before joining the meeting on Chrome or Safari.
2. Enable video. Join from another Chrome device.
3. Join a third attendee from Firefox that will not have support for these profiles. The sender should fall back.
4. Leave the Firefox attendee. Note that there is a known issue with H.264 High profile on MacOS and there is a bug open with WebRTC, so this will lead to freezes currently. https://bugs.chromium.org/p/chromium/issues/detail?id=1520287 . The docstring indicates that this should not be used with fallbacks.

**Checklist:**

y

yes but the APIs match existing patterns so review is not necessary.

n


demos/browser/app/meetingV2/meetingV2.html
```diff
@@ -180,6 +183,9 @@ <h5 class="modal-title" id="additional-options-modal-label">Additional Options</
             <option value="default">Meeting Default</option>
             <option value="vp8">VP8</option>
             <option value="h264ConstrainedBaselineProfile">H.264 Constrained Baseline Profile</option>
+            <option value="h264BaselineProfile">H.264 Baseline Profile</option>
+            <option value="h264MainProfile">H.264 Main Profile</option>
+            <option value="h264HighProfile">H.264 High Profile</option>
```
shi-su, line 188: So I guess we don't bother putting constrained high in as a demo option

shi-su (approved):

src/transceivercontroller/DefaultTransceiverController.ts
```diff
@@ -435,7 +435,7 @@ export default class DefaultTransceiverController
     }
 
     if (track) {
-      transceiver.direction = 'sendrecv';
+      transceiver.direction = 'sendonly';
```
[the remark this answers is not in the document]
hensmi-amazon: Oh yikes good catch, this was just for testing.

demos/browser/app/meetingV2/meetingV2.html
hensmi-amazon: Actually, I'm looking at an iOS safari SDP now and it seems to include only constrained high, and not high. I think i'll look more into the fallback logic we were discussing.

shi-su (dismissed):

src/task/ReceiveVideoStreamIndexTask.ts
```diff
@@ -220,25 +220,23 @@ export default class ReceiveVideoStreamIndexTask
...
+    }
+
+    if (newMeetingSupportedVideoSendCodecPreferences.length > 0) {
+      // We need to renegotiate if we are currently sending a codec that is no longer supported in the call, or if we can
+      // start sending a higher preference codec again (due to a constrained remote attendee leaving)
       if (
         this.context.currentVideoSendCodec !== undefined &&
-        !codecSupported &&
-        capability.equals(this.context.currentVideoSendCodec)
+        !newMeetingSupportedVideoSendCodecPreferences[0].equals(this.context.currentVideoSendCodec)
```
ltrung, line 236: So the first codec in videoSendCodecPreferences is always the current send codec?

src/task/AttachMediaInputTask.ts
```diff
@@ -18,11 +18,7 @@ export default class AttachMediaInputTask extends BaseTask {
   async run(): Promise<void> {
     const transceiverController = this.context.transceiverController;
     transceiverController.setPeer(this.context.peer);
-    try {
-      transceiverController.setupLocalTransceivers();
-    } catch (error) {
-      throw error;
-    }
+    transceiverController.setupLocalTransceivers();
```
ltrung, line 21: so this would not throw any error ?

ltrung (approved):

demos/browser/app/meetingV2/meetingV2.html
hensmi-amazon: Made change on backend to forward CH as H, and included Constrain High when 'High' option is chosen.

src/task/AttachMediaInputTask.ts
hensmi-amazon: The catch just rethrows it :P. My linter was complaining about it so figured i'd clean it up.

src/task/ReceiveVideoStreamIndexTask.ts
hensmi-amazon: I thought so...but now that I think about it there is a pretty unlikely edge case where the preference list starts with a codec that can be received by all clients (including the sender) but not sent. While this is not possible on any existing browser/codec, let me see if I can harden this.

src/task/AttachMediaInputTask.ts
ltrung: I am blind :))

ltrung (dismissed):

ltrung (approved):

PR #2849, opened 2024-03-01
Avoid unnecessary transceiver creation by using no-video policy for content share

hensmi-amazon:
**Issue #:** N/A

**Description of changes:** 
TLDR: It probably could be reproduced without content but basically,
* The content share peer connection actually subscribes to all remote videos but the backend just marks all sections as inactive (has always been like this).
* On the backend, we label the inactive sections with the static label inactive, as some multiple year old mitigation I added for a Firefox Unsignaled SSRC bug.
* [libwebrtc now rejects the SDP](https://webrtc.googlesource.com/src/+/6a8236617dd9416ea0fe7767166cc8e8c729d5e6) if there are two sections with the same msid like that, since the spec.

I'm pushing a fix on the backend to disambiguate the dummy MSIDs in the next week which fully mitigates this issue, but felt like removing the unnecessary subscriptions from the client side as well.

**Testing:**
Mitigates issue without aforementioned backend fix.

1. Join 2 attendees with video
2. Enable content share

For unrelated demo fix
1. Join replica meeting and promote
2. Enable video
3. Demote
4. Demo should automatically turn off video

**Checklist:**

y

n

n

### DIFF
--- a/demos/browser/app/meetingV2/meetingV2.ts
+++ b/demos/browser/app/meetingV2/meetingV2.ts
@@ -1586,9 +1586,10 @@ export class DemoMeetingApp
       this.toggleButton(button, isPromoted ? 'off' : 'disabled');
     }
 
-    // Additionally mute audio so it's not in an unexpected state when demoted
+    // Additionally mute audio and stop local video so it's not in an unexpected state when demoted
     if (!isPromoted) {
       this.audioVideo.realtimeMuteLocalAudio();
+      this.audioVideo.stopLocalVideoTile();
     }
   }
 
--- a/src/contentsharecontroller/DefaultContentShareController.ts
+++ b/src/contentsharecontroller/DefaultContentShareController.ts
@@ -14,6 +14,7 @@ import DefaultModality from '../modality/DefaultModality';
 import AsyncScheduler from '../scheduler/AsyncScheduler';
 import VideoCodecCapability from '../sdp/VideoCodecCapability';
 import { Maybe } from '../utils/Types';
+import NoVideoDownlinkBandwidthPolicy from '../videodownlinkbandwidthpolicy/NoVideoDownlinkBandwidthPolicy';
 import VideoTile from '../videotile/VideoTile';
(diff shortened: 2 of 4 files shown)


demos/browser/app/meetingV2/meetingV2.ts
```diff
@@ -1586,9 +1586,10 @@ export class DemoMeetingApp
       this.toggleButton(button, isPromoted ? 'off' : 'disabled');
     }
 
-    // Additionally mute audio so it's not in an unexpected state when demoted
+    // Additionally mute audio and stop local video so it's not in an unexpected state when demoted
     if (!isPromoted) {
       this.audioVideo.realtimeMuteLocalAudio();
+      this.audioVideo.stopLocalVideoTile();
```
hensmi-amazon, line 1592: unrelated demo fix :)

ltrung: Can we add a test case for this in the PR description :D

src/contentsharecontroller/DefaultContentShareController.ts
```diff
@@ -39,6 +40,7 @@ export default class DefaultContentShareController
     contentShareConfiguration.credentials.joinToken =
       configuration.credentials.joinToken + ContentShareConstants.Modality;
     contentShareConfiguration.meetingFeatures = configuration.meetingFeatures.clone();
+    contentShareConfiguration.videoDownlinkBandwidthPolicy = new NoVideoDownlinkBandwidthPolicy();
```
ltrung, line 43: My bad. We missed this since the aeon of content share :))

ltrung (approved):

demos/browser/app/meetingV2/meetingV2.ts
hensmi-amazon: too much pain to get integration tests to pass nowadays, I just let these changes sit in my local until i have a more critical PR :P

PR #2860, opened 2024-03-17
Downgrade video codec when encoding CPU usage is high or encoder fail

shi-su:
**Issue #:**

**Description of changes:**
Downgrade video codec when encoding CPU usage is high or encoder fail

**Testing:**
Tested codec degradation from software encoder CPU usage with a lowered threshold.

No

**Checklist:**

y

n

n


src/task/MonitorTask.ts
```diff
@@ -51,6 +52,7 @@ export default class MonitorTask
   private isResubscribeCheckPaused: boolean = false;
   private pendingMetricsReport: ClientMetricReport | undefined = undefined;
   private isMeetingConnected: boolean = false;
+  private videoEncodingMonitor: VideoEncodingMonitor;
```
hensmi-amazon, line 55: Maybe we can make this injectable via `ConnectionHealthPolicyConfiguration`

src/task/MonitorTask.ts
```diff
@@ -200,6 +203,8 @@ export default class MonitorTask
       StreamMetricReport
     >();
 
+    this.videoEncodingMonitor.encodingMonitor(clientMetricReport);
```
hensmi-amazon, line 206: nit: maybe this function can just be named `metricsDidReceive` as well. Maybe in the future we can make that its own interface.

src/videoencodingmonitor/VideoEncodingMonitor.ts
```diff
@@ -0,0 +1,173 @@
...
+  /**
+   * @param context Take context from monitor task
+   */
+  constructor(private context: AudioVideoControllerState) {}
+
+  /**
+   * Check video encoding metrics for high CPU usage with software encoder
+   * software encoder and degrade video codec when necessary.
+   */
+  private checkHighEncodeCpu(videoMetricReport: { [id: string]: { [id: string]: {} } }): void {
```
hensmi-amazon, line 27: nit: we can probably split this up into injectable components `VideoEncodingIssueDetector` and `VideoEncodingIssueObserver` or something like that just so we make it easy for builders to either adjust either of those logics seperately without needed to copy paste other code.

src/videoencodingmonitor/VideoEncodingMonitor.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    const attendeeId = this.context.meetingSessionConfiguration.credentials.attendeeId;
+
+    const metricData: { [id: string]: { [id: string]: {} } } = videoMetricReport[attendeeId];
+    const streams = metricData ? Object.keys(metricData) : [];
+    if (streams.length === 0) {
+      return;
+    }
+
+    for (const ssrc of streams) {
```
hensmi-amazon, line 91: nit: seems like maybe we could create helpers here.

src/videoencodingmonitor/VideoEncodingMonitor.ts
```diff
@@ -0,0 +1,173 @@
...
+            videoEncodeFps = Math.trunc(Number(value));
+            break;
+          case 'videoUpstreamEncoderImplementation':
+            isHardwareEncoder = value as boolean;
+            break;
+        }
+      }
+    }
+
+    const hwEncoderFailed = isHardwareEncoder && videoEncodeFps === 0 && videoInputFps > 0;
```
hensmi-amazon, line 107: I wonder if we need to check `isHardwareEncoder` here, in the case that we did not properly mark a hardware encoder as hardware. I don't see much risk here?

src/videoencodingmonitor/VideoEncodingMonitor.ts
```diff
@@ -0,0 +1,173 @@
...
+    return;
+  }
+
+  /**
+   * Degrade video codec to an alternative codec
+   */
+  private degradeVideoCodec(): void {
+    // Degrade video codec if there are other codec options and current codec is not H264 CBP or VP8
+    if (
+      this.context.meetingSupportedVideoSendCodecPreferences !== undefined &&
```
hensmi-amazon, line 130: Probably out of scope for this, but I feel like we should probably try to collect all the logic related to these arrays into a single unit testable class :)

src/videoencodingmonitor/VideoEncodingMonitor.ts
```diff
@@ -0,0 +1,173 @@
...
+        if (!capability.equals(this.context.meetingSupportedVideoSendCodecPreferences[0])) {
+          newMeetingSupportedVideoSendCodecPreferences.push(capability);
+        }
+      }
+      if (newMeetingSupportedVideoSendCodecPreferences.length > 0) {
+        this.context.logger.info(
+          `Downgrading codec to ${newMeetingSupportedVideoSendCodecPreferences[0].codecName} due to slow encoding`
+        );
+        this.context.meetingSupportedVideoSendCodecPreferences = newMeetingSupportedVideoSendCodecPreferences;
+        this.context.audioVideoController.setVideoCodecSendPreferences(
```
hensmi-amazon, line 150: I think we should just call `update` instead? this will overwrite what the builder puts in, which we probably want to always know (e.g. in the case of recovery, etc).

hensmi-amazon (commented):
good stuff! Maybe we can wire up event reporting here as well, on the decode fallback attempts?

src/videoencodingmonitor/VideoEncodingMonitor.ts
shi-su: Just calling `update` will get `meetingSupportedVideoSendCodecPreferences` refreshed at next index frame. That can happen frequently. I added another array `videoSendCodecsBlocklisted` for blocklisted codecs to avoid constantly getting back to the originally selected codec while not overwriting builder input

src/videoencodingmonitor/VideoEncodingMonitor.ts
shi-su: Agree hardware encoder check is not necessary here. Removed check for hardware.

src/videoencodingmonitor/VideoEncodingMonitor.ts
shi-su: Simplified metric update with `getObservableVideoMetricValue` function

src/videoencodingmonitor/VideoEncodingMonitor.ts
shi-su: Splited components into two injectable `ConnectionHealthPolicy`s

src/task/MonitorTask.ts
shi-su: Removed previous class and use `metricsDidReceive` in connection monitor to update metrics.

shi-su:
> good stuff! Maybe we can wire up event reporting here as well, on the decode fallback attempts? Can provide metadata like what codecs we are switching between.

Agree event reporting can be helpful here. Think I can add that in a different PR and have this focus on the metric monitoring and codec degradation.

src/task/MonitorTask.ts
```diff
@@ -438,4 +462,42 @@ export default class MonitorTask
...
+          `Downgrading codec from ${this.context.meetingSupportedVideoSendCodecPreferences[0].codecName} to ${newMeetingSupportedVideoSendCodecPreferences[0].codecName} due to ${cause}`
+        );
+        this.context.videoSendCodecsBlocklisted.push(
+          this.context.meetingSupportedVideoSendCodecPreferences[0]
+        );
+        this.context.meetingSupportedVideoSendCodecPreferences = newMeetingSupportedVideoSendCodecPreferences;
+        this.context.audioVideoController.update({ needsRenegotiation: true });
+      } else {
+        this.context.logger.warn(
+          'Degrading video codec failed since there is no alternative codec to select'
```
avinashmidathada, line 498: Nit: This may flood the client logs (every second) if the policy evaluates to minimum health consecutively due to some issue. do we need this at warn level?

shi-su: Actually, it's very unlikely this log will be triggered since the condition is fairly strict: 
1. Current `meetingSupportedVideoSendCodecPreferences` should at least have two codecs (normally, the updated `meetingSupportedVideoSendCodecPreferences` will just include current codecs except the first one);
2. Current codec is not H.264 CBP or VP8;
3. Excluding current codec, there should be no overlap between `meetingSupportedVideoSendCodecPreferences` and `videoSendCodecPreferences` (note that `meetingSupportedVideoSendCodecPreferences` is basically derived from `videoSendCodecPreferences`, this is even less likely to happen).

Given the the overlay of these conditions should be extremely rare (if even possible), I don't think this will really flood client log unless something goes very wrong.

avinashmidathada (approved):

src/statscollector/StatsCollector.ts
```diff
@@ -618,4 +621,48 @@ export default class StatsCollector implements RedundantAudioRecoveryMetricsObse
...
+
+  /**
+   * Receive video codec degradation event due to hardware encoder failure
+   * from MonitorTask and increment counter
+   */
+  videoCodecDegradationEncodeFailureDidReceive(): void {
+    this.videoCodecDegradationEncodeFailureCount += 1;
+  }
+
+  private getVideoUpstreamSsrc(): number | null {
```
ltrung, line 641: nit: Isn't this the same function as in `ClientMetricReport`? Can we just modify the one in `getVideoUpstreamSsrc` to make it reusable here?

src/task/MonitorTask.ts
```diff
@@ -438,4 +462,42 @@ export default class MonitorTask
...
+    ) {
+      const newMeetingSupportedVideoSendCodecPreferences: VideoCodecCapability[] = [];
+      for (const capability of this.context.videoSendCodecPreferences) {
+        if (!capability.equals(this.context.meetingSupportedVideoSendCodecPreferences[0])) {
+          newMeetingSupportedVideoSendCodecPreferences.push(capability);
+        }
+      }
+      if (newMeetingSupportedVideoSendCodecPreferences.length > 0) {
+        this.context.logger.info(
+          `Downgrading codec from ${this.context.meetingSupportedVideoSendCodecPreferences[0].codecName} to ${newMeetingSupportedVideoSendCodecPreferences[0].codecName} due to ${cause}`
```
ltrung, line 489: Seem like we just downgrade from [0] to the new codec [0]? Would that always be in [1] of the previous meetingSupportedVideoSendCodecPreferences?

src/task/ReceiveVideoStreamIndexTask.ts
```diff
@@ -223,6 +223,16 @@ export default class ReceiveVideoStreamIndexTask
 
     // Intersect `this.context.videoSendCodecPreferences` with `index.supportedReceiveCodecIntersection`
     for (const capability of this.context.videoSendCodecPreferences) {
+      let isCapabilityBlocklisted = false;
+      for (const blocklistedCapability of this.context.videoSendCodecsBlocklisted) {
```
ltrung, line 227: nit: Wonder if there is a better name than `videoSendCodecsBlocklisted` ? Blocklist seems a bit strong and sounds like a setting someone set.
Can we just say `videoSendCodecsCauseDegration`

src/clientmetricreport/ClientMetricReport.ts
```diff
@@ -530,6 +552,21 @@ export default class ClientMetricReport {
       media: MediaType.VIDEO,
       dir: Direction.UPSTREAM,
     },
+    videoUpstreamEncoderImplementation: {
```
ltrung, line 555: can you remind me if all these comes from getStats API?

src/task/ReceiveVideoStreamIndexTask.ts
hensmi-amazon: agree, maybe `degradedVideoSendCodecs`/`brokenVideoSendCodecs`/etc?

src/statscollector/StatsCollector.ts
shi-su: Removed the function in `StatsCollector`

src/task/MonitorTask.ts
shi-su: Yes, it should be the case. Still deriving the list there to avoid things unexpected.

src/task/ReceiveVideoStreamIndexTask.ts
shi-su: Updated the name to `degradedVideoSendCodecs`

src/clientmetricreport/ClientMetricReport.ts
shi-su: Yes, the metrics are all from `getStats`

src/audiovideocontroller/AudioVideoControllerState.ts
```diff
@@ -148,6 +148,9 @@ export default class AudioVideoControllerState {
   // that is provided to the peer connection, which will be ordered by priority.
   prioritizedSendVideoCodecCapabilities: VideoCodecCapability[] = [];
 
+  // Video codecs caused degradation due to encoding issues
```
hensmi-amazon, line 151: Maybe just could use a bit more context
```
// Video codecs that we have detected encoding issues with that we will
// avoid using even if they are provided in `videoSendCodecPreferences`
```

src/connectionhealthpolicy/ConnectionHealthData.ts
```diff
@@ -13,6 +13,11 @@ export default class ConnectionHealthData {
   packetsReceivedInLastMinute: number[] = [];
   fractionPacketsLostInboundInLastMinute: number[] = [];
   audioSpeakerDelayMs = 0;
+  isVideoEncoderHardware = false;
+  videoEncodingTimeInMs = 0;
+  cpuLimitationDuration = 0;
+  videoInputFps = 0;
+  videoEncodeFps = 0;
```
hensmi-amazon, line 20: if we ever do a major version...we should really just pass the whole metrics report...

src/task/ReceiveVideoStreamIndexTask.ts
```diff
@@ -223,6 +223,16 @@ export default class ReceiveVideoStreamIndexTask
 
     // Intersect `this.context.videoSendCodecPreferences` with `index.supportedReceiveCodecIntersection`
```
hensmi-amazon, line 224:
```
// and filter out any send video codecs we have determined to be degraded
```

src/task/ReceiveVideoStreamIndexTask.ts
```diff
@@ -223,6 +223,16 @@ export default class ReceiveVideoStreamIndexTask
...
     for (const capability of this.context.videoSendCodecPreferences) {
+      let isCapabilityDegraded = false;
+      for (const degradedCapability of this.context.degradedVideoSendCodecs) {
```
hensmi-amazon, line 227:
```
if (this.context.degradedVideoSendCodecs.some(
    degradedCapability => capability.equals(degradedCapability))) {
  continue;
}
```

src/task/ReceiveVideoStreamIndexTask.ts
```diff
@@ -223,6 +223,16 @@ export default class ReceiveVideoStreamIndexTask
...
+          isCapabilityDegraded = true;
+          break;
+        }
+      }
+      if (isCapabilityDegraded) {
+        continue;
+      }
       for (const signaledCapability of index.supportedReceiveCodecIntersection) {
         if (capability.equals(VideoCodecCapability.fromSignaled(signaledCapability))) {
           newMeetingSupportedVideoSendCodecPreferences.push(capability);
```
hensmi-amazon, line 238: There are comments and log lines below that need to be updated to call out degraded video codecs.

src/task/MonitorTask.ts
```diff
@@ -438,4 +462,42 @@ export default class MonitorTask
...
+        );
+        this.context.degradedVideoSendCodecs.push(
+          this.context.meetingSupportedVideoSendCodecPreferences[0]
+        );
+        this.context.meetingSupportedVideoSendCodecPreferences = newMeetingSupportedVideoSendCodecPreferences;
+        this.context.audioVideoController.update({ needsRenegotiation: true });
+      } else {
+        this.context.logger.warn(
+          'Degrading video codec failed since there is no alternative codec to select'
```
hensmi-amazon, line 498: Could we dump all the currently degraded video codecs here?

src/task/MonitorTask.ts
```diff
@@ -249,6 +259,20 @@ export default class MonitorTask
       }
     );
 
+    for (const policy of this.videoEncodingHealthPolicies) {
+      this.applyHealthPolicy(policy, connectionHealthData, () => {
+        this.degradeVideoCodec(policy.name);
+        switch (policy.name) {
+          case 'Video Encoding CPU Health':
```
hensmi-amazon, line 266: This feels a bit off. Could we try casting? Or maybe we can add a unit test on the name with a comment mentioning that there is a dependency?

src/connectionhealthpolicy/ConnectionHealthPolicyConfiguration.ts
```diff
@@ -39,4 +42,15 @@ export default class ConnectionHealthPolicyConfiguration {
...
   sendingAudioFailureInitialWaitTimeMs = 3000;
+
+  /**
+   * Policies and parameters related to video encoding health montoring
+   */
+  videoEncodingHealthPolicies = [
+    VideoEncodingCpuConnectionHealthPolicy,
+    VideoEncodingFramerateConnectionHealthPolicy,
+  ];
+  consecutiveHighEncodeCpuThreshold = 10;
```
hensmi-amazon, line 53: could these have a short description + effects of increasing or decreasing?

src/audiovideocontroller/AudioVideoControllerState.ts
shi-su: Updated comment in the latest revision.

src/connectionhealthpolicy/ConnectionHealthData.ts
shi-su: Agree. We may need to do a cleanup here later

src/task/ReceiveVideoStreamIndexTask.ts
shi-su: Updated comment in the latest revision.

src/task/ReceiveVideoStreamIndexTask.ts
shi-su: Updated as suggested

src/task/ReceiveVideoStreamIndexTask.ts
shi-su: Added a log line to log video codec being degraded.

src/task/MonitorTask.ts
shi-su: Changed log line to dump degraded codec here

src/task/MonitorTask.ts
shi-su: Yeah. I added a enum for video encoding health policy names and added comments there to note the names are used in monitor task to determine degradation reason and emit metrics

src/connectionhealthpolicy/ConnectionHealthPolicyConfiguration.ts
shi-su: Added comments to describe effects for these parameters.

hensmi-amazon (approved):
🚀

CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add automatic codec degradation logic when CPU usage of software encoder is high or video encoding of encoder fails.
+
```
brycepj, line 34: @shi-su @hensmi-amazon If I'm not mistaken, this change is actually going out with 3.22.0, correct?

hensmi-amazon: Yes, good catch! this happens only like, 1/10 PRs 😅

PR #2868, opened 2024-03-28
Fixed packets received check on Safari 17.3 and below

hensmi-amazon:
**Issue #:** None

**Description of changes:** Safari only has bytes received on 17.3 and below, so switching to that should be sufficient. Ping-pong check doesn't work in these cases to trigger reconnect itself even though websocket is snapped, but will fix that another day. I made a new minor version since I am just going to hotfix this out.

This original change was released in 3.20, and lead to re-connections not occurring on network changes.

**Testing:**
Tested on Safari 17.3 ios, 17.4 macos, firefox, chrome with forced toggle.

1. Join with Safari 17.3 browser
2. Toggle WiFi off and on. You may need to delay how long you keep it off.
3. Should reconnect after around 15 seconds

**Checklist:**

y

n

n

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+## [3.21.1] - 2024-03-28
+
+### Added
+
+### Removed
+
+### Changed
+
+### Fixed
+- Fixed packets received check on Safari 17.3 and below
+
 ## [3.21.0] - 2024-02-12
 
 ### Added
--- a/src/connectionmonitor/SignalingAndMetricsConnectionMonitor.ts
+++ b/src/connectionmonitor/SignalingAndMetricsConnectionMonitor.ts
@@ -15,8 +15,8 @@ import ConnectionMonitor from './ConnectionMonitor';
 export default class SignalingAndMetricsConnectionMonitor
   implements ConnectionMonitor, PingPongObserver, AudioVideoObserver {
   private isActive = false;
-  private hasSeenValidPacketMetricsBefore = false;
-  private lastTotalPacketsReceived = 0;
+  private hasSeenValidCandidatePairMetricsBefore = false;
+  private lastTotalBytesReceived = 0;
 
   constructor(
     private audioVideoController: AudioVideoController,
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+## [3.21.1] - 2024-03-28
```
ltrung, line 18: You plan to hotfix this?

src/connectionmonitor/SignalingAndMetricsConnectionMonitor.ts
```diff
@@ -87,28 +87,28 @@ export default class SignalingAndMetricsConnectionMonitor
...
     // The stats collector currently doesn't account for candidate pair stats
     // so we just use the raw report for now.
     const webrtcReport = clientMetricReport.getRTCStatsReport();
 
-    let totalPacketsReceived = 0;
+    let totalBytesReceived = 0;
     webrtcReport.forEach(stat => {
-      if (stat.type === 'candidate-pair' && stat.packetsReceived) {
-        totalPacketsReceived += stat.packetsReceived;
+      if (stat.type === 'candidate-pair' && stat.bytesReceived) {
```
ltrung, line 99: In the PR description, you mention Safari 17.3 or below so it never worked before since we introduced this code for safari right?
Can we mention in the PR description what the issue is (reconnect) and what NPM version we start to have this issue (3.20.0?)?

ltrung (commented):
> Ping-pong check is also broken in these cases but will fix that another day

Isn't ping-pong broken would cause reconnect?

ltrung:
Do we need to update here too ? https://github.com/aws/amazon-chime-sdk-js/blob/60eeb35745ec6e0869017edbfab1b032279ecdd0/src/meetingreadinesschecker/DefaultMeetingReadinessChecker.ts#L274
Not sure how this work before if Safari does not have packetReceived?

src/connectionmonitor/SignalingAndMetricsConnectionMonitor.ts
hensmi-amazon: Yes this was broken on arrival, will update description!

hensmi-amazon:
> > Ping-pong check is also broken in these cases but will fix that another day
> 
> Isn't ping-pong broken would cause reconnect?

Different from Chrome (at least in my testing), in safari (though i don't have this issue with 17.4) the websocket seems to shut down in a different way, that leads to the ping/pong logic to be disabled. I'm not sure why we would disable that any time besides the end of call since the signaling connection should always be established, but the code is pretty old.

> Do we need to update here too ? 
This is looking at `inbound-rtp` stats reports which do have the `packetsReceived` metric. The `candidate-pair` (which includes RTCP so can be used when attendee caps is None) did not have `packetsReceived`

ltrung (approved):

PR #2889, opened 2024-05-03
Make server side network adaptation the default when creating `VideoPriorityBasedPolicy`

hensmi-amazon:
**Issue #:** None

**Description of changes:** We are no longer iterating on the priority policy without server side network adaptation. So it makes sense that it is the default.

**Testing:**
Confirmed that SSNA is used when not configured.

N/A no behavior is changed.

**Checklist:**

y

n

n

### DIFF
--- a/demos/browser/app/meetingV2/video/PaginationManager.ts
+++ b/demos/browser/app/meetingV2/video/PaginationManager.ts
@@ -20,7 +20,9 @@ export default class PaginationManager<Type> {
   }
 
   remove(toRemove: Type) {
-    this.all.splice(this.all.indexOf(toRemove));
+    if (this.all.includes(toRemove)) {
+        this.all.splice(this.all.indexOf(toRemove));
+    }
   }
 
   removeIf(toRemoveFn: (value: Type) => boolean) {
--- a/demos/browser/app/meetingV2/video/PaginationManager.ts
+++ b/demos/browser/app/meetingV2/video/PaginationManager.ts
@@ -5,7 +5,7 @@ export default class PaginationManager<Type> {
   private currentPageStart: number = 0;
 
   private all = new Array<Type>();
-  
+
   constructor(private pageSize: number) {}
 
   currentPage(): Array<Type> {
@@ -21,7 +21,7 @@ export default class PaginationManager<Type> {
 
   remove(toRemove: Type) {
     if (this.all.includes(toRemove)) {
-        this.all.splice(this.all.indexOf(toRemove));
+      this.all.splice(this.all.indexOf(toRemove), 1);
     }
   }
 
@@ -30,7 +30,7 @@ export default class PaginationManager<Type> {
     if (index === -1) {
       return;
     }
-    this.all.splice(index,1);
+    this.all.splice(index, 1);
   }
 
   hasNextPage(): boolean {
@@ -38,7 +38,7 @@ export default class PaginationManager<Type> {
   }
 
   nextPage(): void {
-    if (!this.hasNextPage) {
+    if (!this.hasNextPage()) {
       return;
     }
     this.currentPageStart += this.pageSize;
@@ -49,6 +49,8 @@ export default class PaginationManager<Type> {
   }
 
   previousPage(): void {
-    this.currentPageStart -= this.pageSize;
+    if (this.hasPreviousPage()) {
+      this.currentPageStart -= this.pageSize;
+    }
   }
 }
--- a/demos/browser/app/meetingV2/video/VideoTileCollection.ts
+++ b/demos/browser/app/meetingV2/video/VideoTileCollection.ts
@@ -130,7 +130,8 @@ export default class VideoTileCollection implements AudioVideoObserver {
   constructor(private videoTileController: VideoTileControllerFacade,
     private logger: Logger,
     private remoteVideoManager: RemoteVideoManager,
-    private pageSize: number) {
+    private pageSize: number,
+    private localAttendeeId: string) {
     this.setupVideoTiles();
 
     if (!this.remoteVideoManager.supportsRemoteVideoPreferences()) {
@@ -162,7 +163,7 @@ export default class VideoTileCollection implements AudioVideoObserver {
   }
 
   videoTileDidUpdate(tileState: VideoTileState): void {
-    console.log(`video tile updated: ${JSON.stringify(tileState, null, '  ')}`);
+    this.logger.info(`video tile updated: ${JSON.stringify(tileState, null, '  ')}`);
     if (!tileState.boundAttendeeId) {
       return;
     }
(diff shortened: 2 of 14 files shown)


demos/browser/app/meetingV2/video/PaginationManager.ts
```diff
@@ -20,7 +20,9 @@ export default class PaginationManager<Type> {
   }
 
   remove(toRemove: Type) {
-    this.all.splice(this.all.indexOf(toRemove));
+    if (this.all.includes(toRemove)) {
```
hensmi-amazon, line 23: Some unrelated pagination fixes here

src/videodownlinkbandwidthpolicy/VideoAdaptiveProbePolicy.ts
```diff
@@ -3,15 +3,30 @@
...
 import VideoStreamIndex from '../videostreamindex/VideoStreamIndex';
 import VideoPreference from './VideoPreference';
 import { VideoPreferences } from './VideoPreferences';
 import VideoPriorityBasedPolicy from './VideoPriorityBasedPolicy';
+import VideoPriorityBasedPolicyConfig from './VideoPriorityBasedPolicyConfig';
 
+/** [[VideoAdaptiveProbePolicy]] wraps [[VideoPriorityBasedPolicy]] with customized behavior to automatically
+ * assign a high preference to content share.
+ *
+ * @deprecated This class is not compatible with latest priority policy features (i.e. server side network adaptation),
```
devalevenkatesh, line 17: 1. Since we are trying to help the builder with the documenation, probably a link to follow when we say "can be done trivially by the application" would be helpful.
2. This seems like a good detail for changelog entry as a build can understand the change better.

src/videodownlinkbandwidthpolicy/VideoAdaptiveProbePolicy.ts
```diff
@@ -3,15 +3,30 @@
...
+ * [[VideoPriorityBasedPolicy]] should be used instead. If high priority for content is desired it can be done trivially
+ * by the application.
+ */
 export default class VideoAdaptiveProbePolicy extends VideoPriorityBasedPolicy {
+  private static createConfig(): VideoPriorityBasedPolicyConfig {
```
devalevenkatesh, line 22: Hmm why static? Do we need to make this singleton for multiple `VideoAdaptiveProbePolicy` objects? If so checking if this approach is useful or do we need explicit object checking to preserve a singleton use of the config.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicyConfig.ts
```diff
@@ -26,10 +26,11 @@ export default class VideoPriorityBasedPolicyConfig {
   private referenceBitrate: number = 0;
 
   /**
-   * Additional server side features to be enabled for network adaption. This
-   * may be overridden by the server. The default may be changed in future releases.
+   * Additional server side features to be enabled for network adaption. The only currently
+   * recommended value is `ServerSideNetworkAdaption.BandwidthProbingAndRemoteVideoQualityAdaption`
    */
-  serverSideNetworkAdaption = ServerSideNetworkAdaption.Default;
+  serverSideNetworkAdaption =
```
devalevenkatesh, line 32: Just checking if this is a backward compatible change from builder APIs + behavior perspective and whether we need something to revert this back at runtime if builders face issues?

src/videodownlinkbandwidthpolicy/VideoAdaptiveProbePolicy.ts
hensmi-amazon: I'll add a comment. This is just a workaround for the silly typescript rule that `super` calls must be first in constructor. We don't need singleton usage (even though i'm not sure what customer would be creating multiple policies lol).

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicyConfig.ts
hensmi-amazon: TBH I don't think many customers use priority policy without SSNA. Unfortunately since it is client side we don't actually know the number, but given that we don't get any complaints about it (and there is plenty to complain about) I think the number isn't high (and they probably aren't updating the SDK quickly anyways). Customers can of course easily switch to None if they are happy with what they got, but TBH I want to eventually remove the ability to set None (since it isn't technically a change in behavior, I don't see much issue with this).

That being said, I do technically have that knob on the backend, but it's not really a button i want to press given the existing policy has a lot of issues.

src/videodownlinkbandwidthpolicy/VideoAdaptiveProbePolicy.ts
hensmi-amazon: Will update!

src/videodownlinkbandwidthpolicy/VideoAdaptiveProbePolicy.ts
hensmi-amazon: Actually i misunderstood comments i wrote before. We don't need to deprecate this class, my bad.

src/videodownlinkbandwidthpolicy/VideoAdaptiveProbePolicy.ts
hensmi-amazon: actually i misunderstood a comment i wrote before/how this is used (when simulcast is enabled but a policy is not set). I don't need to deprecate this yet.

src/videodownlinkbandwidthpolicy/VideoPriorityBasedPolicyConfig.ts
devalevenkatesh: Cool I am fine with this approach, probably we can leverage client events to help us here in future.

devalevenkatesh (dismissed):

demos/browser/app/meetingV2/video/VideoTileCollection.ts
```diff
@@ -452,10 +454,10 @@ export default class VideoTileCollection implements AudioVideoObserver {
     // We need to manually control visibility of paused tiles anyways so we just do
     // everything here, even though the preference manager adding/removing will
     // result in tile callbacks as well.
-    for (let [index, videoTile] of this.tileIndexToDemoVideoTile.entries()) {
+    for (let videoTile of this.tileIndexToDemoVideoTile.values()) {
         if (attendeesToShow.includes(videoTile.attendeeId)) {
             videoTile.show(false);
-        } else if (this.tileIndexToTileId[index] !== this.findContentTileId()) { // Always show content
```
hensmi-amazon: This breaks because of the local content that automatically makes a tile. It doesn't really need to make UX decisions like this (not like UX eng ever assisted with demo) so simpler is better IMO

PR #2893, opened 2024-05-10
Fixed reconnections not triggering `audioVideoWasDemotedFromPrimaryMeeting`

hensmi-amazon:
**Issue #:** N/A

**Description of changes:** Reconnections were not triggering `audioVideoWasDemotedFromPrimaryMeeting`, only full disconnects. This change fixes that. I also made a new status code for this because the previous didn't always make sense since reconnects can occur due to client side issues.

**Testing:**
Reconnections now trigger the callback. 
1. Join a replica meeting
2. Promote
3. Switch wifi and wait for reconnect
4. You should see a toast describing that you were disconnected

**Checklist:**

y

n

n


src/audiovideoobserver/AudioVideoObserver.ts
```diff
@@ -97,9 +97,8 @@ export default interface AudioVideoObserver {
    * * `MeetingSessionStatusCode.OK`: `demoteFromPrimaryMeeting` was used to remove the attendee.
    * * `MeetingSessionStatusCode.AudioAuthenticationRejected`: `chime::DeleteAttendee` was called on the Primary
    *   meeting attendee used in `promoteToPrimaryMeeting`.
-   * * `MeetingSessionStatusCode.SignalingBadRequest`: Other failure, possibly due to disconnect
-   *   or timeout. These failures are likely retryable. Any disconnection will trigger an automatic
-   *   demotion to avoid unexpected or unwanted promotion state on reconnection.
+   * * `MeetingSessionStatusCode.AudioVideoDisconnectedWhilePromoted`: The client disconnected (and may be reconnecting).
```
devalevenkatesh, line 100: I am not sure if we maintain a posture that a reconnect is idempotent and currently when a reconnect happens a builder needs to not do anything. Here, it seems we are deviating from that path no hard requirement. Is there a way to fix the promotion/demotion and not disconnect on the reconnect and upon reconnect the attendee gets promoted back up, or this is complex and not needed from UX?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -1353,6 +1342,19 @@ export default class DefaultAudioVideoController
   }
 
   reconnect(status: MeetingSessionStatus, error: Error | null): boolean {
+    if (this.promotedToPrimaryMeeting) {
```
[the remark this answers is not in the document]
hensmi-amazon: 1. Yes this is documented in multiple places.
2. That's what I tried when I originally pushed code with this bug :). This is the one place that doesn't require adding logic to two paths. I looked around and it made the most sense to me.

src/audiovideoobserver/AudioVideoObserver.ts
hensmi-amazon: As mentioned above, this is behavior that was intentionally decided on when this feature released, which is why its documented (this comment is just refactored). On the backend, we intentionally do not keep state for connections that do not exist, which is why this would be very complicated to change.  Additionally, this is a safety mechanisms for builders and the end users; you don't have attendees randomly popping back onto the stage when they disconnect, if it takes them a while to reconnect and you move on.

hensmi-amazon: The latter is tbh the more important point. Think of it as the backend intentionally demoting the attendee if they disconnect.

src/audiovideocontroller/DefaultAudioVideoController.ts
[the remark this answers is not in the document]
devalevenkatesh: Ok ok, cool, so just a double check does `reconnect` gets called multiple times when retrying? And if yes, have we checked that this observer does not get triggered multiple times?

src/audiovideoobserver/AudioVideoObserver.ts
devalevenkatesh: Got it, thanks, I missed reading the above earlier documentation.

src/audiovideocontroller/DefaultAudioVideoController.ts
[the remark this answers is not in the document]
hensmi-amazon: I haven't not tried but since it sets
```
this.promotedToPrimaryMeeting = false;
```
it won't be able to hit this path again.

devalevenkatesh (approved):

PR #2906, opened 2024-06-11
Fixed rare race conditions with simulcast + server side network adaptation on third attendee join.

hensmi-amazon:
**Issue #:** N/A

**Description of changes:** Just was testing something else and ran across this. The remote video update logic can happen during subscribe (which is mostly fine) and caused issues to the mappings used in the track mapping.

Additionally, the currently location of `subscribeFrameSent` can cause discrepancies to the track mapping to transceiver. This doesn't cause issues to all highest policy because it will resubscribe anyways, but with SSNA we do not resubscribe in this case.

**Testing:**
See steps below
1. Enable simulcast and SSNA and join 2 attendees and enable video
2. Join a third attendee with SSNA. Depending on timing this can be anywhere from a 0-50% repro (my RTT is pretty low).

**Checklist:**

y

n

n


src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -1039,7 +1039,7 @@ export default class DefaultAudioVideoController
     // Similar to `updateRemoteVideosFromLastVideosToReceive`, we use `subscribeFrameSent` to cache the previous
```
ltrung, line 1039: nit: forget to update comment? Same as above

src/videostreamindex/VideoStreamIndex.ts
```diff
@@ -20,15 +20,21 @@ export default interface VideoStreamIndex {
...
+   * This function caches the currently ingested index frame for subscribe related activities (e.g. mapping tracks).
+   *
+   * The index will also use the stored index as a backup for map functions like `groupIdForStreamId`so that we can maintain information
+   * about previous senders simulcast streams to enable stream switching, e.g. if a sender is transmitting low/hi and
+   * switches to just low, receivers need the previous index to be able to know what group ID the high stream corresponds to.
    */
   subscribeFrameSent(): void;
 
+  /**
+   * This function caches the currently ingested index frame specifically as a backup for map functions like `groupIdForStreamId`so that we can maintain information
```
ltrung, line 32: Maybe I have not looked at video code for a while but both these methods sound very similar :D

src/audiovideocontroller/DefaultAudioVideoController.ts
hensmi-amazon: Good nit :) will update these

src/videostreamindex/VideoStreamIndex.ts
hensmi-amazon: The issue is that two different sections of code want it to be cached at different times:
* On subscribe ack, we need the index when we subscribed, because it is the only one guaranteed to have the stream IDs used in the track mapping.
* On simulcast stream updates, we need the index of the last update, because that is the only one guranteed to have the stream IDs of the 'existing' subscription'
In a rewrite these probably should just be cached by what needs them (which is unfortunately the big context glob). But I just followed the existing pattern.

ltrung (approved):

PR #2911, opened 2024-06-20
Change getDisplayMedia audio params to exclude noiseSuppression, echo…

michhyun1:
**Issue #:**
Customer issue with sharing browser tab audio and loud video.

**Description of changes:**

Based on the article: https://medium.com/@trystonperry/why-is-getdisplaymedias-audio-quality-so-bad-b49ba9cfaa83

We don't need these parameters for content sharing. Removing them as they cause unnecessary audio processing.

**Testing:**

Tested on a serverless demo.

**Checklist:**

yes

no

### DIFF
--- a/src/contentsharecontroller/ContentShareMediaStreamBroker.ts
+++ b/src/contentsharecontroller/ContentShareMediaStreamBroker.ts
@@ -66,7 +66,11 @@ export default class ContentShareMediaStreamBroker implements MediaStreamBroker 
     return {
       audio:
         !sourceId && new DefaultBrowserBehavior().getDisplayMediaAudioCaptureSupport()
-          ? true
+          ? {
+              echoCancellation: false,
+              noiseSuppression: false,
+              autoGainControl: false,
+            }
           : false,
       video: {
         ...(!sourceId && {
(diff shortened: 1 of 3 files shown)


georgezy-amzn (dismissed):

src/contentsharecontroller/ContentShareMediaStreamBroker.ts
```diff
@@ -66,7 +66,11 @@ export default class ContentShareMediaStreamBroker implements MediaStreamBroker
     return {
       audio:
         !sourceId && new DefaultBrowserBehavior().getDisplayMediaAudioCaptureSupport()
-          ? true
+          ? {
+              echoCancellation: false,
+              noiseSuppression: false,
```
ltrung, line 71: Just want to point out that it seems we use `goog` prefix for the audio stream https://github.com/aws/amazon-chime-sdk-js/blob/main/src/devicecontroller/DefaultDeviceController.ts#L1303. 
I think Google has migrated to use the standard one so this is probably fine but wonder if we need to update the other code later too.

ltrung (dismissed):

xuesichao (dismissed):

georgezy-amzn (approved):

PR #2917, opened 2024-07-02
Do not no-op server side network adaptation pauses that happen immediately after subscribe

hensmi-amazon:
**Issue #:** None

**Description of changes:** The new tile may not be added when subscribe ack is received since it relies on Peer Connection track callbacks. Therefore we extend `ReceiveRemoteVideoPauseResumeTask` to also pre-emptively create tiles. Note some of this code could have also gone in `DefaultAudioVideoController` but I figured I could just put it in the task and rename it to `ManageServerSideNetworkAdaptationTask` in a future version. 

**Testing:**
See steps below

1. Join one client with SSNA, enable video, and use Network Link Conditioner to tank the downlink bitrate to 20kbps or lower
2. Join another client and enable video
3. Check pause on first client. Toggling video on second client should always result in pause in first (assumeing downlink bitrate is low enough).

**Checklist:**

Y

N

N


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed rare race conditions with simulcast + server side network adaptation on third attendee join.
 - Make redundant audio worker code generation script work on Windows
+- Do not no-op server side network adaptation pauses that happen immediately after subscribe
```
devalevenkatesh, line 20: If not a no-op would be good to explain the expectation now. Might help a builder's understanding from changelog if checked.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -1062,6 +1065,7 @@ export default class DefaultAudioVideoController
     const groupIdsToReceive = context.videosToReceive.array().map((streamId: number) => {
       return context.videoStreamIndex.groupIdForStreamId(streamId);
     });
+    this.receiveRemotePauseResumeTask.updateSubscribedGroupdIds(new Set(groupIdsToReceive));
```
devalevenkatesh, line 1068: 1. Are there any existing observers that `receiveRemotePauseResumeTask` itself can implement and on update, call the `updateSubscribedGroupdIds`? Will help avoid direct dependency on DAVController.
2. When cleaning up probably as part of retry as well as under the [destroy](https://github.com/aws/amazon-chime-sdk-js/blob/bf22ce0e55b69348391144c6f739e12c10933970/src/audiovideocontroller/DefaultAudioVideoController.ts#L183) method in here, should we set the `receiveRemotePauseResumeTask` to `undefined`?

src/task/ReceiveRemoteVideoPauseResumeTask.ts
```diff
@@ -36,6 +37,32 @@ export default class ReceiveRemoteVideoPauseResume
     this.context.removableObservers.push(this);
   }
 
+  updateSubscribedGroupdIds(groupIds: Set<number>): void {
+    const existingVideoTileIds = new Set<number>();
+    for (const groupId of groupIds) {
+      const attendeeId = this.context.videoStreamIndex.attendeeIdForGroupId(groupId);
+      if (attendeeId.length === 0) {
```
devalevenkatesh, line 44: Can `attendeeId` be `null` here if not found, sorry I dint check `attendeeIdForGroupId` implementation, so should a general falsy check be more appropriate here?

src/task/ReceiveRemoteVideoPauseResumeTask.ts
```diff
@@ -36,6 +37,32 @@ export default class ReceiveRemoteVideoPauseResume
...
+      }
+      let videoTile = this.context.videoTileController.getVideoTileForAttendeeId(attendeeId);
+      if (videoTile === undefined) {
+        this.logger.info(
+          `No existing video tile for attendee ID ${attendeeId} with new group ID ${groupId}. Creating new one.`
+        );
+        videoTile = this.context.videoTileController.addVideoTile();
+        videoTile.bindVideoStream(attendeeId, false, null, 0, 0, 0, null);
+      }
+      existingVideoTileIds.add(videoTile.id());
```
devalevenkatesh, line 56: A verifying check: This should result in `videoTileDidUpdate` and the builder app can accordingly handle this right?

test/task/ReceiveRemoteVideoPauseResumeTask.test.ts
```diff
@@ -286,5 +286,168 @@ describe('ReceiveRemoteVideoPauseResumeTask', () => {
...
+      expect(context.videoTileController.getVideoTileForAttendeeId('attendee-8')).to.not.be
+        .undefined;
+      expect(context.videoTileController.getVideoTileForAttendeeId('attendee-8').state().paused).to
+        .be.true;
+
+      // Additional test to allow reseting paused groups
+      task.updateSubscribedGroupdIds(new Set([]));
+
+      // Manually pause
+      context.videoTileController.getVideoTileForAttendeeId('attendee-7').stateRef().paused = true;
```
devalevenkatesh, line 442: Should we use actual API used to pause if paused by the builder application? It may happen so that approach is not correctly pausing the video tile and help us catch it.

CHANGELOG.md
hensmi-amazon: Will reword.

src/audiovideocontroller/DefaultAudioVideoController.ts
hensmi-amazon: 1. This set needs to be synchronized with the state in DAVC so setting it elsewhere just leads to risk of race conditions, especially if we refactor to reuse transceivers. The existence of injectable downlink policies is why we originally had to shove this logic in DAVC but it would probably be refactored out when/if we remove that pattern.
2. It would survive since it would be reset on the first subscribe, but it doesn't hurt to be more explicit, good point. This honestly makes me wonder if we should have a checklist/gotcha list in PR template around surviving reconnects lol.

src/task/ReceiveRemoteVideoPauseResumeTask.ts
hensmi-amazon: It does not return `null` (even though other functions in that class do). Dreaming of strict mode :D. Just going to leave this as-is since its technically correct as long as no-one makes the return type of `attendeeIdForGroupId` inaccurate.

src/task/ReceiveRemoteVideoPauseResumeTask.ts
hensmi-amazon: Yes, but there will not be any video stream bound to it until the negotiation completes.

test/task/ReceiveRemoteVideoPauseResumeTask.test.ts
hensmi-amazon: sure, i forgot that API existed.

src/audiovideocontroller/DefaultAudioVideoController.ts
hensmi-amazon: I did not bother setting to `undefined`, but I reset it on reconnect. TBH i think the logic we have is out of sync between reconnect and initial connect...not clear why that path wasn't shared.

devalevenkatesh: I see, probably a v4 tracking point to sync it 😅 .

devalevenkatesh (approved):

Issue #2968, opened 2024-10-07
No audio packets sent on iPadOS after unplugging the earphones and switching to the internal microphone if the Web Audio is enabled


pzubar:
### What happened and what did you expect to happen?

I am experiencing an issue  on iPadOS when creating the `DefaultDeviceController` with the `enableWebAudio: true` after the earphones are unplugged. The specific scenario occurs when the following conditions are met.

1. The iPad is connected to earphones that have a built-in microphone (e.g., AirPods, but the issue was reproduced with another wireless earphones).
2. The earphones are subsequently unplugged.
3. The `startAudioInput` method is called inside `audioInputsChanged` callback with a default input deviceId:
```js
audioInputsChanged(freshAudioInputDeviceList) {
    let device = freshAudioInputDeviceList[0]?.deviceId; // i.e. XYZ
    await this.audioVideo.startAudioInput(device);
    // > ... API/DefaultDeviceController/startAudioInputDevice "XYZ" -> "success"
}
```

After that the iPad attendee is not sending audio packets and cannot be heard by other participants. There are no error messages or exceptions in the console (moreover, the DefaultDeviceController's `startAudioInputDevice` is logged with `"success"` postfix). 
As a result, the attendee appears "muted" and `signalStrength` drops to 0.

However, when `enableWebAudio` is set to `false`, the problem does not occur, and `startAudioInput` functions correctly. This suggests the issue likely lies in how the `useWebAudio` [case](https://github.com/aws/amazon-chime-sdk-js/blob/1a9f8f5e123e0a3289472d60a6bb4a3cf8483491/src/devicecontroller/DefaultDeviceController.ts#L247) is managed. _Additionally_, invoking `startAudioInput` with a `voiceFocusDevice` created from the same "default" microphone afterward seems to resolve the issue, allowing audio packets to be sent again.

I would appreciate for any guidance on troubleshooting or a potential fix for this behavior. Please let me know if you need any further details regarding this issue.

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

[...]

Issue #1721, opened 2021-10-26
Plug off headphones while in call interrupts the input stream with Smartphones ( Android - iOS ).


felipenandz:
### What happened and what did you expect to happen?

Trying to plug off the headphones while in the call but the switch doesn't work automatically with mobile phones and the input audio is interrupted. I choose the new device after **audioInputsChanged** is triggered by calling chooseAudioInputDevice with the device ID. 

I expect that plugging off the headphones and choosing the new Audio stream works fluid. 

- The switch works automatically with **desktop browsers** and if I use **Airpods.** ( plug in and plug of many times ) and audio continues working for remote. 

- **Weird Thing:** It works if I start the call without headphones and then plug it in (Input stream continues working ),  but then when I plug it off it stops working.

- **Call keep dropping**: If I start the call with headphones and plug it off, the input audio stream stops, and chime keeps trying to reconnect.

- **_Fun fact_:** If I switch between apps in my phone and come back to safari the input audio starts to work again. 

**Worth attention:** **In the app demo of chime sdk the behaviour is the same (plugging off interrupts the input stream), so I don't think it's a problem in my code**

Devices Tested: 

Android
Iphone SE - v.14.2 
Iphone 12 - v.15.2

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

(Mobile Phone - Not working) Enter the call with headphones and then plug it off -> Input stream interrupts
(Desktop - Working) Enter the call with headphones and then plug it off -> Input stream continues working

### Amazon Chime SDK for JavaScript version

^2.18.0

### What browsers are you seeing the problem on?

Chrome, Firefox, Safari

### Browser version

Chome - 95.0.4638.50

### Browser console logs

choosing the new device works well while looking in the console log, all works successfully.

xuesichao:
@felipenandz Hi, thanks for providing the detailed reproduce steps. In our previous test we didn't enable the WebAudio, and it worked well. This time we followed the detailed steps you provided, and confirm the issue exists.

[...]

PR #2971, opened 2024-10-25
Avoid breaking audio input when external devices are disconnected on iOS browsers when using Web Audio

hensmi-amazon:
**Issue #:** https://github.com/aws/amazon-chime-sdk-js/issues/2968 https://github.com/aws/amazon-chime-sdk-js/issues/1721

**Description of changes:** I'm not sure if that WebKit bug will ever get traction, but it appears that suspending and resuming the audio context allows it to recover from the failure. This path will only be hit when the active device is disconnected.

**Testing:**
Reproduced, no longer reproduces after change.
1. Join on iOS with WebAudio enabled.
2. Select airpods or other external headphones
3. Disconnect headphones
4. Your roster should not have a permanent orange dot/other users should be able to hear you after the device switches to built in microphones.

**Checklist:**

y

n, ignoring browser behavior extension

n


georgezy-amzn (approved):

ltrung (approved):
Very nice :D

xuesichao (approved):

PR #2975, opened 2024-11-01
Prevent dataMessage callback errors from killing a meeting

schontz:
**Issue #:** #2974

**Description of changes:**

Do not create a fatal error when the user-defined callbacks throw an error.

**Testing:**

Unit tested. POC tested in internal application.

**Checklist:**

1. Have you successfully run `npm run build:release` locally?
Yes

2. Do you add, modify, or delete public API definitions? If yes, has that been reviewed and approved?
No

3. Do you change the wire protocol, e.g. the request method? If yes, has that been reviewed and approved?
No

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
Yes

Issue #2974, opened 2024-11-01
User-defined DataMessage callback errors should not be fatal


schontz:
### What happened and what did you expect to happen?

1. `subscribeToTranscriptEvent` callback received unexpected data and threw an error
2. **Actual:** The meeting died (fatal error)
3. **Expected:** The meeting does not die

### Have you reviewed our existing documentation?

- [X] GitHub [issues](https://github.com/aws/amazon-chime-sdk-js/issues)
- [X] FAQs [page](https://aws.github.io/amazon-chime-sdk-js/modules/faqs.html)
- [X] README [page](https://github.com/aws/amazon-chime-sdk-js#amazon-chime-sdk-for-javascript)
- [X] Feature-specific [guides](https://github.com/aws/amazon-chime-sdk-js#guides)

### Reproduction steps

Something like this:

```
subscribeToTranscriptEvent(() => {
  throw new Error('bad callback');
});
```

### Amazon Chime SDK for JavaScript version

main

### What browsers are you seeing the problem on?

Chome

### Browser version

lateset

### Browser console logs

Nothing is logged to the console, because the error is swallowed up. It does exist in the logger because fatal errors are logged there.

ltrung:
I think it is expected that realtimeAPI throw fatal error though

ltrung:
Closing as the PR has been merged.

PR #2976, opened 2024-11-01
Prevent dataMessage callback errors from killing a meeting

hensmi-amazon:
Move from https://github.com/aws/amazon-chime-sdk-js/pull/2975 to make integration tests work. Not trying to steal git coauthorship i swear!

**Issue #:** #2974

**Description of changes:**

Do not create a fatal error when the user-defined callbacks throw an error.

**Testing:**

Unit tested. POC tested in internal application.

**Checklist:**

Yes

No

No

Yes


src/utils/Utils.ts
```diff
@@ -80,3 +80,43 @@ export function getRandomValues(buffer: Uint32Array): void {
...
+  let suppressedError: SuppressedError | undefined;
+  for (const value of iterable) {
+    try {
+      callback(value);
+    } catch (err) {
+      suppressedError = new SuppressedError(err, suppressedError);
+    }
+  }
+  if (suppressedError) {
+    throw suppressedError;
```
ltrung, line 120: So essentially, before we threw fatal error at the first error encountered but now we throw a non fatal error with the last one encountered?

ltrung (approved):

src/utils/Utils.ts
hensmi-amazon: yup!

PR #2986, opened 2024-11-05
Add per-frame encoding time in video encoding CPU connection health policy

shi-su:
**Issue #:**

**Description of changes:**
This change is to add per-frame encoding time as an additional criterion in video encoding CPU connection health policy. Coded will only degrade when both total encode time and per-frame encode time is above threshold. The purpose is to avoid codec degradation with high framerate SVC. Note that video encoder counts each spatial layer of a frame as a frame encoded (e.g., 30 fps video with L3T3 scalability mode has a encoded framerate of 90 fps), the default value of per-frame encode time threshold has this pattern considered. 

**Testing:**
Smoke tested and verified video codec degradation only when both total encode time and per-frame encode time is above threshold. 

No

**Checklist:**

y

n

n


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Add per-frame encoding time in video encoding CPU connection health policy
```
hensmi-amazon, line 16: I messed this up in a previous PR (David's commit added `Unreleased` section) but will fix it myself :).

src/connectionhealthpolicy/VideoEncodingCpuConnectionHealthPolicy.ts
```diff
@@ -12,19 +12,23 @@ export default class VideoEncodingCpuConnectionHealthPolicy
...
   constructor(configuration: ConnectionHealthPolicyConfiguration, data: ConnectionHealthData) {
     super(configuration, data, VideoEncodingConnectionHealthPolicyName.VideoEncodingCpuHealth);
     this.consecutiveHighEncodeCpuThreshold = configuration.consecutiveHighEncodeCpuThreshold;
     this.highEncodeCpuMsThreshold = configuration.highEncodeCpuMsThreshold;
+    this.highEncodeCpuMsPerFrameThreshold = configuration.highEncodeCpuMsPerFrameThreshold;
   }
 
   health(): number {
+    const videoEncodingTimeIsHigh =
+      this.currentData.videoEncodingTimeInMs >= this.highEncodeCpuMsThreshold &&
```
hensmi-amazon, line 27: It _seems_ like we don't need both checks. But I assume you have a good reason already? :)

If so, maybe it can be a comment

shi-su: Having both checks there is to avoid degrading codec in two conditions:
1. When framerate is low but per-frame encoding time is high (e.g., high resolution screen share with low motion on screen, encoder sometimes just encode like one frame every second. There's no need to degrade codec if each of these frames takes 50 ms to encode);
2. When frame rate is high (e.g., high framerate and SVC at the same time).

I don't think these two conditions can be excluded with the same check. So kept both here.

shi-su: Added a comment to explain the logic

hensmi-amazon (approved):

PR #3028, opened 2024-12-20
Match framerate in DefaultVideoFrameProcessorPipeline with input stream

shi-su:
**Issue #:**

**Description of changes:**
In `DefaultVideoFrameProcessorPipeline` the framerate is set as 15 fps. As a result, once video frame processor is applied, video framerate will be capped at 15 fps even if input fps is higher. This change matches this framerate with input video stream.

**Testing:**
Smoke tested and verified video fps matches input stream after applying a video filter.

1. Start a call and add an attendee with 30 fps video;
2. Start video and verify video framerate is 30 fps;
3. Apply background background blur and verify video framerate is still 30 fps (actual framerate may be about 3 fps lower due to video frame processing).

**Checklist:**

y

n

n

### DIFF
--- a/src/videoframeprocessor/DefaultVideoFrameProcessorPipeline.ts
+++ b/src/videoframeprocessor/DefaultVideoFrameProcessorPipeline.ts
@@ -144,6 +144,7 @@ export default class DefaultVideoFrameProcessorPipeline implements VideoFramePro
     this.inputVideoStream = inputMediaStream;
     const settings = this.inputVideoStream.getVideoTracks()[0].getSettings();
     this.logger.info(`processing pipeline input stream settings ${JSON.stringify(settings)}`);
+    this.framerate = settings.frameRate;
     this.canvasOutput.width = settings.width;
     this.canvasOutput.height = settings.height;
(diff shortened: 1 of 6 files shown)


src/videoframeprocessor/DefaultVideoFrameProcessorPipeline.ts
```diff
@@ -144,6 +144,7 @@ export default class DefaultVideoFrameProcessorPipeline implements VideoFramePro
     this.inputVideoStream = inputMediaStream;
     const settings = this.inputVideoStream.getVideoTracks()[0].getSettings();
     this.logger.info(`processing pipeline input stream settings ${JSON.stringify(settings)}`);
+    this.framerate = settings.frameRate;
```
hensmi-amazon, line 147: This does have a 'surprise' performance impact to any customers with input > 15 FPS currently using the pipeline. But i think, at least with V2 bblur pipeline (assuming that is the main usage of this code) that the increase is reasonable, and this is a more 'expected' behavior.

hensmi-amazon (approved):

ltrung (approved):

PR #3034, opened 2025-02-04
Ensure content share resolution is always within limits

shi-su:
**Issue #:**

**Description of changes:**
Video track constraint is checked and applied when the video starts. Because content share resolution may change in the middle of a meeting and the current logic does not apply constraint if resolution is within limits at the beginning, content resolution may go above the limit when content is scaled up and trigger resolution enforcements. To avoid this problem after content resolution change, we always apply constraint for content share and use a constraint that will not unnecessarily scale content when resolution is within the limit.

**Testing:**

1. Start a meeting with `FHD 1080p` selected as `Meeting Maximum Content Share Resolution`
2. Let an attendee join the meeting and start a content share pointing to a window with width and height below 1920 and 1080, respectively
3. Scale the window above width of 1920 and height of 1080
4. Ensure encoding width and height are always within 1920 and 1080, respectively

**Checklist:**

y

n

n

### DIFF
--- a/src/task/ReceiveVideoInputTask.ts
+++ b/src/task/ReceiveVideoInputTask.ts
@@ -40,6 +40,10 @@ export default class ReceiveVideoInputTask extends BaseTask {
     }
 
     const constraint: MediaTrackConstraints = {
+      // Chrome may scale content share to the maximum possible resolution within
+      // max width and height even if input resolution is already under the limits.
+      // Adding ideal resizeMode as none to use the original resoluion when possible
+      // and avoid unexpected scaling.
       resizeMode: { ideal: 'none' },
       width: { max: videoQualitySettings.videoWidth },
       height: { max: videoQualitySettings.videoHeight },
--- a/src/task/ReceiveVideoInputTask.ts
+++ b/src/task/ReceiveVideoInputTask.ts
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import AudioVideoControllerState from '../audiovideocontroller/AudioVideoControllerState';
-import VideoQualitySettings from '../devicecontroller/VideoQualitySettings';
 import DefaultModality from '../modality/DefaultModality';
 import { SdkStreamServiceType } from '../signalingprotocol/SignalingProtocol.js';
 import BaseTask from './BaseTask';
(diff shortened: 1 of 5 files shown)


src/task/ReceiveVideoInputTask.ts
```diff
@@ -24,40 +23,35 @@ export default class ReceiveVideoInputTask extends BaseTask {
...
-          trackSettings
-        )}`
+      // Skip applying constraints if the current video track settings are already
+      // lower than the limit except for content share. Always apply constraint for
+      // content attendees because content share resolution may change.
+      return;
+    }
+
+    const constraint: MediaTrackConstraints = {
+      resizeMode: { ideal: 'none' },
```
hensmi-amazon, line 43: This is interesting, whats the effect of not setting this? The description on MDN is:
```
none means that the user agent uses the resolution provided by the camera, its driver or the OS
```
But seems like it will still scale. Maybe a comment could help clarify other similar confusion in the future.

src/task/ReceiveVideoInputTask.ts
```diff
@@ -24,40 +23,35 @@ export default class ReceiveVideoInputTask extends BaseTask {
...
-        height: { ideal: videoQualitySettings.videoHeight },
-        frameRate: { ideal: videoQualitySettings.videoFrameRate },
-      };
-      this.context.logger.warn(
-        `Video track (content = ${isContentAttendee}) will be constrained to: ${JSON.stringify(
-          constraint
-        )} to remain below configured video quality settings, trackSettings: ${JSON.stringify(
-          trackSettings
-        )}`
+      // Skip applying constraints if the current video track settings are already
```
hensmi-amazon, line 36: Any downside to just setting these on camera as well? Or maybe we can just consider this reducing impact of the change.

src/task/ReceiveVideoInputTask.ts
shi-su: Chrome would scale content share to the maximum resolution within max width and height without this configured. This is not what we want. Adding this resize mode with ideal will make browser not doing scaling if resolution is within limits, otherwise it will scale to meet max requirements. Firefox and safari don't have this option, but the two browsers don't have the unexpected scaling behavior and is not influenced by this.

hensmi-amazon: Gotcha, perfect information.

src/task/ReceiveVideoInputTask.ts
shi-su: Yeah, I would say the reason is to reduce impact. In fact, I noticed applying constraint to camera may make the browser insist to scale it to 640x480 when resolution is configured low. It never happened to content share (I tried a lot of different resolutions).

PR #3068, opened 2025-04-25
Degrade from VP9/AV1 with more than 15 concurrent video sender

shi-su:
**Issue #:**

**Description of changes:**
Some devices may have performance issues when receiving a high number of VP9/AV1 stream. This change degrades video streams to video codecs with low computation cost in such cases to avoid decoding issues.

**Testing:**
Tested and verified codec is degraded with more than 15 video senders

1. Have 2 attendees join a meeting with VP9 as video send codec preference and turn on video
2. Add 15 more attendee to join the meeting with VP9 as preferred codec and turn on video
3. Verify video codec for all attendees are degraded to H.264

**Checklist:**

y

n

n


src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -862,6 +868,18 @@ export default class DefaultAudioVideoController
...
 
+  private updateNumberOfPublishedVideoSources(): void {
+    const selfAttendeeId = this.configuration.credentials.attendeeId;
+    const hasLocalVideo = this._videoTileController.hasStartedLocalVideoTile();
+    if (this.meetingSessionContext.videoStreamIndex !== null) {
+      const videoCount =
+        this.meetingSessionContext.videoStreamIndex.numberOfVideoPublishingParticipantsExcludingSelf(
+          selfAttendeeId
+        ) + (hasLocalVideo ? 1 : 0);
+      this.connectionHealthData.setNumberOfPublishedVideoSources(videoCount);
```
hensmi-amazon, line 879: For this change it's fine, but theoretically i feel like for future flexibility, 'health data' would separate out send and receive stream count. What do you think?

shi-su: Yeah, the count of receive stream might be a more precise one to determine codec fallback in conditions like this. It does require the mechanism to feedback info from receiver to sender. It will bring back the problem of controlling encoding from decoding and it have to be done in a systematic way (together with feedback of other information). I agree it would be a good improvement for future, but for this stage, I think we can proceed with send stream count.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [3.28.0] - 2025-02-13
 
 ### Added
+- Degrade video codec from VP9/AV1 when there are more than 15 video senders
```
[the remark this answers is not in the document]
shi-su: There might be. But we already use the word in for such behaviors and it sorta accurately reflects what we are doing, so I think we keep using it for consistency

hensmi-amazon (approved):

PR #3103, opened 2025-09-16
Add RetryCount to AttendeePresenceReceived event

ltrung:
**Description of changes:**
- Add RetryCount attributes to AttendeePresenceReceived to know how many times it takes to connect so we can differentiate whether the duration is caused by reconnection

**Testing:**

Deployed demo and viewed that the attendee presence meeting events now have RetryCount attributes.

**Checklist:**

Yes

No

No


src/audiovideocontroller/AudioVideoControllerState.ts
```diff
@@ -201,6 +201,8 @@ export default class AudioVideoControllerState {
   // and subscribe-ack)
   isSessionConnected: boolean = false;
 
+  retryCount: number = 0;
```
devalevenkatesh, line 204: Since we are moving to meeting session context now, do we have to reset this once reconnected? Earlier it was tied to the DefaultACController, not the shared context hence checking.

devalevenkatesh (dismissed):

PR #3118, opened 2025-11-20
Remove unload/load signaling client behavior so that we signal 1001 (Going Away) properly

hensmi-amazon:
**Issue #:** None

**Description of changes:** 

The existing code is supposed to allow backend to see a gracefully closed connection, but doesn't work because the leave is not received by the backend unless the client is extremely close to backend region. This is likely due to the data in flight causing chrome to non-gracefully close the connection,  and leads to the backend seeing a snap (1006), instead of a 1001 (which can be considered graceful).

**Testing:**

Confirmed that backend sees 1001 on refresh.

No

**Checklist:**

y

n

n


ltrung:
Were you able to test and verify backend receive 1001?

hensmi-amazon:
> Doesn't it complain you need to update CHANGELOG?

Gosh yes this gets me every time.

> Were you able to test and verify backend receive 1001?

Yes Chrome closes with 1001 in all RTT environments.

ltrung (approved):

PR #3131, opened 2026-01-08
Refactored Encoded Transform management into it's own component for better support of non-redundant audio transforms

hensmi-amazon:
**Issue #:** N/A

**Description of changes:** 

Note that given the size of this PR i am first skipping unit tests to save myself some time in case i need to do significant additional refactoring.

We would like to add transforms for metric/monitoring, and at some point E2EE. This change does that while adding a simple transform PPS metric. First packet metrics will occur in a future PR.

**Testing:**

* Confirmed media/RED metrics report and the RED applies. Have tested legacy path by disabling encoded stream transform. Have tested chrome/FF/safari. Have tested init failure, and failure during connection.

**Checklist:**

n

not really that public :)

n

### DIFF
--- a/demos/browser/package-lock.json
+++ b/demos/browser/package-lock.json
@@ -38,10 +38,10 @@
         "sass-loader": "^12.1.0",
         "style-loader": "^3.2.1",
         "ts-loader": "^9.3.0",
-        "typescript": "4.2.3",
+        "typescript": "~5.8.0",
         "webpack": "^5.94.0",
         "webpack-cli": "^4.8.0",
-        "webpack-dev-server": "^4.9.0"
+        "webpack-dev-server": "^5.2.3"
       }
     },
     "../..": {
@@ -59,48 +59,52 @@
         "ua-parser-js": "^1.0.40"
       },
       "devDependencies": {
+        "@eslint/js": "^9.39.2",
         "@fluffy-spoon/substitute": "^1.89.0",
         "@types/chai": "^4.3.5",
         "@types/chai-as-promised": "^7.1.0",
         "@types/mocha": "^5.2.6",
-        "@types/sinon": "^7.0.12",
+        "@types/node": "^20.19.30",
+        "@types/sinon": "^17.0.0",
         "@types/uuid": "^8.3.1",
-        "@typescript-eslint/eslint-plugin": "^4.18.0",
-        "@typescript-eslint/parser": "^4.18.0",
+        "@typescript-eslint/eslint-plugin": "^8.22.0",
+        "@typescript-eslint/parser": "^8.22.0",
         "chai": "^4.3.7",
         "chai-as-promised": "^7.1.1",
         "cross-env": "^6.0.3",
-        "eslint": "^7.22.0",
-        "eslint-config-prettier": "^6.15.0",
-        "eslint-plugin-prettier": "^3.1.4",
-        "eslint-plugin-simple-import-sort": "^5.0.3",
+        "eslint": "^8.57.1",
+        "eslint-config-prettier": "^9.1.2",
+        "eslint-plugin-prettier": "^5.5.5",
+        "eslint-plugin-simple-import-sort": "^12.1.1",
         "esm": "^3.2.25",
-        "fetch-mock": "9.10.7",
+        "fetch-mock": "^12.6.0",
         "fs-extra": "^8.1.0",
         "git-rev-sync": "^3.0.2",
         "longjohn": "^0.2.12",
         "mocha": "^10.0.0",
         "node-fetch": "^2.6.1",
         "nyc": "^17.1.0",
-        "prettier": "^2.1.2",
+        "prettier": "^3.0.0",
         "protobufjs-cli": "^1.1.2",
         "rimraf": "^3.0.2",
-        "sinon": "^7.3.2",
+        "sinon": "^17.0.0",
         "spawn-wrap": "^2.0.0",
         "ts-node": "^9.1.1",
-        "typedoc": "0.21.2",
-        "typedoc-plugin-merge-modules": "^3.1.0",
-        "typescript": "^4.2.3"
+        "typedoc": "^0.27.9",
+        "typedoc-plugin-merge-modules": "^6.1.0",
+        "typescript": "~5.8.0",
+        "typescript-eslint": "^8.54.0"
       },
       "engines": {
-        "node": ">=18",
-        "npm": ">=8"
+        "node": ">=20",
+        "npm": ">=10"
       }
     },
     "../../node_modules/@aws-crypto/sha256-browser": {
       "version": "5.2.0",
       "resolved": "https://registry.npmjs.org/@aws-crypto/sha256-browser/-/sha256-browser-5.2.0.tgz",
       "integrity": "sha512-AXfN/lGotSQwu6HNcEsIASo7kWXZ5HYWvfOmSNKDsEqC4OashTp8alTmaz+F7TC2L083SFv5RdB+qU3Vs1kZqw==",
+      "license": "Apache-2.0",
       "dependencies": {
         "@aws-crypto/sha256-js": "^5.2.0",
         "@aws-crypto/supports-web-crypto": "^5.2.0",
@@ -115,6 +119,7 @@
       "version": "5.2.0",
       "resolved": "https://registry.npmjs.org/@aws-crypto/sha256-js/-/sha256-js-5.2.0.tgz",
       "integrity": "sha512-FFQQyu7edu4ufvIZ+OadFpHHOt+eSTBaYaki44c+akjg7qZg9oOQeLlk77F6tSYqjDAFClrHJk9tMf0HdVyOvA==",
+      "license": "Apache-2.0",
       "dependencies": {
         "@aws-crypto/util": "^5.2.0",
         "@aws-sdk/types": "^3.222.0",
@@ -128,6 +133,7 @@
       "version": "5.2.0",
       "resolved": "https://registry.npmjs.org/@aws-crypto/util/-/util-5.2.0.tgz",
       "integrity": "sha512-4RkU9EsI6ZpBve5fseQlGNUWKMa1RLPQ1dnjnQoe07ldfIzcsGb5hC5W0Dm7u423KWzawlrpbjXBrXCEv9zazQ==",
+      "license": "Apache-2.0",
       "dependencies": {
         "@aws-sdk/types": "^3.222.0",
         "@smithy/util-utf8": "^2.0.0",
@@ -138,6 +144,7 @@
       "version": "2.2.0",
       "resolved": "https://registry.npmjs.org/@smithy/is-array-buffer/-/is-array-buffer-2.2.0.tgz",
       "integrity": "sha512-GGP3O9QFD24uGeAXYUjwSTXARoqpZykHadOmA8G5vfJPK0/DC67qa//0qvqrJzL1xc8WQWX7/yc7fwudjPHPhA==",
+      "license": "Apache-2.0",
       "dependencies": {
         "tslib": "^2.6.2"
       },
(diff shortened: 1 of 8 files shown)


script/generate-media-transform-worker-code.js
```diff
@@ -0,0 +1,149 @@
...
+ */
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { execSync } = require('child_process');
+const fs = require('fs');
+const path = require('path');
+
+// Create a temporary tsconfig file to transpile media transform worker files.
+const configDir = 'config';
+const workerTsconfig = 'tsconfig.mediatransformworker.json';
```
devalevenkatesh: So no options using vite or Webpack to get the `.js` files from `.ts`? Looks like the file was renamed or the script mainly.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -462,6 +484,22 @@ export default class DefaultAudioVideoController
   private async actionConnect(reconnecting: boolean): Promise<void> {
     this.initSignalingClient();
 
+    if (this._encodedTransformWorkerManager?.isEnabled()) {
```
devalevenkatesh: Should we create a Task for this and let it run when sessions tasks run? InitEncodedTransformWorkerTask?

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -462,6 +484,22 @@ export default class DefaultAudioVideoController
...
+      // If this fails the manager should disable itself, so that when a reconnect is triggered
+      // by the exception, it will disable redundancy
+      await this._encodedTransformWorkerManager.initialize();
```
devalevenkatesh: If audioProfile can set the observer on the worker managers disabled/enabled state can the profile internally disable itself and send observer callback upstream if required? Seems like one is dependent on another hence checking.

src/encodedtransformmanager/EncodedTransformWorkerManager.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * Remove an observer
+   */
+  removeObserver(observer: EncodedTransformWorkerManagerObserver): void;
+
+  /**
+   * Reset all transform managers to their initial state
+   */
+  reset(): void;
```
devalevenkatesh: Does this need to be a promise or all operations for resetting are synchronous?

src/encodedtransformmanager/MediaMetricsEncodedTransformManager.ts
```diff
@@ -0,0 +1,199 @@
...
+  /**
+   * Synchronization source identifier for the stream.
+   */
+  ssrc: number;
+  /**
+   * Number of packets processed in the reporting interval.
+   */
+  packetCount: number;
+  /**
+   * Unix timestamp (ms) when metrics were collected.
```
devalevenkatesh: Can drop (ms)?

src/encodedtransformmanager/MediaMetricsEncodedTransformManager.ts
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    if (message.type !== COMMON_MESSAGE_TYPES.METRICS) return;
+
+    const metrics = message.message?.metrics ? JSON.parse(message.message.metrics) : null;
+    if (!metrics) return;
+
+    switch (message.transformName) {
+      case TRANSFORM_NAMES.AUDIO_SENDER:
+        this.audioSender = metrics as Record<number, EncodedTransformMediaStreamMetrics>;
```
devalevenkatesh: Just to clarify: The metrics received also abide by the `Record<number, EncodedTransformMediaStreamMetrics>;`? Checking for runtime type conversion safety here.

src/encodedtransformmanager/MediaMetricsEncodedTransformManager.ts
```diff
@@ -0,0 +1,199 @@
...
+  private videoSender: Record<number, EncodedTransformMediaStreamMetrics> = {};
+  private videoReceiver: Record<number, EncodedTransformMediaStreamMetrics> = {};
+
+  private observers: Set<EncodedTransformMediaMetricsObserver> = new Set();
+  private metricsReportingScheduler: IntervalScheduler | null = null;
+  private readonly METRICS_REPORTING_INTERVAL_MS = 1000;
+
+  constructor(worker: Worker, logger: Logger) {
+    super(worker, logger);
+    this.startMetricsReporting();
```
devalevenkatesh: So constructor is creating and also starting, so the moment this is created it will start reporting even if worker fails or some setup goes wrong. So should we add a start API on these Manager's? That way one can control when things start. Also is stop needed or destroy is sufficient?

src/encodedtransformmanager/RedundantAudioEncodedTransformManager.ts
```diff
@@ -0,0 +1,405 @@
...
+    // to calculate the loss for the given time window
+    return 0;
+  }
+
+  /**
+   * Returns the number of encodings based on packetLoss value. This is used by `DefaultTransceiverController` to
+   * determine when to alert the encoder to update the number of encodings. It also determines if we need to
+   * turn off red in cases of very high packet loss to avoid congestion collapse.
+   */
+  static getNumRedundantEncodingsForPacketLoss(packetLoss: number): [number, boolean] {
```
devalevenkatesh: This ties this manager to the DefaultTranceiverController. No observers could be introduced for this?

src/encodedtransformworker/RedundantAudioEncodedTransform.ts
```diff
@@ -0,0 +1,1425 @@
...
+
+  protected transformName(): string {
+    return TRANSFORM_NAMES.REDUNDANT_AUDIO;
+  }
+
+  /**
+   * Calculates recommended redundancy level based on packet loss percentage.
+   * @returns [numEncodings, shouldDisableRED] - encodings (0-2) and whether to disable RED entirely
+   */
+  static getNumRedundantEncodingsForPacketLoss(packetLoss: number): [number, boolean] {
```
devalevenkatesh: How is this different than the manager (`RedundantAudioEncodedTransformManager`)?

src/transceivercontroller/SimulcastTransceiverController.ts
```diff
@@ -99,19 +101,29 @@ export default class SimulcastTransceiverController extends DefaultTransceiverCo
         streams: [this.defaultMediaStream],
       });
 
-      if (this.meetingSessionContext?.audioProfile?.hasRedundancyEnabled()) {
-        // This will perform additional necessary setup for the audio transceiver.
-        this.setupAudioRedWorker();
+      if (this.encodedTransformWorkerManager.isEnabled()) {
```
pracheth: Since encodedTransformWorkerManager is optional, should this be `encodedTransformWorkerManager?.isEnabled()`?

src/encodedtransformmanager/MediaMetricsEncodedTransformManager.ts
pracheth: Hmm observers won't be setup at this point no?

script/barrelize.js
```diff
@@ -52,6 +52,16 @@ const ignoredTypes = [
   'index',
   'ScreenSignalingProtocol',
 
+  // Worker transform classes - these are bundled into the worker and shouldn't be exported directly
```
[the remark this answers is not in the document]
hensmi-amazon: We aren't technically changing any of the existing generated files. My purpose was actually to allow breaking changes here in the future :) . Does that answer?

script/generate-media-transform-worker-code.js
hensmi-amazon: We don't currently have a dependency on either outside of the demo, I just followed previous logic. Is that reasonable?

src/audioprofile/AudioProfile.ts
```diff
@@ -68,4 +68,15 @@ export default class AudioProfile {
   hasRedundancyEnabled(): boolean {
     return this.enableAudioRedundancy;
   }
+
+  /**
+   * Disable audio redundancy in this configuration. Setting this
+   * value will not impact an ongoing session, and a reconnect must
```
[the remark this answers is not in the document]
hensmi-amazon: I think i elaborated too much in this docstring for our internal usage. Customers are unlikely to assume that if they hold onto the config, that the SDK will respond to changes. I added this because the audio video controller has a pattern of disabling audio redundancy in some paths. I just wanted to make it clear that those paths always need to trigger a reconnect to have any impact. But I can move this note to the paths themselves.

src/audiovideocontroller/DefaultAudioVideoController.ts
hensmi-amazon: feels a bit overkill? especially since the task can't fail. But maybe I'm missing some of the rationale here.

src/audiovideocontroller/DefaultAudioVideoController.ts
hensmi-amazon: Ergh, you are reminding me i didn't add the path for disabling audio redundancy (which TBH I'm not sure why customers should be able to do at this point, or if any customers use it). I will extend `initialize` with a toggle for each transform to enable. 

I think if I understand the comment correctly you mean to wire up the profile to the failure callback? I think since the audio video controller needs to respond anyways it's more explicit *however*, I think i can update the other paths that check AudioProfile's redundant audio field to just check the transform manager enablement status. Then we don't need to modify the AudioProfile that is passed in! think this should be cleaner.

src/encodedtransformmanager/EncodedTransformWorkerManager.ts
hensmi-amazon: Doesn't hurt to do, will change!

src/encodedtransformmanager/MediaMetricsEncodedTransformManager.ts
hensmi-amazon: i don't think a `start`/`stop` would hurt here. I can replace `destroy`, `stop` implies it can be restarted. Can reuse audio video controller without keeping the worker alive i guess :D.

src/encodedtransformmanager/EncodedTransformWorkerManager.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Observer interface for media transform manager events.
+ */
+export interface EncodedTransformWorkerManagerObserver {
+  /**
+   * Called when the encoded transform manager encounters a failure.
+   */
+  onEncodedTransformWorkerManagerFailed?(error: Error): void;
+}
```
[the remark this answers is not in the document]
hensmi-amazon: I don't like the existing pattern :(. I think years ago Richard said we should break this pattern of one class per file. @devalevenkatesh any opinions here?

src/encodedtransformmanager/MediaMetricsEncodedTransformManager.ts
hensmi-amazon: > Hmm observers won't be setup at this point no?

Technically observers can be added at any time right? Might be missing something

src/encodedtransformmanager/EncodedTransformWorkerManager.ts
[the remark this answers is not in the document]
xuesichao: Looks good to me. It's just a different way of organizing the code. I'd also prefer keeping similar types together if they're small pieces.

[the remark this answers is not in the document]
devalevenkatesh: Yup no problem from, yes I remember, one of the reason was to keep our exports small and API surface smaller clubbed by relevant files.

script/generate-media-transform-worker-code.js
devalevenkatesh: Yes I see you moved things around, no problem. But if workers get adding, might be good to have a worker folder with its own vite/webpack config to make it easy to bundle.

hensmi-amazon: Yeah it's a fair assumption. I think I'm going to hold off for now though, mainly to keep the build time short. In an alternative world this is all pulled remotely :P

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -1720,4 +1746,20 @@ export default class DefaultAudioVideoController
...
+  onEncodedTransformWorkerManagerFailed(error: Error): void {
+    this.logger.error(`Encoded transform worker failed: ${error.message}`);
+
+    if (this.sessionStateController.state() !== SessionStateControllerState.Connected) {
+      this.logger.info('Skipping reconnect for worker failure - session not connected');
+      return;
+    }
+
+    // Trigger reconnect to recreate peer connection without transforms
+    const status = new MeetingSessionStatus(MeetingSessionStatusCode.TaskFailed);
```
devalevenkatesh: Thought: Should we add EncodedTransformManagerFailed or StartEncodedTransformWorkerTaskFailed?

PR #3180, opened 2026-03-14
Fix transform PPS always zero and content share worker crash

hensmi-amazon:
Issue #: N/A

Description of changes:

Fix two encoded transform bugs:

1. AUDIO_RECEIVED_TRANSFORM_PPS and AUDIO_SENT_TRANSFORM_PPS always reported as 0. The timestamp guard in StatsCollector.maybeAddTransformMetrics skipped injecting the cumulative packetCount when the worker hadn't 
reported new data between stats collection cycles. This caused currentMetrics == previousMetrics, making countPerSecond return 0. Removed the guard so the cumulative value is always injected — countPerSecond 
correctly computes the delta between intervals.

2. "Worker not initialized" crash when toggling content share between attendees. DefaultMeetingSession shared one EncodedTransformWorkerManager between the main session and content share AudioVideoController. When 
content share stopped, it terminated the shared worker. A subsequent track event on the main session's peer connection then called setupVideoReceiverTransform on the stopped manager. Fixed by giving content share 
its own DefaultEncodedTransformWorkerManager.

Testing:

Can these tested using a demo application? Please provide reproducible step-by-step instructions.

Bug 1 (Transform PPS):
1. Run the browser demo with log level set to Debug
2. Join two attendees to the same meeting
3. Observe the sending: debug logs — AUDIO_RECEIVED_TRANSFORM_PPS and AUDIO_SENT_TRANSFORM_PPS should report ~50 (previously always 0)

Bug 2 (Content share worker crash):
1. Join two attendees (Alice and Bob) to the same meeting
2. Alice starts content share
3. Alice stops content share
4. Bob starts content share
5. Previously this caused "Worker not initialized" error on Alice's side — now it works without error

Checklist:

1. Have you successfully run npm run build:release locally?

Yes — 2928 passing, 0 errors.

No.

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added PPS metrics on the Encoded Transform when used.
+- Added PPS metrics on Encoded Transforms when used.
 - Added meeting session lifecycle timing tracking and signaling via `MeetingSessionTimingManager`.
 
 ### Removed
--- a/src/meetingsession/DefaultMeetingSession.ts
+++ b/src/meetingsession/DefaultMeetingSession.ts
@@ -34,6 +34,7 @@ export default class DefaultMeetingSession implements MeetingSession, Destroyabl
   private _deviceController: DeviceController;
   private audioVideoFacade: AudioVideoFacade;
   private encodedTransformWorkerManager: EncodedTransformWorkerManager;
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added PPS metrics on the Encoded Transform when used.
+- Added PPS metrics on Encoded Transforms when used.
```
hensmi-amazon, line 12: Avoiding changelog bloat

shi-su (approved):

PR #3245, opened 2026-07-16
fix(statscollector): Skip CPU pressure observe when Permissions Policy disallows it

hensmi-amazon:
**Issue #:**

None

**Description of changes:**

The Compute Pressure API's PressureObserver constructor exists in Chrome even when the compute-pressure Permissions Policy feature is denied, which is the default in cross-origin iframes such as embedded CCP. Calling observe('cpu', ...) there makes Chrome emit a "[Violation] Permissions policy violation: compute-pressure is not allowed in this document." console message that the surrounding try/catch cannot suppress.

Gate start() on document.featurePolicy.allowsFeature('compute-pressure') so the observer is never created when the feature is disallowed, and drop to a debug log instead. When the feature-policy check is unavailable, keep the prior behavior and let observe()'s try/catch handle any rejection.

**Testing:**
N/A
N/A

**Checklist:**

y

n

n

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a Chrome `[Violation] Permissions policy violation: compute-pressure is not allowed in this document.` console message by skipping CPU pressure observation when the `compute-pressure` Permissions Policy feature is not allowed (e.g. in cross-origin iframes such as embedded CCP).
+
 ## [3.32.0] - 2026-05-14
 
 ### Added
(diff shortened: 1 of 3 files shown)


ltrung (approved):

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a Chrome `[Violation] Permissions policy violation: compute-pressure is not allowed in this document.` console message by skipping CPU pressure observation when the `compute-pressure` Permissions Policy feature is not allowed (e.g. in cross-origin iframes such as embedded CCP).
```
ziyiz-amzn, line 18: feel like we may not mention "embeded CCP" in the js sdk lib. Also, no one will know what's CCP here. `cross-origin iframes` should be good.

PR #3251, opened 2026-08-13
Do not pass custom video file if not needed

ltrung:
**Description of changes:**
Browser compatiblity test failed since we pass the custom video file link locally which SauceLabs cannot access remotely which made video could not be rendered.

Conditionally pass the flag only for video processing test now.

Also update so that all tests run to the end and do not quit if one test fails.

**Testing:**

**Checklist:**

### DIFF
--- a/integration/configs/WebDriverBaseConfig.js
+++ b/integration/configs/WebDriverBaseConfig.js
@@ -3,6 +3,23 @@ const getConfig = () => {
   const macVersionMatch = platformName.match(/^macOS\s+(\d+)/);
   const isMacOS14OrHigher = macVersionMatch && parseInt(macVersionMatch[1], 10) >= 14;
 
+  // Only tests that verify pixel output (e.g. background segmentation filters) need a
+  // deterministic fake video source. The file lives on the machine running the test, so it is
+  // only reachable when the browser is on the same host (`--host local`). On remote hosts like
+  // SauceLabs the path does not exist, which breaks video device enumeration for every test.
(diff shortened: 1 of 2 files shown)


integration/configs/WebDriverBaseConfig.js
```diff
@@ -3,6 +3,23 @@ const getConfig = () => {
...
+  } catch (e) {
+    testName = '';
+  }
+  const testsNeedingCustomVideoFile = ['VideoProcessingTest'];
+  const useCustomVideoFile =
+    process.env.HOST === 'local' && testsNeedingCustomVideoFile.includes(testName);
+  const customVideoFileArgs = useCustomVideoFile
+    ? ['--use-file-for-fake-video-capture=' + require('path').resolve(__dirname, '../fake_stream/output.y4m')]
+    : [];
+
```
dinmin-amzn, line 22: Does this mean that the `VideoProcessingTest` will always fail on Saucelabs since the custom video file isn't available?

hensmi-amazon (approved):

integration/configs/WebDriverBaseConfig.js
ltrung: Yep. We need to upload it to s3 so SauceLabs can read it remotely.

ziyiz-amzn: How did the resolution (set a file for local test) fix the SauceLabs compatibility test failure?

Issue #3247, opened 2026-07-22
Default AV1 content codec (3.28+) silently disables the simulcast high layer when enableSimulcastForContentShare(true) is used


paul365:
## What happened?

After upgrading from SDK 3.26 to 3.29, all our content shares (screen shares) dropped to an effective 960×540. With `enableSimulcastForContentShare(true)` (used per the Simulcast guide), the content sender allocates the usual two streams (low: 300 kbps / 5 fps / scale 2, high: 1200 kbps / full resolution), but **the high stream never transmits**. Server-side stream indexes report the high stream with framerate 0 and ~0 bitrate (or omit it entirely), while the low stream carries all traffic. Every receiver — attendees and media capture — gets 960×540.

Comparing production logs before and after the upgrade isolates the change:

- **SDK 3.26 (H.264 content codec):** the high stream transmitted continuously during shares (hundreds of kbps, framerate tracking content activity), across presenters and machines.
- **SDK 3.29/3.30 (AV1 content codec):** the high stream never transmitted in any share we measured (7 shares, 4 machines, Chrome 149–150 on Windows), including through multi-minute periods of heavy motion where the low stream saturated its 300 kbps cap.

## Why it happens (our analysis)

`DefaultAudioVideoController`'s constructor only applies the AV1-first default codec preferences when simulcast is **off**:

```ts
if (!this.configuration.enableSimulcastForUnifiedPlanChromiumBasedBrowsers) {
  // We do not set default video codec for simulcast until simulcast support for VP9/AV1 is added
  const isContentShare = ...
  if (isContentShare && ...hasChromiumWebRTC()) {
    this.videoSendCodecPreferences = [ av1Main(), vp9Profile0(), h264ConstrainedBaselineProfile(), vp8() ];
  }
  ...
}
```

However, the content share controller (and its `AudioVideoController`) is constructed when the `MeetingSession` is created — before the application calls `enableSimulcastForContentShare(true)` (which the Simulcast guide presents as callable any time before starting the share). At construction time the simulcast flag is still false, so the AV1 preference is applied; enabling simulcast afterwards produces the AV1 + simulcast combination the comment above describes as unsupported. The result is that the top simulcast encoding never activates in Chrome.

## What we expected

[...]

PR #3260, opened 2026-09-14
Use SVC instead of simulcast for AV1 content share

hensmi-amazon:
**Issue #:**  https://github.com/aws/amazon-chime-sdk-js/issues/3247

**Description of changes:**

Since 3.28 the default content share codec on Chromium-based browsers is AV1, which does not support simulcast. When simulcast was enabled for an AV1 content share (enableSimulcastForContentShare is typically called after session creation, so the AV1-first defaults are already applied), the high simulcast layer silently never transmitted, capping all viewers at the low layer (~960x540/5fps).

Detect this combination at connect time and use SVC instead, which provides equivalent adaptation for AV1, with a warning log. Simulcast is preserved when an H.264/VP8 preference is set via setContentShareVideoCodecPreferences. Also document the codec restrictions in the simulcast/video codec guides and facade docs.

**Testing:**

**Steps**
3. Join a meeting as `Sender`. Leave **Content Codec** on **"Meeting Default"** — this is what makes the SDK apply its AV1-first content share defaults, and is the case in the issue report.
4. Join the same meeting in a second tab as `Viewer`.
5. In the Sender tab: content-share dropdown → **Configure Content Sharing** → check **Enable Simulcast** → **Save**.
6. Content-share dropdown → pick a source (**Test Video** is deterministic; screen capture behaves the same) → click the **content share** button.
7. On the Sender, read the content-share peer connection's `outbound-rtp` video stats in `chrome://webrtc-internals` (or `getStats()`), and check the resolution the Viewer receives. Watch the console for SDK warnings.

**Before this change**: two encodings are created (`rid: low`, `rid: hi`) but `hi` stays at `framesEncoded: 0` / `bytesSent: 0` forever, so every viewer only ever gets the low layer. No error, no warning.

**After this change**: a warning is logged, simulcast is replaced by SVC, and viewers receive full resolution.

**Results.** The demo's Test Video source is 640×360, so a real screen share shows larger numbers. `framesEncoded` is sampled at t=0 and t=+6s. To vary the codec, change **Content Codec** on the join screen (step 3) — note the demo expands each selection into a fallback list, shown in the "content codec preferences" column.

| # | content codec preferences | negotiated | simulcast / SVC | sender `outbound-rtp` | viewer `inbound-rtp` | warning |
|---|---|---|---|---|---|---|
| 1 | `[AV1, VP9, H264, VP8]` (Meeting Default) — **on `main`** | AV1 | simulcast | `low` 320×180@5fps 70→100 frames; **`hi` 0 frames, 0 bytes** | 320×180 @4fps | none |
| 2 | `[AV1, VP9, H264, VP8]` (Meeting Default) | AV1 | SVC | one AV1 **L1T3** 640×360@30fps, 410→590 frames | 640×360 @30fps, 583 frames | `Simulcast is not supported for AV1 content share. Using SVC instead.` |
| 3 | `[VP9, H264, VP8]` (Content Codec = VP9) | VP9 | SVC | one VP9 L1T1 640×360@31fps, 414→594 frames | 640×360 @30fps, 594 frames | `Simulcast is not supported for VP9 content share. Using SVC instead.` |
| 4 | `[H264 CBP, VP8]` (Content Codec = H.264 CBP) | H264 | simulcast | `hi` 640×360@5fps 70→100 **and** `low` 320×180@5fps 70→99 | 640×360 @5fps | none |
| 5 | `[H264 High, AV1]` (see note) | **AV1** | simulcast | `low` 320×180@5fps 69→99; **`hi` 0 frames, 0 bytes** | 320×180 @5fps | `Simulcast is enabled but the negotiated video send codec AV1 does not support simulcast, so only the lowest layer will transmit.` |

Row 1 is the bug, measured on unmodified `main` with the same steps. Row 2 is the same configuration after the change. Row 3 covers VP9, which has the same dead-high-layer behavior and is the second entry of the default content list, i.e. what a sender falls back to when AV1 is unavailable. Row 4 confirms simulcast still works for the codecs that support it.

Row 5 is the case that motivated the review feedback and is **not** reachable from the Content Codec dropdown (selecting "H.264 High" yields `[High, CBP, VP8]`, all simulcast-capable). Reproduce it by setting the preferences directly before starting the share:

```js
meetingSession.audioVideo.setContentShareVideoCodecPreferences([
  VideoCodecCapability.h264HighProfile(),
  VideoCodecCapability.av1Main(),
]);

**Checklist:**

y

n

n

### CODE REVIEW

**Review by pracheth — DISMISSED** [review]

File: src/audiovideocontroller/DefaultAudioVideoController.ts [hunk 1]
```diff
@@ -526,15 +527,38 @@ export default class DefaultAudioVideoController
...
+    const isContentAttendee = new DefaultModality(
+      this.configuration.credentials.attendeeId
+    ).hasModality(DefaultModality.MODALITY_CONTENT);
+    let overrideSimulcastWithSVCForAv1Content = false;
+    if (
+      this.enableSimulcast &&
+      isContentAttendee &&
+      this.videoSendCodecPreferences !== undefined &&
+      this.videoSendCodecPreferences.length > 0 &&
+      this.videoSendCodecPreferences[0].equals(VideoCodecCapability.av1Main()) &&
```
devalevenkatesh: Checking if AV1 could be in any other positions than first and whether we need to handle that case.

devalevenkatesh (dismissed):

src/audiovideocontroller/DefaultAudioVideoController.ts
hensmi-amazon: good catch lol

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -143,6 +144,11 @@ export default class DefaultAudioVideoController
   private static MIN_VOLUME_DECIBELS = -42;
   private static MAX_VOLUME_DECIBELS = -14;
   private static PING_PONG_INTERVAL_MS = 10000;
+  // Chromium browsers only implement simulcast for these codecs.
+  private static SIMULCAST_CAPABLE_CODEC_NAMES = [
```
devalevenkatesh, line 148: Can create a utility for this or export it and reuse as this is getting reused in multiple places now.

src/audiovideocontroller/DefaultAudioVideoController.ts
```diff
@@ -526,15 +532,40 @@ export default class DefaultAudioVideoController
       this.configuration.enableSimulcastForUnifiedPlanChromiumBasedBrowsers &&
       new DefaultBrowserBehavior().hasChromiumWebRTC();
 
+    // Chromium browsers only support simulcast for H.264 and VP8. A content share preferring
+    // AV1 (the default since 3.28) or VP9 with simulcast enabled would silently never transmit
+    // the high simulcast layer, so fall back to SVC for this connection instead.
+    const isContentAttendee = new DefaultModality(
```
devalevenkatesh: For understanding: Why this is only a limitation for content share and not video in general?

devalevenkatesh (approved):

ltrung (approved):

